PR #1, opened 2010-09-09
Rackspace enhancements

tpherndon:
For the Rackspace driver, create_node() and deploy_node() return different node references unnecessarily.  The create_node() method returns a node instance with password populated in extra, while deploy_node() returns a node from list_nodes(), which does not include passwords.  I've modified base.NodeDriver.deploy_node() to return the node instance it gets from calling create_node(), thus including the password.

I've also modified drivers.rackspace.RackspaceNodeDriver and added set_password() and set_server_name() methods, along with a private implementation method and modifying request() to make PUT use xml.  These changes expand the API to make available the Rackspace API methods for changing the password and server name.


mig5:
plus 1. I needed this feature for deploy_node and can confirm it works.

benatkin:
It's been almost 10 months since the last commit in this pull request. How about closing this pull request and encouraging the participants to submit smaller pull request(s) that are based on the current code base?

tpherndon:
I'd say it's beyond obvious that this repository is not working within the standard Github workflow, and is really meant just as a mirror of the official repository. I am not currently using libcloud for anything, and don't care to maintain the fork or update the patch to work with the current code base.

PR #3, opened 2010-11-05
Trunk

tobami:
The Dummy Driver list_locations() method returns 3 dummy locations. They all have the same id=1, which is not good(tm)


jcsalterego:
Thanks for your patch! This has been fixed in r1054119.

https://issues.apache.org/jira/browse/LIBCLOUD-70

tobami:
Great, thanks!

PR #16, opened 2011-06-15
OpenStack driver fixes

azhuchkov:
I have faced with 2 issues in the OpenStack driver:
1. the driver uses Rackspace pricing - it looks weird and causes error if OpenStack cloud contains 8+ flavors. From my point of view it's better to set the price 0.0 for all flavors by default (user is allowed to provide custom pricing).
2. Sometimes response doesn't contain valid XML - it contains plaint text like 'Request accepted'. It causes MalformedResponseError. So I added content-type header examination.


Kami:
Hello,

Can you please open a new ticket on Libcloud JIRA - https://issues.apache.org/jira/browse/LIBCLOUD.

For now you don't even need to attach the patch there if you don't want, just include the link to this pull request.

Thanks

azhuchkov:
Surely. 
Here is the link for the ticket: https://issues.apache.org/jira/browse/LIBCLOUD-92

benatkin:
From the [issue on JIRA](https://issues.apache.org/jira/browse/LIBCLOUD-92), which has been marked as fixed:

> I have made some style fixes to your patch and merged it into trunk.

I think this pull request can be closed.

Kami:
Sadly we (libcloud developers) don't have admin access to the Github repository and can't manage the pull requests which means it can only be closed by the author.

PR #17, opened 2011-06-18
When I tried to use ScriptDeployment i found that the permissions were sc

Jc2k:
When I tried to use ScriptDeployment i found that the permissions were scrambled after the script had been copied to the node. This resolved the issue.


Kami:
Committed into trunk. Thanks.

Next time please also open a ticket on our bug tracker (https://issues.apache.org/jira/browse/LIBCLOUD).

The problem is that we don't have admin access to the Github repo and we also don't get notified when a new pull request is opened.

PR #18, opened 2011-06-26
Support for CloudStack and thus Ninefold's compute platform

jeamland:
These commits add support for cloud.com's CloudStack API and thus allow the use of Ninefold (ninefold.com) as a compute/loadbalancer provider.


libcloud/common/cloudstack.py
```diff
@@ -0,0 +1,101 @@
...
+
+        return params
+
+    def pre_connect_hook(self, params, headers):
+        params['signature'] = self._make_signature(params)
+
+        return params, headers
+
+    def _sync_request(self, command, **kwargs):
+        "Make a synchronous API request. These return immediately."
```
Kami: Hm, you have probably meant "These **don't** return immediately.", because a synchronous request blocks :)

libcloud/common/cloudstack.py
```diff
@@ -0,0 +1,101 @@
...
+        success = True
+
+        while True:
+            result = self._sync_request('queryAsyncJobResult', jobid=job_id)
+            if result.get('jobstatus', 0) != 0:
+                break
+            time.sleep(self.driver.async_poll_frequency)
+
+        if result['jobstatus'] == 2:
+            raise Exception(result)
```
Kami: Hm, what does jobstatus 2 means? We should probably throw a more granular exception.

libcloud/compute/drivers/cloudstack.py
```diff
@@ -0,0 +1,250 @@
...
+            networkids=network_id,
+        )
+
+        node = result['virtualmachine']
+
+        return Node(
+            id=node['id'],
+            name=node['displayname'],
+            state=self.NODE_STATE_MAP[node['state']],
+            public_ip=[],
```
Kami: Hm, by default Node has no public IP address assigned? We should probably request a public ip address here as well so it's consistent with other drivers.

Also, if Node has no public IP address, `deploy_node` can't be used.

libcloud/compute/drivers/cloudstack.py
```diff
@@ -0,0 +1,250 @@
...
+
+            nodes.append(node)
+
+        return nodes
+
+    def list_sizes(self, location=None):
+        szs = self._sync_request('listServiceOfferings')
+        sizes = []
+        for sz in szs['serviceoffering']:
+            sizes.append(NodeSize(sz['id'], sz['name'], sz['memory'], 0, 0,
```
Kami, line 215: So Ninefold doesn't exposes prices through the API. Do they have them listed on the website? If they do, you should put them in `data/pricing.json`...

Kami:
Hm, I  have clone your repository and two tests fail:

```
...
ERROR: test_async_request_successful (test.common.test_cloudstack.CloudStackCommonTest)
  File "/Users/kami/Projects/libcloud/libcloud-ninefold/test/common/test_cloudstack.py", line 49, in test_async_request_successful
    success, result = self.connection._async_request('fake')
ValueError: need more than 1 value to unpack

ERROR: test_async_request_unsuccessful (test.common.test_cloudstack.CloudStackCommonTest)
  File "/Users/kami/Projects/libcloud/libcloud-ninefold/test/common/test_cloudstack.py", line 55, in test_async_request_unsuccessful
    success, _ = self.connection._async_request('fake')
  File "/Users/kami/Projects/libcloud/libcloud-ninefold/libcloud/common/cloudstack.py", line 81, in _async_request
    raise Exception(result)
Exception: {u'jobstatus': 2, u'jobresult': {u'fake': u'failresult'}}

Ran 684 tests in 5.109s

FAILED (errors=2)
```

libcloud/loadbalancer/drivers/cloudstack.py
```diff
@@ -0,0 +1,106 @@
...
+        return [self._to_balancer(balancer) for balancer in balancers]
+
+    def get_balancer(self, balancer_id):
+        balancer = self._sync_request('listLoadBalancerRules', id=balancer_id)
+        balancer = balancer.get('loadbalancerrule', [])
+        if not balancer:
+            raise Exception("no such load balancer: " + str(balancer_id))
+        return self._to_balancer(balancer[0])
+
+    def create_balancer(self, name, members, protocol='http', port=80,
```
Kami, line 53: Hm, yeah, location should probably be part of the base API, similar to the base compute API...

Will ask what other ppl think about it.

libcloud/common/cloudstack.py
jeamland: Commit af73bfe addresses this. Thanks!

libcloud/common/cloudstack.py
jeamland: Commit 13b7fbf addresses this.  The job can either be still pending, have succeeded or have failed.  Thanks.

libcloud/compute/drivers/cloudstack.py
jeamland: As discussed, Ninefold would rather the act of assigning a public IP be explicit.

Maybe NodeDriver could grow a feature to indicate whether a public IP is assigned by default and the deploy_node could call a hook to allocate the IP if that feature wasn't present?

libcloud/compute/drivers/cloudstack.py
jeamland, line 215: I can sort that out.  What units is the pricing in though? Ninefold charge in AUD rather than USD...

Kami, line 215: Currently we just convert all the prices to USD which is obviously a bad idea.

I was thinking about adding a new `Price` class which would also have a `currency` attribute.

jeamland, line 215: Ok, it's even more complicated than that actually.  Ninefold's pricing consists of a component for the instance size and then another for the template image, both per hour.  The template image cost varies depending on the hypervisor you're using (they support XenServer as well as VMware) and the OS (Linux flavours or Windows).

I'm really not sure if I can fit that into your current pricing model.

Kami, line 215: Yeah, it's probably not worth the hassle. At least for the first version.

Kami:

@jeamland - Those changes have been merged a long time ago, can you please close this pull request?

Thanks!

PR #19, opened 2011-07-03
Support for EMC Atmos and thus Ninefold Storage

jeamland:
These commits add support for the EMC Atmos storage API and add Ninefold as a storage provider.


libcloud/storage/drivers/atmos.py
```diff
@@ -0,0 +1,435 @@
...
+        if params:
+            if type(params) is dict:
+                params = params.items()
+            pathstring += '?' + urllib.urlencode(params)
+        pathstring = pathstring.lower()
+
+        xhdrs = [(k, v) for k, v in headers.items() if k.startswith('x-emc-')]
+        xhdrs.sort(key=lambda x: x[0])
+
+        signature = [
```
Kami, line 97: It's probably a good idea to refactor signature part into a separate method. It will also make testing easier.

jeamland, line 97: Signature method split out in 6d68c6b, tests for signature method added in 488965b.

libcloud/storage/drivers/atmos.py
```diff
@@ -0,0 +1,438 @@
...
+        headers['x-emc-signature'] = self._calculate_signature(params, headers)
+
+        return params, headers
+
+    def _calculate_signature(self, params, headers):
+        pathstring = self.action
+        if pathstring.startswith(self.driver.path):
+            pathstring = pathstring[len(self.driver.path):]
+        if params:
+            if type(params) is dict:
```
Kami, line 89: Hm, I don't really like this check...

Maybe we should just thrown a `TypeError` if users passes in anything other than dict as the `params` argument.

libcloud/storage/drivers/atmos.py
```diff
@@ -0,0 +1,438 @@
...
+        method = 'PUT'
+
+        extra = extra or {}
+        request_path = container.name + '/' + object_name
+        request_path = self._namespace_path(request_path)
+        content_type = extra.get('content_type', None)
+
+        try:
+            self.connection.request(request_path + '?metadata/system')
+        except Exception, e:
```
Kami: Hm, you do a similar check in multiple places.

You should probably override the `request` method and handle this there so we can only do it once?

Method signature could for example look something like this: `def request(self, path, expected_codes)` ...

libcloud/storage/drivers/atmos.py
```diff
@@ -0,0 +1,438 @@
...
+                                          content_type=content_type,
+                                          upload_func=upload_func,
+                                          upload_func_kwargs=upload_func_kwargs,
+                                          request_path=request_path,
+                                          request_method=method,
+                                          headers={}, file_path=file_path)
+
+        response = result_dict['response'].response
+        bytes_transferred = result_dict['bytes_transferred']
+        range_hdr = str(bytes_transferred) + '-' + str(bytes_transferred)
```
Kami: Hm, this variable is defined but unused?

libcloud/storage/drivers/atmos.py
```diff
@@ -0,0 +1,438 @@
...
+            meta_data = extra.get('meta_data', {})
+        meta_data['md5'] = result_dict['data_hash']
+        user_meta = ', '.join([k + '=' + str(v) for k, v in meta_data.items()])
+        self.connection.request(request_path + '?metadata/user', method='POST',
+                                headers={'x-emc-meta': user_meta})
+        result = self.connection.request(request_path + '?metadata/system')
+        meta = self._emc_meta(result)
+        del meta_data['md5']
+        extra = {
+            'object_id': meta['objectid']
```
Kami: You never add `meta_data` dictionary to the `extra` dictionary?

libcloud/storage/drivers/atmos.py
```diff
@@ -0,0 +1,438 @@
...
+                      extra, meta_data, container, self)
+
+    def upload_object_via_stream(self, iterator, container, object_name,
+                                 extra=None):
+        if isinstance(iterator, file):
+            iterator = iter(iterator)
+
+        data_hash = hashlib.md5()
+
+        def chunkify(source):
```
Kami: There already is a `read_in_chunks` utility function in `libcloud/utils.py`. We should modify it to support this use case instead of doing this here.

libcloud/storage/drivers/atmos.py
```diff
@@ -0,0 +1,438 @@
...
+
+        while True:
+            end = bytes_transferred + len(chunk) - 1
+            data_hash.update(chunk)
+            headers = {
+                'x-emc-meta': 'md5=' + data_hash.hexdigest(),
+            }
+            if len(chunk) > 0:
+                headers['Range'] = 'Bytes=%d-%d' % (bytes_transferred, end)
+            result = self.connection.request(path, method='PUT', data=chunk,
```
Kami, line 268: So Atmos doesn't support chunked encoding?

libcloud/storage/drivers/atmos.py
```diff
@@ -0,0 +1,438 @@
...
+        except StopIteration:
+            chunk = ''
+
+        path = self._namespace_path(container.name + '/' + object_name)
+
+        while True:
+            end = bytes_transferred + len(chunk) - 1
+            data_hash.update(chunk)
+            headers = {
+                'x-emc-meta': 'md5=' + data_hash.hexdigest(),
```
Kami, line 264: Hm, is this correct? Shouldn't you only send md5 hash for the current chunk and not for all the chunks which have been transfered?

libcloud/storage/drivers/atmos.py
```diff
@@ -0,0 +1,438 @@
...
+
+    DEFAULT_CDN_TTL = 60 * 60 * 24 * 7 # 1 week
+
+    def __init__(self, key, secret=None, secure=True, host=None, port=None):
+        host = host or self.host
+        super(AtmosDriver, self).__init__(key, secret, secure, host, port)
+
+    def list_containers(self):
+        result = self.connection.request(self._namespace_path(''))
+        entries = self._list_objects(result.object)
```
Kami: Maybe you should add `type` argument (directory / object / both) to the `_list_objects` method and perform filtering there?

libcloud/storage/drivers/atmos.py
```diff
@@ -0,0 +1,438 @@
...
+        objects = []
+        for entry in entries:
+            if entry['type'] != 'regular':
+                continue
+            metadata = {
+                'object_id': entry['id']
+            }
+            objects.append(Object(entry['name'], 0, '', {}, metadata, container,
+                                  self))
+        return objects
```
Kami: To be consistent with other drivers, this method should return a `LazyList` object. For example you can look at the S3 / CloudFiles driver - https://github.com/apache/libcloud/blob/trunk/libcloud/storage/drivers/s3.py#L178

libcloud/storage/drivers/atmos.py
```diff
@@ -0,0 +1,438 @@
...
+            protocol = 'http'
+
+        expiry = str(expiry or int(time.time()) + self.DEFAULT_CDN_TTL)
+        params = [
+            ('uid', self.key),
+            ('expires', expiry),
+        ]
+
+        key = base64.b64decode(self.secret)
+        signature = '\n'.join(['GET', path.lower(), self.key, expiry])
```
Kami, line 370: We should move signature calculation to a separate method.

Kami:
Any updates? It would be nice to merge it.

libcloud/storage/drivers/atmos.py
jeamland, line 89: urllib allows either a dict or an iterable of 2-tuples.  The check is to force it to the latter so that the order remains the same.  I'm happy to code in an assumption that it's dict but it depends on whether you want to impose that constraint.

libcloud/storage/drivers/atmos.py
jeamland: Done in a550f24.

libcloud/storage/drivers/atmos.py
jeamland: Well the issue is more that when an Exception is raised, I'm looking for specific codes.  In abe318b I've changed it somewhat to make it a bit less brittle but I can't think of an obvious way to make it any better abstracted than that.

libcloud/storage/drivers/atmos.py
jeamland: Yep, gone in 0628edd.

libcloud/storage/drivers/atmos.py
jeamland, line 268: Nope.

libcloud/storage/drivers/atmos.py
jeamland: Done in 242577e.

libcloud/storage/drivers/atmos.py
jeamland: Done in 82d710e.  Also added tests for read_in_chunks.

libcloud/storage/drivers/atmos.py
jeamland, line 264: It's a range request.  I'm basically updating the md5 for what's up there.

libcloud/storage/drivers/atmos.py
jeamland, line 370: Done in 4075f67.

libcloud/storage/drivers/atmos.py
jeamland: Done in 7373668.

jeamland:
Sorry about the delay.  Our second child arrived on Monday and strangely I haven't had much time to hack on things. =)

libcloud/storage/drivers/atmos.py
Kami, line 264: Ah, missed that.

Kami:
No problem and congratulations!

I will try to merge this branch over the weekend :)

libcloud/utils.py
```diff
@@ -41,13 +45,30 @@ def read_in_chunks(iterator, chunk_size=None):
...
+                    empty = True
+            except StopIteration:
+                empty = True
+
+        if len(data) == 0:
             raise StopIteration
 
-        yield chunk
+        if fill_size:
+            if empty or len(data) >= chunk_size:
```
Kami, line 66: Equality comparison here doesn't really make sense because it's the same as the else case.

test/test_utils.py
```diff
@@ -88,5 +88,37 @@ class TestUtils(unittest.TestCase):
...
+                self.remaining -= 1
+                if self.remaining == 0:
+                    return ''
+                return 'b' * size
+
+        for result in libcloud.utils.read_in_chunks(FakeFile(), chunk_size=10,
+                                                    fill_size=False):
+            self.assertEqual(result, 'bbbbbbbbbb')
+
+        for result in libcloud.utils.read_in_chunks(FakeFile(), chunk_size=10,
```
Kami: This doesn't really test that the fill_size works properly.

Something like this should work better:

``` python
    def test_read_in_chunks_filelike(self):
        class FakeFile(file):
            def __init__(self):
                self.remaining = 500

            def read(self, size):
                self.remaining -= 1
                if self.remaining == 0:
                    return ''
                return 'b' * (size + 1)

        for index, result in enumerate(libcloud.utils.read_in_chunks(
                                       FakeFile(), chunk_size=10,
                                       fill_size=False)):
            self.assertEqual(result, 'b' * 11)

        self.assertEqual(index, 498)

        for index, result in enumerate(libcloud.utils.read_in_chunks(
                                       FakeFile(), chunk_size=10,
                                       fill_size=True)):
            if index != 548:
                self.assertEqual(result, 'b' * 10)
            else:
                self.assertEqual(result, 'b' * 9)

        self.assertEqual(index, 548)
```

libcloud/utils.py
jeamland, line 66: The else case is against if fill_size though.  Basically that test doesn't bother sending a chunk unless it's the right size or unless we've run out of new data.

test/test_utils.py
jeamland: Done in 27e2896.

libcloud/utils.py
Kami, line 66: Ah, oops, I missed that.

Kami:
Looks OK. Can you please create a ticket on jira and submit a patch there and I'll try to merge it today?

Thanks!

Kami:

@jeamland - Those changes have been merged a long time ago, can you please close this pull request?

Thanks!

PR #20, opened 2011-09-06
Fix inheritance inversion

mikenerone:
In the current state of the computer drivers, the exists an inverted inheritance in that OpenStack inherits from Rackspace, where Rackspace (and potentially other specific OpenStack providers) should rather inherit from OpenStack. This inversion complicated several things: 
- Support for multiple OpenStack providers. 
- Support for multiiple OpenStack API versions in general. 
- Separately, future migration of the Rackspace driver to a new OpenStack version (e.g. migrating Rackspace to OpenStack 1.1 would migrate all OpenStack subclasses to 1.1).

Upstream issue: https://issues.apache.org/jira/browse/LIBCLOUD-110


mikenerone:
This was accepted and merged into libcloud svn trunk. See upstream JIRA issue.

PR #21, opened 2011-09-08
add base support for cloud.com

gigimon:
Base support for cloud.com


Kami:
Hi, the patch looks OK.

Can you please also add some tests? 

Basically, just subclass the base CloudStack test class (test/compute/test_cloudstack.py) and change the `setUp` method to create an instance of `CloudComNodeDriver` driver.

pquerna:
Can we use a DNS name of some kind?

hard coding an IP address in host = '72.52.126.24' just seems insane?

gigimon:
Hello

I'm approve more functions later and create test.
cloud.com not provide DNS host for api (or I'm bad see)

libcloud/compute/drivers/cloudstack.py
```diff
@@ -173,17 +173,15 @@ class CloudStackNodeDriver(CloudStackDriverMixIn, NodeDriver):
...
         network_id = networks['network'][0]['id']
-
         result = self._async_request('deployVirtualMachine',
             name=name,
             displayname=name,
-            serviceofferingid=size.id,
-            templateid=image.id,
-            zoneid=location.id,
-            networkids=network_id,
+            serviceOfferingId=size.id,
```
Kami: Hm, why did you change those? This will break existing drivers based on the CloudStack platform (unless you can provide a reference that it was wrong before).

gigimon: It's was my trooble, sorry. I'm return this back and use create_node with changes in my Node

PR #22, opened 2011-09-08
Compute demo

mikenerone:
I found myself in need of live testing in working on an OpenStack driver. Though writing a driver-specific test/demo tool was an option, it's no more difficult to make one that works for any driver.

Apache JIRA issue: https://issues.apache.org/jira/browse/LIBCLOUD-111


mikenerone:
Note: at the moment, an extra commit, 18000e7ee92286945612630d099acce2f99dac93, shows up in the above list. That's already in libcloud's trunk, but GitHub hasn't updated.

demos/secrets.py.dist
```diff
@@ -17,5 +17,5 @@
 # Copy this file to secrets.py for use with provided examples
 #
 
-EC2_ACCESS_ID=''
```
Kami, line 20: Those credentials are still used in the tests (test/compute/test_ec2.py) so we can't remove them.

demos/compute_demo.py
```diff
@@ -33,8 +34,44 @@ from libcloud.providers import get_driver
...
+            image=images[0],
+            size=sizes[0],
+        )
+    >>> node.destroy()
+    """
+    provider_name = provider_name.upper()
+
+    DriverClass = get_driver(getattr(Provider, provider_name))
+
+    if not args:
```
Kami, line 71: I would throw an exception in case both `args` and `kwargs` are None.

The exception message should say something like: "Missing credentials / arguments for provider the  X. Please edit secrets.py and put the credentials there."

mikenerone, line 71: That occurred to me, but I wanted to leave the door open for unconventional configurations. E.g. someone might be working with a private installation that needs no credentials, or something where their IP is already blessed, or any other situation we didn't think of wherein no credentials are necessary. In any case, for "normal" cases, both blank _will_ raise an authentication failure, which is pretty accurate.

demos/secrets.py.dist
mikenerone, line 20: Ah, yes, I didn't consider that the secrets.py in test and demo should have the same format, but that does make sense. It's pretty hard to make a solution "general" with the existing arrangement, though, so I would have to say that we either change the format of _both_ of the secrets.py's or give up on the reuseability having a single interface (i.e. a tuple) would give us. I'd suggest the change. I know that makes this a little more invasive, but I'm hoping you agree that something more universal is called for here, and it's a big step toward having an easy live test for any driver during development. It immediately serves as such (though admittedly, it wants for increased coverage for that purpose).

I'd be happy to make the changes with your leave - it would only take a few minutes to touch the necessary files.

demos/compute_demo.py
mikenerone, line 71: Oh, best of both worlds: catch that auth failure if it happens, _then_ point them to secrets.py. Done.

demos/compute_demo.py
```diff
@@ -0,0 +1,114 @@
...
+        print(">> Loading nodes...")
+        pprint(driver.list_nodes())
+
+        print(">> Loading images... (showing up to 10)")
+        pprint(driver.list_images()[:10])
+
+        print(">> Loading sizes... (showing up to 10)")
+        pprint(driver.list_sizes()[:10])
+    except Exception as ex:
+        print("A fatal error occurred: {ex}".format(ex=ex))
```
Kami: I don't think `string.format` is available in Python 2.5 (2.5 is the lowest version we officially support as of 0.4.2).

demos/secrets.py.dist
Kami, line 20: Yeah, using a tuple would make it more flexible.

The problem is that in some cases we also want to test different values for host / port / secure argument in a single test file which means we can't always just unpack the tuple. In cases like that we only care about the key and secret argument which comes from secrets.py.

demos/compute_demo.py
mikenerone: Right...fixed.

demos/secrets.py.dist
mikenerone, line 20: I took a look - the only test I see for which those values come from secrets.py is test.compute.test_openstack. In that one, no such iteration is done, and in fact, the values from secrets.py are immediately packed into a tuple for use as *args in the instantiation. So I think that at the present, this needn't be a concern.

For possible future cases where such iteration might be desired though, I don't think it's a problem:
1. Most commonly, those values wouldn't come after the credential params in the constructor, so you just tack them on.
2. In cases where they don't, and one decides that it's an oddball case where they want the separate-variables behavior similar to what we have now, it takes only one line to get it back. E.g. username, api_key = secrets.DRIVER_PARAMS
3. Another way to solve it is to use the kwargs functionality I put in there - that's effectively the same as separate values.

mikenerone:
Just noticed this still needs to be closed. It was merged a while back.

PR #24, opened 2011-09-16
Nova client with novaclient - Proof of Concept

mikenerone:
This pull request represents a new OpenStack driver based on python-novaclient. There are no tests at this time, so it's not ready to merge. I wanted to spark some discussion on this possibility, which I am very much for. python-novaclient is actually used internally by Rackspace and other OpenStack providers, so this is likely to be extremely well-maintained, plus novaclient already covers a lot of corner cases. To do it from scratch in libcloud results in a great deal of duplication of work and code, and would probably still lag on robustness unless we were to actually tediously crib nearly all of the novaclient logic.

BTW, the first commit collapses the OpenStack 1.0 driver into the Rackspace driver (the only public implementation of OpenStack 1.0), as has been discussed recently on the mailing list and in IRC, resulting in a general consensus that this is a good thing to do. Aside from that, the bit you really want to look at for the novaclient-based driver is _all_ in [libcloud/compute/drivers/openstack.py](https://github.com/racker/libcloud/blob/nova-client-with-novaclient/libcloud/compute/drivers/openstack.py). Seriously, the diff is useless for that file.

This is a proposed solution for https://issues.apache.org/jira/browse/LIBCLOUD-83


pquerna:
libcloud is not a collection of dependencies on other python code.  it is a library interacting with rest APIs.

this has been a core part of its design from the very beginning.

-1 to merging.

Kami:

@Manganeez - Those changes are not relevant anymore, can you please close this pull request?

Thanks!

PR #26, opened 2011-09-20
OpenStack Nova 1.1 driver - Feedback only - not ready for merge

mikenerone:
Still have tests to add and clean up, but wanted to open a pull request in case anyone has feedback on any of the refactoring or the factory use, which, as I said in the commit, "effectively makes the OpenStack driver appear as a single driver that supports all versions, specified during instantiation (except it's much easier to maintain)."

This is for [LIBCLOUD-83](https://issues.apache.org/jira/browse/LIBCLOUD-83).

PR #28, opened 2011-09-26
Openstack 1 1 support

mikenerone:
Another proposal for [LIBCLOUD-83](https://issues.apache.org/jira/browse/LIBCLOUD-83).

If there are any concerns, feedback would be appreciated.


libcloud/compute/drivers/openstack.py
```diff
@@ -676,3 +688,276 @@ class OpenStack_1_0_NodeIpAddresses(object):
     def __init__(self, public_addresses, private_addresses):
         self.public_addresses = public_addresses
         self.private_addresses = private_addresses
+
+
+class OpenStack_1_1_Response(Response):
+
+    def success(self):
```
Kami: This could be re-used from 1.0 class (inherit from 1.0 response class)?

libcloud/compute/drivers/openstack.py
```diff
@@ -676,3 +688,276 @@ class OpenStack_1_0_NodeIpAddresses(object):
...
+            return None
+
+        if self.has_content_type('application/json'):
+            try:
+                return json.loads(self.body)
+            except:
+                raise MalformedResponseError(
+                    'Failed to parse JSON',
+                    body=self.body,
+                    driver=OpenStack_1_0_NodeDriver)
```
Kami: Should be 1_1 :)

libcloud/compute/drivers/openstack.py
```diff
@@ -676,3 +688,276 @@ class OpenStack_1_0_NodeIpAddresses(object):
...
+            return self.body
+
+    def parse_error(self):
+        text = self.body
+
+        if self.has_content_type('application/json'):
+            try:
+                text = ';'.join([fault_data['message'] for fault_data in self.parse_body().values()])
+            except MalformedResponseError:
+                pass
```
Kami: Shouldn't MalformedResponseError be re-thrown here?

libcloud/compute/drivers/openstack.py
```diff
@@ -676,3 +688,276 @@ class OpenStack_1_0_NodeIpAddresses(object):
...
+    def has_content_type(self, content_type):
+        content_type_value = self.headers.get('content-type') or ''
+        content_type_value = content_type_value.lower()
+        return content_type_value.find(content_type.lower()) > -1
+
+    def parse_body(self):
+        if self.status == httplib.NO_CONTENT or not self.body:
+            return None
+
+        if self.has_content_type('application/json'):
```
Kami: Actually, looks like the only real difference between 1.0 and 1.1 class is the default content type (xml vs json) and api version.

Imo we should add class two attributes to the base class and handle different content types there. Then both 1.0 and 1.1 would inherit from the base class and just specify different content_type and api_version.

Same goes for the Connection class.

libcloud/compute/drivers/openstack.py
```diff
@@ -676,3 +688,276 @@ class OpenStack_1_0_NodeIpAddresses(object):
...
+
+    NODE_STATE_MAP = OpenStack_1_0_NodeDriver.NODE_STATE_MAP
+
+    def __init__(self, *args, **kwargs):
+        self._ex_force_base_url = kwargs.pop('ex_force_base_url', None)
+        self._ex_force_auth_url = kwargs.pop('ex_force_auth_url', None)
+        self._ex_force_auth_version = kwargs.pop('ex_force_auth_version', None)
+        super(OpenStack_1_1_NodeDriver, self).__init__(*args, **kwargs)
+
+    def _ex_connection_class_kwargs(self):
```
Kami: This could be re-used from the 1.0 class.

Kami:
It looks like a lot of functionality between 1.0 and 1.1 could be re-used.

I would vote for creating a base `OpenStackNodeDriver` class where we would put common functionality. Then both 1.0 and 1.1 class drivers should inherit from it and implement just the differences.

test/compute/fixtures/openstack_v1.1/README
```diff
@@ -0,0 +1,7 @@
```
pquerna, line 0: This is actually not how most of our other fixtures are created.  Fixtures should be created from live captures whenever possible, not documentation. Documentation has been wrong _many_ times.

libcloud/compute/drivers/openstack.py
mikenerone: One parses XML and the other JSON. They could be combined since they look at the content type, but I thought it might be better to make sure the XML-based driver only has XML processing associated, and ditto for JSON. Plus, I've had difficulty getting refactorings of existing things accepted, so I was trying to touch existing stuff as little as possible. Do you want me to go ahead and combine them?

libcloud/compute/drivers/openstack.py
mikenerone: Whoops! 64cd5ab409ff6940b519fd1a09ef94a1ea0ac47c

libcloud/compute/drivers/openstack.py
Kami: I would say +1, do it, but lets wait for other ppls opinion.

libcloud/compute/drivers/openstack.py
mikenerone: Indeed. 653b6c40045c96c893d62245256e4a875c03b50c

libcloud/compute/drivers/openstack.py
mikenerone: I did something similar in my [previous submission](https://github.com/apache/libcloud/pull/26), except I only handled JSON in the base class because based on conversations, it seemed like future version were likely to stick to JSON, so might as well keep the base class cleaner. I just overrode with XML stuff in the 1.0 driver instead. That pull request was rejected for too much refactoring, however; so again, I was trying to touch as little existing code as possible.

Please confirm that you specifically want me to combine these things. Also, do you mean in NodeDriver (which both of these drivers currently inherit from), or to make a new base class for OpenStack node drivers?

test/compute/fixtures/openstack_v1.1/README
mikenerone, line 0: Cool, I'll remember that for the future. As it happens, the docs and reality match with just one exception: in server responses, the API currently returns tenant_id and user_id, where the docs say tenantId and userId. I believe we should continue to use the documented versions, because when I asked OpenStack peeps about this, I learned that the reality in alpha is incorrect - the docs are correct, and reality will be adjusted to match theory. :)

As far as how this difference affects current code: For tenantId, the code handles whichever is present. For userId, we don't actually use that, so no effect.

mikenerone:
Kami: what you suggest about having an `OpenStackNodeDriver` is exactly what I did in my [previous pull-request](https://github.com/apache/libcloud/pull/26); but that was rejected.

libcloud/compute/drivers/openstack.py
mikenerone: And less sillily (is that a word?) in 49659dc7f5b1ae64d9928a9cf0547661896f7b43.

Kami: heh, yeah, better :)

libcloud/compute/drivers/openstack.py
Kami: 1. +1 for only handling JSON in a base class and overring with XML in the 1.0 driver if this means you won't need to duplicate most of the logic from the base one.
2. I mean a new base NodeDriver class for OpenStack - e.g. BaseOpenStackNodeDriver

mikenerone: Yeah, I realized that when I read comments further down.

libcloud/compute/drivers/openstack.py
```diff
@@ -554,11 +589,14 @@ class OpenStack_1_0_NodeDriver(NodeDriver):
                      driver=self.connection.driver)
         return s
 
-    def _to_images(self, object):
+    def _to_images(self, object, ex_filter_active):
```
Kami: `ex_only_active` might be a better name for this arg?

libcloud/compute/drivers/openstack.py
```diff
@@ -107,17 +133,86 @@ class OpenStack_1_0_Connection(OpenStackBaseConnection):
...
+                                              .object, ex_filter_active)
+
+    def ex_get_node_details(self, node_id):
+        # @TODO: Remove this if in 0.6
+        if isinstance(node_id, Node):
+            node_id = node_id.id
+
+        uri = '/servers/%s' % (node_id)
+        resp = self.connection.request(uri, method='GET')
+        if resp.status == 404:
```
Kami: `httplib.NOT_FOUND` (consistency)

Kami:
Hm, I noticed another thing. How do you currently instantiate 1.1 driver via `get_provider`? Looks like we only have a provider constant for version 1.0.

I like the idea me and @Manganeez originally talked about. Each provider driver which has support for multiple API versions should only have a single provider constant and a base class which constructor takes `version` argument.

Based on this version argument, base class would then instantiate a correct version-specific class - constructor method would basically be a factory method. As far as I know @Manganeez has already implemented something like this in his original pull request.

Kami:

@Manganeez - Those changes have been merged a long time ago, can you please close this pull request?

Thanks!

PR #29, opened 2011-09-28
Improvement of gandi.net compute driver

gdchamal:
Here a pull request for improvement of gandi.net compute driver in libcloud. Methods added are specifics ones to manipulate network interfaces and also disks attached to nodes. All methods are covered by a unittest.

Base objects for connections, errors are now in common/gandi.py.

### DIFF
--- a/test/secrets.py-dist
+++ b/test/secrets.py-dist
@@ -36,4 +36,4 @@ VOXEL_PARAMS = ('key', 'secret')
 VPSNET_PARAMS = ('user', 'key')
 
 # DNS
-# DNS_PARAMS_LINODE = ('user', 'key')
+DNS_PARAMS_LINODE = ('user', 'key')


libcloud/common/gandi.py
```diff
@@ -0,0 +1,234 @@
...
+    """
+    Gandi base driver
+
+    """
+    connectionCls = GandiConnection
+    name = 'Gandi'
+    api_name = 'gandi'
+    friendly_name = 'Gandi.net'
+    country = 'FR'
+    type = Provider.GANDI
```
Kami: Provider constants are not the same among different APIs (compute / storage / dns / ...).

libcloud/common/gandi.py
```diff
@@ -0,0 +1,234 @@
...
+
+
+class BaseGandiDriver(object):
+    """
+    Gandi base driver
+
+    """
+    connectionCls = GandiConnection
+    name = 'Gandi'
+    api_name = 'gandi'
```
Kami: api_name is compute specific

libcloud/common/gandi.py
```diff
@@ -0,0 +1,234 @@
...
+
+class IPAddress(BaseObject):
+    """
+    Provide a common interface for ip addresses
+    """
+
+    uuid_prefix = 'inet:'
+
+    def __init__(self, id, state, inet, driver, version=4, extra=None):
+        BaseObject.__init__(self, id, state, driver)
```
Kami: Since `BaseObject` is a new style class, it would be correct to use `super` here. Same goes for other objects which inherit from `BaseObject`.

libcloud/compute/drivers/gandi.py
```diff
@@ -365,3 +262,131 @@ class GandiNodeDriver(NodeDriver):
     def list_locations(self):
         res = self.connection.request("datacenter.list")
         return [self._to_loc(l) for l in res]
+
+    def _to_iface(self, iface):
+        ips = []
+        for ip in iface.get('ips'):
```
Kami: Should be `iface.get('ips', [])`, otherwise it will return `None` if the ips key is not present.

libcloud/common/gandi.py
```diff
@@ -0,0 +1,234 @@
...
+
+        # Connect only with an api_key generated on website
+        self.api_key = user
+
+        try:
+            self._proxy = self.proxyCls(self._user_agent())
+        except xmlrpclib.Fault, e:
+            raise GandiException(1000, e)
+
+    def _user_agent(self):
```
Kami: Hm, is there a special reason why you put this method here? It looks the same as the one in the base `Connection` class (libcloud.common.base) which you should be able to re-use.

Same goes for the `user_agent_append` method bellow.

PR #31, opened 2011-10-19
Add extra info to images for EC2 driver

juancarlosm:


libcloud/compute/base.py
```diff
@@ -266,7 +266,7 @@ class NodeImage(object):
 
     def __repr__(self):
         return (('<NodeImage: id=%s, name=%s, driver=%s  ...>')
-                % (self.id, self.name, self.driver.name))
+                % (self.id, self.name, self.driver.name, self.extra))
```
Kami, line 269: This will throw an exception since there are only three placeholders in the string.

Kami:
I have merged your patch (with some changes) into trunk in [1187584](http://svn.apache.org/viewvc?rev=1187584&view=rev).

Next time please make sure `pep8` tool returns no warnings for modified files.

Also, can you please close this pull request now (we can't do it)?

juancarlosm:
ok, thank you. I Hope to contribute more in the future, and I yes, I will check warnings :)

PR #32, opened 2011-10-21
Response hierarchy - Closes LIBCLOUD-97

caio:
Hello,

This patchset implements two base response classes XmlResponse and JsonResponse in order to reduce code duplication and adapts the common and specialized drivers to use the new classes.

Tests passing for each commit separately.

### DIFF
--- a/libcloud/compute/drivers/bluebox.py
+++ b/libcloud/compute/drivers/bluebox.py
@@ -80,13 +80,6 @@ NODE_STATE_MAP = { 'queued': NodeState.PENDING,
                    'unknown': NodeState.UNKNOWN }
 
 class BlueboxResponse(JsonResponse):
-    def parse_body(self):
-        try:
-            js = super(BlueboxResponse, self).parse_body()
-            return js
-        except MalformedResponseError:
-            return self.body
-
     def parse_error(self):
         if int(self.status) == 401:
             if not self.body:
--- a/libcloud/compute/drivers/bluebox.py
+++ b/libcloud/compute/drivers/bluebox.py
@@ -26,14 +26,10 @@ import copy
 import urllib
 import base64
 
-try:
-    import simplejson as json
-except ImportError:
-    import json
-
-from libcloud.common.base import Response, ConnectionUserAndKey
+from libcloud.common.base import JsonResponse, ConnectionUserAndKey
 from libcloud.compute.providers import Provider
 from libcloud.compute.types import NodeState, InvalidCredsError
+from libcloud.common.types import MalformedResponseError
 from libcloud.compute.base import Node, NodeDriver
 from libcloud.compute.base import NodeSize, NodeImage, NodeLocation
 from libcloud.compute.base import NodeAuthPassword, NodeAuthSSHKey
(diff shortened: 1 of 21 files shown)


libcloud/compute/drivers/bluebox.py
```diff
@@ -83,12 +79,12 @@ NODE_STATE_MAP = { 'queued': NodeState.PENDING,
                    'error': NodeState.TERMINATED,
                    'unknown': NodeState.UNKNOWN }
 
-class BlueboxResponse(Response):
+class BlueboxResponse(JsonResponse):
     def parse_body(self):
```
Kami: Looks like it's safe to remove `parse_body` here.

`parse_error` doesn't call `parse_body` here and `parse_body` always gets called after `parse_error` if `success()` returns `True`. Also looking at the code all API calls seem to return JSON.

libcloud/compute/drivers/brightbox.py
```diff
@@ -31,18 +31,18 @@ except ImportError:
...
 
 
-class BrightboxResponse(Response):
+class BrightboxResponse(JsonResponse):
     def success(self):
         return self.status >= 200 and self.status < 400
 
     def parse_body(self):
         if self.headers['content-type'].split('; ')[0] == 'application/json' and len(self.body) > 0:
-            return json.loads(self.body)
```
Kami: Hm, thinking if we should add `parse_zero_length_body` attribute to the base XMLResponse & JSONResponse class.

Then we could so something like this there:

``` python
if len(self.body) == 0 and not self.parse_zero_length_body:
  return self body
...
```

libcloud/compute/drivers/elasticstack.py
```diff
@@ -121,15 +121,7 @@ class ElasticStackResponse(Response):
     def parse_body(self):
         if not self.body:
```
Kami: Yeah, if we add `parse_zero_length_body` attribute we could also get rid of overriding `parse_body` here...

libcloud/compute/drivers/slicehost.py
```diff
@@ -35,26 +34,13 @@ class SlicehostResponse(Response):
         # application/xml" header. booya.
         if not self.body or len(self.body) <= 1:
```
Kami: Same here with regards to the `parse_zero_length_body`attribute.

caio: The `len(self.body) == 0` would fail here, should we change the `parse_zero_length_body` logic to test the stripped body? I.e.: 

``` python
if len(self.body.strip()) == 0 and not self.parse_zero_length_body:
    return self body
```

Or maybe we could strip it directly on `Response.__init__`

libcloud/compute/drivers/brightbox.py
caio: In this case in particular (brightbox driver) this wouldn't help much since there's still the need to check for the content-type, right?

This makes me wonder if we shouldn't move this content-type verification into the {Json,Xml}Response classes and create a behavior toggle just like this `parse_zero_length_body` to let the specialized response class decide what to do if the header is not as expected (`ignore_content_type` to allow parsing regardless of the content-type sent back).

This is probably not the best place for brainstorming, though :)

libcloud/compute/drivers/slicehost.py
Kami: Yeah, I would be OK with something like that in the `Response` class.

caio: Done that in the two new commits. More lines removed :)

caio:
Pull request merged as single blob: https://github.com/apache/libcloud/commit/211a6b5219a94f0d11e42e5198f7729cc361cb2f
Thanks for the reviews!

PR #33, opened 2011-10-28
Set metadata - OpenStack driver

mynnx:
From http://docs.openstack.org/cactus/openstack-compute/developer/openstack-compute-api-1.1/content/ServerUpdate.html:

This operation updates the editable attributes of a server: the name of the server, the server metadata, and the IPv4 and IPv6 access address. Note that while the server name is editable, the operation does not change the server host name. Note also that server names are not guaranteed to be unique.

### DIFF
--- a/libcloud/compute/drivers/openstack.py
+++ b/libcloud/compute/drivers/openstack.py
@@ -912,12 +912,11 @@ class OpenStack_1_1_NodeDriver(OpenStackNodeDriver):
         """
         return self._update_node(node, name=name)
 
-    def ex_get_metadata(self, node, metadata):
+    def ex_get_metadata(self, node):
         """
         Get a Node's metadata.
 
-        @return     Key/Value metadata to associate with a node
-        @type       C{dict}
+        @return C{dict} : Key/Value metadata to associate with a node
         """
         return self.connection.request(
                 '/servers/%s/metadata' % (node.id,), method='GET',
--- a/libcloud/compute/drivers/openstack.py
+++ b/libcloud/compute/drivers/openstack.py
@@ -916,11 +916,22 @@ class OpenStack_1_1_NodeDriver(OpenStackNodeDriver):
         """
         Sets the Node's metadata.
 
-        @param metadata server metadata, a dict with string keys and values
+        @keyword    metadata: Key/Value metadata to associate with a node
+        @type       metadata: C{dict}
         """
         return self._update_node(node, metadata=metadata)
 
     def ex_update_node(self, node, **node_updates):
+        """
+        Update the Node's editable attributes.  Currently this driver
+        supports editing name and metadata.
+
+        @keyword    ex_metadata: Key/Value metadata to associate with a node
+        @type       ex_metadata: C{dict}
+
+        @keyword    name:   New name for the server
+        @type       name:   C{str}
+        """
         potential_data = self._create_args_to_params(node, node_updates)
         updates = {'name': potential_data['name'],
                    'metadata': potential_data['metadata']}
--- a/libcloud/compute/drivers/openstack.py
+++ b/libcloud/compute/drivers/openstack.py
@@ -912,6 +912,17 @@ class OpenStack_1_1_NodeDriver(OpenStackNodeDriver):
         """
         return self._update_node(node, name=name)
 
+    def ex_get_metadata(self, node, metadata):
+        """
+        Get a Node's metadata.
+
+        @return     Key/Value metadata to associate with a node
+        @type       C{dict}
+        """
+        return self.connection.request(
+                '/servers/%s/metadata' % (node.id,), method='GET',
+            ).object['metadata']
+
     def ex_set_metadata(self, node, metadata):
         """
         Sets the Node's metadata.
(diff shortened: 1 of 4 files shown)


test/compute/test_openstack.py
```diff
@@ -597,6 +597,18 @@ class OpenStack_1_1_Tests(unittest.TestCase, TestCaseMixin):
...
 
+    def test_ex_set_server_name(self):
+        try:
+            self.driver.ex_set_server_name(self.node, 'new_name')
+        except Exception, e:
+            self.fail('An error was raised: ' + repr(e))
+
+    def test_ex_set_metadata(self):
+        try:
+            self.driver.ex_set_metadata(self.node, {'a-key': 'a-value'})
```
Kami: This doesn't really test much. Ideally you would also mock the response and they verify the `metadata` attribute on the returned Node object.

libcloud/compute/drivers/openstack.py
```diff
@@ -910,18 +910,44 @@ class OpenStack_1_1_NodeDriver(OpenStackNodeDriver):
...
         Sets the Node's name.
         """
-        return self.ex_update_node(node, name=name)
+        return self._update_node(node, name=name)
+
+    def ex_get_metadata(self, node, metadata):
+        """
+        Get a Node's metadata.
+
+        @return     Key/Value metadata to associate with a node
```
mynnx: How do we document return types?

Kami: `@return C{dict} : Key/Value metadata to associate with a node` should work.

test/compute/test_openstack.py
```diff
@@ -597,6 +597,23 @@ class OpenStack_1_1_Tests(unittest.TestCase, TestCaseMixin):
...
+        new_node = self.driver.ex_set_server_name(old_node, 'Bob')
+        self.assertEqual('Bob', new_node.name)
+
+    def test_ex_set_metadata(self):
+        old_node = Node(
+            id='12063', name=None, state=None,
+            public_ip=None, private_ip=None, driver=self.driver,
+        )
+        metadata = {'doo': 'wop'}
+        new_node = self.driver.ex_set_metadata(old_node, metadata)
```
Kami: ex_set_metadata returns the metadata dictionary not the Node object.

test/compute/test_openstack.py
```diff
@@ -597,6 +597,23 @@ class OpenStack_1_1_Tests(unittest.TestCase, TestCaseMixin):
...
+        )
+        new_node = self.driver.ex_set_server_name(old_node, 'Bob')
+        self.assertEqual('Bob', new_node.name)
+
+    def test_ex_set_metadata(self):
+        old_node = Node(
+            id='12063', name=None, state=None,
+            public_ip=None, private_ip=None, driver=self.driver,
+        )
+        metadata = {'doo': 'wop'}
```
Kami: You also need to add a method to the `OpenStack_1_1_MockHttp` class which loads the appropriate response fixture.

For example:

``` python
def _servers_12063_metadata(self, method, url, body, headers):
        if method == "PUT":
            body = self.fixtures.load('_servers_12063_metadata_two_keys.json')
            return (httplib.OK, body, self.json_content_headers, httplib.responses[httplib.OK])
```

`metadata` dictionary must also match data located in the fixtures. In this case:

``` python
metadata = {'Image Version': '2.1', 'Server Label': 'Web Head 1'}
```

test/compute/test_openstack.py
```diff
@@ -597,6 +597,23 @@ class OpenStack_1_1_Tests(unittest.TestCase, TestCaseMixin):
         else:
             self.fail('An expected error was not raised')
 
+    def test_ex_set_server_name(self):
+        old_node = Node(
+            id='12062', name=None, state=None,
+            public_ip=None, private_ip=None, driver=self.driver,
+        )
+        new_node = self.driver.ex_set_server_name(old_node, 'Bob')
```
Kami: `id` here should be `12064`. "bob" fixture is loaded for this id:

``` python

def _servers_12064(self, method, url, body, headers):
        if method == "GET":
            body = self.fixtures.load('_servers_12064.json')
            return (httplib.OK, body, self.json_content_headers, httplib.responses[httplib.OK])
        elif method == "PUT":
            **body = self.fixtures.load('_servers_12064_updated_name_bob.json')**
            return (httplib.OK, body, self.json_content_headers, httplib.responses[httplib.OK])
        elif method == "DELETE":
            return (httplib.ACCEPTED, "", {}, httplib.responses[httplib.ACCEPTED])
        else:
            raise NotImplementedError()
```

libcloud/compute/drivers/openstack.py
```diff
@@ -910,18 +910,44 @@ class OpenStack_1_1_NodeDriver(OpenStackNodeDriver):
         """
         Sets the Node's name.
         """
-        return self.ex_update_node(node, name=name)
+        return self._update_node(node, name=name)
+
+    def ex_get_metadata(self, node, metadata):
```
Kami: `metadata` argument shouldn't be here.

Kami:
I've committed your patch with a bunch of fixes and a test for `ex_get_metadata` in `r1196577` (http://svn.apache.org/viewvc?view=revision&revision=1196577).

mynnx:
Thanks, Kami.  I made most of these changes last night, but for some reason I only pushed the fixtures.  Sucks, too, because I was proud I'd figured it out and written good tests  =P

mynnx:
also, this is handy for verifying that it works: https://gist.github.com/6a3f611a80b84b3b1b77

Kami:
No problem. And yeah I know it takes some time to figure out how the whole tests fixture stuff works :)

My goal before 1.0 release is to have at least some basics docs up which explain how our tests work and how to use the fixtures.

Kami:
Can you please close this pull request (I can't do it)?

Thanks.

PR #35, opened 2011-11-04
Patched the ec2 driver to include instance start-stop functionality

vishy1618:
- Added ex_start_node method to the ec2 compute driver
- Added ex_stop_node method to the ec2 compute driver
- Added a new _get_state_boolean function to get the state of the node while starting or stopping

### DIFF
--- a/libcloud/dns/drivers/linode.py
+++ b/libcloud/dns/drivers/linode.py
@@ -35,6 +35,8 @@ VALID_RECORD_EXTRA_PARAMS = ['Priority', 'Weight', 'Port', 'Protocol',
                              'TTL_sec']
 
 RECORD_TYPE_MAP = {
+    RecordType.NS: 'NS',
+    RecordType.MX: 'MX',
     RecordType.A: 'A',
     RecordType.AAAA: 'AAAA',
     RecordType.CNAME: 'CNAME',
--- a/test/dns/test_linode.py
+++ b/test/dns/test_linode.py
@@ -41,7 +41,7 @@ class LinodeTests(unittest.TestCase):
 
     def test_list_record_types(self):
         record_types = self.driver.list_record_types()
-        self.assertEqual(len(record_types), 5)
+        self.assertEqual(len(record_types), 7)
         self.assertTrue(RecordType.A in record_types)
 
     def test_list_zones_success(self):


libcloud/compute/drivers/ec2.py
```diff
@@ -295,6 +295,19 @@ class EC2NodeDriver(NodeDriver):
         tag = "{%s}%s" % (NAMESPACE, 'return')
         return element.findtext(tag) == 'true'
 
+    def _get_state_boolean(self, element):
+        iterobj = element.iter()
+        while 1:
```
Kami: Lets not do this and use xpath.

Since we will always operate just on a single instance something like this should work:

``` python
state = findall(element=element, xpath='instancesSet/item/currentState/name', namespace=NAMESPACE)[0].text
return state in ('pending', 'stopping', 'starting')
```

Kami:
Lets also add some tests for the new functionality. 

For the examples you can see `test/compute/test_ec2.py`. You will also need two test fixtures (one for start and one for stop instance response).

vishy1618:
OK, will do it accordingly. Thanks for the inputs!

2011/11/4 Tomaž Muraus <
reply@reply.github.com>

> Lets also add some tests for the new functionality.
> 
> For the examples you can see `test/compute/test_ec2.py`. You will also
> need two test fixtures (one for start and one for stop instance response).
> 
> ---
> 
> Reply to this email directly or view it on GitHub:
> https://github.com/apache/libcloud/pull/35#issuecomment-2630471

## 

Suvish V.T.

vishy1618:
Hi, will the patches be accepted for the next release?

Kami:
I will take a look at the updated patch later on today.

test/compute/test_ec2.py
```diff
@@ -125,6 +125,16 @@ class EC2Tests(LibcloudTestCase, TestCaseMixin):
         ret = self.driver.reboot_node(node)
         self.assertTrue(ret)
 
+    def test_ex_start_node(self):
```
Kami, line 128: You need to add and load fixtures otherwise this tests will fail.

libcloud/compute/drivers/ec2.py
```diff
@@ -295,6 +295,16 @@ class EC2NodeDriver(NodeDriver):
         tag = "{%s}%s" % (NAMESPACE, 'return')
         return element.findtext(tag) == 'true'
 
+    def _get_state_boolean(self, element):
+        """
+        Checks for the instances's state
+        """
+        elem_xpath_list = ['instancesSet/', 'item/', 'currentState/', 'name/']
```
Kami: Lets use `libcloud.utils.findall` (`state = findall(element=element, xpath='instancesSet/item/currentState/name', namespace=NAMESPACE)[0].text`), it knows how to handle namespaces.

vishy1618:
I hope I have written the tests appropriately now. I ran the tests, and they work. I have also added some more functionality to the driver (elastic address association, describe all the addresses, list only specific nodes).

Kami:
I have merged your changes without the `ex_list_nodes` for now (r120273 - http://svn.apache.org/viewvc?view=revision&revision=1202734).

Currently `ex_list_nodes` contains too much copy and pasted code from `list_nodes` and I'm not even sure if this is the best approach. For this function it might even make more sense to take a list of node ids as an argument instead of a list of Node objects.

Some thing to keep in mind for future contributions:
- work on your changes in a separate branch
- always make sure your branch with changes is synced with the latest trunk
- don't include trailing whitespace
- run pep8 on the changed file (line-length, line breaks)

Thanks.

vishy1618:
Thanks a lot, Tomaz. I'll definitely take note of your suggestions.

 Regarding the `ex_list_nodes`, I had previously added a list of node_ids as the parameter, but reverted to using a list of nodes as I saw that was the pattern used elsewhere in the code. How about adding a parameter of `node_ids=None` to the `list_nodes` method itself?

Kami:
Yeah, I would be ok with adding `ex_node_ids` argument to the `list_nodes` method.

It also needs to be clearly documented what it does.

PR #37, opened 2011-11-09
Add optional "list_images" filters for EC2 driver

juancarlosm:


libcloud/compute/drivers/ec2.py
```diff
@@ -464,11 +464,26 @@ class EC2NodeDriver(NodeDriver):
             sizes.append(NodeSize(driver=self, **attributes))
         return sizes
 
-    def list_images(self, location=None):
+    def list_images(self, location=None, ex_image_filter=False):
```
Kami, line 467: Also need to add a docstring + tests :)

juancarlosm, line 467: yes, you are right!. Ok, I'will add it soon :)

Ei Kami, a question about providers, why there is not a function to retrieve volume information? ec2 have is own function, and I supose openstack/rackspace too. It's not a generic function for providers? or has been not developed yet?

Kami, line 467: With volume I assume you mean EBS in the EC2 case?

There was already some debate about this in the past and I even think there is like a year+ old ticket on our JIRA somewhere. Creating a standard API for managing volumes / NAS would require more research and I'm not even convinced it belongs in a base API, because many providers don't support that.

Kami:

@JuanCarlosM - Those changes are not relevant anymore, can you please close this pull request?

Thanks!

PR #38, opened 2011-11-22
Linode API Driver Fix: NS and MX were missing from RECORD_TYPE_MAP

brhelwig:


Kami:
Added in r1205005 - http://svn.apache.org/viewvc?view=revision&revision=1205005.

Thanks

PR #40, opened 2011-11-30
Fix OpenStack Alpha No Longer Returning Node Data on Server Create

tildedave:
WHY THIS HAD TO BE DONE

OpenStack Alpha at http://alpha.auth.api.rackspacecloud.com:5000/v2.0/ has started conforming to the spec and not returning any node data back except for new password on node create.

See: http://docs.openstack.org/cactus/openstack-compute/developer/openstack-compute-api-1.1/content/CreateServers.html

"Note that when creating a server only the server ID, its links, and the admin password are guaranteed to be returned in the request. Additional attributes may be retrieved by performing subsequent GETs on the server."

Specifically the OpenStack 1.1 node driver was failing on create_node as it was attempting to fill the information to create a libcloud node from the server response (which is not just returning ID and admin password).

WHAT

The implementation I took here was to do the follow-up GET to get the node information.  There is enough information returned there to make a libcloud node object, though some of it won't be filled in until the server has gone from BUILD to ACTIVE (for example, ip addresses).

I changed some of the create_node test data to match the existing asserts.

### DIFF
--- a/libcloud/compute/drivers/openstack.py
+++ b/libcloud/compute/drivers/openstack.py
@@ -805,7 +805,12 @@ class OpenStack_1_1_NodeDriver(OpenStackNodeDriver):
                                        method='POST',
                                        data={'server': server_params})
 
-        return self._to_node(resp.object['server'])
+        create_response = resp.object['server']
+        server_resp = self.connection.request('/servers/%s' % create_response['id'])
+        server_object = server_resp.object['server']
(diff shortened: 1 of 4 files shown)


Kami:
Yeah, that's an acceptable solution. We do a similar thing in some other drivers.

Merged in r1208595 - http://svn.apache.org/viewvc?view=revision&revision=1208595.

tildedave:
Thanks Kami!

PR #41, opened 2011-12-08
Update Cloud Files and Load Balancer Rackspace Connection Classes to pass on kwargs

tildedave:
WHY

Rackspace is currently rolling out Auth 2.0.  The existing libcloud connection classes for Load Balancers and Cloud Files do not pass on kwargs to their superclass, essentially forcing these drivers to always use Auth 1.0/Auth 1.1 (whatever their particular defaults are).

The OpenStack Nova Driver does not define a constructor so we get this behavior for free there.

WHAT

Just pass on any kwargs that are passed into the superclass.

Tests pass.  Let me know if you'd like me to write some tests around this specific change.  When Load Balancers are in the Auth 2.0 service catalog we will aim at a more complete patch.  (Would we prefer extra connection classes for CloudFiles_Auth_2_0 in this situation?)

### DIFF
--- a/libcloud/loadbalancer/drivers/rackspace.py
+++ b/libcloud/loadbalancer/drivers/rackspace.py
@@ -45,8 +45,9 @@ class RackspaceConnection(OpenStackBaseConnection):
     auth_url = AUTH_URL_US
     _url_key = "lb_url"
 
-    def __init__(self, user_id, key, secure=True):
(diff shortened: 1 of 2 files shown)


Kami:
Merged in 1213945.

tildedave:
thanks!

PR #42, opened 2011-12-09
RackspaceLBDriver: vips and sourceaddresses

tildedave:
WHAT

Currently Rackspace has the option of returning multiple virtual IP addresses as part of the return load balancer details.

This adds extra["privateVips"] and extra["publicVips"] containing all the SERVICENET/PUBLIC virtual IPs (respectively).

This also adds extra["ipv6PublicSource"], extra["ipv4PublicSource"], extra["ipv4PrivateSource"].

WHY

Information not otherwise accessible.

I used 'extra' to mirror what was done for the Node object.

Tests all pass + added tests for all new functionality.


tildedave:
merged into trunk with modifications, thanks!

PR #43, opened 2011-12-09
Expand RackspaceLBDriver attributes, add error page/access list getters

tildedave:
This patch adds a number of new attributes to the Rackspace LB Driver's "extra" dict, and two new extended actions for getting the load balancer access list and error page.
- Weighted algorithms
- New LB states (error and deleted) + translation into Libcloud "Pending" for PENDING_UPDATE/PENDING_DELETE Rackspace states
- Health Monitors (Connect, HTTP, HTTPS)
- Connection Throttle
- Session Persistence Type
- Connection Logging
- Error page
- Access list


libcloud/loadbalancer/drivers/rackspace.py
```diff
@@ -42,6 +42,49 @@ def success(self):
         return 200 <= int(self.status) <= 299
 
 
+class RackspaceHealthMonitor(object):
+
+    def __init__(self, type, delay, timeout, attempts_before_deactivation):
```
Kami, line 68: Some docstrings documenting all the attributes would be nice.

libcloud/loadbalancer/drivers/rackspace.py
```diff
@@ -167,6 +221,18 @@ def balancer_list_members(self, balancer):
         return self._to_members(
                 self.connection.request(uri).object)
 
+    def ex_balancer_error_page(self, balancer):
```
Kami: Maybe this should be called `ex_get_balancer_error_page`?

libcloud/loadbalancer/base.py
```diff
@@ -40,6 +40,8 @@ class Algorithm(object):
     RANDOM = 0
     ROUND_ROBIN = 1
     LEAST_CONNECTIONS = 2
+    WEIGHTED_ROUND_ROBIN = 3
```
Kami, line 43: Previously all the providers supported all the algorithms listed here, but now that's not the case anymore.

Imo we should add `list_algorithms` / `list_supported_algorithms` method to the `libcloud.loadbalancer.base.Driver` class which should return a list of supported algorithms by the provider driver (`list(self._ALGORITHM_TO_VALUE_MAP.keys())`).

tildedave:
Thanks for the feedback, I'll update this with the suggested changes.

libcloud/loadbalancer/base.py
tildedave, line 43: Added list_supported_algorithms to Base LB driver and ex_list_algorithms to the Rackspace LB Driver (for getting algorithms from the API endpoint).

I noticed during this Currently Gogrid and Cloudstack do not support RANDOM algorithm types, making this idea even more timely!

libcloud/loadbalancer/drivers/rackspace.py
tildedave: fixed

libcloud/loadbalancer/drivers/rackspace.py
tildedave, line 68: added docstrings to RackspaceHealthMonitor, RackspaceHTTPHealthMonitor, RackspaceConnectionThrottle, and RackspaceAccessRule

tildedave:
This should be ready for another look.

libcloud/loadbalancer/drivers/rackspace.py
```diff
@@ -167,6 +270,26 @@ def balancer_list_members(self, balancer):
         return self._to_members(
                 self.connection.request(uri).object)
 
+    def ex_list_algorithms(self):
```
Kami: We already have `list_supported_algorithm` now, do we really need this function?

Can we we combine those two methods?

libcloud/loadbalancer/drivers/rackspace.py
```diff
@@ -42,6 +42,98 @@ def success(self):
         return 200 <= int(self.status) <= 299
 
 
+class RackspaceHealthMonitor(object):
+    """
+    type: type of load balancer.  currently CONNECT (connection monitoring),
```
Kami: Looks OK, but can you please use epytext syntax - e.g. https://github.com/apache/libcloud/blob/trunk/libcloud/compute/drivers/opennebula.py#L246

tildedave: Updated

libcloud/loadbalancer/drivers/rackspace.py
tildedave: changed to ex_list_algorithm_names based on our discussion in #libcloud

tildedave:
Thanks for the feedback so far.  This should address the remaining comments.

Kami:
Merged in r1213364 (https://svn.apache.org/viewvc?view=revision&revision=1213364).

Thanks!

PR #46, opened 2011-12-19
Rackspace Load Balancer 'members' - weight and condition

mynnx:
This patch adds support for "weight" and "condition" attributes of Rackspace load balancer members, as [documented](http://docs-beta.rackspace.com/loadbalancers/api/v1.0/clb-devguide/content/List_Nodes-d1e2218.html)

In addition, calls to the Rackspace `driver.get_balancer()` will now return a LoadBalancer with the list of members in `extra['members']`.

### DIFF
--- a/libcloud/loadbalancer/base.py
+++ b/libcloud/loadbalancer/base.py
@@ -26,10 +26,11 @@ __all__ = [
 
 class Member(object):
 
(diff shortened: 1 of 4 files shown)


Kami:
Merged in r1221010.

Thanks.

PR #47, opened 2011-12-20
Rackspace Load Balancer - simple edits

mynnx:
Enable PUTs on `/loadbalancer/<id>`, as documented here:

http://docs-beta.rackspace.com/loadbalancers/api/v1.0/clb-devguide/content/List_Load_Balancer_Details-d1e1522.html

### DIFF
--- a/libcloud/common/base.py
+++ b/libcloud/common/base.py
@@ -618,14 +618,16 @@ class PollingConnection(Connection):
                                          context=context)
         response = request(**kwargs)
         kwargs = self.get_poll_request_kwargs(response=response,
-                                              context=context)
+                                              context=context,
+                                              request_kwargs=kwargs)
 
         end = time.time() + self.timeout
         completed = False
         while time.time() < end and not completed:
             response = request(**kwargs)
             completed = self.has_completed(response=response)
-            time.sleep(self.poll_interval)
+            if not completed:
+                time.sleep(self.poll_interval)
 
         if not completed:
             raise LibcloudError('Job did not complete in %s seconds' %
@@ -643,13 +645,17 @@ class PollingConnection(Connection):
                   'headers': headers, 'method': method}
         return kwargs
 
-    def get_poll_request_kwargs(self, response, context):
+    def get_poll_request_kwargs(self, response, context, request_kwargs):
         """
         Return keyword arguments which are passed to the request() method when
         polling for the job status.
 
         @param response: Response object returned by poll request.
         @type response: C{HTTPResponse}
+
+        @param request_kwargs: Kwargs previously used to initiate the
+                                  poll request.
+        @type response: C{dict}
 
         @return C{dict} Keyword arguments
         """
--- a/libcloud/loadbalancer/base.py
+++ b/libcloud/loadbalancer/base.py
@@ -165,7 +165,7 @@ class Driver(BaseDriver):
 
     def update_balancer(self, balancer, **kwargs):
         """
-        Sets the name, algorithm, protocol, or port on a Rackspace load balancer.
+        Sets the name, algorithm, protocol, or port on a load balancer.
 
         @keyword    name: New load balancer name
         @type       metadata: C{str}
--- a/libcloud/loadbalancer/base.py
+++ b/libcloud/loadbalancer/base.py
@@ -163,6 +163,25 @@ class Driver(BaseDriver):
         raise NotImplementedError(
                 'get_balancer not implemented for this driver')
 
+    def update_balancer(self, balancer, **kwargs):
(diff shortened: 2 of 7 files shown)


libcloud/loadbalancer/drivers/rackspace.py
```diff
@@ -292,6 +292,45 @@ def balancer_list_members(self, balancer):
...
+        @type       metadata: C{str}
+
+        @keyword    port: New load balancer port
+        @type       metadata: C{int}
+        """
+        attrs = self._to_update_attrs(kwargs)
+        resp = self.connection.request('/loadbalancers/%s' % balancer.id,
+                method='PUT',
+                data=json.dumps(attrs))
+        return resp.status == 202
```
Kami, line 324: `== httplib.ACCEPTED`

libcloud/loadbalancer/drivers/rackspace.py
```diff
@@ -292,6 +292,45 @@ def balancer_list_members(self, balancer):
...
+        update_attrs = {}
+        if "name" in attrs:
+            update_attrs['name'] = attrs['name']
+
+        if "algorithm" in attrs:
+            algorithm_value = self._algorithm_to_value(attrs['algorithm'])
+            update_attrs['algorithm'] = algorithm_value
+
+        if "protocol" in attrs:
+            update_attrs['protocol'] = attrs['protocol']
```
Kami: create function does `protocol.upper()` so we should do the same thing here.

Ideally we would isolate this into a function.

libcloud/loadbalancer/drivers/rackspace.py
mynnx, line 324: 8a01c5e

libcloud/loadbalancer/drivers/rackspace.py
mynnx: 13fe9e4

libcloud/loadbalancer/drivers/rackspace.py
```diff
@@ -285,13 +286,52 @@ def balancer_detach_member(self, balancer, member):
...
 
-        return resp.status == 202
+        return resp.status == httplib.ACCEPTED
 
     def balancer_list_members(self, balancer):
         uri = '/loadbalancers/%s/nodes' % (balancer.id)
         return self._to_members(
                 self.connection.request(uri).object)
 
+    def update_balancer(self, balancer, **kwargs):
```
Kami, line 310: This method should go to the base class. It should also follow other API's (e.g. DNS) and return a modified `LoadBalancer` object.

mynnx, line 310: c22f8f65, a90472b1, and 8d627e2c.  thanks for the feedback, @Kami!

libcloud/loadbalancer/base.py
```diff
@@ -163,6 +163,25 @@ def get_balancer(self, balancer_id):
         raise NotImplementedError(
                 'get_balancer not implemented for this driver')
 
+    def update_balancer(self, balancer, **kwargs):
+        """
```
Kami, line 167: Need to update the docstring...

mynnx, line 167: fcb37c3a

libcloud/loadbalancer/drivers/rackspace.py
Kami, line 310: Update call is asynchronous which means currently solution won't really work (you will most likely get a non-modified object back).

One solution is to use a `PollingConnection` class. This class periodically polls and blocks until a job has completed. For more examples see Rackspace DNS driver.

libcloud/common/base.py
```diff
@@ -618,14 +618,16 @@ def async_request(self, action, params=None, data='', headers=None,
...
+                                              context=context,
+                                              request_kwargs=kwargs)
 
         end = time.time() + self.timeout
         completed = False
         while time.time() < end and not completed:
             response = request(**kwargs)
             completed = self.has_completed(response=response)
-            time.sleep(self.poll_interval)
+            if not completed:
```
mynnx, line 629: Also, we don't need to sleep if we're done polling.  This should speed up all `async_request`s.

libcloud/loadbalancer/drivers/rackspace.py
mynnx, line 310: Check it out now.  I had to tweak `PollingConnection` slightly, but I like this solution.  I also added an `ex_update_balancer_no_poll()`.

Kami:
Merged in r1222444 .

Thanks.

PR #48, opened 2011-12-21
Rackspace LB - Adds status to member, enumerates conditions

mynnx:
Rackspace load balancer members have a [status attribute](http://docs-beta.rackspace.com/loadbalancers/api/v1.0/clb-devguide/content/Update_Load_Balancer_Attributes-d1e1812.html).  This pull request adds it to the RS driver's `extra` Member dict.

Also, 'condition' should be an enumeration like `base.STATUS` because we'll eventually have an action to set a LB node's condition, and we'll need to do some kwarg transformation when we do.


libcloud/loadbalancer/types.py
```diff
@@ -52,3 +52,12 @@ class State(object):
     UNKNOWN = 2
     ERROR = 3
     DELETED = 4
+
+class MemberCondition(object):
```
mynnx, line 56: I'm definitely willing to move this to another module if it belongs elsewhere.

Kami:
Merged in r1221946 (with some style changes).

Next time please make sure your patch doesn't introduce any new pep8 issues in the edited files.

mynnx:
Thanks!  I'll make sure to run pep8 next time.

PR #50, opened 2012-01-04
Rackspace LB Driver: Bulk deletes

tildedave:
The Rackspace LB API supports bulk deletes of balancers and members (up to 10).  

This pull request adds two new functions to the Rackspace LB Driver:

ex_destroy_balancers
ex_balancer_detach_members

For testing I just chose arbitrary fixtures and wrote tests for the 'delete all balancers/members on that fixture' behavior.

Let me know what can be improved -- I think I imported urllib in the correct way in the tests but I may have messed that up.


tildedave:
Going to merge this into another patch.

PR #51, opened 2012-01-04
Rackspace LB Driver: update member attributes

tildedave:
This adds the following functions to the Rackspace LB Driver for updating the extra attributes on a Rackspace Member (condition and weight).
- ex_balancer_update_member

This makes the request, then waits for the parent balancer to go from PENDING_UPDATE (Libcloud 'Build' status) into ACTIVE (Libcloud 'Running' status.)
- ex_balancer_update_member_no_poll

This makes the request and immediately returns.


tildedave:
Merging this into another patch.

PR #55, opened 2012-01-26
Add NodeBase base class for Node, NodeImage, and NodeSize

keiko713:
Add NodeBase base class for Node, NodeImage, and NodeSize to inherit from in order to share the get_uuid function and common attributes.

### DIFF
--- a/libcloud/compute/base.py
+++ b/libcloud/compute/base.py
@@ -66,16 +66,10 @@ __all__ = [
     "LibcloudHTTPConnection"
     ]
 
-class NodeBase(object):
+class UuidMixin(object):
     """
-    Base class for Node, NodeSize, and NodeImage.
+    Mixin class for get_uuid function.
     """
-
-    def __init__(self, id, name, driver):
-        self.id = str(id) if id else None
-        self.name = name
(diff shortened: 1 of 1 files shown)


Kami:
Hey, @tildedave actually had a good suggestion on the IRC.

It would be better for `NodeBase` to be a mixin instead of a base class. Here is an example of a Mixin class we use in tests - https://github.com/apache/libcloud/blob/trunk/test/compute/__init__.py#L20. Also, better name for this class would be `UuidMixin`.

Edit: Basically, I would do the following:
1. Name the base class `UuidMixin`
2. Remove `__init__` and `__repr__` method and just keep `get_uuid` method
3. Update affected code

keiko713:
Hi, Tomaz. Thank you for your advice. I agree with that suggestion, I'm gonna change and I'll commit again.

Kami:
This patch has been merged, you can close this pull request now.

Thanks

Kami:

@keiko713 - Those changes have been merged a long time ago, can you please close this pull request?

Thanks!

keiko713:
Sure thing!

PR #58, opened 2012-02-06
Openstack Service Catalog (Discussion)

morgabra:
I've changed the openstack base connection and inheriting drivers a bit to make using the service catalog a little easier (hopefully).

Now, every connection class that inherits from the openstack base connection must implement get_endpoint(), who's job is to return the correct endpoint out of the service catalog.

The upside to this is we don't have to generalize parsing the service catalog anymore, which makes it much easier to support things like regions or multiple endpoints of the same type.

The downside is that no other drivers behave this way, although you'll get a pretty obvious NotImplementedError if you do it wrong.

I've also added a new driver for the rackspace nova private beta, which is a pretty simple reference implementation of how you would properly subclass the base openstack infrastructure to make a driver for a specific implementation. 

Future TODO: Split out rackspace-specific code from the openstack base classes/drivers and create new providers for openstack implementations/regions.


libcloud/common/openstack.py
```diff
@@ -202,65 +203,130 @@ def authenticate_2_0_with_body(self, reqbody):
...
+            self._parse_auth_v2(service_catalog)
+        elif ('1.1' in self._auth_version) or ('1.0' in self._auth_version):
+            self._parse_auth_v1(service_catalog)
+        else:
+            raise LibcloudError('auth version "%s" not supported' % (self._auth_version))
+
+    def get_endpoint(self, service_type=None, name=None, region=None):
+
+        if '2.0' in self._auth_version:
+            endpoint = self._service_catalog.get(service_type, {}).get(name, {}).get(region)
```
Kami: Minor styling issue:

I would do `get(region, [])` here and this down bellow:

```
if len(endpoints) == 1:
  return endpoints[0]

return None
```

libcloud/common/openstack.py
```diff
@@ -202,65 +203,130 @@ def authenticate_2_0_with_body(self, reqbody):
...
+            service_type = service['type']
+            service_name = service.get('name', None)
+
+            if service_type not in self._service_catalog:
+                self._service_catalog[service_type] = {}
+
+            if service_name not in self._service_catalog[service_type]:
+                self._service_catalog[service_type][service_name] = {}
+
+            for endpoint in service.get('endpoints'):
```
Kami: Should probably be `service.get('endpoints, [])`

libcloud/common/openstack.py
```diff
@@ -202,65 +203,130 @@ def authenticate_2_0_with_body(self, reqbody):
...
+        """
+        raise NotImplementedError
+
     def add_default_headers(self, headers):
         headers['X-Auth-Token'] = self.auth_token
         headers['Accept'] = self.accept_format
         return headers
 
     def morph_action_hook(self, action):
-        if self._force_base_url:
```
Kami, line 235: Hm, why did you remove this? This way it looks like it won't work with custom base urls anymore.

libcloud/common/openstack.py
```diff
@@ -288,39 +354,8 @@ def _populate_hosts_and_request_paths(self):
 
             self.auth_token = osa.auth_token
 
-            # TODO: Multi-region support
-            if self._auth_version in ['2.0', '2.0_apikey', '2.0_password']:
-                self.tenant_ids = {}
```
Kami, line 293: Can we still assign this variable after parsing the auth API response?

Some code relies in `self.tenant_ids` being present.

morgabra, line 293: grepping returns nothing for 'tenant_ids' and all tests pass. 

LC doesn't really need to care about tenant_ids in general right? We only need the endpoint out of the service catalog.

libcloud/common/openstack.py
morgabra, line 235: Still works because of this:

https://github.com/morgabra/libcloud/blob/5586e622207170bac11f57f9c04aa65196a1af81/libcloud/common/openstack.py#L361

libcloud/common/openstack.py
morgabra: 4c5c0233

libcloud/common/openstack.py
morgabra: 4c5c0233

libcloud/common/openstack.py
Kami, line 293: Yeah, Libcloud doesn't care but some external code cares (e.g. rackspace-monitoring library).

I'm actually OK with removing it from Libcloud and moving this logic into our library.

libcloud/common/openstack.py
Kami, line 235: OK, cool.

If we don't have any existing tests for `ex_force_base_url` functionality it would be nice to add some.

libcloud/common/openstack.py
```diff
@@ -202,65 +203,125 @@ def authenticate_2_0_with_body(self, reqbody):
...
+
+        # check this way because there are a couple of different 2.0_* auth types
+        if '2.0' in self._auth_version:
+            self._parse_auth_v2(service_catalog)
+        elif ('1.1' in self._auth_version) or ('1.0' in self._auth_version):
+            self._parse_auth_v1(service_catalog)
+        else:
+            raise LibcloudError('auth version "%s" not supported' % (self._auth_version))
+
+    def get_endpoint(self, service_type=None, name=None, region=None):
```
pquerna, line 232: add function template_from_tenantId.....

pass in a base url, and this combines it with the tenantId, returning a new url?

morgabra, line 232: I know we talked in meatspace, but now that I think about it...

This returns the endpoint dict, not the url. You'll be able to pull out the tenant_id on your own in your subclasses 'get_endpoint()' implementation and do whatever you like with it. 

Here's the loadbalancer connection's get_endpoint() that does a dirty hack kind of similar to this:

https://github.com/morgabra/libcloud/blob/4c5c023378f66e89ff8b716ad42dff284a33d497/libcloud/loadbalancer/drivers/rackspace.py#L265

libcloud/common/openstack.py
morgabra, line 235: bb4a50ed

libcloud/compute/drivers/openstack.py
```diff
@@ -121,6 +121,30 @@ def parse_error(self):
...
+        elif ('1.1' in self._auth_version) or ('1.0' in self._auth_version):
+            ep = self.service_catalog.get_endpoint(name='cloudServers')
+
+        if 'publicURL' in ep:
+            return ep['publicURL']
+        else:
+            if '2.0' in self._auth_version:
+                ep = self.service_catalog.get_endpoint(service_type='compute',
+                                                       name='nova',
+                                                       region='ORD')
```
pquerna: this really sucks to hard code ORD region into the generate base openstack connection, this belongs in RackspaceNovaConnection?

libcloud/common/openstack.py
morgabra, line 232: get_endpoint on the service catalog return the whole endpoint, so you can get whatever you want out of it in the connection class. 

For example:

``` python
(Pdb) driver.connection.service_catalog.get_endpoint(name='cloudServersOpenStack', service_type='compute', region='DFW')
{'region': 'DFW', 'tenantId': '1337', 'versionId': '2', 'versionList': 'https://dfw.servers.api.rackspacecloud.com/', 'versionInfo': 'https://dfw.servers.api.rackspacecloud.com/v2/', 'publicURL': 'https://dfw.servers.api.rackspacecloud.com/v2/1337'}
(Pdb) ep = driver.connection.service_catalog.get_endpoint(name='cloudServersOpenStack', service_type='compute', region='DFW')
(Pdb) ep['publicURL']
'https://dfw.servers.api.rackspacecloud.com/v2/1337'
(Pdb) ep['tenantId']
'1337'
```

libcloud/common/openstack.py
```diff
@@ -202,65 +203,125 @@ def authenticate_2_0_with_body(self, reqbody):
                 e = sys.exc_info()[1]
                 raise MalformedResponseError('Auth JSON response is missing required elements', e)
 
+
+class OpenStackServiceCatalog(object):
+    """
+    http://docs.openstack.org/api/openstack-identity-service/2.0/content/
+
+    This class should be instanciated with the contents of the 'serviceCatalog'
```
mynnx, line 211: "instantiated"

libcloud/common/openstack.py
```diff
@@ -202,65 +203,125 @@ def authenticate_2_0_with_body(self, reqbody):
...
+        if '2.0' in self._auth_version:
+            endpoint = self._service_catalog.get(service_type, {}).get(name, {}).get(region, [])
+        elif ('1.1' in self._auth_version) or ('1.0' in self._auth_version):
+            endpoint = self._service_catalog.get(name, {}).get(region, [])
+
+        # ideally an endpoint either isn't found or only one match is found.
+        if len(endpoint) == 1:
+            return endpoint[0]
+        else:
+            return []
```
Kami: Wait, this should be `None`, not empty list, right?

Kami:
LGTM, can you please open a ticket on JIRA and attach a patch there? Thanks.

Kami:
This has been merged, can you please close this pull request now? Thanks.

PR #59, opened 2012-02-10
Specify condition weight when adding lb members

adampickeral:
Weight or condition couldn't be specified when adding members to a rackspace load balancer. This fixes that.


Kami:
This has been merged. Can you please close the pull request?

PR #61, opened 2012-03-02
Openstack images - adding serverId extra

mynnx:
Images in Openstack may have a `server` dictionary with an ID, just like in the Rackspace API.  This patch adds a `serverId` extra to Openstack images.

### DIFF
--- a/libcloud/compute/drivers/openstack.py
+++ b/libcloud/compute/drivers/openstack.py
@@ -843,7 +843,7 @@ class OpenStack_1_1_NodeDriver(OpenStackNodeDriver):
         return images
 
     def _to_image(self, api_image):
-        server = api_image.get('server') or {}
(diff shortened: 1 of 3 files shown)


libcloud/compute/drivers/openstack.py
```diff
@@ -843,6 +843,7 @@ def _to_images(self, obj, ex_only_active):
         return images
 
     def _to_image(self, api_image):
+        server = api_image.get('server') or {}
```
Kami: Should be `server = api_image.get('server', {})`.

mynnx: oh yeah =P

8e78da4

mynnx:
Kami merged this with https://github.com/apache/libcloud/commit/edcb55e2771a804f0d6f93a7e4365c7fd6475877

PR #64, opened 2012-05-09
Return password in deploy_node

juancarlosm:


Kami:
I committed a more robust version to trunk - http://svn.apache.org/viewvc/libcloud/trunk/libcloud/compute/base.py?view=diff&r1=1336475&r2=1336476&pathrev=1336476 :
- Only put a password in the extra dictionary if the password is actually provided (aka password, not key auth is used)

Thanks

juancarlosm:
ok!! ;)

PR #65, opened 2012-05-14
Fix typo in compute/drivers/cloudstack.py

daemianmack:
Add an R!


novel:
Thanks for noticing that, fix committed!

However, pull requests are not the preferred way of submitting patches to libcloud, please use mailling list or JIRA next time.

daemianmack:
Happy to help with such a crucial fix! Will check out JIRA for next commit.

PR #66, opened 2012-06-07
Added doc tags to joyent driver

benr:


Kami:
Committed into trunk (http://svn.apache.org/viewvc/libcloud/trunk/libcloud/compute/drivers/joyent.py?view=diff&r1=1348253&r2=1348254&pathrev=1348254), thanks.

PR #67, opened 2012-07-02
Modified AtmosError so that the message argument is actually useful.

freakboy3742:
As currently implemented, AtmosError (in the Atmos storage backend) extends Exception, but then doesn't pass the message argument to the superclass. As a result, when an AtmosError is raised, the printed error message is useless.

Simply passing the message argument up to the superclass means that the message handling of the base Exception class is used, which means printed exceptions actually make sense.

### DIFF
--- a/libcloud/storage/drivers/atmos.py
+++ b/libcloud/storage/drivers/atmos.py
@@ -44,8 +44,8 @@ def collapse(s):
 
 class AtmosError(Exception):
     def __init__(self, code, message):
+        super(AtmosError, self).__init__(message)
         self.code = code
-        self.message = message
 
 class AtmosResponse(XmlResponse):
     def success(self):


Kami:
Thanks for the patch. I've modified it to follow Libcloud conventions and inherit from LibcloudError - http://svn.apache.org/viewvc/libcloud/trunk/libcloud/storage/drivers/atmos.py?view=diff&r1=1356063&r2=1356064&pathrev=1356064

Kami:

@freakboy3742 - Those changes have been merged a long time ago, can you please close this pull request?

Thanks!

freakboy3742:
Not sure why this wasn't automatically closed as a result of the commit, or why you can't close it yourself as a result of knowing that it's been committed... but sure - I can close it.

PR #74, opened 2012-11-02
LIBCLOUD-252 : Implement a local storage driver

mahendra:
The idea is to provide a 'local' storage driver. This will provide developers with the ability to implement storage features with libcloud such that it can be deployed on a cloud platform or over local file-systems (for eg: over a local disk or a n/w mounted storage). This will allow for hybrid deployments where the storage layer can be altered via config changes. 

In addition, this also let's users play around with the library without the need to access a cloud system 

A sample implementation could 
- use the 'key' argument as the base file-system where the objects will be stored 
- use a file-system hierarchy to represent containers and objects

### DIFF
--- a/libcloud/storage/drivers/local.py
+++ b/libcloud/storage/drivers/local.py
@@ -1,0 +1,583 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Provides storage driver for working with local filesystem
+"""
+
+import os
+import shutil
+
+from lockfile import mkdirlockfile
+from pwd import getpwuid
+from libcloud.utils.py3 import next
+from libcloud.common.base import Connection
+from libcloud.storage.base import Object, Container, StorageDriver
+from libcloud.common.types import LibcloudError, LazyList
+from libcloud.storage.types import ContainerAlreadyExistsError
+from libcloud.storage.types import ContainerDoesNotExistError
+from libcloud.storage.types import ContainerIsNotEmptyError
+from libcloud.storage.types import ObjectError
+from libcloud.storage.types import ObjectDoesNotExistError
+from libcloud.storage.types import InvalidContainerNameError
+
+CHUNK_SIZE = 16384
+IGNORE_FOLDERS = ['.lock', '.hash']
+
+class LockLocalStorage:
+    """
+    A class to help in locking a local path before being updated
+    """
+    def __init__(self, path):
+        self.path = path
+        self.lock = mkdirlockfile.MkdirLockFile(self.path, threaded=True)
+
+    def __enter__(self):
+        try:
+            self.lock.acquire(timeout=0.1)
+        except lockfile.LockTimeout:
+            raise LibcloudError('Lock timeout')
+
+    def __exit__(self, type, value, traceback):
+        if self.lock.is_locked():
+            self.lock.release()
+
+        if value is not None:
+            raise value
+
+class LocalStorageDriver(StorageDriver):
+    """
+    Implementation of local file-system based storage. This is helpful
+    where the user would want to use the same code (using libcloud) and
+    switch between cloud storage and local storage
+    """
+
+    connectionCls = Connection
+    name = 'Local Storage'
+
+    def __init__(self, key, secret=None, secure=True, host=None, port=None,
+                 **kwargs):
+
+        # Use the key as the path to the storage
+        self.base_path = key[0]
+
+        if not os.path.isdir(self.base_path):
+            raise LibcloudError('The base path is not a directory')
+
+        super(StorageDriver, self).__init__(key=key, secret=secret,
+                                            secure=secure, host=host,
+                                            port=port, **kwargs)
+
+    def _make_path(self, path, ignore_existing=True):
+        """
+        Create a path by checking if it already exists
+        """
+
+        try:
+            os.makedirs(path)
+        except OSError, exp:
+            if exp.errno == 17 and not ignore_existing:
+                raise exp
+
+    def _make_container(self, container_name):
+        """
+        Create a container instance
+
+        @param container_name: Container name.
+        @type container_name: C{str}
+
+        @return: Container instance.
+        @rtype: L{Container}
+        """
+
+        if '/' in container_name:
+            raise InvalidContainerNameError(value=None, driver=self,
+                                            container_name=container_name)
+
+        full_path = os.path.join(self.base_path, container_name)
+
+        try:
+            stat = os.stat(full_path)
+            if not os.path.isdir(full_path):
+                raise ValueError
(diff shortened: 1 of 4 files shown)


libcloud/storage/drivers/local.py
```diff
@@ -0,0 +1,583 @@
...
+from libcloud.storage.types import ContainerDoesNotExistError
+from libcloud.storage.types import ContainerIsNotEmptyError
+from libcloud.storage.types import ObjectError
+from libcloud.storage.types import ObjectDoesNotExistError
+from libcloud.storage.types import InvalidContainerNameError
+
+CHUNK_SIZE = 16384
+IGNORE_FOLDERS = ['.lock', '.hash']
+
+class LockLocalStorage:
```
Kami: Use new style classes (inherit from `object`).

libcloud/storage/drivers/local.py
```diff
@@ -0,0 +1,583 @@
...
+
+    def _make_path(self, path, ignore_existing=True):
+        """
+        Create a path by checking if it already exists
+        """
+
+        try:
+            os.makedirs(path)
+        except OSError, exp:
+            if exp.errno == 17 and not ignore_existing:
```
Kami: Use a constant - `errno.EEXIST`

libcloud/storage/drivers/local.py
```diff
@@ -0,0 +1,583 @@
...
+        Create a container instance
+
+        @param container_name: Container name.
+        @type container_name: C{str}
+
+        @return: Container instance.
+        @rtype: L{Container}
+        """
+
+        if '/' in container_name:
```
Kami: Should also check for `\`.

libcloud/storage/drivers/local.py
```diff
@@ -0,0 +1,583 @@
...
+            raise InvalidContainerNameError(value=None, driver=self,
+                                            container_name=container_name)
+
+        full_path = os.path.join(self.base_path, container_name)
+
+        try:
+            stat = os.stat(full_path)
+            if not os.path.isdir(full_path):
+                raise ValueError
```
Kami: Include an error message `raise ValueError('Target path is not a directory')`.

libcloud/storage/drivers/local.py
```diff
@@ -0,0 +1,583 @@
...
+                                            port=port, **kwargs)
+
+    def _make_path(self, path, ignore_existing=True):
+        """
+        Create a path by checking if it already exists
+        """
+
+        try:
+            os.makedirs(path)
+        except OSError, exp:
```
Kami: This way of catching exceptions won't work with all the Python versions we support. Please have a look at how we do it in other places - https://github.com/apache/libcloud/blob/trunk/libcloud/compute/drivers/gandi.py#L66

libcloud/storage/drivers/local.py
```diff
@@ -0,0 +1,583 @@
...
+        if '/' in container_name:
+            raise InvalidContainerNameError(value=None, driver=self,
+                                            container_name=container_name)
+
+        path = os.path.join(self.base_path, container_name)
+
+        try:
+            self._make_path(path, ignore_existing=False)
+        except OSError, exp:
+            if exp.errno == 17:
```
Kami: Same here (use a constant).

libcloud/storage/drivers/local.py
```diff
@@ -0,0 +1,583 @@
...
+            raise InvalidContainerNameError(value=None, driver=self,
+                                            container_name=container_name)
+
+        full_path = os.path.join(self.base_path, container_name)
+
+        try:
+            stat = os.stat(full_path)
+            if not os.path.isdir(full_path):
+                raise ValueError
+        except Exception:
```
Kami: You should try to catch more specific error, `OSError` should work in this case I think.

libcloud/storage/drivers/local.py
```diff
@@ -0,0 +1,583 @@
...
+        """
+
+        full_path = os.path.join(self.base_path, container.name, object_name)
+
+        if os.path.isdir(full_path):
+            raise ObjectError(value=None, driver=self, object_name=object_name)
+
+        try:
+            stat = os.stat(full_path)
+        except Exception:
```
Kami, line 160: Same here.

libcloud/storage/drivers/local.py
```diff
@@ -0,0 +1,583 @@
...
+            raise ObjectDoesNotExistError(value=None, driver=self,
+                                          object_name=object_name)
+
+        extra = {}
+        extra['creation_time'] = stat.st_ctime
+        extra['access_time'] = stat.st_atime
+        extra['modify_time'] = stat.st_mtime
+
+        meta_data = {}
+        meta_data['owner'] = getpwuid(stat.st_uid).pw_name
```
Kami: getpwuid stuff is not supported on Window...

libcloud/storage/drivers/local.py
```diff
@@ -0,0 +1,583 @@
...
+
+        @param container_name: Container name.
+        @type container_name: C{str}
+
+        @return: L{Container} instance.
+        @rtype: L{Container}
+        """
+        return self._make_container(container_name)
+
+    def get_container_cdn_url(self, container, check=False):
```
Kami, line 244: Missing a docstring for `check` argument.

libcloud/storage/drivers/local.py
```diff
@@ -0,0 +1,583 @@
...
+
+        if os.path.exists(file_path) and not overwrite_existing:
+            raise LibcloudError(
+                value='File %s already exists, but ' % (file_path) +
+                'overwrite_existing=False',
+                driver=self)
+
+        try:
+            shutil.copy(obj_path, file_path)
+        except Exception:
```
Kami: Please catch more specific exception.

libcloud/storage/drivers/local.py
```diff
@@ -0,0 +1,583 @@
...
+        @rtype: C{object}
+        """
+
+        path = self.get_object_cdn_url(obj)
+        obj_file = open(path)
+
+        if chunk_size is None:
+            chunk_size = CHUNK_SIZE
+
+        data = obj_file.read(chunk_size)
```
Kami: There is an existing method called `libcloud.utils.files.read_in_chunks` which already does what you are doing here.

Kami: And this whole thing also isn't wrapped in with which could potentially mean that file gets left open in case an exception thrown.

libcloud/storage/drivers/local.py
```diff
@@ -0,0 +1,583 @@
...
+            except Exception, exp:
+                return False
+
+        # Check and delete the folder if required
+        path = os.path.dirname(path)
+
+        try:
+            if path != obj.container.get_cdn_url():
+                os.rmdir(path)
+        except Exception, exp:
```
Kami: Same here - get rid of `, exp` otherwise it won't work with all the supported Python versions.

libcloud/storage/drivers/local.py
```diff
@@ -0,0 +1,583 @@
...
+        Create a new container.
+
+        @type container_name: C{str}
+        @param container_name: Container name.
+
+        @return: C{Container} instance on success.
+        @rtype: L{Container}
+        """
+
+        if '/' in container_name:
```
Kami: Same here, also check for `\`. You should probably make a method called `_verify_container_name` or something like that.

Kami:
I also noticed some pep8 issues, please run pep8 on your files.

libcloud/storage/drivers/local.py
mahendra: For the time being I am removing the need for 'owner'. Will explore this and get back to you with a fix later.

mahendra:
Thanks for your review. Have addressed them.

Kami:
@mahendra Looks good. Can you please attach an updated patch in JIRA ticket and I'll merge it.

mahendra:
done!

mahendra:
Fixed by https://issues.apache.org/jira/browse/LIBCLOUD-252

PR #75, opened 2012-11-07
LIBCLOUD-254 : Generator based iteration instead of LazyList

mahendra:
`LazyList` was implemented by issue (LIBCLOUD-78) for more efficient iteration over objects stored in a container (S3, CloudFiles etc. limit the maximum number of objects returned in a single call). 

`LazyList` solved this problem, but I think it might have the following issues while handling containers with large number of objects 
1. It loads the entire list to memory 
2. caller has to wait for the entire list to be loaded in memory before any operation can be done 
3. The api invocation using `get_more()` and value_list is a bit complex (as-in, it can be simplified) 

By using python generators, the above problems can be alleviated. Results can be returned to the caller as and when it is returned from the server. 

The following changes were done to the libcloud apis 
- A new api called - `iterate_container_objects()` was introduced. The storage drivers need to implement this instead of `list_container_objects()`. This API now returns a generator. Usage of this API will alleviate the above three problems. 
- `list_container_objects()` will simply do - `list(self.iterate_container_objects(container))` - this is maintained for backwards compatibility. It would be better if users can start using `iterate_**()` api instead. 
- The same changes have been made for the DNS base class also. 
- `LazyList()` can be removed from libcloud if it is OK with everyone. 
- The generator based interface can be used (WIP) for providing paginated access to objects - This will be useful for webpages/apps where the user has to paginate through the results. The same can be implemented by providing `start_key` and `count` parameters (similar to CouchDB) instead of generating the entire list and then doing an offset. This will be more performance/memory efficient than generating the entire list for every request.

### DIFF
--- a/libcloud/dns/base.py
+++ b/libcloud/dns/base.py
@@ -164,14 +164,34 @@ class DNSDriver(BaseDriver):
         """
         return list(self.RECORD_TYPE_MAP.keys())
 
+    def iterate_zones(self):
+        """
+        Return a generator to iterate over available zones.
+
+        @rtype: C{generator} of L{Zone}
+        """
+        raise NotImplementedError(
+            'iterate_zones not implemented for this driver')
+
     def list_zones(self):
         """
         Return a list of zones.
 
         @rtype: C{list} of L{Zone}
         """
+        return list(self.iterate_zones())
+
+    def iterate_records(self, zone):
+        """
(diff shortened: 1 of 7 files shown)


mahendra:
Fixed by https://issues.apache.org/jira/browse/LIBCLOUD-254

PR #76, opened 2012-11-07
LIBCLOUD-255 : Driver registration and handling improvements

mahendra:
The following limitations can come up while using libcloud in production - because of usage of C/C++ style `ENUM` based mechanism of registering drivers.
- In a s/w setup which supports multiple cloud environments (or local), the admin/dev must be able to change the driver via a conf file. For eg: 

```
storage_driver = 's3'
username = 'pqrs'
key = 'abcd'
```

   In the current mechanism, the mapping of the driver name to number must be done externally.
- There is no mechanism for dynamically registering a driver. It is quite possible that user's might have drivers which are for private cloud setups. They may want to install libcloud via PYPI, but use their own private drivers. Currently, there is no mechanism to do this without changing libcloud code.

Also, IMHO, the C/C++ style enumerations need not be there in python.

The solution for this is pretty simple to implement (given the good design of libcloud).
1. Replace the `enum` numbers with strings. Given the usage of a dictionary for registering drivers, this will work out of the box.
2. Implement a `set_driver()` API which can be used for registering new drivers.

With this change, libcloud usage can be controlled externally (via a config file or a db field etc.) as the admin sees fit.


mahendra:
Fixed by https://issues.apache.org/jira/browse/LIBCLOUD-255

PR #80, opened 2012-12-20
LIBCLOUD-269 : Multipart upload for amazon S3

mahendra:
This patch adds support for streaming data upload using Amazon's multipart upload support as listed in [S3 docs](http://docs.amazonwebservices.com/AmazonS3/latest/dev/UsingRESTAPImpUpload.html)

As per current behaviour, the `upload_object_via_stream()` API will collect the entire object in memory, and then upload it to S3. This can turn problematic with large files (think HD videos around 4GB). This will be a huge hit in performance and memory of the python application.

With this patch, the API `upload_object_via_stream()` will use the S3 multipart upload feature to upload data in 5MB chunks, thus reducing the overall memory impact of the application.
## Design of this feature:
- The `S3StorageDriver()` is not used just for Amazon S3. It is sub-classed for being used with other S3 compliant cloud storage providers like Google Storage.
- The Amazon S3 multipart upload is not (or may not be) supported by other storage providers (who will prefer the chunked upload mechanism)

We can solve this problem in two ways:
- Create a new subclass of `S3StorageDriver` (say `AmazonS3StorageDriver`), which implements this new multipart upload mechanism. Other storage providers will subclass `S3StorageDriver`. This is a more cleaner approach.
- Introduce an attribute `supports_s3_multipart_upload` and based on it's value, control the callback function passed to `_put_object()` API. This makes the code look a bit hacky, but this approach is better for supporting such features in the future. We don't have to keep making sub-classes for each feature.

In the current patch, I have implemented the latter approach, though I prefer the former. After discussions with the community and knowing their preferences, we can select a final approach.
## Design notes:
- Implementation has three steps
  1. `POST` request to `/container/object_name?uploads`. This returns an XML with a unique `uploadId`. This is handled as part of `_put_object()`. Doing it via `_put_object()` ensures that all S3 related parameters are set correctly.
  2. Upload each chunk via `PUT` to `/container/object_name?partNumber=X&uploadId=***` - This is done via the callback that is passed to `_put_object()` named `_upload_multipart()`
  3. `POST` an XML containing part-numbers and etag headers returned for each part to `/container/object_name?uploadId=***`, implemented via `_commit_multipart()`
  4. In case of any failures in steps (2) or (3), the upload is deleted from S3 through a `DELETE` request to `/container/object_name?uploadId=****`, implemented via `_abort_multipart()`
- The chunk size for upload was set as 5MB - This is the minimum allowed size as per Amazon S3 docs.
## Other changes:
- Did some PEP8 cleanup on s3.py
- `s3.get_container()` would iterate through the list of containers for finding the requested entry. This can be simplified by making a `HEAD` request. The only downside is that `created_time` is not available for the container. Let me know if this approach is OK or if I must revert it.
- Introduced the following APIs for the `S3StorageDriver()`, to make some functionality easier.
  - `get_container_cdn_url()`
  - `get_object_cdn_url()`
- In `libcloud.common.base.Connection`, the `request()` method is used as the basis for all HTTP requests made by libcloud. This method had a limitation, which became apparent in S3 multipart upload implementation. For initializing an upload, the API invoked was  `/container/object_name?uploads`.  The `uploads` parameter had to be passed as-is, without any values. If we made use of `params` argument in `request()` method, it would have come up as `uploads=***`. To prevent this, the `action` was set to `/container/object_name?uploads` and slight modifications were made to how parameters were appended. This also forced a change in `BaseMockHttpObject._get_method_name()`
## Bug fixes in test framework
- While working on the test cases, I noticed a small issue. Not sure if it was a bug or as per design. `MockRawResponse._get_response_if_not_availale()` would return two different values on subsequent invocations.

```
     if not self._response:
         ...
         return self  <----- this was inconsistent.
     return self._response
```

  While adding test cases for the Amazon S3 functionality, I noticed that instead of getting back `MockResponse`, I was getting `MockRawResponse` instance (which did not have methods like `read()` or `parse_body()`). So, I fixed this issue. Because of this other test cases started failing and they were subsequently fixed. Not sure if this has to be fixed or if it was done on purpose. If someone can throw some light on it, I can work on it further. As of now, all test cases pass.
- In test_s3.py, the driver was being set everywhere to `S3StorageDriver`. This same test case is used for `GoogleStorageDriver`, where the driver turns up as `S3StorageDriver` instead of `GoogleStorageDriver`. This was fixed by changing code to `driver=self.driver_type`


libcloud/common/base.py
```diff
@@ -574,7 +574,10 @@ def request(self,
         params, headers = self.pre_connect_hook(params, headers)
 
         if params:
-            url = '?'.join((action, urlencode(params)))
+            if '?' in action:
```
Kami, line 578: Would be nice to document this behavior in the docstring.

libcloud/storage/drivers/s3.py
```diff
@@ -234,6 +241,42 @@ def get_object(self, container_name, object_name):
...
+        @param check: Indicates if the path's existance must be checked
+        @type check: C{bool}
+
+        @return: A CDN URL for this container.
+        @rtype: C{str}
+        """
+        if check:
+            self.get_container(container.name)
+
+        return '/%s' % (container.name)
```
Kami, line 291: This method shouldn't really be called `get_container_cdn_url` because it just returns an internal path to the object.

`get_container_cdn_url` is supposed to return a full CDN url of a container. In AWS CDN is handled by a separate service (CloudFront) so adding CDN support is more involved.

libcloud/storage/drivers/s3.py
```diff
@@ -234,6 +241,42 @@ def get_object(self, container_name, object_name):
...
+
+    def _get_object_cdn_url(self, container, object_name):
+        container_url = self.get_container_cdn_url(container)
+        object_name_cleaned = self._clean_object_name(object_name)
+        object_path = '%s/%s' % (container_url, object_name_cleaned)
+        return object_path
+
+    def get_object_cdn_url(self, obj):
```
Kami: Same here.

libcloud/storage/drivers/s3.py
```diff
@@ -337,6 +373,133 @@ def upload_object(self, file_path, container, object_name, extra=None,
...
+
+        try:
+            # Upload the data through the iterator
+            result = self._upload_from_iterator(iterator, object_path,
+                                                upload_id, calculate_hash)
+            (chunks, data_hash, bytes_transferred) = result
+
+            # Commit the chunk info and complete the upload
+            etag = self._commit_multipart(object_path, upload_id, chunks)
+        except Exception:
```
Kami, line 447: Hm, can we catch more specific exception here?

In any case we should at least include the original exception message in the custom exception we re-throw.

libcloud/storage/drivers/s3.py
```diff
@@ -337,6 +373,133 @@ def upload_object(self, file_path, container, object_name, extra=None,
...
+
+            etag_id = SubElement(part, 'ETag')
+            etag_id.text = str(etag)
+
+        if PY3:
+            encoding = 'unicode'
+        else:
+            encoding = None
+
+        data = tostring(root, encoding=encoding)
```
Kami: Hm, utility function which wraps original string and uses correct encoded based on the Python version should probably live in `libcloud.utils.py3`.

libcloud/storage/drivers/s3.py
```diff
@@ -337,6 +373,133 @@ def upload_object(self, file_path, container, object_name, extra=None,
...
+
+        # Modify the response header of the first request. This is used
+        # by other functions once the callback is done
+        response.headers['etag'] = etag
+
+        return (True, data_hash, bytes_transferred)
+
+    def _upload_from_iterator(self, iterator, object_path, upload_id,
+                              calculate_hash=True):
+        """Uploads data from an interator in fixed sized chunks to S3"""
```
Kami: Please add docstrings for all the arguments (http://libcloud.apache.org/docstring-conventions.html).

libcloud/common/base.py
mahendra, line 578: Done.

libcloud/storage/drivers/s3.py
mahendra, line 291: Agreed. Changed it to _get_container_path()

libcloud/storage/drivers/s3.py
mahendra: Changed it to _get_object_path()

libcloud/storage/drivers/s3.py
mahendra, line 447: I thought about it. In Amazon multipart upload, the uploadId keeps lingering if the upload is not aborted explicitly. 

Using boto, we can observe all the pending multipart uploads.

```
connection = boto.s3.S3Connection(...)
bucket = connection.get_bucket(container_name)
uploads = bucket.get_all_multipart_uploads()
```

So, Ideally we should abort the multipart upload in case of any exception. Else, we will leave lingering uploads pending in S3. That is why, I decided to catch a generic exception.

Will change the code to ensure that the specific exception is thrown back

libcloud/storage/drivers/s3.py
mahendra: Done

libcloud/common/base.py
Kami, line 578: I don't see any new commits yet. Looks like you still need to push the changes :)

libcloud/storage/drivers/s3.py
mahendra: Hmmm, could not find any... Or do you want me to create one there?

Kami: Sorry, I wasn't clear in my previous comment. One doesn't exist yet so it would be great if you can create it there.

libcloud/common/base.py
mahendra, line 578: Sorry. I was addressing comments one-by-one. Did a push after I answered all the comments. You should be able to see one now.

libcloud/storage/drivers/s3.py
mahendra: OK. So, you want it to be like this in utils.py3

```
if sys.version_info >= (3, 0):
    encoding = 'unicode'
else:
    encoding = None
```

and it being used as 
`ET.tostring(data, encoding=encoding)`

OR

```
if sys.version_info >= (3,0):
    ...
    def tostring(data):
        return ET.tostring(data)

else:
    tostring = ET.tostring
```

(ideally the second one should be in utils.xml.py
OR, we can use a combination of both. (using 'xml_encoding' in py3) and tostring() in xml.py

Kami: Number two or a combination of both.

mahendra: Actually, this is turning a bit tricky because of import conflict among `utils.xml` and `xml.extree.ElementTree`

mahendra: Is it ok, if we rename the xml.py to xmlutils.py or something?

Kami: I'm a bit confused, can you please clarify what's conflicting?

Can't you just use an alias when importing it?

libcloud/storage/drivers/s3.py
Kami, line 447: Looks good now :)

Kami, line 447: Another thing - can we also add an extension method for listing lingering multipart uploads (e.g. `ex_list_multipart_uploads)?

libcloud/test/storage/test_s3.py
```diff
@@ -617,6 +736,34 @@ def test_upload_object_via_stream(self):
...
+        object_name = 'foo_test_stream_data'
+        iterator = _faulty_iterator()
+        extra = {'content_type': 'text/plain'}
+
+        try:
+            obj = self.driver.upload_object_via_stream(container=container,
+                                                       object_name=object_name,
+                                                       iterator=iterator,
+                                                       extra=extra)
+        except Exception:
```
Kami, line 792: Need to use sys "hack" otherwise it won't work with all the supported Python versions.

libcloud/storage/drivers/s3.py
Kami, line 447: Besides this and a couple of other minor things I have mentioned, the patch looks great.

Please update JIRA ticket with the latest patch when you are done with addressing those things and I will proceed with testing and merging it into trunk :)

mahendra, line 447: Yeah sure. Will add few more methods

```
ex_list_multipart_uploads
ex_abort_multipart_upload
ex_cleanup_all_multipart_uploads
```

Also, need to check how to expose these in the CLI commands. Have some ideas. Will work on a future ticket for this.
Once I implement this, will upload patch to JIRA

libcloud/storage/drivers/s3.py
mahendra: In the code, if I use

`from xml.etree import ....`

it clashes with

```
libcloud.utils.xml
```

libcloud/storage/drivers/s3.py
mahendra, line 447: Taken care of all your comments. Thanks for the detailed review.

Implemented the following extensions

```
ex_iterate_multipart_uploads()
ex_cleanup_all_multipart_uploads()
```

mahendra:
Fixed by https://issues.apache.org/jira/browse/LIBCLOUD-269

PR #83, opened 2013-01-04
Support Amazon Route53 DNS

Jc2k:
Here is a set of improvements to the Amazon Route53 Driver.

In includes tests for the existing code and also adds the methods required to actually make changes to a zone.


libcloud/dns/drivers/route53.py
```diff
@@ -71,19 +65,28 @@ def error(self):
...
-            context = self.connection.context
-            messages = []
-            if context['InvalidChangeBatch']['Messages']:
-                for message in context['InvalidChangeBatch']['Messages']:
-                    messages.append(message['Message'])
+        try:
+            body = ET.XML(self.body)
+        except:
+            raise MalformedResponseError("Failed to parse XML",
+                                         body=self.body, driver=EC2NodeDriver)
```
Kami: Should be `driver=self.driver`.

libcloud/dns/drivers/route53.py
```diff
@@ -71,19 +65,28 @@ def error(self):
...
 
-        elif status == 400:
-            context = self.connection.context
-            messages = []
-            if context['InvalidChangeBatch']['Messages']:
-                for message in context['InvalidChangeBatch']['Messages']:
-                    messages.append(message['Message'])
+        try:
+            body = ET.XML(self.body)
+        except:
```
Kami: `except Exception`

libcloud/dns/drivers/route53.py
```diff
@@ -71,19 +65,28 @@ def error(self):
...
+            raise MalformedResponseError("Failed to parse XML",
+                                         body=self.body, driver=EC2NodeDriver)
+
+        errs = findall(element=body, xpath='Error', namespace=NAMESPACE)
+        if errs:
+            t, code, message = errs[0].getchildren()
+            print t.text, code.text, message.text
```
Kami: Extra print statement.

libcloud/dns/drivers/route53.py
```diff
@@ -71,19 +65,28 @@ def error(self):
...
+            if code.text == "NoSuchHostedZone":
+                raise ZoneDoesNotExistError(value=message.text, driver=self, zone_id=context.get('zone_id',''))
```
Kami: `driver=self.driver`

Kami: Usually the convention is to use `None` instead of an empty string so it should be `context.get('zone_id', None)`

libcloud/dns/drivers/route53.py
```diff
@@ -123,36 +126,122 @@ class Route53DNSDriver(DNSDriver):
...
 
     def get_record(self, zone_id, record_id):
         zone = self.get_zone(zone_id=zone_id)
-        data = ET.XML(self.connection.request(API_ROOT + 'hostedzone/'
-                      + zone_id + '/rrset?maxitems=1&name=' + record_id)
-                      .object)
+        record_type, name = record_id.split(":",1)
+        self.connection.set_context({'zone_id': zone_id})
+        data = self.connection.request(API_ROOT + 'hostedzone/'
+                      + zone_id + '/rrset?maxitems=1&name=' + name + '&type=' + record_type).object
```
Kami: Please use `libcloud.utils.py3.urlencode` to construct query parameter string for the URL instead of concatenating strings.

Jc2k:
I have implemented these suggestions.

libcloud/dns/drivers/route53.py
```diff
@@ -123,36 +125,123 @@ class Route53DNSDriver(DNSDriver):
...
         zone = self._to_zone(elem=findall(element=data, xpath='HostedZone',
                                           namespace=NAMESPACE)[0])
         return zone
 
     def get_record(self, zone_id, record_id):
         zone = self.get_zone(zone_id=zone_id)
-        data = ET.XML(self.connection.request(API_ROOT + 'hostedzone/'
-                      + zone_id + '/rrset?maxitems=1&name=' + record_id)
-                      .object)
+        record_type, name = record_id.split(":",1)
```
Kami: Is it possible for this call to return multiple items?

libcloud/dns/drivers/route53.py
```diff
@@ -123,36 +125,123 @@ class Route53DNSDriver(DNSDriver):
...
 
-        record = self._to_records(data=data, zone=zone)
         return record
 
+    def create_zone(self, domain, type='master', ttl=None, extra=None):
+        zone = ET.Element("CreateHostedZoneRequest", {'xmlns': NAMESPACE})
+        ET.SubElement(zone, "Name").text = domain
+        ET.SubElement(zone, "CallerReference").text = str(uuid.uuid4())
+        if extra and "Comment" in extra:
+            ET.SubElement(ET.SubElement(zone, "HostedZoneConfig"), "Comment").text = extra['Comment']
```
Kami: You can remove this method. If absent, parent class method defaults to raising UnimplementedError.

libcloud/dns/drivers/route53.py
```diff
@@ -123,36 +125,123 @@ class Route53DNSDriver(DNSDriver):
...
+
+        response = self.connection.request(API_ROOT+'hostedzone', method="POST", data=ET.tostring(zone)).object
+
+        return self._to_zone(elem=findall(element=response, xpath='HostedZone', namespace=NAMESPACE)[0])
```
Kami: Default behavior should be "try to delete a zone without deleting records", because that's what we do in other drivers and generally in Libcloud we try to do "1 http request per method call" by default.

Then a method can take `ex_delete` keyword argument and if it's true (defaults to false) it proceeded with deleting records before trying to delete a zone.

libcloud/dns/drivers/route53.py
```diff
@@ -123,36 +125,123 @@ class Route53DNSDriver(DNSDriver):
...
+        for r in zone.list_records():
+            if r.type in (RecordType.NS, RecordType.SOA):
+                continue
+            deletions.append(("DELETE", r.name, r.type, r.data, r.extra))
+        if deletions:
+            self._post_changeset(zone, deletions)
+
+        # Now delete the zone itself
+        response = self.connection.request(API_ROOT+'hostedzone/%s' % zone.id, method="DELETE").object
+        return True
```
Kami: Hm, is this how AWS uniquely identifies a record?

libcloud/dns/drivers/route53.py
```diff
@@ -123,36 +125,123 @@ class Route53DNSDriver(DNSDriver):
...
+        changeset = ET.Element("ChangeResourceRecordSetsRequest", {'xmlns': NAMESPACE})
+        batch = ET.SubElement(changeset, "ChangeBatch")
+        changes = ET.SubElement(batch, "Changes")
+
+        for action, name, type_, data, extra in changes_list:
+            change = ET.SubElement(changes, "Change")
+            ET.SubElement(change, "Action").text = action
+
+            rrs = ET.SubElement(change, "ResourceRecordSet")
+            ET.SubElement(rrs, "Name").text = name + "." + zone.domain
```
Kami, line 244: Line is too long and there are some other minor pep8 style issues in this file. Please run pep8 tool on this file and fix the issues :)

Kami:
@Jc2k Thanks! Added some more comments (was AFK yesterday and couldn't fully review the patch).

In general it's looking good. There are some more minor issues but once you address those it should be ready to be merged.

libcloud/dns/drivers/route53.py
Jc2k: No - there is a limit of 1 return value because of the maxitems=1 bit.

libcloud/dns/drivers/route53.py
Jc2k: AWS doesn't have a record id. So yes, this is the best we can do.

Jc2k:
I've dealt with most of the pep8 issues - a few remain in the mock due to the length of the mock names. I've made changes for any review suggests that i've not replied to.

Kami:
@Jc2k Looks good. 

Can you please a attach a patch of the latest version to JIRA ticket and then I will perform a final review and merge it.

Thanks.

PR #84, opened 2013-01-04
Support Amazon ELB

Jc2k:
This adds a new loadbalancer driver and tests.

One important thing to note is that AWS only lets you add members that are compute nodes - you can't add IP addresses directly.


libcloud/loadbalancer/drivers/elb.py
```diff
@@ -0,0 +1,214 @@
...
+
+
+class ELBResponse(AWSBaseResponse):
+    """
+    Amazon ELB response class.
+    """
+    def success(self):
+        return self.status in [httplib.OK, httplib.CREATED, httplib.ACCEPTED]
+
+    def parse_error(self):
```
Kami, line 51: Can we move this functionality into `AWSBaseResponse` class?

libcloud/loadbalancer/drivers/elb.py
```diff
@@ -0,0 +1,214 @@
...
+        status = int(self.status)
+
+        if status == 403:
+            if not self.body:
+                raise InvalidCredsError(str(self.status) + ': ' + self.error)
+            else:
+                raise InvalidCredsError(self.body)
+
+
+class ELBConnection(ConnectionUserAndKey):
```
Kami, line 61: It would be great if this can be refactored to `AWSBaseConnection` class and re-used with other drivers.

libcloud/loadbalancer/drivers/elb.py
```diff
@@ -0,0 +1,214 @@
...
+    name = 'ELB'
+    website = 'http://aws.amazon.com/elasticloadbalancing/'
+    connectionCls = ELBConnection
+
+    def list_protocols(self):
+        return ["tcp", "ssl", "http", "https"]
+
+    def list_balancers(self):
+        params = {
+            "Action": "DescribeLoadBalancers",
```
Kami: Minor style thing - please be consistent and use single quotes around strings.

libcloud/loadbalancer/drivers/elb.py
```diff
@@ -0,0 +1,214 @@
...
+
+    def create_balancer(self, name, port, protocol, algorithm, members):
+        params = {
+            "Action": "CreateLoadBalancer",
+            "LoadBalancerName": name,
+            "Listeners.member.1.InstancePort": str(port),
+            "Listeners.member.1.InstanceProtocol": protocol.upper(),
+            "Listeners.member.1.LoadBalancerPort": str(port),
+            "Listeners.member.1.Protocol": protocol.upper(),
+            "AvailabilityZones.member.1": "eu-west-1a",
```
Kami: Is it required to specify an availability zone for the members?

If it is, it should be an extension argument  (e.g. `ex_members_availability_zone`)   which defaults to this zone.

Kami:
I think there are also some pep8 issues in this file so please fix those.

Otherwise (besides the comments I have posted), the patch looks good and should be ready to be merged as soon as you address those comments.

Thanks!

Jc2k:
I've done the PEP8 changes, the string changes, and the ex_members_availabilty_zone thing.

Can we do the connection/response refactoring after merging both my AWS patches as it impacts both of them?

Jc2k:
Here's a gist with what I think we can make common:

https://gist.github.com/4468065

But i'm reluctant to put it in the same patch as the add-ELB patch.

Kami:
I'm fine with doing this refactoring in a separate patch / pr after this has been merged.

Also, sorry for the delay, I'm pretty busy, but I do plan to review and hopefully merge this patch at some point today.

Kami:
@Jc2k Patch mostly looks good.

Can you please attach final version to the JIRA ticket so I can do final review and merge it? Thanks.

PR #85, opened 2013-01-07
LIBCLOUD-169: Refactor AWS connection and response to reduce code duplication

Jc2k:
As promised in the elb pr, here is a follow up patch to reduce code duplication. I haven't run pep8 over it yet, but the tests pass. I'm AFK for a few days, but wanted to at least get something up here.

I think this error parsing can be applied to the EC2 backend too but wanted to get some feedback on the general approach first.

### DIFF
--- a/libcloud/common/aws.py
+++ b/libcloud/common/aws.py
@@ -17,14 +17,76 @@ import base64
 import hmac
 import time
 from hashlib import sha256
+from xml.etree import ElementTree as ET
 
-from libcloud.utils.py3 import b, urlquote
 from libcloud.common.base import ConnectionUserAndKey, XmlResponse
+from libcloud.common.types import InvalidCredsError, MalformedResponseError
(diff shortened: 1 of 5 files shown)


Kami:
General approach looks good to me.

Kami:
It might take a while to get this fully reviewed and merged, because I also plan to test everything manually.

Currently we don't have a complete test coverage for all the authentication stuff so a problem could easily slip through and cause a regression.

One problem I see immediately is `version` attribute not being set in the Route53 connection class which would cause authentication to fail.

Jc2k:
Good idea.

That specific example - Route53 doesn't use SignedAWSConnection (yay for consistent authentication API's amazon!!) so doesn't need `self.version`.

Kami:
@Jc2k Also, I'm already on it so no need for you to update the PR at the moment.

Kami:
@Jc2k I've tested it, made some changes and merged patch into master. Thanks!

PR #86, opened 2013-01-09
Fix broken DummyDNSDriver.list_records

Jc2k:
Also adds test coverage for the method in question.

### DIFF
--- a/libcloud/dns/drivers/dummy.py
+++ b/libcloud/dns/drivers/dummy.py
@@ -67,7 +67,14 @@ class DummyDNSDriver(DNSDriver):
         return [zone['zone'] for zone in list(self._zones.values())]
 
     def list_records(self, zone):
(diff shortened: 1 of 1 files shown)


libcloud/dns/drivers/dummy.py
```diff
@@ -67,7 +67,14 @@ def list_zones(self):
         return [zone['zone'] for zone in list(self._zones.values())]
 
     def list_records(self, zone):
-        return self.get_zone(zone_id=zone.id).records
+        """
```
Kami, line 70: Thanks.

While we are at it we should also add test cases for other cases - e.g. if there are some records.

Kami:
I've improved your change (fixed it so it works with 3.2, added another test case) and merged patch into trunk - https://svn.apache.org/viewvc?view=revision&revision=1431200. Thanks.

PR #87, opened 2013-01-12
[#LIBCLOUD-281] Support Gandi DNS

Jc2k:
This pull request adds a new DNS Driver for Gandi. It includes tests. It is pep8 compliant and also brings `libcloud/common/gandi.py` into pep8 compliance.

Note that whilst Gandi's API does expose a record id field it changes for each version of the zone (effectively each 'transaction' renumbers all record id's that libcloud sees). So we use the same approach as Route53 and use type:name as the record id.


libcloud/common/gandi.py
```diff
@@ -93,7 +93,10 @@ def request(self, method, *args):
             return getattr(self._proxy, method)(self.key, *args)
         except xmlrpclib.Fault:
             e = sys.exc_info()[1]
-            raise GandiException(1001, e)
+            self.parse_error(e.faultCode, e.faultString)
+
+    def parse_error(self, code, message):
```
Kami, line 99: Hm, you pass in `code`, but you never use it.

Method name is also slightly misleading to me - it actually throws an exception and doesn't do any parsing.

libcloud/dns/drivers/gandi.py
```diff
@@ -0,0 +1,249 @@
...
+from libcloud.dns.types import RecordError
+from libcloud.dns.types import ZoneDoesNotExistError, RecordDoesNotExistError
+from libcloud.dns.base import DNSDriver, Zone, Record
+
+
+TTL_MIN = 30
+TTL_MAX = 2592000  # 30 days
+
+
+class NewZoneVersion(object):
```
Kami, line 31: Would be nice to add a short docstring which describes what this context manager does.

libcloud/dns/drivers/gandi.py
```diff
@@ -0,0 +1,249 @@
...
+        zid = int(self.zone.id)
+        self.connection.set_context({'zone_id': self.zone.id})
+        vid = self.connection.request("domain.zone.version.new", zid)
+        self.vid = vid
+        return vid
+
+    def __exit__(self, type, value, traceback):
+        if not traceback:
+            zid = int(self.zone.id)
+            c = self.connection
```
Kami: Minor code style thing: We have an (atm) unwritten rule about the shortest variable name being 3 characters long

libcloud/dns/drivers/gandi.py
```diff
@@ -0,0 +1,249 @@
...
+            c.set_context({'zone_id': self.zone.id})
+            c.request("domain.zone.version.set", zid, self.vid)
+
+
+class GandiDNSConnection(GandiConnection):
+
+    def parse_error(self, code, message):
+        if code == 581042:
```
Kami, line 67: Doesn't have en else, is this expected? Or should we throw an unexpected code error if we don't recognize the code?

libcloud/dns/drivers/gandi.py
```diff
@@ -0,0 +1,249 @@
...
+            zone_id = str(self.context.get('zone_id', None))
+            raise ZoneDoesNotExistError(value='', driver=self.driver,
+                                        zone_id=zone_id)
+
+
+class GandiDNSDriver(BaseGandiDriver, DNSDriver):
+    type = Provider.GANDI
+    name = 'Gandi DNS'
+    website = 'http://doc.rpc.gandi.net/domain/reference.html'
```
Kami: `website` attribute should point to the product page, not the docs.

Feel free to include link to the docs at the module level docstring though.

libcloud/dns/types.py
```diff
@@ -57,7 +58,9 @@ class RecordType(object):
     REDIRECT = 13
     GEO = 14
     URL = 15
-
+    WKS = 16
```
Kami, line 61: Heh, I actually had to google what `WKS` record type is since I never heard about it before http://tools.ietf.org/html/rfc1123#page-55.

It sounds like it's a deprecated record type, does it make sense to still support it here?

libcloud/test/dns/test_gandi.py
```diff
@@ -0,0 +1,287 @@
...
+    def _xmlrpc__domain_zone_version_set_RECORD_DOES_NOT_EXIST(self, method, url, body, headers):
+        body = self.fixtures.load('new_version.xml')
+        return (httplib.OK, body, {}, httplib.responses[httplib.OK])
+
+
+class DummyTransport(MockGandiTransport):
+    mockCls = GandiMockHttp
+
+
+class GandiTests(BaseGandiTests):
```
Kami, line 114: Minor style thing - common convention is to move test class at the top and MockHttp class at the bottom.

libcloud/dns/types.py
Jc2k, line 61: It might exist in a Gandi zone as they are exposed by its API. It seems a shame for libcloud to not work on that zone for the sake of a single enum.

Kami, line 61: Yeah, I'm in this case I'm fine with supporting it. Supporting it doesn't really make other code any harder to maintain for us.

libcloud/test/dns/test_gandi.py
Jc2k, line 114: I thought you might say that :)

GandiTests needs to point at a DummyTransport with a class attribute, which needs to point at a mock. Class attributes need to point at things that are defined, so this would explode.

I can do this other ways but they are ugly. My next patch for the 2 gandi drivers is to use the xmlrpc parsing but atop the libcloud HTTP stack. This will fix the some other issues we know the Gandi drivers have but also make the tests a bit more like the other tests, and I can fix this then. Or I can find some other way of setting the transport class up now?

libcloud/dns/drivers/gandi.py
Jc2k, line 67: The base class calls parse_error and is meant to throw an exception if the subclass doesn't throw a more specific one. This emulated the behaviour of parse_error in the normal connection/response objects.

It looks like I didn't get that right in the base class though!

libcloud/common/gandi.py
Jc2k, line 99: Intention is for it to look like a normal Response class. It should be subclassed (and is, in the DNS case). The `code` attribute is used in the DNS driver. I was mightily annoyed to raise GandiException 1001, e but thats what the existing compute driver expects :(

Kami, line 99: Ah, I see...

Would be nice to at least add a docstring which explains this then.

libcloud/dns/drivers/gandi.py
Kami, line 67: Yeah, I'm still slightly confused though because you never actually call a parent class method and you override it so the parent method will never end up being called in this case.

Jc2k, line 67: Yeah, what i should have done is put the raise GandiExeption in the method that called parse_error, like is done in a normal response object :-(

Jc2k:
I've added the comments, nerfed the 1 character variables, and the parse_error thing does as intended.

I also noticed there was an unused import and removed it.

I'm started to investigate a follow up piece of work which will as a consequence fix the test/mock order, but in the mean time this is fully functional, the tests are passing (including an out-of-tree check i run against a real account) as is pep8 :-)

Kami:
@Jc2K Looks good.

Please attach a final patch to the jira ticket and I will perform a final review and merge it into trunk.

Jc2k:
Merged.

PR #91, opened 2013-01-29
LIBCLOUD-80: Windows Azure Storage driver

mahendra:
Implement storage driver for Windows Azure Blob Storage.

http://msdn.microsoft.com/en-us/library/windowsazure/dd135733.aspx

The driver supports the following
- Azure Block Blobs support (default)
- Azure Page Blobs (ex_page_type option)
- Support for leasing blobs while being updated (ex_use_lease option)


libcloud/common/azure.py
```diff
@@ -0,0 +1,183 @@
...
+            # Some APIs respond with an XML error. Others just dump HTML
+            body = self.parse_body()
+
+            if type(body) == ElementTree.Element:
+                code = body.findtext(fixxpath(xpath='Code'))
+                message = body.findtext(fixxpath(xpath='Message'))
+                message = message.split('\n')[0]
+                error_msg = '%s: %s' % (code, message)
+
+        except MalformedResponseError:
```
Kami, line 64: This could potentially also throw a different exception if the returned XML is messed up (e.g. can't access [0] index), right?

I know this shouldn't happen, but In Libcloud we deal a lot with stuff which is not documented and shouldn't happen, but it actually happens in real life so we try to program as defensively as possible :)

libcloud/common/azure.py
```diff
@@ -0,0 +1,183 @@
...
+        headers['x-ms-date'] = time.strftime(AZURE_TIME_FORMAT, time.gmtime())
+        headers['x-ms-version'] = API_VERSION
+
+        # Add the authorization header
+        headers['Authorization'] = self._get_azure_auth_signature(
+            method=self.method, headers=headers, params=params,
+            account=self.user_id, secret_key=self.key, path=self.action)
+
+        # Azure cribs about this in 'raw' connections
+        headers.pop('Host', None)
```
Kami, line 105: Usually in cases like this I prefer copying the headers before modifying it or at least documenting it in the docstring that `headers` gets mutated in a way that `Host` gets removed if it's present.

libcloud/storage/base.py
```diff
@@ -134,13 +134,13 @@ def get_object(self, object_name):
         return self.driver.get_object(container_name=self.name,
                                       object_name=object_name)
 
-    def upload_object(self, file_path, object_name, extra=None):
```
Kami, line 137: Any reason you replaced `extra` with `kwargs`? Is it at least possible to keep extra and ad kwargs after it?

We try to avoid using _args or *_kwargs if it's not really necessary, because it makes parsing the code (among many other things) harder. This affects projects such as Libcloud.REST for example (cc: @islamgulov).

libcloud/storage/base.py
```diff
@@ -615,7 +615,7 @@ def _upload_object(self, object_name, content_type, upload_func,
             file_size = os.path.getsize(file_path)
             upload_func_kwargs['chunked'] = False
 
-        if file_size is not None:
+        if file_size is not None and 'Content-Length' not in headers:
```
Kami, line 619: Yeah, probably a good check :)

Would be great to add a test case for it if you haven't already.

libcloud/storage/drivers/azure_blobs.py
```diff
@@ -0,0 +1,996 @@
...
+RESPONSES_PER_REQUEST = 100
+
+# Pattern for validating the azure container name
+# Refer : http://msdn.microsoft.com/en-us/library/windowsazure/dd135715.aspx
+# * Must start with [a-z][0-9]
+# * 3-63 characters
+# * Only [a-z][0-9][-] - No upper case allowed
+# * No consecutive --
+# * '$root' is a valid container name
+AZURE_CONTAINER = '^(([a-z\\d]((-(?=[a-z\\d]))|([a-z\\d])){2,62})|(\\$root))$'
```
Kami: Can we delegate this check to the API? 

We usually do this in other places and don't do a whole lot of validation on the client side. It means we don't need to update the code if the business rules change and APIs usually already performs a validation like this on the server side and returns a semi-decent error message.

libcloud/storage/drivers/azure_blobs.py
```diff
@@ -0,0 +1,996 @@
...
+
+# The time period (in seconds) for which a lease must be obtained.
+# If set as -1, we get an infinite lease, but that is a bad idea. If
+# after getting an infinite lease, there was an issue in releasing the
+# lease, the object will remain 'locked' forever, unless the lease is
+# released using the lease_id (which is not exposed to the user)
+AZURE_LEASE_PERIOD = 60
+
+
+class AzureBlobLease(object):
```
Kami, line 77: Just thinking out loudly - IIRC context managers and `with` are supported in all the Python versions Libcloud supports, right?

libcloud/storage/drivers/azure_blobs.py
```diff
@@ -0,0 +1,996 @@
...
+            chunk_hash = base64.b64encode(chunk_hash.digest())
+
+            headers['Content-MD5'] = chunk_hash
+            headers['Content-Length'] = content_length
+
+            if blob_type == 'BlockBlob':
+                # Block id can be any unique string that is base64 encoded
+                # A 10 digit number can hold the max value of 50000 blocks
+                # that are allowed for azure
+                block_id = base64.b64encode('%10d' % (count))
```
Kami: IIRC, you need to decode this (.decode('utf-8')) otherwise you will get bytes back in Python 3.

Kami:
Did a quick first pass at reviewing it.

Besides some of the minor things I've mentioned it looks solid overall.

libcloud/storage/drivers/azure_blobs.py
Jc2k: I quite like this. It'd be quite sweet if there was a dry run mode that validated the arguments for an action without doing it or if there was a validate function one could call.

I actually do wrap the libcloud API with a layer that can do dry-runs in one of my tools.

If you were a defensive coder consuming libcloud i guess what you might like to do is first of all inspect the current state of your cloud app and work out what actions you want to perform. Then validate your inputs as much as you can. Then actually try and make the changes. Obviously you still need to code for the worst case, but it's a lot nicer if you can detect the faulty container name before you start making lots of changes to other parts of a system.

A half way house might be some standard way of collecting and organising information like the AZURE_CONTAINER regex within libcloud but don't enforce it? Some sort of .validators dictionary on the driver?

libcloud/storage/drivers/azure_blobs.py
Jc2k, line 77: They were added in python 2.5 :)

libcloud/storage/drivers/azure_blobs.py
Kami: Yeah, my main problem with pushing more validation to the client side is that API's change. In theory most of the APIs are versioned and things like this shouldn't happen, but in real-life they do happen quite a lot.

Pushing validation to the client side means we can detect a potential problem and throw sooner which is great, but it also means we need to update a driver when the API changes which is not all that great because we currently support many different providers.

A standard way of defining this validation rules would be an interesting thing to explore. If no no other reason it could be useful from the parsing / documentation perspective.

libcloud/common/azure.py
mahendra, line 64: Yeah, weird stuff can happen. Specially with Microsoft stuff. I have seen very inconsistent behaviour with Azure!

However, if the response is XML, message will be a valid string and split('\n')[0] will work even if there is no '\n' in the string.

We can catch a generic exception while parsing

libcloud/common/azure.py
mahendra, line 105: Agree. Will do a copy.deepcopy() of headers before modifying it.

libcloud/storage/base.py
mahendra, line 137: Agree with your point. Keeping extra and adding kwargs after it. Adding kwargs was needed because without it it was not possible to pass the `ex_**` options while doing `container.upload_object()`. We had to resort to `driver.upload_object()` for doing it.

libcloud/storage/drivers/azure_blobs.py
mahendra: OK. I will remove it. I agree with @Kami - I added it to avoid an extra request to the Azure service (which is slower than other cloud providers).

libcloud/storage/drivers/azure_blobs.py
mahendra: yeah! actually I have missed the b() thingy in quite a few places. I forgot to run the tests for python 3. Fixed that now.

Kami: Yeah, supporting all those versions is pretty big pita, especially 2.5 stuff. I plan to start a new thread on the mailing list soon about dropping support for Python 2.5 in the future releases.

mahendra:
Have updated the patch with all comments.

Also tested it thoroughly with python 2.7 and python 3.2

libcloud/storage/drivers/azure_blobs.py
mahendra: Hi,

I updated the diffs after taking care of the review comments. Do have a
look when you have time.

Regards,
Mahendra

On Wed, Jan 30, 2013 at 11:52 AM, Tomaz Muraus notifications@github.comwrote:

> In libcloud/storage/drivers/azure_blobs.py:
> 
> > +
> > -            if calculate_hash:
> > -                data_hash.update(data)
> >   +
> > -            chunk_hash = self._get_hash_function()
> > -            chunk_hash.update(data)
> > -            chunk_hash = base64.b64encode(chunk_hash.digest())
> >   +
> > -            headers['Content-MD5'] = chunk_hash
> > -            headers['Content-Length'] = content_length
> >   +
> > -            if blob_type == 'BlockBlob':
> > -                # Block id can be any unique string that is base64 encoded
> > -                # A 10 digit number can hold the max value of 50000 blocks
> > -                # that are allowed for azure
> > -                block_id = base64.b64encode('%10d' % (count))
> 
> Yeah, supporting all those versions is pretty big pita, especially 2.5
> stuff. I plan to start a new thread on the mailing list soon about dropping
> support for Python 2.5 in the future releases.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/apache/libcloud/pull/91/files#r2821968.

## 

Mahendra

http://twitter.com/mahendra

Kami:
Yay, finally got to this PR.

Can you please attach the latest patch to JIRA ticket? Thanks.

mahendra:
@Kami - done

Kami:
This patch has been merged. Can you please close the PR now?

Thanks

mahendra:
Fixed by https://issues.apache.org/jira/browse/LIBCLOUD-80

PR #92, opened 2013-01-31
LIBCLOUD-288 Gandi on libcloud http

Jc2k:
This makes the Gandi xmlrpclib driver work much more like a normal driver, and it makes full use of the libcloud http features.

### DIFF
--- a/libcloud/common/gandi.py
+++ b/libcloud/common/gandi.py
@@ -23,11 +23,10 @@ import sys
 from libcloud.utils.py3 import xmlrpclib
 from libcloud.utils.py3 import b
 
-from libcloud.common.base import ConnectionKey
+from libcloud.common.base import Response, ConnectionKey
 
 # Global constants
 
-API_URL = "https://rpc.gandi.net/xmlrpc/"
 
 DEFAULT_TIMEOUT = 600   # operation pooling max seconds
 DEFAULT_INTERVAL = 20   # seconds between 2 operation.info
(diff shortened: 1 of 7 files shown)


libcloud/common/gandi.py
```diff
@@ -104,6 +67,27 @@ def parse_error(self, code, message):
         pass
 
 
+class GandiConnection(ConnectionKey):
+    """
+    Connection class for the Gandi driver
+    """
+
+    responseCls = GandiResponse
+    host = "rpc.gandi.net"
```
Kami: Minor style thing - please use single quotes around the strings everywhere besides the docstrings :)

Kami:
Just curious, do the tests pass with Python 3.2?

I know there were some changes to `xmlrpclib` in 3.2 which are usually causing troubles.

libcloud/common/gandi.py
Jc2k: Gah, I will get used to this. I will get used to this. I will get use to this.

It's for my own good, everything other bit of code I write must randomly hop between the two styles :-)

Kami: heh :)

Jc2k:
Good call.

The tests do seem to pass on 3.2, as does my out-of-tree script that goes and pokes around uncomfortably close to my production DNS :D

But i'll pay extra attention for 3.2 related issues.

Kami:
Looks good to me then. Feel free to attach patch to the JIRA ticket when you are done with addressing all the comments.

Jc2k:
Patch attached to JIRA, closing PR.

PR #98, opened 2013-03-11
Starting and stopping cloudstack nodes

xyzjace:
Added the ability to start and stop cloudstack nodes with synchronous requests, and by extension Ninefold nodes.

### DIFF
--- a/libcloud/compute/drivers/cloudstack.py
+++ b/libcloud/compute/drivers/cloudstack.py
@@ -275,6 +275,41 @@ class CloudStackNodeDriver(CloudStackDriverMixIn, NodeDriver):
         self._async_request('rebootVirtualMachine', id=node.id)
         return True
 
+
+    def start_node(self, node):
+        """
+        @type node: L{CloudStackNode}
+        """
+        if type(node) is not CloudStackNode:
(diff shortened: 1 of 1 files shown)


libcloud/compute/drivers/cloudstack.py
```diff
@@ -275,6 +275,37 @@ def reboot_node(self, node):
...
 
+
+    def start_node(self, node):
+        """
+        @type node: L{CloudStackNode}
+        """
+        if type(node) is not CloudStackNode:
+            raise LibcloudError("Must pass a valid node")
+
+        available_nodes = self.list_nodes();
```
Kami, line 286: This method can take `node_id` instead of `node` argument. This way you can avoid (expensive) request for retrieving a list of all the nodes and doing client side filtering.

Kami:
This pull requests is also missing tests - please have a look at http://libcloud.apache.org/contributing.html.

If you need help with tests look at the existing examples and let us know.

Kami:
When all the comments have been address you should open an issue and attach a final patch on our JIRA - https://issues.apache.org/jira/browse/LIBCLOUD

Thanks

PR #99, opened 2013-03-13
Implement list_locations() in new Rackspace provider

cpsaltis:
this solves https://issues.apache.org/jira/browse/LIBCLOUD-305


cpsaltis:
The way this works is about to be refactored so no need to add it at this point.

Check out the discussion at https://issues.apache.org/jira/browse/LIBCLOUD-305

PR #100, opened 2013-04-12
Trunk

juancarlosm:
Replace common/linode.py constant LINODE_PLAN_IDS by updated plan ids get function. Currently there are no plan 512 or 768, and plan id 3 is returns 1024 plan, but now is 2048

### DIFF
--- a/libcloud/common/linode.py
+++ b/libcloud/common/linode.py
@@ -31,19 +31,6 @@ __all__ = [
 API_HOST = 'api.linode.com'
 API_ROOT = '/'
 
-# Constants that map a RAM figure to a PlanID (updated 6/28/10)
-LINODE_PLAN_IDS = {512: '1',
-                   768: '2',
-                  1024: '3',
-                  1536: '4',
-                  2048: '5',
-                  4096: '6',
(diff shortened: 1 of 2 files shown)


Kami:
This patch seems reasonable to me. What do other people think - @jedsmith.

@JuanCarlosM Can you please open a ticket at https://issues.apache.org/jira/browse/LIBCLOUD and attach a final patch there?

Thanks

juancarlosm:
yes, of course. thanks

jedsmith:
I think it would be better if it remained static, as you're now incurring an (agreed, minimal) overhead for what ultimately is static data; for perspective, that data hasn't changed since I wrote it in 2010, and I doubt it will change again any time soon. If the overhead is genuinely minimal, I guess it's fine, but I think you're adding a bit of startup delay to the driver unnecessarily.

I had this on my list for this weekend, so thanks for getting to it, either way!

Kami:
@jedsmith Thanks for looking into this.

Kami:

@JuanCarlosM - Those changes are not relevant anymore, can you please close this pull request?

Thanks!

PR #105, opened 2013-05-07
Add Homebrew support to CA_CERTS_PATH

robert-chiniquy:
Same patch I added to [LIBCLOUD-324](https://issues.apache.org/jira/browse/LIBCLOUD-324), just offering  you a PR in case that's more convenient.


Kami:
Don't watch, don't comment.

robert-chiniquy:
Heh.

PR #108, opened 2013-06-04
Change create_loadBalancer add multi port and protocol

Gestator:
Add create_healthcheck

### DIFF
--- a/libcloud/loadbalancer/drivers/elb.py
+++ b/libcloud/loadbalancer/drivers/elb.py
@@ -74,8 +74,8 @@ class ElasticLBDriver(Driver):
             'Listeners.member.1.InstancePort': str(port),
             'Listeners.member.1.InstanceProtocol': protocol.upper(),
             'Listeners.member.1.LoadBalancerPort': str(port),
-            'Listeners.member.1.Protocol': protocol.upper(),
-        }
+            'Listeners.member.1.Protocol': protocol.upper()
+            }
 
         for i, z in enumerate(ex_members_availability_zones):
             zone = ''.join((self.region, z))
@@ -93,7 +93,7 @@ class ElasticLBDriver(Driver):
         )
         balancer._members = []
         return balancer
-
+   
     def destroy_balancer(self, balancer):
         params = {
             'Action': 'DeleteLoadBalancer',
--- a/libcloud/loadbalancer/drivers/elb.py
+++ b/libcloud/loadbalancer/drivers/elb.py
@@ -63,19 +63,24 @@ class ElasticLBDriver(Driver):
         data = self.connection.request(ROOT, params=params).object
         return self._to_balancers(data)
 
-    def create_balancer(self, name, port, protocol, algorithm, members,
+    def create_balancer(self, name, ports, protocols, algorithm, members,
                         ex_members_availability_zones=None):
         if ex_members_availability_zones is None:
             ex_members_availability_zones = ['a']
+        if not (len(ports) == len(protocols)):
+            raise SizeFit('ports and protocols not same size')
 
         params = {
             'Action': 'CreateLoadBalancer',
-            'LoadBalancerName': name,
-            'Listeners.member.1.InstancePort': str(port),
-            'Listeners.member.1.InstanceProtocol': protocol.upper(),
-            'Listeners.member.1.LoadBalancerPort': str(port),
-            'Listeners.member.1.Protocol': protocol.upper()
+            'LoadBalancerName': name
             }
+        iter = 1
+        for port in ports:
+            params['Listeners.member.{0}.InstancePort'.format(str(iter))] = str(port)
+            params['Listeners.member.{0}.InstanceProtocol'.format(str(iter))] = protocols[iter-1].upper()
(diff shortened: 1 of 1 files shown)


Kami:
I don't see any actual changes besides the whitespace stuff.

When you have actual changes, please attach patch with your changes to your JIRA ticket - https://issues.apache.org/jira/browse/LIBCLOUD-336

Gestator:
non i don't commit good elb this new commit is ok

libcloud/loadbalancer/drivers/elb.py
```diff
@@ -63,19 +63,24 @@ def list_balancers(self):
         data = self.connection.request(ROOT, params=params).object
         return self._to_balancers(data)
 
-    def create_balancer(self, name, port, protocol, algorithm, members,
+    def create_balancer(self, name, ports, protocols, algorithm, members,
```
Jc2k: This method is meant to be compatible with the other LoadBalancer drivers. This changes the signature of create_balancer method and makes it incompatible with the other load balancer drivers. Libcloud aims to share the same API between different cloud providers, and changes like this need to consider the other LoadBalancer types.

libcloud/loadbalancer/drivers/elb.py
```diff
@@ -88,12 +93,30 @@ def create_balancer(self, name, port, protocol, algorithm, members,
...
             state=State.PENDING,
             ip=findtext(element=data, xpath='DNSName', namespace=NS),
-            port=port,
+            port=ports,
             driver=self.connection.driver
         )
         balancer._members = []
         return balancer
+    
+    def create_HealthCheck(self, name, port, protocol, interval = 30,timeout = 5, unhealthyThreshold = 2,
```
Jc2k: create_health_check would be a more PEP8 name. In `def` declarations we do have spaces between arguments, the `=` and the default value.

libcloud/loadbalancer/drivers/elb.py
```diff
@@ -88,12 +93,30 @@ def create_balancer(self, name, port, protocol, algorithm, members,
...
+                            healthyThreshold = 10, path = ''):
+        if path is None:
+            path = ['']
```
Jc2k: The default value is `''` - a string, but in this case it is a list. Should path be a string or a list, or can it really be either?

libcloud/loadbalancer/drivers/elb.py
```diff
@@ -88,12 +93,30 @@ def create_balancer(self, name, port, protocol, algorithm, members,
...
+        params = {
+            'Action': 'ConfigureHealthCheck',
+            'LoadBalancerName': name,
+            'HealthCheck.HealthyThreshold': str(healthyThreshold),
+            'HealthCheck.UnhealthyThreshold': str(unhealthyThreshold),
+            'HealthCheck.Interval': str(interval),
+            'HealthCheck.Timeout': str(timeout),
+            'HealthCheck.Target': '{0}:{1}/{2}'.format(protocol,str(port),path)
+            }
+        healtCheck = self.connection.request(ROOT, params=params).object
```
Jc2k: `health_check` would be a more PEP8 name for this.

libcloud/loadbalancer/drivers/elb.py
```diff
@@ -63,19 +63,24 @@ def list_balancers(self):
...
-            'LoadBalancerName': name,
-            'Listeners.member.1.InstancePort': str(port),
-            'Listeners.member.1.InstanceProtocol': protocol.upper(),
-            'Listeners.member.1.LoadBalancerPort': str(port),
-            'Listeners.member.1.Protocol': protocol.upper(),
-        }
+            'LoadBalancerName': name
+            }
+        iter = 1
+        for port in ports:
```
Jc2k: You could use `enumerate` here instead of the `iter` variable.

```
for i, port in enumerate(ports, 1):
    params['Listeners.member.{0}.InstancePort'.format(str(i))] = str(port)
```

In libcloud the convention is to use `%` substitution rather than `.format` substitution, so this would be even better:

```
for i, port in enumerate(ports, 1):
    params['Listeners.member.%d.InstancePort' % i] = str(port)
```

libcloud/loadbalancer/drivers/elb.py
```diff
@@ -88,12 +93,30 @@ def create_balancer(self, name, port, protocol, algorithm, members,
             name=name,
             state=State.PENDING,
             ip=findtext(element=data, xpath='DNSName', namespace=NS),
-            port=port,
+            port=ports,
```
Jc2k: Exposing the list of ports here is inconsistent with the other LoadBalancers, and it is also inconsistent with the `get_balancer` and `list_balancers` methods of this driver where it won't currently be a list, if i'm reading the code right?

Gestator: Wath are your issus for support ssl,tcp and http in your loadbalancer

Gestator:
The base interface don't work if you can't add function after loadbalancer creation to add more port and protocol 
it's imposible for ...all to push http and not https or tcp

Gestator:
CreateLoadBalancerListeners this function exist I work for implement that but in the createLoadbalancer you are parametre(members)
 but is not implemented in ELB

PR #112, opened 2013-06-19
LIBCLOUD-344: vCloud: ex_query() should return result parameters (total, page, page_size, etc.), add more metadata to Vdc object

galet:


libcloud/compute/drivers/vcloud.py
```diff
@@ -1290,14 +1315,19 @@ def ex_query(self, type, filter=None, page=1, page_size=100, sort_asc=None,
...
+        records = []
         res = self.connection.request(url)
         for elem in res.object:
             if not elem.tag.endswith('Link'):
                 result = elem.attrib
                 result['type'] = elem.tag.split('}')[1]
-                results.append(result)
-        return results
+                records.append(result)
+        return QueryResult(id=res.object.get('href'),
```
Kami: Ideally, this method should transparently handle the pagination in the background and return a generator.

Here are a couple of examples of how we do this in other drivers:
- https://github.com/apache/libcloud/blob/trunk/libcloud/storage/base.py#L182
- https://github.com/apache/libcloud/blob/trunk/libcloud/storage/drivers/azure_blobs.py#L359
- https://github.com/apache/libcloud/blob/trunk/libcloud/dns/drivers/zerigo.py#L146

galet: Right. I will look into it. Thanks.

galet:
QueryResult class rewritten to return generator. Please review.

libcloud/test/compute/test_vcloud.py
```diff
@@ -255,11 +257,21 @@ def test_ex_power_off(self):
...
-        self.assertEqual(results[0]['type'], 'UserRecord')
-        self.assertEqual(results[0]['name'], 'jrambo')
-        self.assertEqual(results[0]['isLdapUser'], 'true')
+        result = self.driver.ex_query('user', filter='name==jrambo', page_size=3, sort_desc='startDate')
+
+        self.assertEqual(result.page_size, 3)
+        self.assertEqual(result.total, 5)
+        self.assertEqual(result.name, 'user')
+        self.assertEqual(len(result), 5)
+        items = [x for x in result]
```
Kami: You can just do `items = list(result)` to exhaust the generator.

libcloud/compute/drivers/vcloud.py
```diff
@@ -143,6 +145,73 @@ def __repr__(self):
...
+                            e.args[0].get('minorErrorCode') == 'BAD_REQUEST' and
+                            re.match('Invalid parameter page=\\d+ must be less or equal to \\d+', e.args[0].get('message'))):
+                        raise StopIteration
+                    else:
+                        raise
+
+    def _fetch_page(self):
+        self._records = []
+        self._page += 1
+        res = self.driver.connection.request('{0}&page={1}'.format(self.url, self._page))
```
Kami: `.format` won't work with all the versions of Python we support, please use old style string interpolation (%s).

libcloud/compute/drivers/vcloud.py
```diff
@@ -143,6 +145,73 @@ def __repr__(self):
                 % (self.type, self.name, self.access_level))
 
 
+class QueryResult(object):
+    """
+    Query result object. Is iterable with lazy loading.
+    """
+    def __init__(self, driver, type, filter, sort_asc, sort_desc, page_size):
```
Kami, line 170: Would be nice to add a docstring and explain what each argument means.

libcloud/compute/drivers/vcloud.py
```diff
@@ -143,6 +145,73 @@ def __repr__(self):
...
+                    self._fetch_page()
+                except Exception:
+                    e = sys.exc_info()[1]
+                    if (isinstance(e.args[0], _ElementInterface) and
+                            e.args[0].tag.endswith('Error') and
+                            e.args[0].get('minorErrorCode') == 'BAD_REQUEST' and
+                            re.match('Invalid parameter page=\\d+ must be less or equal to \\d+', e.args[0].get('message'))):
+                        raise StopIteration
+                    else:
+                        raise
```
Kami: You should raise original exception (`raise e`)

libcloud/compute/drivers/vcloud.py
```diff
@@ -143,6 +145,73 @@ def __repr__(self):
...
+    def iterate_records(self):
+        while self._records:
+            item = self._records.pop(0)
+            yield item
+            if not self._records:
+                try:
+                    self._fetch_page()
+                except Exception:
+                    e = sys.exc_info()[1]
+                    if (isinstance(e.args[0], _ElementInterface) and
```
Kami: This if looks pretty fragile. Not sure what's a better way to do it though.

libcloud/compute/drivers/vcloud.py
```diff
@@ -143,6 +145,73 @@ def __repr__(self):
...
+                            e.args[0].tag.endswith('Error') and
```
Kami, line 218: Slightly confused - is the `endswith` check necessary?

I would imagine `element.tag` attribute already contains just the tag name so doing just the `tag == 'Error'` check should be enough?

libcloud/compute/drivers/vcloud.py
```diff
@@ -143,6 +145,73 @@ def __repr__(self):
...
+            self.params['sortDesc'] = sort_desc
+
+        self.url = '/api/query?' + urlencode(self.params)
+        if filter:
+            if not filter.startswith('('):
+                filter = '(' + filter + ')'
+            self.url += '&filter=' + filter.replace(' ', '+')
+
+        self.id = self.url
+        self._page = 0
```
Kami, line 191: You probably also want to initialize `self.total` to `0`, right?

Kami:
@galet Thanks - I've added some comments :)

libcloud/test/compute/test_vcloud.py
galet: Of course. Thanks.

libcloud/compute/drivers/vcloud.py
galet: Sure, will fix it.

libcloud/compute/drivers/vcloud.py
galet, line 170: Will do.

libcloud/compute/drivers/vcloud.py
galet: Will do.

libcloud/compute/drivers/vcloud.py
galet: You are right. We can check the end by counting yielded items and comparing the number to the "total":

``` python
...
current_index = 0
while self._records:
    item = self._records.pop(0)
    yield item

    # Are we at the end?
    current_index += 1
    if current_index >= self.total:
        raise StopIteration
    if not self._records:
        try:
            self._fetch_page()
        except Exception:
            e = sys.exc_info()[1]
            if (len(e.args) > 0 and
...
```

libcloud/compute/drivers/vcloud.py
galet, line 218: Actually it is prefixed with a long namespace. The `element.tag` content is `'{http://www.vmware.com/vcloud/v1.5}Error'` therefore `endswith`.

libcloud/compute/drivers/vcloud.py
galet, line 191: `self.total` is set right after this line in the `_fetch_page` method. However setting this to `0` here won't harm.

galet:
Thanks Kami for your comments. I will push an update soon.

Just a question: It could be written also as an iterator. This would simplify use cases when you just need the first item.
Example:

``` python
res = ex_query(...)
first_record = res.iterate_records().next()
```

vs.

``` python
res = ex_query(...)
first_record = res.next()
```

Any reason to prefer generator?

Kami:
@galet Yeah, you can implement `next` method (and `__next__` since we also support Python 3). This way you can call `.next()` on it directly.

E.g. https://github.com/apache/libcloud/blob/trunk/libcloud/storage/drivers/cloudfiles.py#L856

galet:
I see. Should we support both ways?

If so, what should be the correct behavior for:

``` python
res = ex_query(...)
res.next()
res.iterate_records().next()
res.next()
```

?

Should the generator start always from the beginning?

PR #114, opened 2013-06-20
Ported the Rackspace DNS and its tests to the one-class approach.

alex:
This probably isn't merge ready:
- [ ] totally breaks backwards compatibility (I'll reintroduce the compatibility aliases once I verify this is the right approach)
- [ ] I'm not convinced this actually works (maybe I'm wrong though!)
- [ ] It could really use a thorough review

So let me know what you think and I'll add the backwards compat aliases (if needed) once everything else looks ok.


libcloud/dns/drivers/rackspace.py
```diff
@@ -119,18 +119,21 @@ def _construct_dns_endpoint_from_servers_endpoint(self, ep):
...
     website = 'http://www.rackspace.com/'
+    connectionCls = RackspaceDNSConnection
 
     def __init__(self, *args, **kwargs):
+        datacenter = kwargs.pop('datacenter', 'dfw')
+        if datacenter not in ['dfw', 'ord', 'lon']:
+            raise ValueError('Invalid datacenter: %s' % (datacenter))
+
+        if datacenter in ['dfw', 'ord']:
+            self.connectionCls.auth_url = AUTH_URL_US
```
Kami, line 132: I assume you looked at how this is done for compute in trunk.

Approach in trunk is not OK, but sadly I haven't had the time to fix it yet (sorry). `connectionCls` is a class and not an instance attribute which means it won't work correctly if you instantiate multiple drivers with a different region.

There are multiple ways to fix and improve this, one way is just using `_ex_connection_class_kwargs`. If we go with this approach we can also move this functionality into `OpenStackDriverMixim` class.

Kami, line 132: Just FYI, I've started fixing this in https://github.com/apache/libcloud/pull/116

This is still some nasty stuff going on (e.g. https://github.com/apache/libcloud/pull/116/files#L2R73), but overall it should still be an improvement.

alex, line 132: Ok cool, yes you're correct this was based on the other work in `trunk`.

PR #115, opened 2013-06-22
Initial commit of support for Google Compute Engine.

wrigri:
The unit tests are still in development and will be coming in a separate commit.

This currently works, and does not rely on any additional third-party packages, with one exception:  If you want to use the "Service Account" OAUTH authentication, you will need the PyCrypto package installed.  However, it fails gracefully if it's not there.

I'm particularly interested to see if you notice any places where I broke the established API.  I believe that I followed it as closely as possible, however.


wrigri:
Also, I should note that this has currently only been fully tested with Python 2.7.  It's likely that some minor adjustments will need to be made for Python 3.X (and maybe 2.5 if 2.5 support isn't officially dropped before this is merged in).

demos/secrets.py-dist
```diff
@@ -22,6 +22,10 @@ DREAMHOST_PARAMS = ('key',)
 EC2_PARAMS = ('access_id', 'secret')
 ECP_PARAMS = ('user_name', 'password')
 GANDI_PARAMS = ('user',)
+GCE_PARAMS = ('email_address', 'key') # Service Account Authentication
+#GCE_PARAMS = ('client_id', 'client_secret') # Installed App Authentication
+GCE_KEYWORD_PARAMS = ({'project': 'project_name', 'ssh_username': None,
```
Kami: Parenthesis are unnecessary around the dict which is defined over multiple lines.

libcloud/common/google.py
```diff
@@ -0,0 +1,466 @@
...
+        Return an updated token from a token request body.
+
+        @param  request_body: A dictionary of values to send in the body of the
+                              token request.
+        @type   request_body: C{dict}
+
+        @return:  A dictionary with updated token information
+        @rtype:   C{dict}
+        """
+        data = urllib.urlencode(request_body)
```
Kami: You need to use `urlencode` from `libcloud.utils.py3` (https://github.com/apache/libcloud/blob/trunk/libcloud/utils/py3.py), otherwise it won't work will all the supported versions

libcloud/common/google.py
```diff
@@ -0,0 +1,466 @@
...
+        """
+        auth_params = {'response_type': 'code',
+                       'client_id': self.user_id,
+                       'redirect_uri': self.redirect_uri,
+                       'scope': self.scope,
+                       'state': 'Libcloud Request'}
+        if self.login_hint:
+            auth_params['login_hint'] = self.login_hint
+
+        data = urllib.urlencode(auth_params)
```
Kami: Same here (use `urlencode` from `libcloud.utils.py3`.

libcloud/common/google.py
```diff
@@ -0,0 +1,466 @@
...
+                authentication.
+        @type   user_id: C{str}
+
+        @param  key: The RSA Key or path to file containing the key.
+        @type   key: C{str}
+        """
+        if SHA256 is None:
+            raise GoogleAuthError('PyCrypto library required for '
+                                  'Service Accout Authentication.')
+        keypath = os.path.expanduser(key)
```
Kami, line 275: I'm not a big fan of a single argument containing two different type of values (either a path or a string), but if we proceed with this approach I would rather check if `key` is a file file path like this:

``` python
is_file_path = os.path.exists(keypath) and os.path.isfile(keypath)
...
```

libcloud/common/google.py
```diff
@@ -0,0 +1,466 @@
...
+        try:
+            key = open(keypath).read()
```
Kami: Please use `with` context manager when opening and reading a file (e.g. https://github.com/apache/libcloud/blob/trunk/libcloud/compute/deployment.py#L188).

This way you can be sure file is always closed (even if the exception happens) and you don't need to manually wrap it in `try / except / finally` block and make sure you always close file in the `finally` block.

libcloud/common/google.py
```diff
@@ -0,0 +1,466 @@
...
+        """
+        Read credential file and return token information.
+
+        @return:  Token information dictionary, or None
+        @rtype:   C{dict} or C{None}
+        """
+        token_info = None
+        filename = os.path.realpath(os.path.expanduser(self.credential_file))
+        try:
+            f = open(filename)
```
Kami: Same here, please use `with`.

libcloud/compute/drivers/gce.py
```diff
@@ -0,0 +1,1854 @@
...
+        self.extra = extra
+        country = name.split('-')[0]
+        super(GCEZone, self).__init__(id=str(id), name=name, country=country,
+                                      driver=driver)
+
+    def _get_next_maint(self):
+        """
+        Returns the next Maintenance Window.
+
+        @return:  A dictionary containing maintenance window info
```
Kami, line 242: Would be nice to document which keys returned value contains because now just by looking at the docstring I have no idea what kind of dictionary this function returns.

libcloud/compute/drivers/gce.py
```diff
@@ -0,0 +1,1854 @@
...
+        now = datetime.datetime.utcnow()
+        next_begin = timestamp_to_datetime(next_window['beginTime'])
+        return next_begin - now
+
+    def _get_next_mw_duration(self):
+        """
+        Returns the duration of the next maintenance window.
+
+        @return:  Duration of next maintenance window
+        @rtype:   C{datetime.timedelta}
```
Kami, line 278: Cool, I like that you are using `timedelta` instead of seconds.

libcloud/compute/drivers/gce.py
```diff
@@ -0,0 +1,1854 @@
...
+                             on value of user_id.
+        @type     auth_type: C{str}
+        """
+        self.auth_type = auth_type
+        self.project = project
+        super(GCENodeDriver, self).__init__(user_id, key, **kwargs)
+        if project:
+            self.project = project
+        else:
+            print "Please specify the project in your Driver's constructor"
```
Kami: Please throw `ValueError` instead of exiting here.

libcloud/compute/drivers/gce.py
```diff
@@ -0,0 +1,1854 @@
...
+
+        @keyword  region: If True, find a region instead of a zone.
+        @keyword  region: C{bool}
+
+        @return:  Name of zone (or region) that the resource is in.
+        @rtype:   C{str}
+        """
+        request = '/aggregated/%s' % res_type
+        res_list = self.connection.request(request).object
+        for k, v in res_list['items'].items():
```
Kami, line 409: We could probably use multiple list comprehensions to make this slightly more readable and remove multiple levels of nested.

libcloud/compute/drivers/gce.py
```diff
@@ -0,0 +1,1854 @@
...
+            request = '/aggregated/addresses'
+        else:
+            request = '/regions/%s/addresses' % region
+
+        response = self.connection.request(request, method='GET').object
+
+        if 'items' in response:
+            # The aggregated result returns dictionaries for each region
+            if region is None:
+                for k, v in response['items'].items():
```
Kami: Same here (reducing nesting)

Kami:
I've glanced over the code and posted some comments in-line. There are some issues, but it shouldn't be too hard to address those:
- Making sure it works with all the supported Python versions (after you add the tests it should be easier to test that using `tox`)
- Some code readability can be improved by making the code more flat and reducing the levels of nesting (either by using multiple list comprehensions or by splitting code in multiple functions)

As far as the high level stuff goes - overall all the class relationships and driver look solid.

demos/secrets.py-dist
wrigri: Fixed.

libcloud/common/google.py
wrigri: Thanks for pointing that out.  I made the changes for urlencode and also for urlparse.

libcloud/common/google.py
wrigri: Fixed

libcloud/common/google.py
wrigri, line 275: OK, I updated the check.  I don't have a strong opinion about the best way to pass this key in, so I'm certainly open to suggestions about the best way to handle it.  The key file itself is a little less than 1KB in size, so copying and pasting it as a string it a little cumbersome.  I would assume that most people would want to read it out of a file.  Of course, we could leave it to the end user to read the file themselves and pass the string.  Anyway, like I said, I'm open to suggestions about the UI for this.

libcloud/common/google.py
wrigri: Done. (I prefer 'with' myself, but wasn't sure if you were using it with your 2.5 support.)

libcloud/common/google.py
wrigri: Done.

libcloud/compute/drivers/gce.py
wrigri, line 242: Done.

libcloud/compute/drivers/gce.py
wrigri: Done

libcloud/compute/drivers/gce.py
wrigri, line 409: OK, I cleaned things up a bit, replaced several for loops with list comprehensions and for the more complex ones like this one, got it down to a single for loop.

libcloud/compute/drivers/gce.py
wrigri: Done.

libcloud/common/google.py
Kami: Thanks.

`with` is also available in Python 2.5, you just need to add `from __future__ import with_statement` to the top of the file where you use `with` (e.g. https://github.com/apache/libcloud/blob/trunk/libcloud/test/file_fixtures.py#L17)

libcloud/common/google.py
```diff
@@ -0,0 +1,467 @@
...
+        # Adds some retry logic for the occasional
+        # "Connection Reset by peer" error.
+        retries = 4
+        tries = 0
+        while tries < (retries - 1):
+            try:
+                return super(GoogleBaseConnection, self).request(
+                    *args, **kwargs)
+            except socket.error as e:
+                if e.errno == 104:
```
Kami: Please use a constant - `errno.ECONNRESET`

libcloud/common/google.py
```diff
@@ -0,0 +1,467 @@
...
+        """
+        # Adds some retry logic for the occasional
+        # "Connection Reset by peer" error.
+        retries = 4
+        tries = 0
+        while tries < (retries - 1):
+            try:
+                return super(GoogleBaseConnection, self).request(
+                    *args, **kwargs)
+            except socket.error as e:
```
Kami: Need to use this way of extracting an exception - https://github.com/apache/libcloud/blob/trunk/libcloud/utils/misc.py#L68

Otherwise it won't work with all the supported Python versions.

libcloud/common/google.py
```diff
@@ -0,0 +1,467 @@
...
+                if e.errno == 104:
+                    tries = tries + 1
+                else:
+                    raise
```
Kami: Please raise the original exception - `raise e`

Kami:
@wrigri Spotted a couple of minor things, otherwise looks good.

Should be ready to be merged once you finish the tests.

wrigri:
OK, I've added unittests.  I've tested (unittests and integration test (gce_demo.py)) against Python 2.7 and Python 3.2.  I'll run tox on the whole thing tomorrow and make sure there aren't any other cross-version issues that need to be addressed.

Other than that, the code should all be PEP8 compliant (except for some long lines due to long method names in test_gce.py), and epydoc doesn't show any errors (except that it complains about 'inherits').

Kami:
@wrigri I've just checked the tests and they look good - thanks!

I still need to run them locally and make sure they pass with all the supported versions before +1'ing it though.

When the pull request is +1'ed, please attach final attach to the corresponding JIRA ticket.

Since this is a slightly larger patch, please also to file an ICLA ((https://libcloud.apache.org/contributing.html#contributing-bigger-changes)) if you haven't already.

wrigri:
I sent a signed ICLA to Apache this morning.  In addition, I made a minor change based on the recent change to add list_volumes to the API.  Let me know if you find any additional issues that need to be addressed.

Kami:
I've opened a PR which fixes some 2.5 compatibility issues - https://github.com/wrigri/libcloud/pull/1

Once those issues are addresses, please attach a final patch to the JIRA ticket. I will merge it as soon as I receive a notification about ICLA filling from the ASF secretary.

Kami:
This has been merged. You can close this pull request now.

Thanks!

Kami:
@wrigri Hey, I just noticed Google added support for load balancing*. Do you also plan to add support for it to libcloud?

It would be great if you could :)
- http://googlecloudplatform.blogspot.com/2013/08/google-cloud-platform-enhances-compute_7.html
- https://developers.google.com/compute/docs/load-balancing/

wrigri:
Yes, load balancing support is in progress and I should have a pull request out for that in the near future.  In addition, I have some misc enhancements and bug fixes in the compute engine driver coming shortly.

Kami:
@wrigri Awesome. :)

PR #119, opened 2013-06-28
Libcloud 353 implement storage volume methods for openstack

bernieke:
Implements storage volume methods for the openstack driver.

### DIFF
--- a/libcloud/compute/drivers/openstack.py
+++ b/libcloud/compute/drivers/openstack.py
@@ -34,10 +34,10 @@ from xml.etree import ElementTree as ET
 
 from libcloud.common.openstack import OpenStackBaseConnection
 from libcloud.common.openstack import OpenStackDriverMixin
-from libcloud.common.types import MalformedResponseError
+from libcloud.common.types import MalformedResponseError, ProviderError
 from libcloud.compute.types import NodeState, Provider
 from libcloud.compute.base import NodeSize, NodeImage
-from libcloud.compute.base import NodeDriver, Node, NodeLocation
+from libcloud.compute.base import NodeDriver, Node, NodeLocation, StorageVolume
 from libcloud.pricing import get_size_price
 from libcloud.common.base import Response
 from libcloud.utils.xml import findall
@@ -57,6 +57,10 @@ __all__ = [
 ATOM_NAMESPACE = "http://www.w3.org/2005/Atom"
 
 DEFAULT_API_VERSION = '1.1'
+
+
+class OpenStackException(ProviderError):
+    pass
 
 
 class OpenStackResponse(Response):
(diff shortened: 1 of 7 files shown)


Kami:
Thanks.

I quickly glanced over it and this patch looks good and fits into the current model.

I still need to do a proper review, but I plan to merge it if there aren't any objections after I perform the review.

@alex is currently also looking into improving block storage support in Libcloud so we should also figure out how this fits into his stuff.

alex:
I think this fits in well with the model we're discussing now on the list, so I'll try to give this a review.

Kami:
@alex Sweet, thanks

libcloud/compute/drivers/openstack.py
```diff
@@ -206,6 +206,55 @@ def list_nodes(self):
         return self._to_nodes(
             self.connection.request('/servers/detail').object)
 
+    def create_volume(self, size, name, location=None, snapshot=None):
```
alex, line 213: This doesn't seem to handle snapshots, so it should raise an error if a snapshot is provided.

libcloud/compute/drivers/openstack.py
```diff
@@ -206,6 +206,55 @@ def list_nodes(self):
...
+                                               },
+                                               'availability_zone': location,
+                                           }
+                                       }).success()
+
+    def destroy_volume(self, volume):
+        return self.connection.request('/os-volumes/%s' % volume.id,
+                                       method='DELETE').success()
+
+    def attach_volume(self, node, volume, device=None):
```
alex: What happens when device=None, I think this should document teh default mount location.

libcloud/compute/drivers/openstack.py
```diff
@@ -206,6 +206,55 @@ def list_nodes(self):
...
+            data={
+                'volumeAttachment': {
+                    'volumeId': volume.id,
+                    'device': device,
+                }
+            }).success()
+
+    def detach_volume(self, volume):
+        success = True
+        for attachment in volume.extra['attachments']:
```
alex, line 252: Should there be a way to specify to only detach from some nodes?

libcloud/compute/drivers/openstack.py
```diff
@@ -206,6 +206,55 @@ def list_nodes(self):
...
+            if not self.connection.request(
+                '/servers/%s/os-volume_attachments/%s' %
+                (attachment['serverId'], attachment['id']),
+                method='DELETE').success():
+                success = False
+        return success
```
alex: And should this return which nodes failed, or raise an exception with them, or something?

alex:
Besides these comments, this looks good to me.

bernieke:
Awesome.

I'm on vacation next week, but as soon as I get back I'll check the behavior of openstack in these cases, and make the necessary changes.

Thanks for the quick turnaround!

bernieke:
I've addressed your remarks in a second commit: https://github.com/bernieke/libcloud/commit/456261fb379f5629ce0f047a757b8307ce181dca

libcloud/compute/drivers/openstack.py
```diff
@@ -59,6 +59,14 @@
 DEFAULT_API_VERSION = '1.1'
 
 
+class CloudStackException(Exception):
```
Kami: Here, this should probably be `OpenStackException` and not `CloudStackException` :)

We recently also added a new `ProviderError` base exception class (https://github.com/apache/libcloud/blob/cb9d374732fca01d488e117221baf1fbf6603032/libcloud/common/types.py#L67) for error like this, so please make sure this class inherits from it.

bernieke: I've switched to the ProviderError exception class, and fixed the name of the exception (good catch :))

This is the commit with the change: https://github.com/bernieke/libcloud/commit/c72e6bafc3b75716094454fa404236644d11df20

Kami:
Looks good, please attach patch to your JIRA ticket and I'll get it merged.

bernieke:
commits squashed, patch attached to https://issues.apache.org/jira/browse/LIBCLOUD-353

PR #118, opened 2013-06-28
Issue LIBCLOUD-352: Add list_volumes to compute.

bernieke:
Add list_volumes, to accompany create_volume, destroy_volume, attach_volume and detach_volume.

PR #120, opened 2013-07-01
Issue LIBCLOUD-354: Add support for volume-related functions to OpenNebula compute driver

ema:


Kami:
Patch is missing `list_volumes` method (I've just merged this pr into trunk https://github.com/apache/libcloud/pull/118), but besides that it looks good to me.

libcloud/test/compute/test_opennebula.py
```diff
@@ -631,6 +631,52 @@ def test_list_sizes(self):
...
+        nodes = self.driver.list_nodes()
+        node = nodes[0]
+
+        images = self.driver.list_images()
+        image = images[0]
+
+        ret = self.driver.attach_volume(node, image, 'sda')
+        self.assertTrue(ret)
+
+    def test_detach_volume(self):
```
Kami, line 673: Would be nice to add a test case which xercises a scenario when a node only has a single image associated with it (https://github.com/apache/libcloud/pull/120/files#L0R1174).

libcloud/compute/drivers/opennebula.py
```diff
@@ -868,28 +870,35 @@ def _extract_images(self, compute):
...
-        if len(disks) > 0:
+        # Return all disks when the Node type accepts multiple attached disks
+        # per node.
+        if len(disks) > 1:
+            return disks
+        
+        if len(disks) == 1:
             return disks[0]
-        else:
-            return None
```
Kami, line 892: FYI - I just spotted while merging merging patch into trunk. I prefer being explicit (Zen of Python agrees with me) so I've changed it to explicitly return "None".

Kami:
I've merged this into trunk, so you can close this pull request now.

Thanks!

libcloud/compute/drivers/opennebula.py
ema, line 892: On 08/07 04:07, Tomaz Muraus wrote:

> > ```
> >          return disks[0]
> > ```
> > -        else:
> > -            return None
> 
> FYI - I just spotted while merging merging patch into trunk. I prefer
> being explicit (Zen of Python agrees with me) so I've changed it to
> explicitly return "None".

Sounds good to me. Thanks for your quick feedback!

Cheers,
    Emanuele

PR #121, opened 2013-07-02
Issue LIBCLOUD-356: Provide user-data for Openstack driver

mpuel:
This implements a mimic of Nova client behavior: fails if the provided user-data does not point to a readable file, provides the content of the file as user-data if the file is readable.

### DIFF
--- a/libcloud/compute/drivers/openstack.py
+++ b/libcloud/compute/drivers/openstack.py
@@ -1169,8 +1169,13 @@ class OpenStack_1_1_NodeDriver(OpenStackNodeDriver):
             server_params['key_name'] = kwargs['ex_keyname']
 
         if 'ex_userdata' in kwargs:
-            server_params['user_data'] = base64.b64encode(
-                b(kwargs['ex_userdata'])).decode('ascii')
+            f = open(kwargs['ex_userdata'], 'r')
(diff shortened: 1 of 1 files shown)


libcloud/compute/drivers/openstack.py
```diff
@@ -1169,8 +1169,13 @@ def _create_args_to_params(self, node, **kwargs):
             server_params['key_name'] = kwargs['ex_keyname']
 
         if 'ex_userdata' in kwargs:
-            server_params['user_data'] = base64.b64encode(
-                b(kwargs['ex_userdata'])).decode('ascii')
+            f = open(kwargs['ex_userdata'], 'r')
```
Kami, line 1172: I would do something like this here https://github.com/apache/libcloud/pull/115#discussion_r4831237

Please also use `with` context manager to make sure file is always closed.

Kami:
Also needs a test case which tests a behavior when `ex_userdata` points to a file.

Kami:
@mpuel bump^

Kami:
@mpuel Is this still relevant?

If so, please synchronize your repository and this branch with the upstream one and add test cases. If not, please close this pull request.

Thanks!

PR #123, opened 2013-07-15
Added some as-yet-unimplemented methods to the base compute driver.

alex:
This is so drivers can have a common design to implement against.

### DIFF
--- a/libcloud/compute/base.py
+++ b/libcloud/compute/base.py
@@ -415,7 +415,7 @@ class StorageVolume(UuidMixin):
 
         @returns C{VolumeSnapshot}
         """
-        return self.driver.snapshot_volume(volume=self, name=name)
+        return self.driver.create_volume_snapshot(volume=self, name=name)
 
     def destroy(self):
         """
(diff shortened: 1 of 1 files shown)


libcloud/compute/base.py
```diff
@@ -780,6 +808,33 @@ def list_volumes(self):
...
+    def snapshot_volume(self, volume, name):
+        """
+        Creates a snapshot of the storage volume.
+
+        @rtype: L{VolumeSnapshot}
+        """
+        raise NotImplementedError(
+            'snapshot_volume not implemented for this driver')
+
+    def destroy_snapshot(self, snapshot):
```
Kami: This one should probably be called something like `destroy_volume_snapshot` so users knows it's related to volume and not virtual machine snapshots.

libcloud/compute/base.py
```diff
@@ -780,6 +808,33 @@ def list_volumes(self):
...
+    def list_volume_snapshots(self, volume):
+        """
+        List snapshots for a storage volume.
+
+        @rtype: C{list} of L{VolumeSnapshot}
+        """
+        raise NotImplementedError(
+            'list_volume_snapshots not implemented for this driver')
+
+    def snapshot_volume(self, volume, name):
```
Kami: Maybe `create_volume_snapshot` to be consistent with other method names?

I'm not too opinionated about it though.

libcloud/compute/base.py
Kami: Besides that, LGTM

alex:
Updated to resolve the issues.

Kami:
Merged, thanks.

PR #126, opened 2013-07-27
Brightbox compute Driver: Fixed attribute name, Added check for null data

sankroh:
1. 'user_data' changed to 'account'
2. When there are deleted servers in the node_list, zone was None and it was throwing a TypeError when trying to parse it.

### DIFF
--- a/libcloud/compute/drivers/brightbox.py
+++ b/libcloud/compute/drivers/brightbox.py
@@ -20,6 +20,7 @@ from libcloud.utils.py3 import httplib
 from libcloud.utils.py3 import b
 
 from libcloud.common.brightbox import BrightboxConnection
+from libcloud.common.types import LibcloudError
 from libcloud.compute.types import Provider, NodeState
 from libcloud.compute.base import NodeDriver
 from libcloud.compute.base import Node, NodeImage, NodeSize, NodeLocation
@@ -63,7 +64,7 @@ class BrightboxNodeDriver(NodeDriver):
                                                   **kwargs)
 
     def _to_node(self, data):
-        if data is not None:
+        if data:
             extra_data = _extract(data, ['fqdn', 'account', 'status',
                                      'interfaces', 'snapshots',
                                      'server_groups', 'hostname',
@@ -90,6 +91,8 @@ class BrightboxNodeDriver(NodeDriver):
                 image=self._to_image(data['image']),
                 extra=extra_data
             )
+        else:
+            raise LibcloudError('No data was provided.')
 
     def _to_image(self, data):
         extra_data = _extract(data, ['arch', 'compatibility_mode',
@@ -123,7 +126,7 @@ class BrightboxNodeDriver(NodeDriver):
     def _to_location(self, data):
         if data:
             return NodeLocation(
-               id=data['id'],
+                id=data['id'],
                 name=data['handle'],
                 country='GB',
                 driver=self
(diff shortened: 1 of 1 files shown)


libcloud/compute/drivers/brightbox.py
```diff
@@ -63,32 +63,33 @@ def __init__(self, key, secret=None, secure=True, host=None, port=None,
...
 
     def _to_node(self, data):
-        extra_data = _extract(data, ['fqdn', 'user_data', 'status',
+        if data is not None:
+            extra_data = _extract(data, ['fqdn', 'account', 'status',
                                      'interfaces', 'snapshots',
                                      'server_groups', 'hostname',
                                      'started_at', 'created_at',
                                      'deleted_at'])
-        extra_data['zone'] = self._to_location(data['zone'])
```
Kami, line 71: Instead of not returning `Node` object at all if `zone` is not available, we should just skip inclusion of `zone` attribute in `extra_data` dictionary.

libcloud/compute/drivers/brightbox.py
```diff
@@ -120,12 +121,15 @@ def _to_size(self, data):
         )
 
     def _to_location(self, data):
-        return NodeLocation(
-            id=data['id'],
-            name=data['handle'],
-            country='GB',
-            driver=self
-        )
+        if data:
```
Kami, line 127: I would get rid of this check - it just hides an actual error which is passing invalid data to this function (`None`, `''`, ...).

Kami:
@roadhead Any updates on this?

It would be great to see this merged into trunk.

libcloud/compute/drivers/brightbox.py
sankroh, line 71: hmm, no. It returns `None` for `zone` if it's not available. If there is no `data` passed in then it does not return a `Node` object.

libcloud/compute/drivers/brightbox.py
sankroh, line 127: The only `data` thats passed in to this function is the `zone` info. If there is no `zone` info, then it returns `None`

sankroh:
Sorry for the late response, totally forgot about this... See my comments above.

libcloud/compute/drivers/brightbox.py
Kami, line 71: Yeah, sorry, maybe I wasn't clear in my previous comment.

What I wanted to say is that we should get rid of this check all together. If someone passes `None` for `data` argument, this means it's not using the method incorrectly and the method should throw.

Imo, a better and correct solution would be to fix methods which are calling `_to_node` and `_to_location` to never call this method with `None` as an argument.

Kami:
@roadhead No problem, I've replied to your comments.

libcloud/compute/drivers/brightbox.py
sankroh, line 71: Yep, so I've made it raise an exception when there's no data passed in, that should take care of your use-case.

libcloud/compute/drivers/brightbox.py
```diff
@@ -63,32 +64,35 @@ def __init__(self, key, secret=None, secure=True, host=None, port=None,
...
+                            for interface in data['interfaces']
+                            if 'ipv6_address' in interface],
+
+                driver=self.connection.driver,
+                size=self._to_size(data['server_type']),
+                image=self._to_image(data['image']),
+                extra=extra_data
+            )
+        else:
+            raise LibcloudError('No data was provided.')
```
Kami, line 95: This should be `ArgumentError`. Also, to reduce nesting and make it more clear, please move this check to the top of the function:

``` python

if not data:
    raise ArgumentError...

...
```

Can you please also do the same in `_to_location`?

PR #127, opened 2013-07-30
[RFC] Command line utility for updating pricing file

Kami:
This branch adds a command line tool for updating pricing file (as discussed in the last design day). It includes the following changes:
1. Modifies the pricing code, to read pricing data from `~/.libcloud/pricing.json` file, if it exists. If it doesn't it reads it from a pricing file which is bundled with a libcloud distribution.
2. Adds a command line tool to download the latest pricing file from our git repository and saves it to `~/.libcloud/pricing.json` 

Example CLI tool usage:

``` bash
$ libcloud update-pricing [--file-path=~/.libcloud/pricing.json] [--file-url=...]
```

The tool is non-interactive (so it can be used in a cronjob, etc) and it's very simple. It doesn't perform any update checks and it simply overwrites  the file at `~/.libcloud/pricing.json`. 

Since the pricing file is small, the current approach is imo totally fine. Adding extra code to handle "only update if necessary" by using hash comparison / modified-at or similar approach would imo just add unnecessary complexity with very little benefit.


Jc2k:
I thought we were on the whole against having a CLI as part of libcloud? A CLI is just noise for people using libcloud as a lib?

libcloud/cli/pricing.py
```diff
@@ -0,0 +1,84 @@
...
+    parser.add_argument('--file-path', dest='file_path', action='store',
+                        default=CUSTOM_PRICING_FILE_PATH,
+                        help='Path where the file will be saved')
+    parser.add_argument('--file-url', dest='file_url', action='store',
+                        default=DEFAULT_FILE_URL,
+                        help='URL to the pricing file')
+    return parser
+
+
+def update_pricing(file_url, file_path):
```
Jc2k: It would be good if this function was in libcloud.common and raise exceptions so that users who have their own GUI or console UI can use it.

setup.py
```diff
@@ -228,6 +228,13 @@ def run(self):
         cov.save()
         cov.html_report()
 
+if pre_python26:
```
Jc2k: If the CLI is going to be part of libcloud, we could use extra_requires to make CLI dependencies such as argparse optional.

libcloud/cli/pricing.py
```diff
@@ -0,0 +1,84 @@
...
+from libcloud.pricing import CUSTOM_PRICING_FILE_PATH
+from libcloud.utils.connection import get_response_object
+
+__all__ = [
+    'add_subparser',
+    'update_pricing'
+]
+
+# Default URL to the pricing file
+DEFAULT_FILE_URL = 'https://git-wip-us.apache.org/repos/asf?p=libcloud.git;a=blob_plain;f=libcloud/data/pricing.json'
```
Jc2k: There are a bunch of "settings" in libcloud now - here, pricing.py, the ssl cert settings. If I was using libcloud and shipping it inside an app I would likely want to customize them. But they are spread out. Is it a bit ugly? Maybe we should have a libcloud.settings?

Kami:
@Jc2k Yeah, I do believe an actual CLI for managing all the libcloud resources should be available as a separate (optional) package. Actual management CLI would probably also depend on more libraries (probably `cliff` which has tons of dependencies by itself) and have different versions requirements.

`update-pricing` command is a bit special case though. I would imagine a lot of people using just the library would also like to use this command.

libcloud/cli/pricing.py
Kami: I agree, I'll update it accordingly.

libcloud/cli/pricing.py
Kami: I would treat this variable more of as a constant and less so as a setting. I do agree there are some other "setting" variables though:
- `LIBCLOUD_DEBUG`
- `SSL_CERT_FILE`
- `VERIFY_CERT`

Currently those values can be set either by changing the module variable value or through environment variables. 

I can see value in unifying all the settings and moving them in a single file / module, but I don't want to make everything more complex as necessary and require initialization phase, etc.

I would assume that with `libcloud.settings` you mean Django like settings?

Jc2k: I didn't mean as complicated as django settings (which has a wrapper in django.conf etc), I just meant it might be nice to unify them in a single module.

I think `CUSTOM_PRICING_FILE_PATH` is one too? (If my tool updates its pricing.json automatically I would think it "polite" to not stomp on a shared pricing.json?)

Jc2k:
@Kami I still disagree about `libcloud update-pricing`.

If I was writing a Django app that needed this data, I would write a celery task to update the pricing routinely.

If I was writing a command line tool, I would build the update pricing call into it. I wouldn't want the users to have to call `libcloud update-pricing` manually.

Even if I was writing a hacky script that needed to use pricing data I would probably want to make sure it was always up to date whenever I ran that script. So I would want to use the API.

For prototypers and tinkerers we could just use use the "run module" style - `python -m libcloud.pricing`. That would seem more appropriate?

Kami:
First a quick recap to make sure everyone is one the same page and aware what is the end goal / desired output:
1. We want to allow users to use a custom pricing file without modifying any code (I think default to `~/.libcloud/pricing.json file if available is a good and straight-forward approach)
2. Provide a simple mechanism to update the pricing file - it's still up to the user to reload the pricing file / invalidate the cache if they use libcloud in a long running process

@Jc2k I do agree about the "pricing always up to date in my app / script" part - we need to provide programmatic API which users can call when they run their script or initialize the app.

As far as the celery goes - I personally think it's an overkill for a thing like this. If a CLI / "run module" style is available you can simply use a cron job.

As far as the "run module" style goes - I'm not to opinionated about this, although I'm not a big fun of `python -m libcloud.pricing` since it's less explicit than `libcloud update-pricing`.

I would like to hear more opinion on this one.

Kami:
P.S. @Jc2k - thanks for your feedback so far!

Jc2k:
No problem!

My feedback so far has been from the angle of "Libcloud is a library". I find it weird that a library would have a binary as a default part of consuming that library. By all means have example scripts, or modules that you can run. But its a library. The programatic API should be very much "the right way" to do it.

None of my use cases are hindered by the presence of the command line API, but I strongly feel that a command line tool that appears in your bin/ dir sends the wrong message, that its the "first class" way to do things. When really the first class way to do it should be via the API.

I think _every_ tool/app/script using pricing data should use the programatic API. I don't see how its any different to wanting a CLI to create a compute node with libcloud.

cpsaltis:
I don't think that `python -m update.pricing` is any better than `libcloud update-pricing`.  The script by @Kami offers a simple and straightforward solution, however I do understand the concerns raised by @Jc2k.

I don't know the story behind the management CLI @Kami mentioned, but it doesn't sound like a very bad idea. It could be used to update `pricing.json` before each release. By default libcloud would try to get `pricing.json` from SVN and fallback to the local copy. It is a bit more complicated, but if there are more use cases for such a tool, why not. 

Any other ideas?

Kami:
@Jc2k after more thought, I agree with you. As a first step, I'll merge those the following two changes which allow users to more easily update pricing file.
1. Read pricing from ~/.libcloud/pricing.json file (if available)
2. Programmatic way to update pricing file

I'll leave command line utility for updating the pricing file out for now and work on bootstrapping "libcloud CLI" as a separate package (apache-libcloud-cli). 

First command libcloud-cli will contain will be a command for updating a pricing file.

Kami:
I've merged this PR into trunk.

I'll open a separate PR for the CLI stuff soon.

Kami:
@cpsaltis, @Jc2k and others - I've added some docs about pricing and the new `download_pricing_file` functionality which is available in trunk - https://ci.apache.org/projects/libcloud/docs/compute/pricing.html

If something is missing or not clear, please let me know and I'll update it.

cpsaltis:
Seems clear. Let me know when you start the libcloud CLI package and me or @mgogoulos will commit the price scraping part.

PR #130, opened 2013-08-01
LIBCLOUD-367: improve feature metadata

Jc2k:
This pull request updates the metadata of `self.features['create_node']`.

It refactors `deploy_node` to not care about the metadata when authentication information is provided, this means EC2 and other users of deploy_node can continue to function even though they don't really support the NodeAuthSSHKey interface.

It turns out that `deploy_node` will generate a password for drivers that support `password` if auth is not provided. In order to not break that i've had to move the feature into the `create_node` implementations. So now any driver that supports `password` will behave like `generates_password` if no auth information is provided.

It adds a generic helper for making sure the right kind of NodeAuth object has been pased to `create_node`.

The PR also contains a few minor PEP8 cleanups in files i've touched.

### DIFF
--- a/libcloud/compute/base.py
+++ b/libcloud/compute/base.py
@@ -492,20 +492,20 @@ class NodeDriver(BaseDriver):
         """
 
         if isinstance(auth, NodeAuthPassword):
-            if "password" in self.features['create_node']:
+            if 'password' in self.features['create_node']:
                 return auth
             raise ProviderError(
-                "Password provided as authentication information, but password"
-                "not supported")
+                'Password provided as authentication information, but password'
+                'not supported')
 
         if isinstance(auth, NodeAuthSSHKey):
-            if "ssh_key" in self.features['create_node']:
+            if 'ssh_key' in self.features['create_node']:
                 return auth
             raise ProviderError(
-                "SSH Key provided as authentication information, but SSH Key"
-                "not supported")
+                'SSH Key provided as authentication information, but SSH Key'
+                'not supported')
 
-        if "password" in self.features['create_node']:
+        if 'password' in self.features['create_node']:
             value = os.urandom(16)
             return NodeAuthPassword(binascii.hexlify(value), generated=True)
 
@@ -693,17 +693,14 @@ class NodeDriver(BaseDriver):
                 raise NotImplementedError(
                     'If providing auth, only NodeAuthSSHKey or'
                     'NodeAuthPassword is supported')
-
         elif 'ssh_key' in kwargs:
             # If an ssh_key is provided we can try deploy_node
             pass
-
         elif 'create_node' in self.features:
             f = self.features['create_node']
             if not 'generates_password' in f and not "password" in f:
                 raise NotImplementedError(
                     'deploy_node not implemented for this driver')
-
         else:
             raise NotImplementedError(
                 'deploy_node not implemented for this driver')
--- a/libcloud/compute/base.py
+++ b/libcloud/compute/base.py
@@ -494,25 +494,25 @@ class NodeDriver(BaseDriver):
         if isinstance(auth, NodeAuthPassword):
             if 'password' in self.features['create_node']:
                 return auth
-            raise ProviderError(
+            raise LibcloudError(
                 'Password provided as authentication information, but password'
-                'not supported')
+                'not supported', driver=self)
 
         if isinstance(auth, NodeAuthSSHKey):
             if 'ssh_key' in self.features['create_node']:
                 return auth
-            raise ProviderError(
+            raise LibcloudError(
                 'SSH Key provided as authentication information, but SSH Key'
-                'not supported')
+                'not supported', driver=self)
 
         if 'password' in self.features['create_node']:
             value = os.urandom(16)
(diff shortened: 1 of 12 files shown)


libcloud/compute/base.py
```diff
@@ -478,6 +479,36 @@ def __init__(self, key, secret=None, secure=True, host=None, port=None,
...
+    def _get_and_check_auth(self, auth):
+        """
+        Helper function for providers supporting L{NodeAuthPassword} or
+        L{NodeAuthSSHKey}
+
+        Validates that only a supported object type is passed to the auth
+        parameter and raises an exception if it is not.
+
+        If no L{NodeAuthPassword} object is provided but one is expected then a
+        password is automatically generated.
```
Kami, line 491: Currently this method returns an implicit `None` if you pass in invalid value for `auth`.

Would probably be better to ensure that user can only pass in NodeAuthPassword, NodeAuthSsKey and None.

libcloud/compute/base.py
```diff
@@ -478,6 +479,36 @@ def __init__(self, key, secret=None, secure=True, host=None, port=None,
...
+        """
+
+        if isinstance(auth, NodeAuthPassword):
+            if "password" in self.features['create_node']:
```
Kami: Minor style thing - we use single quotes around regular strings.

libcloud/compute/drivers/bluebox.py
```diff
@@ -197,6 +195,10 @@ def create_node(self, **kwargs):
         result = self.connection.request('/api/blocks.json', headers=headers,
                                          data=params, method='POST')
         node = self._to_node(result.object)
+
+        if getattr(auth, "generated", False):
```
Kami, line 199: I think we still want to set `extra['password']` even if `generated` is `True`. Otherwise there is no way for user to access the generated password, right?

libcloud/compute/base.py
```diff
@@ -656,29 +687,36 @@ def deploy_node(self, **kwargs):
...
+                    'If providing auth, only NodeAuthSSHKey or'
+                    'NodeAuthPassword is supported')
 
-        if 'create_node' not in self.features:
-            raise NotImplementedError(
-                'deploy_node not implemented for this driver')
-        elif 'generates_password' not in self.features["create_node"]:
-            if 'password' not in self.features["create_node"] and \
-               'ssh_key' not in self.features["create_node"]:
+        elif 'ssh_key' in kwargs:
```
Kami, line 701: Another minor style thing - we don't use blank lines between different if, elif and else.

libcloud/compute/drivers/abiquo.py
```diff
@@ -40,7 +40,6 @@ class AbiquoNodeDriver(NodeDriver):
     name = 'Abiquo'
     website = 'http://www.abiquo.com/'
     connectionCls = AbiquoConnection
-    features = {'create_node': ['password']}
```
Kami, line 43: Just want to confirm - this driver was incorrectly indicating that it supports password authentication, right?

libcloud/compute/base.py
```diff
@@ -478,6 +479,36 @@ def __init__(self, key, secret=None, secure=True, host=None, port=None,
                                          host=host, port=port,
                                          api_version=api_version, **kwargs)
 
+    def _get_and_check_auth(self, auth):
```
Kami, line 482: Now that this functionality is split out it should be easier to test it, so it would be nice to add some test cases for it.

Kami:
Thanks.

Once this has been merged our goal should also be to fully document and explain (with examples) how `deploy_node` and `auth` stuff works.

libcloud/compute/drivers/abiquo.py
Jc2k, line 43: Correct. It definitely doesn't look at the auth argument, so there is no way to provide a password.

libcloud/compute/drivers/bluebox.py
Jc2k, line 199: I think you misread this expression? Or am i being dumb? The intended logic is to only return the password if it is a generated password. It won't return the password if it is one you manually provided to the API.

The thinking was to minimize leaking a password that you have chosen manually - but i'm more than happy to remove the `if` guard.

libcloud/compute/base.py
Jc2k, line 491: Aha! That implicit None had been bothering me, and your suggestion is an ideal fix.

libcloud/compute/drivers/bluebox.py
Kami, line 199: Yeah, you can ignore my comment - I did in fact misread it :)

Jc2k:
I have implemented your suggestions.

Kami:
Thanks!

Can you please squash all the commits into a single and attach a patch to the JIRA ticket?

Jc2k:
Done

Kami:
Thanks, I've merged this PR into trunk.

This branch also exposed some encoding related Python 3.x bugs which I fixed in commit 05e8559d4321c27f32cdbe0c913b334f0c1b60ae.

PR #132, opened 2013-08-01
Initial pass at documenting how auth works for create_node/deploy_node

Jc2k:
This fleshes out the description for how NodeAuth\* objects work in the compute drivers.


Kami:
Thanks, looks good.

Now that we are moving to Sphinx and having documentation in the same repository as code we should also start putting more stuff with examples to the actual documentation.

Kami:
Merged into trunk, thanks!

PR #135, opened 2013-08-18
Test LoadBalancers with a common set of tests to verify they are API compliant

Jc2k:
If we are happy with the approach demonstrated in this PR I still have to port RackSpace to the scheme.

This PR also adds more intelligent mocks. Whilst there are some downsides to that it has elevated the amount of input validation and round-trip testing that happens.

This PR would be followed by a second to add `feature` metadata to expose which load balancers manage members by IP and which manage them by instance id. And also the backend port policy (a good few ignore Member.port). I'd love to write tests for these that validate consistency over the entire API, so i'd like to finish this PR first.


Kami:
First some  feedback on the whole approach.
1. I like the explicit route to method mapping in the Brightbox mock. I think we can definitely improve current "implicit" mapping by making all the mappings explicit. Explicit mappings will make writing tests easier and it will also be more obvious to see what is going on.

I think we could make it even better (and less error-prone) if a route would also contain an HTTP method (GET, etc.).
1. I'm not a fan of `_to_json`, `_from_balancer` and other methods. It means we need to keep multiple things in sync (fixtures + those methods). The more things a developer needs to update, the easier it is to forget to update some, which means it more cumbersome and error prone.

I would prefer to stick with just the response fixtures. This way it's easier to maintain and generate the fixtures.

Overall I'm not too sure if I like the "more intelligent" / dynamic mocks approach. It basically means we are re-implementing dummy-like driver and logic for each provider. This has the following disadvantages (imo):
- More things to keep up to date
- It's easier to incorrectly implement a mock
- We are actually just testing the mock API implementation and not the provider API which doesn't add any value imo.

Imo, tests should test the following things:
- Test that each provider method calls the correct provider endpoint
- Test that each method sends the right data (body + query string) along to the endpoint
- Test that data returned by the endpoint is correctly parsed

I would also love to hear other people's opinion.

Jc2k:
I agree with you, I think, and you have stated the other half of my internal dialogue quite accurately :-)

I was initially reluctant to stick with the fixtures because we would have to manually update the fixtures to get test data that was consistent enough to allow re-usable tests, but equally effectively requiring devs to implement both sides of the protocol is a bit extreme.

Most of all I want to test that each method is sending the right data. The right arguments. The trouble is, if i parse the entire body all I can do is validate it is a valid request, not that it is the correct request for the arguments that were passed. I can see 2 ways to factor this.

The first is this branch. By "round-tripping" a request I can test that what i put in one end comes out of the other. I must be passing all the data i need to if i can get it back out of the other side.

However i've just thought of a second that fits more naturally with keeping the fixtures. I can create a generic adaptor that records the arguments it was invoked with in a way that lets the mock peek at them, and then calls the driver. I can then parse the incoming body in my mock and make sure it matches the arguments passed to the driver, then return the hard-coded fixture file for that request.

A secondary concern is duplication. If there are 4 variants of the `/balancer/list` API call then it's hard to factor the current mocks to avoid duplicate checking logic. (Is this where you see the route stuff in the Brightbox mock helping?).

PR #138, opened 2013-08-27
Merge Elastichosts NodeDriver classes into a single class that accepts a region argument

bennettaur:


libcloud/compute/drivers/elastichosts.py
```diff
@@ -128,91 +127,36 @@
...
-    """
-    host = API_ENDPOINTS['ca-1']['host']
-
-
-class ElasticHostsCA1NodeDriver(ElasticHostsBaseNodeDriver):
-    """
-    ElasticHosts node driver for the Toronto Peer 1 end-point
-    """
-    connectionCls = ElasticHostsCA1Connection
+    def __init__(self, *args, **kwargs):
```
Kami: Please use normal arguments and keyword argument instead of _args and *_kwargs so it's more explicit and easier to see what is going on.

This approach also wouldn't work because parent class doesn't take `region` argument and you don't use `kwargs.pop('host')`.

libcloud/compute/providers.py
```diff
@@ -47,16 +47,8 @@
         ('libcloud.compute.drivers.ec2', 'EC2APSESydneyNodeDriver'),
     Provider.ECP:
         ('libcloud.compute.drivers.ecp', 'ECPNodeDriver'),
-    Provider.ELASTICHOSTS_UK1:
```
Kami, line 50: Ideally, we could leave old constants here for a while for backward compatibility reasons so it makes upgrading less painful and users don't need to do it asap.

You can have a look at how this is done for the EC2 driver.

libcloud/compute/drivers/elastichosts.py
```diff
@@ -128,91 +127,36 @@
...
+        if 'region' in kwargs:
+            self.region = kwargs['region']
+            if self.region not in API_ENDPOINTS:
+                raise ElasticHostsException("Invalid region specified")
```
Kami: `ValueError` would be more appropriate. `ElasticHostsException` should only be used for HTTP exceptions.

libcloud/compute/drivers/elastichosts.py
```diff
@@ -128,91 +127,36 @@
...
+        else:
+            raise ElasticHostsException("ElasticHosts Driver requires a region to be specified")
+        super(ElasticHostsNodeDriver, self).__init__(*args, **kwargs)
+
+    def _ex_connection_class_kwargs(self):
+        """
+        Return the host value based the user supplied region
+        """
+        return {'host': API_ENDPOINTS[self.region]['host']}
```
Kami: Need to make sure this method only returns `host` argument if `host` kwarg is not passed to the constructor. The reason is that `host` has precedence over `region`.

Another option would be to make those arguments mutually exclusive - require user to pass at least one, but not both at once.

Kami:
Thanks.

It would also be great to add some tests as noted in the JIRA ticket.

Otherwise the patch looks good to me and should be ready to be merged once the comments in this PR are addressed.

bennettaur:
I've made some changes that I'll be pushing up soon. I've modelled it to be more like the EC2 driver, which allows the ElasticHostsDriver to work for any region but still be backwards compatible. 

Just thought I'd let you know I'm still working on this

Kami:
@bennettaur Sounds good. Ping me once you push those changes and I'll have a look.

Kami:
@bennettaur I saw you pushed a commit.

The PR is still lacking tests, but besides that, is it ready to be reviewed?

bennettaur:
@Kami I found one issue with my code, which I fixed. You can review what I have now if you like.

I was thinking for testing, without requiring any running nodes or drives, the only way to test whether the general Driver connects to the various API endpoints would be to use one region's credentials for another region when instantiating the driver, then catching the invalid credentials exception when trying to run a command with that driver.

If that sounds sufficient for you then I'll go ahead and add those in. It wouldn't require any fixtures, just a valid test.secrets file.

Kami:
@bennettaur Sounds good, I'll have a look shortly.

libcloud/compute/drivers/elastichosts.py
```diff
@@ -128,91 +127,105 @@
...
+            raise ValueError("ElasticHosts Driver requires a region to be specified")
+        super(ElasticHostsNodeDriver, self).__init__(key=key, secret=secret,
+                                                     secure=secure, host=host,
+                                                     port=port, **kwargs)
 
+    def _ex_connection_class_kwargs(self):
+        """
+        Return the host value based the user supplied region
+        """
+        if self.host_set:
```
Kami: Yeah, this looks better and should address issue mentioned in the previous commit. I'm just not sure about the variable name.

Maybe `host_argument_set` would be better, since `host_set` reminds me of a set of hosts (servers) :) Besides that, I would also make variable "private" (prefix it with an underscore).

libcloud/compute/drivers/elastichosts.py
```diff
@@ -128,91 +127,105 @@
...
-class ElasticHostsUK1Connection(ElasticStackBaseConnection):
-    """
-    Connection class for the ElasticHosts driver for
-    the London Peer 1 end-point
-    """
+        if hasattr(self, '_region'):
+            region = self._region
 
-    host = API_ENDPOINTS['uk-1']['host']
+        if region is not None:
```
Kami, line 155: `region` shouldn't be required if user providers `host` argument.

Kami:
@bennettaur As far as the tests go - none of our tests actually connect to the provider endpoint. We use fixture system which returns mocked responses which are stored in files.

To test this, I would do something along those lines for every region:
1. Instantiate a driver for region A
2. Assert that `driver.connection.host` equals host for the provided region
3. Ideally you would also use a mocked `connectionCls` and make sure the `connectionCls` constructor gets passed the correct `host` argument. 

Third step would like something along those lines:

``` python
connectionCls = Mock()
Driver.connectionCls.conn_classes = (connectionCls, connectionCls)
driver = Driver(...)
call_kwargs = connectionCls.call_args[1]
self.assertEqual(call_kwargs['host'], actual_region_host)
```

To make sure host is correctly preserved I would also do that:
1. Test for region A
2. Test for region B
3. Test for region A (make sure host is still set correctly)

bennettaur:
Reviewing this a little more, I renamed the ElasticHostsBaseNodeDriver to just be ElasticHostsNodeDriver, which could break backwards compatibility for anyone using the BaseNodeDriver to connect to the default 'us-1' region. Not sure if this is a concern of yours or not.

I can't commit time to writing the tests right now, but I will as soon as I am able to.

Kami:
@bennettaur Thanks for you work on this so far.

I want this change to be included in the next release which should go out very shortly, so I will merge what we have so far and finalize the patch.

Kami:
This pull request can be closed now.

I've finished your changes and merged them into trunk. Those changes will be included in the next release. Thanks!

PR #139, opened 2013-08-27
Added support to the Rackspace driver for using compute and storage in the new IAD region

alex:


Kami:
I forgot to say that this pull request has been merged - thanks for closing it.

PR #140, opened 2013-09-01
Issue LIBCLOUD-389: Adds accessIPv4 value to the node_list results

johnsocp:
This is a pretty simple fix to expose the public IP that gets assigned to a server built in a Rackspace RackConnect environment. The original public ip stays assigned to the server but the interface is disabled so the value set in accessIPv4 essentially becomes the only working public IP once the post-build automation completes.

### DIFF
--- a/libcloud/compute/drivers/openstack.py
+++ b/libcloud/compute/drivers/openstack.py
@@ -1716,6 +1716,7 @@ class OpenStack_1_1_NodeDriver(OpenStackNodeDriver):
             driver=self,
             extra=dict(
                 hostId=api_node['hostId'],
+                access_ip=api_node.get('accessIPv4'),
                 # Docs says "tenantId", but actual is "tenant_id". *sigh*
                 # Best handle both.
                 tenantId=api_node.get('tenant_id') or api_node['tenantId'],


libcloud/compute/drivers/openstack.py
```diff
@@ -1716,6 +1716,7 @@ def _to_node(self, api_node):
             driver=self,
             extra=dict(
                 hostId=api_node['hostId'],
+                access_ips=api_node.get('accessIPv4'),
```
Kami: Is this value a list? If so, we should probably default to an empty list - `access_ips=api_node.get('accessIPv4', [])`

johnsocp: The rackconnect automation will only assign a single IP but its probably a good idea to create the value as a list to match the other IP fields anyway.

Kami: I see. In this case I would prefer to omit "s" from the variable names and set it to `None` if the value is not available.

Kami:
Besides that the patch looks good to me.

It would also be great to do the thing I've mentioned on the mailing list - let us know if you need any help with implementing that.

Kami:
Can you please squash all the commits into one so I can merge it into trunk?

johnsocp:
Ok, squashed.

Kami:
Thanks.

Patch has been merged, you can closet his pull request now.

Just curious - do you have any plans to work on a separate driver for RackConnect? If so, feel free to open a new pull request for it (we can re-use the jira ticket LIBCLOUD-389).

johnsocp:
Yes, I was just looking at what it would take. I added the wait for the rax_automation_status metadata key to salt-cloud but none of that would be necessary if it were added to the libcloud driver--definitely a better solution.

PR #141, opened 2013-09-02
Sleep for a while to avoid errno 104

s0undt3ch:
On some weird and random occasions the GIL deadlocks while connecting forcing the remote servers to 104 the connection. A simple sleep avoids this issue(after googling, I found http://goo.gl/0LpLvJ which helped)

```
  File "/usr/src/salt-cloud/saltcloud/clouds/openstack.py", line 386, in create
    data = conn.create_node(**kwargs)
  File "/usr/lib/python2.6/site-packages/libcloud/compute/drivers/openstack.py", line 1111, in create_node
    data={'server': server_params})
  File "/usr/lib/python2.6/site-packages/libcloud/compute/drivers/openstack.py", line 145, in request
    method=method, headers=headers)
  File "/usr/lib/python2.6/site-packages/libcloud/common/openstack.py", line 580, in request
    return super(OpenStackBaseConnection, self).request(**kwargs)
  File "/usr/lib/python2.6/site-packages/libcloud/common/base.py", line 632, in request
    else:
  File "/usr/lib64/python2.6/httplib.py", line 990, in getresponse
    response.begin()
  File "/usr/lib64/python2.6/httplib.py", line 391, in begin
    version, status, reason = self._read_status()
  File "/usr/lib64/python2.6/httplib.py", line 349, in _read_status
    line = self.fp.readline()
  File "/usr/lib64/python2.6/socket.py", line 433, in readline
    data = recv(1)
  File "/usr/lib64/python2.6/ssl.py", line 215, in recv
    return self.read(buflen)
  File "/usr/lib64/python2.6/ssl.py", line 136, in read
    return self._sslobj.read(len)
error: [Errno 104] Connection reset by peer
```

### DIFF
--- a/libcloud/common/base.py
+++ b/libcloud/common/base.py
@@ -593,6 +593,12 @@ class Connection(object):
         # Removed terrible hack...this a less-bad hack that doesn't execute a
         # request twice, but it's still a hack.
         self.connect()
+
+        # And because on some weird ocasions the GIL will dead-lock on this
+        # code triggering a 104 connection reset by peer, let's sleep just for
+        # a litle while.
+        time.sleep(0.01)
+
         try:
             # @TODO: Should we just pass File object as body to request method
             # instead of dealing with splitting and sending the file ourselves?


Kami:
@s0undt3ch A couple of questions / comments:
1. Are you sure this is not a bad / flaky local connection or a server side issue?
2. Can you provide more environment details (Python versions, etc.) and post a code which reproduces this issue? 
3. We don't do anything nasty with threads or anything else I'm aware which could cause GIL deadlock so only thing I could see causing this is a bug in CPython which (imo) is unlikely.

@alex is way more familiar with CPython internals than I am so he might have some better ideas of what could be going on here.

s0undt3ch:
1. No, I'm not sure, specially because it's a bit random(the failure), but after this `sleep` is hasn't appeared again and even though it's random it should have appeared by now.
2. See bellow.
3. The issue is, apparently, with httplib, this "patch" just tries not to trigger it.

``` yaml
local:
  cpu_flags: fpu de tsc msr pae cx8 cmov pat clflush mmx fxsr sse sse2 ht syscall nx mmxext fxsr_opt lm up rep_good aperfmperf unfair_spinlock pni pclmulqdq ssse3 fma cx16 sse4_1 sse4_2 popcnt aes f16c hypervisor lahf_lm cmp_legacy extapic cr8_legacy abm sse4a misalignsse 3dnowprefetch xop fma4 tce tbm perfctr_core
  cpu_model: AMD Opteron(tm) Processor 4332 HE
  cpuarch: x86_64
  defaultencoding: UTF8
  defaultlanguage: en_GB
  domain: saltstack.com
  fqdn: REMOVED
  fqdn_ip4:
      REMOVED
      REMOVED
  fqdn_ip6:
      2001:4800:7813:516:c49f:4cd2:ff05:41d9
  gpus:
  host: jenkins
  id: local-jenkins-minion
  ip_interfaces: {'lo': ['127.0.0.1'], 'eth1': ['REMOVED'], 'eth0': ['REMOVED']}
  ipv4:
      REMOVED
      127.0.0.1
      REMOVED
  ipv6:
      REMOVED
      ::1
      REMOVED
      REMOVED
  kernel: Linux
  kernelrelease: 2.6.32-358.14.1.el6.x86_64
  localhost: jenkins.saltstack.com
  lsb_distrib_codename: Final
  lsb_distrib_id: CentOS
  lsb_distrib_release: 6.4
  master: REMOVED
  mem_total: 992
  nodename: REMOVED
  num_cpus: 1
  num_gpus: 0
  os: CentOS
  os_family: RedHat
  osarch: x86_64
  oscodename: Final
  osfinger: CentOS-6
  osfullname: CentOS
  osmajorrelease:
      6
      4
  osrelease: 6.4
  path: /sbin:/bin:/usr/sbin:/usr/bin
  ps: ps -efH
  pythonpath:
      /usr/bin
      /usr/src/salt-cloud
      /usr/src/salt
      /usr/lib64/python26.zip
      /usr/lib64/python2.6
      /usr/lib64/python2.6/plat-linux2
      /usr/lib64/python2.6/lib-tk
      /usr/lib64/python2.6/lib-old
      /usr/lib64/python2.6/lib-dynload
      /usr/lib64/python2.6/site-packages
      /usr/lib/python2.6/site-packages
      /usr/lib/python2.6/site-packages/setuptools-0.6c11-py2.6.egg-info
  pythonversion: 2.6.6.final.0
  saltpath: /usr/src/salt/salt
  saltversion: 0.16.0-2750-g336afd8
  saltversioninfo: (0, 16, 0)
  server_id: 1387667779
  shell: /bin/bash
  virtual: xen
  virtual_subtype: Xen PV DomU
```

Those of you familiar with salt will understand that system grains output above, but it's self explanatory I think. Let me know what additional info you need.

s0undt3ch:
Ah, still for 2, we use salt-cloud which uses libcloud with rackspace, we just spin VM's up and randomly, we hit that 104 error.

s0undt3ch:
The link I added in the initial comment has additional data

Kami:
Thanks for the additional info. Tomorrow, I'll check if salt-cloud does something funky which could cause this.

As far as the link goes - it didn't load in the morning when I initially checked it, but it does load now. The post appears to be from 2008 and you are running 2.6.6 which has been released in 2010. I would be surprised if there was an issue in CPython which hasn't been fixed by then.

In any case, I'll also wait for @alex, I'm sure he has some good insight.

s0undt3ch:
I'm a salt-cloud dev, let me know if you need additional info. We're not doin' anything magical with GIL either ;)

Kami:
@s0undt3ch Ah, ok :)

Just wanted to check if you are doing something with threads.

Kami:
@s0undt3ch I've talked to Alex on the IRC and he said this issue doesn't ring any bells, but a bunch of things which could potentially cause this to happen in the older versions have been fixed in 2.7.

We both agreed that a reasonable thing would be to only run this code under Python <= 2.6. 

One thing we would also like to know is if we can decrease the sleep interval even further (e.g. `time.sleep(0.001)`) and see if it still helps - can you please test this out?

```
[20:12] <       Kami_> | Alex_Gaynor: hey, have you had a chance to look at ] [ ilgiz    ] [ kandinski   ] [ polvi          ] 
[20:12] <       Kami_> | Alex_Gaynor: https://github.com/apache/libcloud/pull/141
[20:12] <       Kami_> | well that was a #failpaste
[20:28] <Alex_Gaynor> | Kami_: I hadn't, looking now
[20:29] <       Kami_> | Alex_Gaynor: thanks
[20:34] <Alex_Gaynor> | Kami_: hmm, this issue doesn't ring any bells, I know that a bunch of signals/ssl/random stuff has been fixed 
                      in 2.7, and I see this 2.6; I wonder if it's related
[20:35] <       Kami_> | Alex_Gaynor: hm, ok
[20:35] <       Kami_> | Alex_Gaynor: do you think it would be reasonable to include his patch and only use it if python <= 2.6?
[20:36] <Alex_Gaynor> | Kami_: I think that's maybe the best idea, can we use an even shorter sleep duration?
[20:36] <       Kami_> | Alex_Gaynor: since they use this code in salt-cloud and I would imagine a lot of ppl still use older distros 
          which ship with 2.6
[20:36] <       Kami_> | Alex_Gaynor: yeah, I was thinking about the same thing (shorter sleep duration)
[20:37] <       Kami_> | Alex_Gaynor: i'll add a comment to the ticket and ask if the person can try to decrease it and see if it 
          still helps
```

s0undt3ch:
This issue was presented to me under python 2.6. Unfortunately, I'm unable to stress test this issue under python 2.7, however, I'll lower the sleep to the value you asked and see how it goes.

Kami:
@s0undt3ch Sounds good. Keep me posted about the lower sleep value.

s0undt3ch:
For the past 2 days, your proposed sleep value has been working fine.

s0undt3ch:
Let's wait 2 more days?

Kami:
@s0undt3ch Yeah, I have no idea how many requests were made during those two days and if this is significant so I leave this decision to you :)

Kami:
@s0undt3ch bumping again since two days have passed.  I'll also be working on a new release today and if everything looks OK, I'd like to include this fix.

s0undt3ch:
Ok, so, I have bad and good news.

The good news is, right now, there's no need to wait for me any more.
The bad news is that I've removed the sleep completely from the code for the past 3 days. No issues. Meaning, my PR is not needed and I've failed to find the point of failure.

Ultimately, you decide, but I think it's best not to include my fix since I'm now unable to get it to fail again meaning libcloud would have an unnecessary sleep on it's code, not cool.

Kami:
@s0undt3ch OK, thanks for keeping me posted. This does seem like an intermediate (networking) issue then.

Out of curiosity, do you happen to know how many requests were made during this last period?

In any case, thanks. Feel free to close the PR.

s0undt3ch:
Approximately 60_10_2.
I'll explain, 60 builds on 10 [Jenkins](http://jenkins.saltstack.com/) build configurations, only counting start and stop vm requests(of course there's more)...

PR #142, opened 2013-09-03
Libcloud 386 GCE Load Balancer Support

wrigri:
This adds Load Balancing support for Google Compute Engine.

Since Load Balancer support for GCE is directly part of the GCE API, all of the API calls were added to the existing compute driver.  Then, wrapper methods were added to the Load Balancer driver to make it compatible with the Libcloud Load Balancer API.

In addition, there are some bug fixes and minor improvements.

Note that I'm still working on the test for the "wrapper" methods in the LoadBalancer driver, but tests have been added for all of the actual API calls that were added to the Compute driver.

### DIFF
--- a/demos/gce_demo.py
+++ b/demos/gce_demo.py
@@ -41,7 +41,13 @@ import sys
 try:
     import secrets
 except ImportError:
-    secrets = None
+    print('"demos/secrets.py" not found.\n\n'
+          'Please copy secrets.py-dist to secrets.py and update the GCE* '
+          'values with appropriate authentication information.\n'
+          'Additional information about setting these values can be found '
+          'in the docstring for:\n'
+          'libcloud/common/google.py\n')
+    sys.exit()
 
 # Add parent dir of this file's dir to sys.path (OS-agnostically)
 sys.path.append(os.path.normpath(os.path.join(os.path.dirname(__file__),
--- a/demos/gce_demo.py
+++ b/demos/gce_demo.py
@@ -58,7 +58,7 @@ MAX_NODES = 5
 DEMO_BASE_NAME = 'libcloud-demo'
 
 # Datacenter to create resources in
-DATACENTER = 'us-central1-a'
+DATACENTER = 'us-central2-a'
 
 # Clean up resources at the end (can be set to false in order to
 # inspect resources at the end of the run). Resources will be cleaned
--- a/demos/gce_demo.py
+++ b/demos/gce_demo.py
@@ -47,7 +47,7 @@ except ImportError:
           'Additional information about setting these values can be found '
           'in the docstring for:\n'
           'libcloud/common/google.py\n')
-    sys.exit()
+    sys.exit(1)
 
 # Add parent dir of this file's dir to sys.path (OS-agnostically)
 sys.path.append(os.path.normpath(os.path.join(os.path.dirname(__file__),
@@ -74,10 +74,14 @@ CLEANUP = True
 args = getattr(secrets, 'GCE_PARAMS', ())
 kwargs = getattr(secrets, 'GCE_KEYWORD_PARAMS', {})
 
+# Add datacenter to kwargs for Python 2.5 compatibility
+kwargs = kwargs.copy()
+kwargs['datacenter'] = DATACENTER
+
 
 # ==== HELPER FUNCTIONS ====
 def get_gce_driver():
-    driver = get_driver(Provider.GCE)(*args, datacenter=DATACENTER, **kwargs)
+    driver = get_driver(Provider.GCE)(*args, **kwargs)
     return driver
 
 
--- a/demos/gce_lb_demo.py
+++ b/demos/gce_lb_demo.py
@@ -26,7 +26,7 @@
 #      libcloud/common/google.py docstring)
 #    - Run 'python' in this directory, then:
 #        import gce_lb_demo
-#        gcelb = gce_lb_demo.get_gce_lb_driver()
+#        gcelb = gce_lb_demo.get_gcelb_driver()
 #        gcelb.list_balancers()
 #        etc.
 #    - Or, to run the full demo from the interactive python shell:
@@ -43,7 +43,13 @@ import time
 try:
     import secrets
 except ImportError:
-    secrets = None
+    print('"demos/secrets.py" not found.\n\n'
+          'Please copy secrets.py-dist to secrets.py and update the GCE* '
+          'values with appropriate authentication information.\n'
+          'Additional information about setting these values can be found '
+          'in the docstring for:\n'
+          'libcloud/common/google.py\n')
+    sys.exit()
 
 # Add parent dir of this file's dir to sys.path (OS-agnostically)
 sys.path.append(os.path.normpath(os.path.join(os.path.dirname(__file__),
--- a/demos/gce_lb_demo.py
+++ b/demos/gce_lb_demo.py
@@ -49,7 +49,7 @@ except ImportError:
           'Additional information about setting these values can be found '
           'in the docstring for:\n'
           'libcloud/common/google.py\n')
-    sys.exit()
+    sys.exit(1)
 
 # Add parent dir of this file's dir to sys.path (OS-agnostically)
 sys.path.append(os.path.normpath(os.path.join(os.path.dirname(__file__),
@@ -82,10 +82,14 @@ CLEANUP = True
 args = getattr(secrets, 'GCE_PARAMS', ())
 kwargs = getattr(secrets, 'GCE_KEYWORD_PARAMS', {})
 
+# Add datacenter to kwargs for Python 2.5 compatibility
+kwargs = kwargs.copy()
+kwargs['datacenter'] = DATACENTER
+
 
 # ==== HELPER FUNCTIONS ====
 def get_gce_driver():
-    driver = get_driver(Provider.GCE)(*args, datacenter=DATACENTER, **kwargs)
+    driver = get_driver(Provider.GCE)(*args, **kwargs)
     return driver
 
 
--- a/demos/gce_lb_demo.py
+++ b/demos/gce_lb_demo.py
@@ -101,8 +101,7 @@ def get_gcelb_driver(gce_driver=None):
     if gce_driver:
         driver = get_driver_lb(Provider_lb.GCE)(gce_driver=gce_driver)
     else:
-        driver = get_driver_lb(Provider_lb.GCE)(*args, datacenter=DATACENTER,
-                                                **kwargs)
+        driver = get_driver_lb(Provider_lb.GCE)(*args, **kwargs)
     return driver
 
 
--- a/demos/gce_lb_demo.py
+++ b/demos/gce_lb_demo.py
@@ -1,0 +1,285 @@
+#!/usr/bin/env python
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+# This example performs several tasks on Google Compute Engine and the GCE
+# Load Balancer.  It can be run directly or can be imported into an
+# interactive python session.  This can also serve as an integration test for
+# the GCE Load Balancer Driver.
+#
+# To run interactively:
+#    - Make sure you have valid values in secrets.py
+#      (For more information about setting up your credentials, see the
+#      libcloud/common/google.py docstring)
+#    - Run 'python' in this directory, then:
+#        import gce_lb_demo
+#        gcelb = gce_lb_demo.get_gce_lb_driver()
+#        gcelb.list_balancers()
+#        etc.
+#    - Or, to run the full demo from the interactive python shell:
+#        import gce_lb_demo
+#        gce_lb_demo.CLEANUP = False               # optional
+#        gce_lb_demo.MAX_NODES = 4                 # optional
+#        gce_lb_demo.DATACENTER = 'us-central1-a'  # optional
+#        gce_lb_demo.main()
+
+import os.path
+import sys
+import time
+
+try:
+    import secrets
+except ImportError:
+    secrets = None
+
+# Add parent dir of this file's dir to sys.path (OS-agnostically)
+sys.path.append(os.path.normpath(os.path.join(os.path.dirname(__file__),
+                                 os.path.pardir)))
+
+from libcloud.utils.py3 import urllib2
+
+# This demo uses both the Compute driver and the LoadBalancer driver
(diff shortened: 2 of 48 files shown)


Kami:
Thanks and sorry for the delay. This pull request is slightly larger then usual so I will take a more in depth look sometime in the next couple of days.

We've recently bootstrap new Sphinx documentation so it would be great if you can add some usage example and docs about the GCE drivers there - https://github.com/apache/libcloud/tree/trunk/docs

wrigri:
No problem.  As you can see, I've been making some additional minor fixes as a colleague of mine has been using the GCE driver.  I'm happy to contribute some documentation, so I'll start looking at that.

demos/gce_lb_demo.py
```diff
@@ -0,0 +1,293 @@
...
+#        gce_lb_demo.main()
+
+import os.path
+import sys
+import time
+
+try:
+    import secrets
+except ImportError:
+    secrets = None
```
Kami: Hm, shouldn't we exit here if `secrets.py` file doesn't exist and print a friendly message which tells the user how to update the secrets file?

libcloud/compute/drivers/gce.py
```diff
@@ -499,6 +711,47 @@ def ex_list_firewalls(self):
...
+        @type     region: C{str} or L{GCERegion} or C{None}
+
+        @return: A list of forwarding rule objects.
+        @rtype: C{list} of L{GCEForwardingRule}
+        """
+        list_forwarding_rules = []
+        # Use provided region or default region
+        region = region or self.region
+        # Setting region to 'all' overrides the default region
+        if region == 'all':
```
Kami: I noticed this pattern repeated a couple of times.

Is it possible to refactor it in a separate function and re-use this function?

libcloud/compute/drivers/gce.py
```diff
@@ -1006,6 +1440,58 @@ def ex_create_multiple_nodes(self, base_name, size, image, number,
...
+        region = region or self.region
+        targetpool_data = {}
+        targetpool_data['name'] = name
+        if not hasattr(region, 'name'):
+            region = self.ex_get_region(region)
+        targetpool_data['region'] = region.extra['selfLink']
+
+        if healthchecks:
+            if not hasattr(healthchecks[0], 'name'):
+                hc_list = [self.ex_get_healthcheck(h).extra['selfLink'] for h
```
Kami, line 1494: I also noticed this repeated a couple of times - it would be good to add a note to documentation which says something along the lines of "Some functions allows you to pass in either a list of objects or a list strings. In some cases passing in a list of strings will result in additional <list length> HTTP requests".

libcloud/compute/drivers/gce.py
```diff
@@ -1051,7 +1539,42 @@ def create_volume(self, size, name, location=None, image=None,
...
+            hc_data['host'] = healthcheck.extra['host']
+        if healthcheck.extra['description']:
+            hc_data['description'] = healthcheck.extra['description']
+
+        request = '/global/httpHealthChecks/%s' % healthcheck.name
+
+        response = self.connection.async_request(request, method='PUT',
+                                                 data=hc_data).object
+        if 'error' in response:
+            self._categorize_error(response['error'])
```
Kami: I also noticed this pattern repeated all over.

Can we use the response status code and centralize the error handling in the `Response` class so we don't need to repeat this if check in many function bodies?

[Here](https://github.com/apache/libcloud/blob/trunk/libcloud/dns/drivers/zerigo.py#L66-L95) is an example of how we do that in the Zerigo driver.

libcloud/compute/drivers/gce.py
```diff
@@ -1086,6 +1609,138 @@ def ex_update_firewall(self, firewall):
...
+        response = self.connection.async_request(request, method='POST',
+                                                 data=targetpool_data).object
+        if 'error' in response:
+            self._categorize_error(response['error'])
+        else:
+            # Remove node object from node list
+            index = None
+            for i, nd in enumerate(targetpool.nodes):
+                if nd.name == node.name:
+                    index = i
```
Kami: Can we break here on the first match?

libcloud/compute/drivers/gce.py
```diff
@@ -1407,17 +2120,36 @@ def ex_get_address(self, name, region=None):
...
 
         @return:  An Address object for the address
         @rtype:   L{GCEAddress}
         """
-        address_region = region or self._find_zone(name, 'addresses',
-                                                   region=True)
-        request = '/regions/%s/addresses/%s' % (address_region, name)
+        region = region or self.region or self._find_zone(name, 'addresses',
+                                                          region=True)
+        if not region:
```
Kami: Sam here - can we refactor the next couple of lines in a separate function?

libcloud/compute/drivers/gce.py
```diff
@@ -1407,17 +2120,36 @@ def ex_get_address(self, name, region=None):
...
+            raise ResourceNotFoundError(
+                'Address %s not found in any region.' % name, 404)
```
Kami: I prefer to always use parenthesis around the argument(s) when doing string formatting. I've seen missing parenthesis cause explosions multiple times - people add another argument, but forgot to add parenthesis.

libcloud/compute/drivers/gce.py
```diff
@@ -1559,7 +2373,12 @@ def ex_get_zone(self, name):
         if short_name in self.zone_dict:
             return self.zone_dict[short_name]
         # Otherwise, look up zone information
-        response = self.connection.request(request, method='GET').object
+        try:
+            response = self.connection.request(request, method='GET').object
+        # If zone is not found, the response is html and not json, so it fails
+        # parsing
+        except MalformedResponseError:
+            return None
```
Kami, line 2328: (off topic)

Google, that's kinda nasty... But then again, I've also seen a lot of worse things when working with other provider APIs :P

libcloud/compute/drivers/gce.py
```diff
@@ -1767,10 +2674,45 @@ def _to_storage_volume(self, volume):
...
+        """
+        extra = {}
+        extra['selfLink'] = targetpool['selfLink']
+        extra['description'] = targetpool.get('description')
+        region = self.ex_get_region(targetpool['region'])
+        healthcheck_list = [self.ex_get_healthcheck(h.split('/')[-1]) for h
+                            in targetpool.get('healthChecks', [])]
+        node_list = []
+        for n in targetpool.get('instances', []):
+            node_path = n.split('/')
```
Kami: It would be good to split this into a separate function which returns a dict - e.g. `get_components_from_node_path.`

libcloud/test/compute/test_gce.py
```diff
@@ -96,13 +109,27 @@ def test_ex_list_addresses(self):
         self.assertEqual(len(address_list_all), 4)
         self.assertEqual(address_list[0].name, 'libcloud-demo-address')
         self.assertEqual(address_list_uc1[0].name, 'libcloud-demo-address')
-        #self.assertEqual(address_list_all[0].name, 'lcaddress')
+
+    def test_ex_list_healthchecks(self):
```
Kami, line 115: Side note about the tests - a bunch of compute tests randomly fail under Python 3.3 in trunk.

I think that is because some of the tests rely on the dictionary being ordered. This is generally a bad practice and won't work in Python 3.3 where hash randomization has been enabled by default (http://docs.python.org/3/whatsnew/3.3.html).

For now I commented some of those tests out, but it would be great if you could change them to not rely on the dictionary ordering.

wrigri:
Thanks for the comments, I should be able to address most of them today. A couple of notes:
- Like you, I wasn't happy with the way that various exceptions were being handled, so I re-wrote the exception handling in a separate branch last week which addresses several of the issues that you noted (including that MalformedResponseError for missing zones that occurred because one of the zones was in a weird state the week before last, and moving it into the Response class).  I'll merge that branch in today (but squash the commits).
- Thanks for the heads-up about the dictionary ordering thing.  I'm actually already aware of that, since I've used previous versions of Python where the randomization has been turned on.  I'm surprised I let any of those slip through, but I'll track them down and get them fixed.

Kami:
Sounds good. Thanks!

wrigri:
Do you have some examples of the tests that are failing?  I can't seem to reproduce the failures when I turn on hash randomization (and I was able to put in a temporary test that _would_ fail intermittently with the randomization turned on, to verify that it was actually randomized).

libcloud/compute/drivers/gce.py
wrigri, line 1494: I added some additional information to the docstring for the GCE Driver.  Let me know if that makes sense to you or not.

wrigri:
Nevermind.  I see what you are referring to on the tests now.  (I hadn't seen that you had committed the change, so I didn't notice when I merged those changes into my branch.)  Anyway, i'll add back some extra tests to replace the removed ones.

libcloud/compute/drivers/gce.py
Kami, line 1494: Looks good.

Need to make sure we also don't forget to add this to Sphinx documentation once it's available.

libcloud/common/google.py
```diff
@@ -101,10 +103,121 @@ def __repr__(self):
...
+        try:
+            body = json.loads(self.body)
+        except:
+            # If there is both a JSON parsing error and an unsuccessful http
+            # response (like a 404), we want to raise the http error and not
+            # the JSON one, so don't raise JsonParseError here.
+            body = self.body
+            json_error = True
+
+        if self.status in [httplib.OK, httplib.CREATED, httplib.ACCEPTED]:
```
Kami, line 188: Slightly confused - does Google always return those codes, even on error?

libcloud/common/google.py
```diff
@@ -101,10 +103,121 @@ def __repr__(self):
...
+        if 'errors' in body['error']:
+            err = body['error']['errors'][0]
+        else:
+            err = body['error']
+
+        code = err.get('code')
+        message = err.get('message')
+        return (code, message)
+
+    def parse_body(self):
```
Kami, line 168: This looks way better, but a the "libcloud way" would be to do the following:
- Modify `success` method to only return `True` if the status code indicates success
- If `success` returns `False`, `parse_error` will be called where you can parse error from the body and raise appropriate exception

This approach relies on provider returning appropriate status codes on error response though. Looking at the code bellow, this might not be the case though.

libcloud/common/google.py
wrigri, line 188: Correct.  In many cases, there can be "errors" that still return a 200 code (which was why I was previously looking for "error" in the JSON response, and then calling that _categorize_error() method.  Except for the 404, most "expected" errors do not return http error codes.

libcloud/common/google.py
wrigri, line 168: I could probably split it up a bit, and add the 404 and other errors to parse_error, but I would still need to leave the ResourceInUse/QuotaExceeded/etc. error parsing in parse_body.

libcloud/common/google.py
Kami, line 188: I see. Yeah, that's unfortunate and means we can't really split it up and move most of the error parsing to `parse_error`.

You should also bug your colleagues to use 4xx error code for error conditions and make everyone's life easier :)

demos/gce_demo.py
```diff
@@ -41,7 +41,13 @@
...
     import secrets
 except ImportError:
-    secrets = None
+    print('"demos/secrets.py" not found.\n\n'
+          'Please copy secrets.py-dist to secrets.py and update the GCE* '
+          'values with appropriate authentication information.\n'
+          'Additional information about setting these values can be found '
+          'in the docstring for:\n'
+          'libcloud/common/google.py\n')
+    sys.exit()
```
Kami: Minor thing - exiting with a non-zero code would be better.

demos/gce_lb_demo.py
```diff
@@ -0,0 +1,299 @@
...
+try:
+    import secrets
+except ImportError:
+    print('"demos/secrets.py" not found.\n\n'
+          'Please copy secrets.py-dist to secrets.py and update the GCE* '
+          'values with appropriate authentication information.\n'
+          'Additional information about setting these values can be found '
+          'in the docstring for:\n'
+          'libcloud/common/google.py\n')
+    sys.exit()
```
Kami: Same here.

demos/gce_lb_demo.py
```diff
@@ -0,0 +1,299 @@
...
+# at the beginning regardless.
+CLEANUP = True
+
+args = getattr(secrets, 'GCE_PARAMS', ())
+kwargs = getattr(secrets, 'GCE_KEYWORD_PARAMS', {})
+
+
+# ==== HELPER FUNCTIONS ====
+def get_gce_driver():
+    driver = get_driver(Provider.GCE)(*args, datacenter=DATACENTER, **kwargs)
```
Kami: Does this pass under Python 2.5? IIRC you can't mix kwargs with arg and kwarg expansion.

Kami: Yeah, just checked it:

``` python

Python 2.5.6 (r256:88840, Aug 21 2013, 17:00:00) 
[GCC 4.6.3] on linux2
>>> def foo(a, b, c): pass

>>> foo(*(1,2), b='foo', **{'c': ''})
  File "<stdin>", line 1
    foo(*(1,2), b='foo', **{'c': ''})
                ^
SyntaxError: invalid syntax
```

Solution would be to only use arg and kwarg expansion. So doing something like:

``` python
kwargs = kwargs.copy()
kwargs['datacenter'] = DATACENTER
...
```

demos/gce_lb_demo.py
```diff
@@ -0,0 +1,299 @@
...
+
+def get_gcelb_driver(gce_driver=None):
+    # The GCE Load Balancer driver uses the GCE Compute driver for all of its
+    # API calls.  You can either provide the driver directly, or provide the
+    # same authentication information so the the LB driver can get its own
+    # Compute driver.
+    if gce_driver:
+        driver = get_driver_lb(Provider_lb.GCE)(gce_driver=gce_driver)
+    else:
+        driver = get_driver_lb(Provider_lb.GCE)(*args, datacenter=DATACENTER,
```
Kami: Same here (with regards to mixing arg expansion + kwargs).

libcloud/compute/drivers/gce.py
```diff
@@ -1353,7 +2004,14 @@ def ex_destroy_multiple_nodes(self, nodelist, ignore_errors=True,
...
             request = '/zones/%s/instances/%s' % (node.extra['zone'].name,
                                                   node.name)
-            response = self.connection.request(request, method='DELETE').object
+            try:
+                response = self.connection.request(request,
+                                                   method='DELETE').object
+            except:
+                e = sys.exc_info()[1]
+                response = None
+                if not ignore_errors:
```
Kami: I won't if we could refactor this `ignore_errors` functionality in a way so it can be reused.

Kami:
I've looked over the pull request again.

It's close - there are some minor issues left + missing load balancer wrapper tests.

wrigri:
OK, I think that I've addressed all of the comments.  Let me know if there are any other final things that need to be done before we can get this merged in.

Kami:
@wrigri Thanks and sorry for the delay.

We don't receive notifications when you add a comment so next time please mention my nickname so I will receive a notification and respond faster.

Anyway, please squash all the commits and attach latest patch to the JIRA ticket and I will get it merged.

wrigri:
@Kami Thanks for taking a look.  I have attached a patch to the ticket.  (And sorry for not including your nickname, I'll make sure I do that in the future.)

Kami:
Forgot to add a comment - this pull request has been merged and it can be closed now.

Thanks again!

wrigri:
Closing pull request.

PR #143, opened 2013-09-03
LIBCLOUD-390: fix missing Content-Length header

ogrisel:
On the Storage API for create_container method on the Azure Storage Service.


Kami:
Thanks, but as noted on JIRA (https://issues.apache.org/jira/browse/LIBCLOUD-390#comment-13757164) this shouldn't be needed, because the actual root cause of this issue has been fixed in trunk.

Kami:
This has been fixed as part of https://issues.apache.org/jira/browse/LIBCLOUD-390.

Can you please close the pull request now? (I don't have permissions to do it since this is a read-only Apache mirror)

ogrisel:
Done.

PR #146, opened 2013-09-10
Add load balancer stats to libcloud

rphillips:


Kami:
Thanks.

Like I've said on the IRC, the pull request looks good but it needs a test case before I can merge it.

PR #147, opened 2013-09-12
Use string for RecordType enum value

Kami:
@Jc2k mentioned this "issue" on the IRC yesterday.

Using integer value for enums is a common practice in Python, but I believe that using a string value is better in this case. It makes debugging and everything else easier.

With this approach, user can also directly pass a string value for `type` argument and it makes getting rid of `RecordType` type in the future easier, if we ever want to do that.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -114,6 +114,12 @@ Changes with Apache Libcloud in development
      the provided key is not set.
      [Tomaz Muraus]
 
+   - Use string instead of integer for RecordType ENUM value.
+     Note: If you directly use an integer instead of RecordType ENUM class you
+     need to update your code to use the RecordType ENUM otherwise the code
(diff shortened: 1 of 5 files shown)


Jc2k:
Nice. I thought it was going to be much harder than this to do 'right', but this is straightforward and pragmatic. And if you were using the constants in your coded properly then its an entirely backwards compatible change.

Still worth a mention in the changelog in case anyone has code that depends on a number somehow?

Otherwise, big +1 from me.

Kami:
@Jc2k Yeah, I've updated the CHANGES file - 8475777c92756be18dacc2dca352747de732175b. I also plan to update the "Upgrade notes page" before releasing a new version.

Kami:
Merged it into trunk.

PR #152, opened 2013-09-27
Issue LIBCLOUD-402: Support OS-DCF:diskConfig

briancurtin:
Rackspace is implementing a change to server creation where disk partitioning will not automatically fill the entire disk, leaving the option to the user to expand the disk or not. When creating nodes, you have the ability to specify if you want a MANUAL disk config, or if you want an AUTO disk config. The former results in quicker server creation,
while the latter results in the entire disk being used.

This change is exposed via an extension to the OpenStack compute APIs, which Rackspace's Cloud Servers are built on top of. "`OS-DCF:diskConfig`" is an attribute to be set in message bodies POST'ed to /servers, and is exposed as the "`ex_disk_config`" parameter to `OpenStack_1_1_NodeDriver.create_node`.


Kami:
Patch looks good to me.

Please attach it to the corresponding JIRA ticket and I'll merge it into trunk.

Kami:
Actually, I just started bootstraping provider / driver specific documentation (https://github.com/apache/libcloud/tree/trunk/docs/compute/drivers) so it would be good to add a section on how to use this option and what kind of consequences it has to the OpenStack / Rackspace driver documentation.

Kami:
Merged, this PR can be closed now.

Thanks!

PR #153, opened 2013-09-30
Issue LIBCLOUD-404: Add nephoscale cloud driver

mgogoulos:
nephoscale driver (with size prices)


Kami:
Thanks!

I'll review it and add some in-line comments.

libcloud/compute/drivers/nephoscale.py
```diff
@@ -0,0 +1,460 @@
...
+from libcloud.utils.py3 import b
+from libcloud.utils.py3 import urlencode
+
+from libcloud.compute.providers import Provider
+from libcloud.common.base import JsonResponse, ConnectionUserAndKey
+from libcloud.compute.types import NodeState, InvalidCredsError
+from libcloud.compute.base import (Node, NodeDriver, NodeImage, NodeSize,
+                                    NodeLocation)
+
+API_HOST = "api.nephoscale.com"
```
Kami: (minor style thing) Please use single quotes around strings.

libcloud/compute/drivers/nephoscale.py
```diff
@@ -0,0 +1,460 @@
...
+                (self.id, self.name))
+
+
+class NephoscaleResponse(JsonResponse):
+    """
+    Nephoscale API Response
+    """
+
+    def parse_error(self):
+        if self.status == 401:
```
Kami: Please use `httplib` constants (`httplib.UNAUTHORIZED`, ...)

libcloud/compute/drivers/nephoscale.py
```diff
@@ -0,0 +1,460 @@
...
+        super(NephoscaleNodeDriver, self).__init__(*args, **kwargs)
+
+    def list_locations(self):
+        result = self.connection.request('/datacenter/zone/').object
+        locations = []
+        for value in result.get('data', []):
+            location = NodeLocation(id=value.get('id'),
+                                    name=value.get('name'),
+                                    country='US',
+                                    driver=self.connection.driver)
```
Kami: You could just do `driver=self` :)

libcloud/compute/drivers/nephoscale.py
```diff
@@ -0,0 +1,460 @@
...
+    def list_nodes(self):
+        """
+        List available Nodes
+        """
+        result = self.connection.request('/server/cloud/').object
+        nodes = [self._to_node(value) for value in result.get('data', [])]
+        return nodes
+
+    def rename_node(self, node, name, hostname=None):
+        "rename a cloud server, optionally specify hostname too"
```
Kami: For consistency, please use triple double quote around docstrings.

libcloud/compute/drivers/nephoscale.py
```diff
@@ -0,0 +1,460 @@
...
+
+    def list_nodes(self):
+        """
+        List available Nodes
+        """
+        result = self.connection.request('/server/cloud/').object
+        nodes = [self._to_node(value) for value in result.get('data', [])]
+        return nodes
+
+    def rename_node(self, node, name, hostname=None):
```
Kami, line 203: Please also document the arguments (node, name and hostname) using Sphinx style docstrings.

libcloud/compute/drivers/nephoscale.py
```diff
@@ -0,0 +1,460 @@
...
+            'public_key': public_key,
+            'key_group': key_group
+            #key_group: The group for the key where Server=1 and Console=4
+        }
+        params = urlencode(data)
+        try:
+            result = self.connection.request('/key/sshrsa/', data=params,
+                                             method='POST').object
+        except Exception:
+            e = sys.exc_info()[1]
```
Kami: Hm, this `try` / `catch` block seems to be redundant since you are re-throwing the same exception.

libcloud/compute/drivers/nephoscale.py
```diff
@@ -0,0 +1,460 @@
...
+        keys = [self._to_ssh_key(value) for value in result.get('data', [])]
+        return keys
+
+    def list_password_keys(self):
+        "list password console and password server keys"
+        result = self.connection.request('/key/password/').object
+        keys = [self._to_ssh_key(value) for value in result.get('data', [])]
+        return keys
+
+    def add_ssh_key(self, name, public_key, key_group=1):
```
Kami: We recently started unifying all the extension methods for managing key pairs. Please follow the guidelines for that outlined in this ticket https://issues.apache.org/jira/browse/LIBCLOUD-392

libcloud/compute/drivers/nephoscale.py
```diff
@@ -0,0 +1,460 @@
...
+            'password': password,
+            'key_group': key_group
+             #key_group: The group for the key, where Server=1 and Console=4
+        }
+        params = urlencode(data)
+        try:
+            result = self.connection.request('/key/password/', data=params,
+                                             method='POST').object
+        except Exception:
+            e = sys.exc_info()[1]
```
Kami: Same here (re-throwing original exception)

libcloud/compute/drivers/nephoscale.py
```diff
@@ -0,0 +1,460 @@
...
+            raise Exception("Failed to create node %s" % e)
+        node = Node(id='', name=name, state='', public_ips='', private_ips='',
+                    driver=self.connection.driver)
+        #try to get the created node public ips, for use in deploy_node
+        #At this point we don't have the id of the newly created Node,
+        #so search name in nodes
+
+        created_node = False
+        while connect_attempts > 0:
```
Kami: The convention for `create_node` method is to not block and immediately return available node information.

If you want to block you should use `deploy_node` - deploy node does exactly that. It waits for node to come online and has at least one IP address assigned.

libcloud/compute/drivers/nephoscale.py
```diff
@@ -0,0 +1,460 @@
...
+        state = NODE_STATE_MAP.get(data.get('power_status'), '4')
+        public_ips = []
+        private_ips = []
+        ip_addresses = data.get('ipaddresses', '')
+        #E.g. "ipaddresses": "198.120.14.6, 10.132.60.1"
+        if ip_addresses:
+            ip_addresses_list = ip_addresses.split(',')
+            for ip in ip_addresses_list:
+                ip = ip.replace(' ', '')
+                if ip.startswith('10.') or ip.startswith('192.168'):
```
Kami: You can use `libcloud.compute.base.is_private_subnet` method to determine if an IP address is private.

libcloud/compute/drivers/nephoscale.py
mgogoulos: Mmm, I've added that as soon as NephoScale receives a successful request for create machine, it immediately sends an OK response. This does not contain anything related to our -to be created- node.

I preferred to implement the check on create_node, so that the generic deploy_node works, rather than write two funcions (create and deploy)

mgogoulos: I can add a nowait=[False, True] parameter in create_node so if specified it either waits for the node to be created, or returns after successfully sending the request. What do you think?

Kami: 1. What does the response contain, empty body? Or does it at least contain an ID?
2. Generic deploy_node should work as long as node object has an ID assigned.
3. I'm fine with adding `ex_wait` argument which defaults to `False`. This way it doesn't break a standard API.

If you decide to do that, you should use `libcloud.compute.base.wait_until_running` method which does all the magic for you.

mgogoulos: I've added an ex_wait flag which defaults to False. The response from NephoScale does not contain an ID of the machine to be created, only a response that the job will be processed. Most probably they have a queue system. 

So if create_node is called without specifying that flag, it returns without waiting for the node to be created. If ex_wait=True, it waits the machine to be created and then returns the node.

Kami: Ah, oh so they have an async API and don't return a node id.

Edit: I just checked the API docs (http://docs.nephoscale.com/#!/reference/server). It looks like they return a job ID which equals to the server id.

Everywhere they reference server id attribute as "The ID of the job object".

mgogoulos:
Committed the changes!

libcloud/compute/drivers/nephoscale.py
```diff
@@ -0,0 +1,474 @@
...
+        """
+        if ssh:
+            result = self.connection.request('/key/sshrsa/%s/' % key_id,
+                                             method='DELETE').object
+        else:
+            result = self.connection.request('/key/password/%s/' % key_id,
+                                         method='DELETE').object
+        return result.get('response') in VALID_RESPONSE_CODES
+
+    def create_node(self, **kwargs):
```
Kami: I just noticed `**kwargs` here.

There doesn't seem to be a reason to use `**kwargs` here so you should define arguments normally. You should only use kwargs if there is real,  because they make many thing hard / impossible (method signature introspection, editor auto-completion, makes things harder to read - can't just glance to see which arguments method takes, etc.).

libcloud/compute/drivers/nephoscale.py
```diff
@@ -0,0 +1,474 @@
...
+    """
+
+    type = Provider.NEPHOSCALE
+    api_name = 'nephoscale'
+    name = 'NephoScale'
+    website = 'http://www.nephoscale.com'
+    connectionCls = NephoscaleConnection
+    features = {'create_node': ['ssh_key']}
+
+    def __init__(self, *args, **kwargs):
```
Kami: This method is redundant - you can remove it. Constructor from the parent class should do the trick.

libcloud/compute/drivers/nephoscale.py
```diff
@@ -0,0 +1,474 @@
...
+            'password': password,
+            'key_group': key_group
+             #key_group: The group for the key, where Server=1 and Console=4
+        }
+        params = urlencode(data)
+        result = self.connection.request('/key/password/', data=params,
+                                         method='POST').object
+        return result.get('data', {}).get('id', '')
+
+    def delete_keypair(self, key_id, ssh=False):
```
Kami: This one should be called `ex_delete_keypair`.

libcloud/compute/drivers/nephoscale.py
```diff
@@ -0,0 +1,474 @@
...
+        if not (ssh or password):
+            result = self.connection.request('/key/').object
+        keys = [self._to_key(value) for value in result.get('data', [])]
+
+        if key_group:
+            keys = [key for key in keys if
+                    key.key_group == key_group]
+        return keys
+
+    def create_ssh_key(self, name, public_key, key_group=1):
```
Kami: This one should be called `ex_create_keypair`.

libcloud/compute/drivers/nephoscale.py
```diff
@@ -0,0 +1,474 @@
...
+                                         % node.id, method='POST').object
+        return result.get('response') in VALID_RESPONSE_CODES
+
+    def destroy_node(self, node):
+        """destroy a node"""
+        result = self.connection.request('/server/cloud/%s/' % node.id,
+                                         method='DELETE').object
+        return result.get('response') in VALID_RESPONSE_CODES
+
+    def list_keypairs(self, ssh=False, password=False, key_group=None):
```
Kami: This one should be called `ex_list_keypairs`.

libcloud/compute/drivers/nephoscale.py
```diff
@@ -0,0 +1,474 @@
...
+        return node
+
+    def _to_key(self, data):
+        return NodeKey(id=data.get('id'),
+                      name=data.get('name'),
+                      password=data.get('password'),
+                      key_group=data.get('key_group'),
+                      public_key=data.get('public_key'))
+
+    def random_password(self, size=8,
```
Kami: To be on the safe side, I would rather use choreographically secure random generator (`os.urandom` should do the trick).

mgogoulos: True but we can't use other than A-Z, a-z, 0-9 for that type of passwords. NephoScale won't accept other chars

Kami: You can use the same approach we use in `libcloud.compute.base`:

``` python
value = os.urandom(16)
value = binascii.hexlify(value).decode('ascii')
```

The resulting value should match the a-z0-9 criteria :)

mgogoulos:
So what's left here for the merge?

Kami:
@mgogoulos Hey, there are still a couple of unresolved comments (using kwargs, etc.).

So those things needs to be addressed + it needs tests for all the functionality and it should be ready to be merged :)

mgogoulos:
I've added tests for all functionality and fixed the kwargs on create_node!

mgogoulos:
Please have a look and let me know if there are pending changes!

libcloud/compute/drivers/nephoscale.py
```diff
@@ -0,0 +1,450 @@
...
+        return result.get('response') in VALID_RESPONSE_CODES
+
+    def ex_list_keypairs(self, ssh=False, password=False, key_group=None):
+        """
+        List available console and server keys
+
+        :keyword ssh: if specified, show ssh keys only
+        :type    ssh: ``bool``
+
+        :keyword password: if specified, show password keys only
```
Kami: Just want to confirm - "password keys" means ssh keys protected with a password?

If that is indeed the case, it would be great to update the docstring to say that.

libcloud/compute/drivers/nephoscale.py
```diff
@@ -0,0 +1,450 @@
...
+        :type    password: ``bool``
+
+        :keyword key_group: if specified, show keys with this key_group only
+                            eg key_group=4 for console password keys
+        :type    key_group: ``int``
+
+        :rtype: ``list`` of :class:`NodeKey`
+        """
+        if ssh:
```
Kami, line 262: It looks like `password` and `ssh` argument are mutually exclusive the way it's currently implemented. It would be good to document that and throw if both arguments are provided.

libcloud/compute/drivers/nephoscale.py
Kami: @mgogoulos Hey, can you please cross-check my last comment / observation?

If what I said is indeed the case, we should update `create_node` to assign an id to the returned `Node` object.

Kami:
Sorry for the delay.

There are also a couple of pep8 issues in this branch.

Once comments posted above are addressed, please squash all the commits, attach latest patch to the JIRA ticket and I will merge it into master.

libcloud/compute/drivers/nephoscale.py
mgogoulos: NephoScale has two types of keys: ssh keys and password keys, that can be used to deploy a server. I will document this further on the code

libcloud/compute/drivers/nephoscale.py
mgogoulos, line 262: The API request differs in case of ssh keys with password keys, so ex_list_keypairs if called with no arguments will show all availabl keys, while if you specify password or ssh it will show these keys only. I'll document that on the code

Kami, line 262: Yes, but the problem is that it won't work as expected if you pass `True` for both arguments.

Currently, in this case it will only return password keys. It should do one of the following:
1. Arguments are mutually exclusive - if you pass both an exception is thrown
2. If you pass `True` for both arguments, this method performs two requests (for regular and password protected keys) and combined the result.

libcloud/compute/drivers/nephoscale.py
Kami: OK, thanks for the clarification. I also checked the API docs and yes, this term is kinda confusing (http://docs.nephoscale.com/#!/reference/key).

It looks like "password key" is simply root password for Linux boxes and administrator password for Windows boxes.

libcloud/compute/drivers/nephoscale.py
```diff
@@ -0,0 +1,455 @@
...
+        }
+
+        params = urlencode(data)
+        try:
+            node = self.connection.request('/server/cloud/', data=params,
+                                           method='POST')
+        except Exception:
+            e = sys.exc_info()[1]
+            raise Exception("Failed to create node %s" % e)
+        node = Node(id='', name=name, state='', public_ips='', private_ips='',
```
Kami: There are a couple more things which needs to be addressed here:
1. public_ips and private_ips should be a list (empty list in this case)
2. Initial state should be set to `NodeState.PENDING`
3. https://github.com/apache/libcloud/pull/153#discussion_r6751811

mgogoulos:
I see the point on 153. Submitted a patch

Kami:
@mgogoulos Sorry, I'm slightly confused, https://github.com/apache/libcloud/pull/153#discussion_r6658032 issue still seems to be open.

If my observation is correct, the code should be updated to set Node.id attribute to the job id. If not, please say so.

libcloud/compute/drivers/nephoscale.py
mgogoulos: Unfortunately this is not true. The id received is the job id most probably, and not the id of the server to be created. 

I've just sent a request to create a server, response from NephoScale is:

{u'subcode': 0, u'data': {u'id': 22091, u'resource_type': u'/cloud/initiator', u'uri': u'https://api.nephoscale.com/cloud/initiator/22091/'}, u'message': u'Your request was processed successfully.', u'success': True, u'response': 202}

Then after a few minutes the server is created, it has the id: 89061

PR #155, opened 2013-10-11
Issue LIBCLOUD-408: Pass kwargs to ex_rebuild to create node args

tildedave:
Added test that verifies this does the expected thing for the
disk configuration option.


Kami:
I've removed unnecessary try / except block (I assume you just followed an existing pattern so that's fine) and merged the patch into trunk.

Thanks!

tildedave:
Thank you!

PR #157, opened 2013-10-14
Additional methods to ec2 driver and some fixes for Cloudstack driver

gigimon:
Hello

I add some new functionality for ec2 driver:
- list_volumes
- create_volume_snapshot
- delete_snapshot
- delete_keypair
- list_images
- delete_image

Moreover I fix some bugs for cloudstack driver and create it more EC2 like (some names)

All of functions has tests.


libcloud/compute/drivers/ec2.py
```diff
@@ -648,7 +668,7 @@ def list_sizes(self, location=None):
             sizes.append(NodeSize(driver=self, **attributes))
         return sizes
 
-    def list_images(self, location=None, ex_image_ids=None):
+    def list_images(self, location=None, ex_image_ids=None, owner=None):
```
Kami: `owner` is an extension argument so please prefix it with `ex_` (`ex_owner`) and add a docstring which explains what this argument does.

libcloud/compute/drivers/ec2.py
```diff
@@ -720,6 +758,67 @@ def detach_volume(self, volume):
         self.connection.request(self.path, params=params)
         return True
 
+    def create_volume_snapshot(self, volume, description=None):
+        """
+        Create snapshot from volume
```
Kami, line 772: We've recently switched to Sphinx style docstrings so please update this code to use Sphinx style docstrings.

You can find some examples in this file.

libcloud/compute/base.py
```diff
@@ -410,13 +410,13 @@ def list_snapshots(self):
...
 
-    def snapshot(self, name):
+    def snapshot(self, name=''):
         """
         Creates a snapshot of this volume.
 
         :return s ``VolumeSnapshot``
         """
-        return self.driver.create_volume_snapshot(volume=self, name=name)
+        return self.driver.create_volume_snapshot(volume=self, description=name)
```
Kami: Confused, why did you change `name` to `description`? `create_volume_snapshot` doesn't take `description` argument.

libcloud/compute/base.py
```diff
@@ -410,13 +410,13 @@ def list_snapshots(self):
         """
         return self.driver.list_volume_snapshots(volume=self)
 
-    def snapshot(self, name):
+    def snapshot(self, name=''):
```
Kami: Why did you change the default value to an empty string?

In the existing API, `name` is a required argument and because of this it has no default value.

libcloud/compute/base.py
```diff
@@ -433,8 +433,12 @@ def __repr__(self):
 
 
 class VolumeSnapshot(object):
-    def __init__(self, driver):
+    def __init__(self, id, size, description, driver, extra=None):
```
Kami: I would add `name` argument and change the method signature to:

``` python
def __init__(self, id, name, description=none, size=None, driver, extra=None):
```

This way only `id` and `name` are a required argument. This seems more reasonable for a default since not all of the providers will expose all those attributes.

Kami: Please also add docstring which explains what each argument means.

libcloud/compute/drivers/cloudstack.py
```diff
@@ -97,13 +97,15 @@ class CloudStackPortForwardingRule(object):
     "A Port forwarding rule for Source NAT."
 
     def __init__(self, node, rule_id, address, protocol, public_port,
-                 private_port):
+                 private_port, public_end_port=None, private_end_port=None):
         self.node = node
```
Kami, line 132: It would be good to add a docstring which explains what each argument means and also how those two new argument behave (e.g. if you don't specify `end` single port is used, otherwise port range is used).

libcloud/compute/drivers/cloudstack.py
```diff
@@ -302,17 +304,21 @@ def list_nodes(self):
...
-                for r in result.get('portforwardingrule', []):
-                    if r['virtualmachineid'] == node.id:
-                        rule = CloudStackPortForwardingRule(node, r['id'],
-                                                            addr,
-                                                            r['protocol']
-                                                            .upper(),
-                                                            r['publicport'],
-                                                            r['privateport'])
-                        rules.append(rule)
+            public_ips = self.ex_list_public_ips()
```
Kami, line 338: I know you haven't worked on the original method and it looks like I missed this during the original review.

Anyway, this method it's gigantic and it would be good to split it into multiple smaller methods. Specifically, those parts could be split out:
1. List IP forwarding rule part should be split into `ex_list_ip_forwarding_rules` method
2. List port forwarding rule part should be split into `ex_list_port_forwarding_rules` method

libcloud/compute/drivers/ec2.py
```diff
@@ -598,11 +599,30 @@ def _to_volume(self, element, name):
...
+        state = findtext(element=element, xpath='status', namespace=NAMESPACE)
+        create_time = findtext(element=element, xpath='createTime', namespace=NAMESPACE)
         return StorageVolume(id=volId,
                              name=name,
                              size=int(size),
-                             driver=self)
+                             driver=self,
+                             extra={'state': state,
+                                    'device': findtext(element=element, xpath='attachmentSet/item/device', namespace=NAMESPACE),
+                                    'create-time': datetime.strptime(create_time, '%Y-%m-%dT%H:%M:%S.000Z')})
```
Kami: You should also be able to use `libcloud.utils.iso8601.parse_date` function.

libcloud/compute/drivers/ec2.py
```diff
@@ -720,6 +758,67 @@ def detach_volume(self, volume):
         self.connection.request(self.path, params=params)
         return True
 
+    def create_volume_snapshot(self, volume, description=None):
```
Kami: EC2 doesn't allow specifying snapshot name?

libcloud/compute/drivers/ec2.py
```diff
@@ -745,6 +844,25 @@ def ex_create_keypair(self, name):
             'keyFingerprint': key_fingerprint,
         }
 
+    def ex_destroy_keypair(self, name):
```
Kami: To be consistent with other drivers this method should be called `ex_delete_keypair` and take `keypair` as an argument.

For example, see https://github.com/apache/libcloud/blob/trunk/libcloud/compute/drivers/openstack.py#L1822

Kami:
Thanks!

I've added some in-line comments to the pull request.

libcloud/compute/drivers/ec2.py
gigimon: No, EC2 has only description for snapshot
http://docs.aws.amazon.com/AWSEC2/latest/APIReference/ApiReference-query-CreateSnapshot.html

libcloud/compute/base.py
gigimon: because snapshot don't has a name only description which not required

libcloud/compute/drivers/cloudstack.py
gigimon, line 338: Yes, it's big, but set to node more information, in next revision I move some parts of code to functions

gigimon:
I have only one question with name/description for snapshotting, what I must use? rename description to name and create it required?

libcloud/compute/base.py
gigimon: But which will name EC2 not support name for snapshots?

gigimon:
I fix all except name for snapshot in StorageVolume.snapshot() and insert documentation to methods.
Will commentaries about name and description for ec2

Kami:
@gigimon OK, in this case I would do the following:
1. Change the constructor to `def __init__(self, id, size, driver, extra=None):`
2. Store `description` and `name` (for providers which expose it) in the `extra` dictionary

libcloud/compute/drivers/cloudstack.py
```diff
@@ -874,12 +923,12 @@ def ex_create_keypair(self, name, **kwargs):
         res = self._sync_request('createSSHKeyPair', name=name, **extra_args)
         return res['keypair']
 
-    def ex_delete_keypair(self, name, **kwargs):
+    def ex_delete_keypair(self, keypair, **kwargs):
```
Kami, line 926: If you look at the example I posted, `keypair` is an actual Keypair object and not a string.

Kami, line 926: @gigimon Once this issue has been addressed, please squash all the commits into a single on and I'll merge the patch.

gigimon, line 926: Sorry, but which object do you mean? In ec2 ex_list_keypairs return list with keypairs.

``` python
            keypair = {
                'keyName': findtext(element=elem, xpath='keyName',
                                    namespace=NAMESPACE),
                'keyFingerprint': findtext(element=elem,
                                           xpath='keyFingerprint',
                                           namespace=NAMESPACE).strip(),
            }
            keypairs.append(keypair)

        return keypairs
```

Kami, line 926: Sorry, I thought we have already created a base class for KeyPair objects, but that's not the case yet (https://github.com/apache/libcloud/blob/trunk/libcloud/compute/drivers/openstack.py#L1128).

Leave it as it is for now and I'll create a base Keypair object and update this code to use it some time in the near future.

gigimon:
Sorry, what do You mean at "please squash all the commits into a single on"? I'm new in pull requests and don't know what is and how :(

Kami:
@gigimon I mean merging all the commits into one.

This can be achieved using git rebase and squash - http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

Basically, you want to do:

git rebase -i <sha revision of the parent of your first commit>
type s or squash for all the commits except the first one
update commit message
done!

gigimon:
Complete, all in one commit, Thank You for help!

Kami:
@gigimon Thanks!

Also forgot to say - can you please also create a ticket on our issue tracker (https://issues.apache.org/jira/browse/LIBCLOUD), describe all the changes you made in the description and attach the patch there?

gigimon:
Ok, complete: https://issues.apache.org/jira/browse/LIBCLOUD-409

Kami:
This has been merged into trunk, you can close this pull request now.

Thanks!

PR #161, opened 2013-10-23
LIBCLOUD-415 try fix keypair delete

gigimon:
Hello

I think this fix delete keypair in Eucalyptus (ticket https://issues.apache.org/jira/browse/LIBCLOUD-415)

### DIFF
--- a/libcloud/compute/drivers/ec2.py
+++ b/libcloud/compute/drivers/ec2.py
@@ -881,7 +881,7 @@ class BaseEC2NodeDriver(NodeDriver):
         """
         params = {
             'Action': 'DeleteKeyPair',
-            'KeyName.1': keypair
+            'KeyName': keypair
         }
         result = self.connection.request(self.path, params=params).object
         element = findtext(element=result, xpath='return',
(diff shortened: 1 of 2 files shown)


Kami:
The pull request looks good, but it needs a test case which verifies that right query parameters are sent to the server (in this case, that is `KeyName`).

Also next time when making a changes like this, please link to a reference guide on the provide page (in this case, that would be http://docs.aws.amazon.com/AWSEC2/latest/APIReference/ApiReference-query-DeleteKeyPair.html).

gigimon:
Test for this function already in trunk, this "patch" fix set up a arguments correspondence to api docs

Kami:
@gigimon Yes, but the existing test case (https://github.com/apache/libcloud/blob/trunk/libcloud/test/compute/test_ec2.py#L382), doesn't test that the right arguments are sent to the API which is the important part in this case.

You can see an example of how other tests do that here - https://github.com/apache/libcloud/blob/trunk/libcloud/test/compute/test_ec2.py#L709

gigimon:
please check, I added test for arguments

Kami:
@gigimon Looks good. Please attach this patch to the JIRA issue and I'll merge it into trunk.

gigimon:
@Kami  complete, please see in JIRA

Kami:
@gigimon Merged, this pr can be closed now.

Thanks.

gigimon:
Ok, Thx!

PR #163, opened 2013-10-23
Issue LIBCLOUD-417: Add support for more EC2 Elastic IP Address operations

oldpatricka:
Add support for:

DisassociateAddress
ReleaseAddress
AllocateAddress

Also fix AssociateAddress

### DIFF
--- a/libcloud/compute/drivers/ec2.py
+++ b/libcloud/compute/drivers/ec2.py
@@ -1300,6 +1300,36 @@ class BaseEC2NodeDriver(NodeDriver):
             'Filter.0.Value.0': node.id
         })
 
+    def ex_allocate_address(self):
+        """
+        Allocate a new Elastic IP address
+
+        :return: String representation of allocated IP address
+        :rtype: ``str``
+        """
+        params = {'Action': 'AllocateAddress'}
+
+        response = self.connection.request(self.path, params=params).object
+        public_ip = findtext(element=response, xpath='publicIp',
+                                namespace=NAMESPACE)
+        return public_ip
+
+    def ex_release_address(self, elastic_ip_address):
+        """
+        Release an Elastic IP address
+
+        :param      elastic_ip_address: Elastic IP address which should be used
+        :type       elastic_ip_address: ``str``
+
+        :return: True on success, False otherwise.
+        :rtype: ``bool``
+        """
+        params = {'Action': 'ReleaseAddress'}
+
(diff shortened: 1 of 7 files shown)


libcloud/compute/drivers/ec2.py
```diff
@@ -1155,6 +1155,34 @@ def _add_instance_filter(self, params, node):
             'Filter.0.Value.0': node.id
         })
 
+    def ex_allocate_address(self):
+        """
+        Allocate a new IP address
+
+        @rtype: C{str}
```
Kami: We've recently migrated to Sphinx style docstrings (yeah, still need to update documentation). Can you please update your code to use Sphinx style docstrings?

libcloud/compute/drivers/ec2.py
```diff
@@ -1155,6 +1155,34 @@ def _add_instance_filter(self, params, node):
...
+        params = {'Action': 'AllocateAddress'}
+
+        response = self.connection.request(self.path, params=params).object
+        public_ip = findtext(element=response, xpath='publicIp',
+                                namespace=NAMESPACE)
+        return public_ip
+
+    def ex_release_address(self, elastic_ip_address):
+        """
+        Release an IP address
```
Kami: I would also update docstrings of all those new methods to clarify that those are elastic ip addresses (Release an elastic IP address and so on).

Kami:
Can you please also add some examples of how to use this functionality to - https://github.com/apache/libcloud/blob/trunk/docs/compute/drivers/ec2.rst

And yes, the file is currently empty. The reason for that is that we just recently started bootstrapping provider specific documentation.

libcloud/compute/drivers/ec2.py
oldpatricka: Good idea! I did it in a new commit (which I'll squash when this pr is ready)

oldpatricka:
Ok, I switched to Sphinx style docstrings, and added an example to the docs. How does it look now?

Kami:
@oldpatricka Looks good.

Please squash all the commits into one and attach patch to the corresponding JIRA ticket and I'll merge it into trunk.

oldpatricka:
Great, the patch is attached to the ticket (https://issues.apache.org/jira/browse/LIBCLOUD-417), so I'm closing this request. Thanks for the prompt responses!

PR #164, opened 2013-10-24
Issue LIBCLOUD-358: Add compute driver for CloudFounders CloudFrames.

bernieke:


libcloud/compute/drivers/cloudframes.py
```diff
@@ -0,0 +1,424 @@
...
+    Cloudapi connection class
+    """
+
+    repsonseCls = XMLRPCResponse
+    base_url = None
+
+    def __init__(self, key=None, secret=None, secure=True,
+                 host=None, port=80, url=None, timeout=None):
+        """
+        @keyword  key:    The username to connect with to the cloudapi
```
Kami: We've recently switched to Sphinx style docstrings so please update this pull request to use that.

libcloud/compute/drivers/cloudframes.py
```diff
@@ -0,0 +1,424 @@
...
+class CloudFramesConnection(XMLRPCConnection, ConnectionKey):
+    """
+    Cloudapi connection class
+    """
+
+    repsonseCls = XMLRPCResponse
+    base_url = None
+
+    def __init__(self, key=None, secret=None, secure=True,
+                 host=None, port=80, url=None, timeout=None):
```
Kami: You should set default value for `port` to `None`.

This way it won't explode and it will use `443` by default when `secure` is `True`.

Kami: Actually, I just noticed the comment says `secure` should always be set to `False`.

Does this means that this provider doesn't work expose https endpoint?

libcloud/compute/drivers/cloudframes.py
```diff
@@ -0,0 +1,424 @@
...
+        @type     port:   C{int}
+
+        @keyword  url:    Url to the cloudapi (can replace all above)
+        @type     url:    C{str}
+        """
+
+        super(CloudFramesConnection, self).__init__(key=key, secure=secure,
+                                                    host=host, port=port,
+                                                    url=url, timeout=timeout)
+        self._auth = base64.b64encode('%s:%s' % (key, secret))
```
Kami: To make this work with all the supported versions, you need to pass `bytes` to `b64encode` function. You also need to make sure that the return value is a string since `b64encode` returns `bytes` by default in Python 3 and above.

You can see how other drivers handle this here - https://github.com/apache/libcloud/blob/trunk/libcloud/compute/drivers/ibm_sce.py#L65

libcloud/compute/drivers/cloudframes.py
```diff
@@ -0,0 +1,424 @@
...
+        self.endpoint = url
+
+    def __getattr__(self, key):
+        return CloudFramesComponent(self, key)
```
Kami, line 148: I'm not sure that I get why this `CloudFramesComponent` abstraction is used, care to explain?

libcloud/compute/drivers/cloudframes.py
```diff
@@ -0,0 +1,424 @@
...
+        @keyword size:           The size of the new node (required)
+        @type    size:           L{NodeSize}
+
+        @keyword location:       The location to create the new node
+        @type    location:       L{NodeLocation}
+
+        @keyword defaultgateway: The default gateway to be used
+        @type    defaultgateway: C{str}
+
+        @keyword additionalinfo: Additional requirements (extra disks fi.)
```
Kami: In libcloud, the convention is to use `extra` dictionary for provider specific arguments.

libcloud/compute/drivers/cloudframes.py
```diff
@@ -0,0 +1,424 @@
...
+        @keyword name:           The name for the new node (required)
+        @type    name:           C{str}
+
+        @keyword size:           The size of the new node (required)
+        @type    size:           L{NodeSize}
+
+        @keyword location:       The location to create the new node
+        @type    location:       L{NodeLocation}
+
+        @keyword defaultgateway: The default gateway to be used
```
Kami: Please use underscore separated variable names (`default_gateway`, etc.).

libcloud/compute/drivers/cloudframes.py
bernieke: Yes, https is not supported by cloudframes.

CloudFramesNodeDriver throws an NotImplementedError when you try to pass secure=true.

libcloud/compute/drivers/cloudframes.py
bernieke, line 148: All cloudframes api calls are made up as /object_type/method, so I use this abstraction to enable me to type connection.machine.list(arg) instead of either having to define all the object_types as separate classes with all methods on them as separate methods, or either typing connection.request('cloud_api_machine.list, [arg, {'wait': False}]) On top of that it also abstracts away the manner in which the cloudframes api handles async calls.

It's nothing more than a more convenient way to access the api.

libcloud/compute/drivers/cloudframes.py
```diff
@@ -0,0 +1,427 @@
...
+
+    def _to_location(self, location_dict):
+        return NodeLocation(id=location_dict['guid'],
+                            name=location_dict['name'],
+                            country=None,
+                            driver=self)
+
+    def _to_node(self, node_dict):
+        # remove autotests and deleted nodes
+        if node_dict['status'] == 'CONFIGURED':
```
Kami, line 300: Can you please add a comment which clarifies why we do that?

libcloud/compute/drivers/cloudframes.py
```diff
@@ -0,0 +1,427 @@
...
+    def _to_snapshot(self, snapshot_dict):
+        return CloudFramesSnapshot(id=snapshot_dict['guid'],
+                                   timestamp=snapshot_dict['timestamp'],
+                                   label=snapshot_dict['backuplabel'],
+                                   description=snapshot_dict['description'],
+                                   driver=self)
+
+
+    # subclassed public methods, and provider specific public methods
+    def list_images(self, location=None):
```
Kami, line 321: So it looks like this results in `1 + numer_of_images` requests.

Is it possible to optimize this to only do 2 requests (get image ids, get images for all the provided ids)?

libcloud/compute/drivers/cloudframes.py
```diff
@@ -0,0 +1,427 @@
...
+        guids = self._machine_find(template=True)
```
Kami: I would find this more readable if the variable would be called `image_ids` so we know guid references image id in this case.

libcloud/compute/drivers/cloudframes.py
```diff
@@ -0,0 +1,427 @@
...
+
+    @property
+    def _languid(self):
+        if not self.__languid:
+            self.__languid = self.connection.lan.find(
+                '', '', 'public_virtual', '', '', '', '', '', '', '', '', '',
+                '', '', '', '', '')[0]
+        return self.__languid
+
+    def _machine_list(self, guid):
```
Kami: Adding a docstring would make it more obvious what this method does.

Also, it seems this method also returns public_ips and other attributes which are specific to node, but it's used in image retrieval context as well.

Maybe the this method should look something like this `_get_machine_data(guid, type=node|image)`?

Edit: Never mind, I just noticed other method references machine_find and not machine_list. In any case, better method name and docstring would also be nice.

libcloud/test/compute/test_cloudframes.py
```diff
@@ -0,0 +1,289 @@
...
+            args = ()
+            kwargs = {'url': self.url}
+        else:
+            CloudFramesNodeDriver.connectionCls.conn_classes = (
+                CloudFramesMockHttp, CloudFramesMockHttp)
+            args = CLOUDFRAMES_PARAMS
+            kwargs = {}
+        self.driver = CloudFramesNodeDriver(*args, **kwargs)
+
+    def _retry_until_up(self, cmd, *args, **kwargs):
```
Kami, line 178: Would be good to add a docstring which clarifies why this is needed (e.g. to test it with a live system).

libcloud/test/compute/test_cloudframes.py
```diff
@@ -0,0 +1,289 @@
...
+        self.assertRaises(
+            NotImplementedError, CloudFramesNodeDriver,
+            secret=secret, url='http://%s/appserver/xmlrpc' % host)
+        self.assertRaises(NotImplementedError, CloudFramesNodeDriver,
+                          key=key, url='http://%s/appserver/xmlrpc' % host)
+        CloudFramesNodeDriver(
+            key=key, secret=secret, url='http://%s/appserver/xmlrpc' % host)
+
+    def test_snapshot(self):
+        for node in self.driver.list_nodes():
```
Kami: Minor style thing - list comprehension would look better - `node = [n for n in driver.list_nodes() if n.state == NodeState.RUNNING]`.

Kami:
It would also be good to add some documentation which explains how to use this driver to `docs/compute/drivers/cloudframes.rst`.

libcloud/compute/drivers/cloudframes.py
```diff
@@ -0,0 +1,427 @@
...
+        d['size'] = None
+        d['image'] = None
+        return d
+
+    def _machine_find(self, template=False, machinetype=None,
+                      machinerole=None):
+        if not machinetype:
+            guids = []
+            for machinetype in ['VIRTUALSERVER', 'VIRTUALDESKTOP']:
+                guids += self.connection.machine.find(
```
Kami, line 271: No offense, but this part of code looks very nasty and confusing.

Is there no better way to do this?

libcloud/compute/drivers/cloudframes.py
```diff
@@ -0,0 +1,427 @@
...
+class CloudFramesNodeDriver(NodeDriver):
+    """
+    CloudFrames node driver
+    """
+
+    connectionCls = CloudFramesConnection
+
+    name = 'CloudFrames'
+    api_name = 'cloudframes'
+    website = 'http://cloudfounders.com'
```
Kami: This should probably link to the product (http://www.cloudframes.net/) and not the company website.

libcloud/compute/drivers/cloudframes.py
bernieke, line 321: Unfortunately not. The cloudframes api call machines.list is not actually capable of accepting a list of guids, it acts like you would expect from a call named machines.info...

libcloud/compute/drivers/cloudframes.py
bernieke, line 271: Unfortunately not, when using the xmlrpc interface the cloudframes api requires you to explicitly pass all args and kwargs as positional args, you can't just give it a few select keyword arguments.

There is also a rest api which _does_ allow you to use keyword arguments, but that one doesn't allow you to pass anything but basic types (so only ints, strings and bools, no lists or dicts as we require for fi. the createFromTemplate call.)
And to start using two different interfaces just so these couple of lines would be more readable seems like a bit of overkill...

I've added a comment to the code explaining why it looks the way it does.

docs/compute/_supported_providers.rst
```diff
@@ -1,67 +1,69 @@
-===================================== ========================================= =================== ============================================== ====================================
```
Kami, line 1: Hm, what happened to this file?

It's auto-generated and the entries are sorted so not sure why such a big diff.

docs/compute/drivers/cloudframes.rst
```diff
@@ -0,0 +1,33 @@
+CloudFrames Compute Driver Documentation
+========================================
+
+Connecting to the Openstack installation
```
Kami: OpenStack -> CloudFrames :P

Kami:
Besides the typo thing, it looks good now.

Please make sure pep8 passes on the new files, squash all the commits and attach a final patch to the corresponding JIRA ticket.

docs/compute/_supported_providers.rst
bernieke, line 1: Seems like the generated documentation link increased the field size a few spaces.

Kami, line 1: Ah, ok, makes sense.

Kami:
Merged into trunk, this pull request can be closed now.

Thanks.

PR #166, opened 2013-10-25
Import LockTimeout before trying to catch it.

briancurtin:
If `lock.acquire` raised any sort of exception, a `NameError` would trump it because the `__enter__` method was trying to catch `lockfile.LockTimeout`, which wasn't imported.

### DIFF
--- a/libcloud/storage/drivers/local.py
+++ b/libcloud/storage/drivers/local.py
@@ -25,7 +25,7 @@ import shutil
 import sys
 
 try:
-    from lockfile import mkdirlockfile
+    from lockfile import LockTimeout, mkdirlockfile
 except ImportError:
     raise ImportError('Missing lockfile dependency, you can install it ' \
                       'using pip: pip install lockfile')
@@ -57,7 +57,7 @@ class LockLocalStorage(object):
     def __enter__(self):
         try:
             self.lock.acquire(timeout=0.1)
-        except lockfile.LockTimeout:
+        except LockTimeout:
             raise LibcloudError('Lock timeout')
 
     def __exit__(self, type, value, traceback):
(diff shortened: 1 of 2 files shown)


alex:
We should consider running flake8 over the code, and making it part of CI to catch these issues

Kami:
@alex I agree.

In the past we had pep8, pylint and pyflakes running on Apache buildbot, but a failure was never marked as fatal. The reason for that is that there were a lot of legacy issues which no-one cleaned up (there were some attempts, but no managed to get through).

I also tried to manually enforce pep8 for new patches, but obviously humans are bad at tasks like this and it should really be part of an automatic process.

To get it passing and hook it up as part of the CI process, we should do the following:
1. Clean up the existing code
2. Ignore line limit errors emitted by pep8 - mock function methods can get really long and there is no nice way to split them. Other, a little better solution would be to only allow longer lines in `test_*` files. Ideal solution would be to change the way mocking works, but that's too big of a task and out of scope for cleaning this up.
3. Add `# noqa` pragma where necessary to silence `imported but unused errors`

Jc2k:
A similar discussion is happening in buildbot land and one of the developers there recently suggested this:

https://pypi.python.org/pypi/autopep8

I’m about to take it for a spin on one of my pet projects.

Kami:
@Jc2k thanks, I'm aware of auto pep8, but in our case, major problem are pyflake violations (besides the line lengths).

Also, what kind of set up did you have in mind? Automatically running it before file save in your editor, pre-commit hook, pre-receive hook?

First two approaches are helpful and good, but not something we can enforce as part of our contribution process (we should still add a sample pre-commit hook script to contrib/ and document how to use it).

Kami:
@briancurtin change looks good, but the tests need to be fixed before it can be merged.

briancurtin:
Updated to use an assertRaises format that works on all versions.

Kami:
@briancurtin merged, thanks.

Kami:
@briancurtin This has been merged so the pull request can be closed now.

Thanks!

PR #167, opened 2013-10-25
Fix F401 errors from flake8

briancurtin:
`flake8` emits warnings for names that are imported but not used. In most cases, these either look like copy/pasted import blocks, or names left in from debugging or after refactoring. Common unused names are the `sys` or `time` modules, and base classes. Some modules were doing an import dance to get `simplejson` or standard `json`, but those modules no longer parse json directly, likely as the result of refactoring it to a lower level.

The `libcloud.utils.py3` module contains a lot of these unused imports, but that's intentional. Because of that, it includes a `flake8: noqa` comment near the top of the file to tell `flake8` to ignore it.

### DIFF
--- a/libcloud/utils/py3.py
+++ b/libcloud/utils/py3.py
@@ -13,6 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Skip flake8 for this file as it raises many warnings about unused imports.
+# flake8: noqa
+
 # Libcloud Python 2.x and 3.x compatibility layer
 # Some methods bellow are taken from Django PYK3 port which is licensed under 3
 # clause BSD license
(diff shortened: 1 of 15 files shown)


libcloud/utils/py3.py
```diff
@@ -13,6 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Skip flake8 for this file as it raises many warnings about unused imports.
+# flake8: noqa
```
alex, line 17: You can fix that by defining `__all__`.

Kami:
@briancurtin I just started working on cleaning up all flake8 issues around 1 hour ago. I'm almost done with a full cleanup.

PR #170, opened 2013-10-28
[RFC] Validate SSH key format and size in the SSHKeyDeployment class

Kami:


Kami:
This is slightly off topic, but one thing I would also eventually like to do is refactor the `libcloud.utils.pubkey` module.

I would add a new `PublicKey` class which would handle all the public key related methods which are currently exported by `pubkey` module

Advantages of this approach are:
- No need to copy and paste "if pycrypto available" stuff
- This class would know how to deal with different key types (rsa, ecdsa, dsa)

Anyway, that's more of a low priority refactoring.

PR #172, opened 2013-10-31
#issue-421 supporting kt ucloud storage driver

charsyam:
kt ucloud storage is based on swift.
but kt customized it.
so default cloudfiles driver doesn't work for it.

so I patched 2 thigns.
1. using auth_uri variable to change auth url address.
 for example:
  currently cloudfiles driver's auth entry point is fixed. so we can't change it.
  to set auth_uri, now we can change it.
1. add ktucloud storage driver.
   It is inherited cloudfiles.py and set some parameters. 
   so it is very easy  to read.

### DIFF
--- a/libcloud/common/openstack.py
+++ b/libcloud/common/openstack.py
@@ -98,9 +98,10 @@ class OpenStackAuthConnection(ConnectionUserAndKey):
     responseCls = OpenStackAuthResponse
     name = 'OpenStack Auth'
     timeout = None
+    auth_uri = None
 
     def __init__(self, parent_conn, auth_url, auth_version, user_id, key,
-                 tenant_name=None, timeout=None):
+                 tenant_name=None, timeout=None, auth_uri=None):
         self.parent_conn = parent_conn
         # enable tests to use the same mock connection classes.
         self.conn_classes = parent_conn.conn_classes
@@ -113,6 +114,7 @@ class OpenStackAuthConnection(ConnectionUserAndKey):
         self.driver = self.parent_conn.driver
         self.tenant_name = tenant_name
         self.timeout = timeout
+        self.auth_uri = auth_uri
 
         self.urls = {}
         self.auth_token = None
@@ -141,12 +143,24 @@ class OpenStackAuthConnection(ConnectionUserAndKey):
             return self
 
         if self.auth_version == "1.0":
+            if self.auth_uri is None:
+                self.auth_uri = '/v1.0'
+
             return self.authenticate_1_0()
         elif self.auth_version == "1.1":
+            if self.auth_uri is None:
+                self.auth_uri = '/v1.1/auth'
+
             return self.authenticate_1_1()
         elif self.auth_version == "2.0" or self.auth_version == "2.0_apikey":
+            if self.auth_uri is None:
+                self.auth_uri = '/v2.0/tokens'
+
             return self.authenticate_2_0_with_apikey()
         elif self.auth_version == "2.0_password":
+            if self.auth_uri is None:
+                self.auth_uri = '/v2.0/tokens'
+
             return self.authenticate_2_0_with_password()
         else:
             raise LibcloudError('Unsupported Auth Version requested')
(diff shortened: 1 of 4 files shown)


Kami:
Can you please provide some reference which describes auth endpoints for ktucloud?

Kami:
Also, those changes are unnecessary.

You can simply set `auth_url` to `https://ssproxy.ucloudbiz.olleh.com/auth/` and it will work as expected - If you set `auth_url` to `https://ssproxy.ucloudbiz.olleh.com/auth/` and `_auth_version` to `1.0`, final request URL will look exactly like you want it to - `https://ssproxy.ucloudbiz.olleh.com/auth/v1.0`

That's also the way Rackspace and other drivers handle that.

libcloud/common/openstack.py
```diff
@@ -157,12 +171,12 @@ def authenticate_1_0(self):
...
         }
 
-        resp = self.request('/v1.0', headers=headers, method='GET')
+        resp = self.request(self.auth_uri, headers=headers, method='GET')
 
         if resp.status == httplib.UNAUTHORIZED:
             # HTTP UNAUTHORIZED (401): auth failed
             raise InvalidCredsError()
-        elif resp.status != httplib.NO_CONTENT:
+        elif resp.status != httplib.NO_CONTENT and resp.status != httplib.OK:
```
Kami, line 179: Can you please provide a link to the reference documentation which says 200 status code can be returned on success?

charsyam, line 179: maybe, I think 200 status code is for KTUCloud.
http://developer.ucloudbiz.olleh.com/export/sites/default/support/pds/download/KTucloudSS_API_V1.220130128_1.7.4.pdf
but it is written in Korean.

page 5. 사용자 인증(User Authentication) Section
page 6. Response Body <- it returns "body" 
page 6. HTTP Status Code <-- 인증 완료(Authentication Complete) -> 200, 인증 실패(Authentication Error) 401

so. it returns response body and it's status code is 200.

charsyam:
@Kami 
and KT UCloud support 2 auth page.
1. http://https://ssproxy.ucloudbiz.olleh.com/auth/v1.0
2. https://api.ucloudbiz.olleh.com/storage/v1/auth

I intent to give an functionality to change auth uri, even it is not ended with "v1.0" or "v1.1/auth" or "v2.0/tokens"

charsyam:
@Kami and I found something wrong in openstack.py

I tried to follow your advice.
but in openstack.py

it just request "/v1.0" or something not combination of end part of url + auth_version.

if we want it to work well. we can change it to like this.

'''python
(host, port, secure, request_path) = \
            self._tuple_from_url(self.auth_url)

uri = "%sv%s"%(request_path, self.auth_version)
'''

but I think it is also weird.

because when version 2.0 or 1.1 it should add "/v1.1/auth" in authenticate function in openstack.py

so. I think setting auth url explicitly is better than now.

what do you think about this?

Kami:
@charsyam Actually, you are right. There seems to be a bug / some nasty stuff going on in OpenStackAuthConnection class.

`OpenStackAuthConnection` class overrides `morph_action_hook` method (https://github.com/apache/libcloud/blob/trunk/libcloud/common/openstack.py#L122) which otherwise appends action to the request path (https://github.com/apache/libcloud/blob/trunk/libcloud/common/base.py#L638).

Kami:
@charsyam I think the best approach would be to allow user to pass in either a base `auth_url` without a path (e.g. `https://192.168.100.100`) or a full `auth_url` with a path (e.g. `http://192.168.100.100/v1.0/auth`, `http://192.168.100.100/`).

If user passes in just a base url, default auth path is appended to it, otherwise, the url is used directly the way it's passed in.

This also makes more sense, since currently the variable is called `auth_url` and now `base_auth_url` so you would expect you can pass in a whole url with a path.

libcloud/common/openstack.py
Kami, line 179: OK, thanks.

I would just change it to `elif resp.status not in [httplib.NO_CONTENT, httplib.OK]` to make it slightly more easier to read, otherwise looks good.

charsyam:
@Kami I pulled request again with your review.

and I will close this issue :) Thank you.

PR #173, opened 2013-11-01
[#issue-421] apply review and fix broken test suite

charsyam:
@kami I applied your review to code with our best approch.

and fix testcase that were broken because of changing.

please review this. Thank you for your advice..

### DIFF
--- a/libcloud/common/openstack.py
+++ b/libcloud/common/openstack.py
@@ -120,7 +120,37 @@ class OpenStackAuthConnection(ConnectionUserAndKey):
         self.auth_user_info = None
 
     def morph_action_hook(self, action):
-        return action
+        (host, port, secure, request_path) = \
+            self._tuple_from_url(self.auth_url)
+
+        # for example auth_url is
+        #
+        # http://192.168.0.100      - assume path is not provided,
+        #                             depending on the auth_version,
+        #                             correct path should be appended
+        #                             to the final url
+        #                             if auth_version is 1.0
+        #                             final is http://192.168.0.100/v1.0
+        #                             if auth_version is 1.1
+        #                             final is http://192.168.0.100/v1.1/auth
+        #                             if auth_version is 2.0
+        #                             final is http://192.168.0.100/v2.0/tokens
+        #
+        # http://192.168.0.100/         - assume path(foo/bar) is already
+        #                                 provided, nothing should be
+        #                                 apppended to the final url
+        #
+        # http://192.168.0.100/foo/bar  - assume path(foo/bar) is already
+        #                                 provided, nothing should be
+        #                                 apppended to the final url
+        #
+        # http://192.168.0.100/foo/bar/ - assume path(foo/bar) is already
+        #                                 provided, nothing should be
+        #                                 appended to the final url
+        if request_path is "":
+            return action
+
+        return request_path
 
     def add_default_headers(self, headers):
         headers['Accept'] = 'application/json'
@@ -162,7 +192,7 @@ class OpenStackAuthConnection(ConnectionUserAndKey):
         if resp.status == httplib.UNAUTHORIZED:
             # HTTP UNAUTHORIZED (401): auth failed
             raise InvalidCredsError()
-        elif resp.status != httplib.NO_CONTENT:
+        elif resp.status not in [httplib.NO_CONTENT, httplib.OK]:
             body = 'code: %s body:%s headers:%s' % (resp.status,
                                                     resp.body,
                                                     resp.headers)
--- a/libcloud/test/compute/test_openstack.py
+++ b/libcloud/test/compute/test_openstack.py
@@ -94,7 +94,7 @@ class OpenStackServiceCatalogTests(unittest.TestCase):
 
     def test_connection_get_service_catalog(self):
         connection = OpenStackBaseConnection(*OPENSTACK_PARAMS)
-        connection.auth_url = "https://auth.api.example.com/v1.1/"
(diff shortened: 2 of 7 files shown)


libcloud/common/openstack.py
```diff
@@ -120,7 +120,13 @@ def __init__(self, parent_conn, auth_url, auth_version, user_id, key,
         self.auth_user_info = None
 
     def morph_action_hook(self, action):
-        return action
```
Kami, line 123: Can you please add a docstring / comment which describes how this function works (as I described it in my previous comment).

Kami:
@charsyam Thanks for working on this!

Besides the comments I already posted, it also needs some test cases which test at least the following `auth_url` values:
1. `http://192.168.0.100/` - assume path (`/`) is already provided, nothing should be appended to the final url
2. `http://192.168.0.100/foo/bar` - assume path (`/foo/bar`) is already provided, nothing should be appended to the final url
3. `http://192.168.0.100/foo/bar/` - assume path (`/foo/bar/`) is already provided, nothing should be appended to the final url
4. `http://192.168.0.100` - assume path is not provided, depending on the`auth_version`, correct path should be appended to the final url

libcloud/common/openstack.py
charsyam, line 123: ok :)

charsyam:
@Kami I have a question about this rule.
I might think many developers don't think about the last "/" of url.
so, they might expect "http://192.168.0.100/" and "http://192.168.0.100" is same.

so how about this. only 1 is changed.
1. http://192.168.0.100/ - assume path is not provided, depending on the auth_version, correct path should be appended to the final url
   1. http://192.168.0.100/foo/bar - assume path (/foo/bar) is already provided, nothing should be appended to the final url
   2. http://192.168.0.100/foo/bar/ - assume path (/foo/bar/) is already provided, nothing should be appended to the final url
   3. http://192.168.0.100 - assume path is not provided, depending on the auth_version, correct path should be appended to the final url

Kami:
@charsyam The problem with your approach is that there is no way to specify a URL where you don't any path to be appended (e.g. you want a final URL to be `http://192.168.0.100`).

Imo the confusion between no slash and `/` shouldn't be too big of a deal, as long as we document it well. This means updating this page (https://github.com/apache/libcloud/blob/trunk/docs/compute/drivers/openstack.rst) and updating method docstrings.

charsyam:
@Kami ok. it makes sense. I will change patch soon :)

charsyam:
@Kami I uploaded new patch. please review my test code(test_auth_url)

libcloud/test/compute/test_openstack.py
```diff
@@ -120,6 +120,58 @@ def setUp(self):
         OpenStackBaseConnection.conn_classes = (OpenStackMockHttp,
                                                 OpenStackMockHttp)
 
+    def test_auth_url(self):
```
Kami: I would give the test function a more verbose name - e.g. `test_auth_url_is_correctly_assembled`.

libcloud/test/compute/test_openstack.py
```diff
@@ -120,6 +120,58 @@ def setUp(self):
...
+                    self._get_mock_connection(mock_http_class=mock_http_class,
+                                              auth_url=_auth_url)
+                auth_url = connection.auth_url
+
+                osa = OpenStackAuthConnection(connection,
+                                              auth_url,
+                                              auth_version,
+                                              user_id, key)
+
+                try:
```
Kami, line 168: Why do we need this try / except block here?

libcloud/test/compute/test_openstack.py
```diff
@@ -120,6 +120,58 @@ def setUp(self):
...
+                    osa = osa.authenticate()
+                except:
+                    pass
+
+                if (status == APPEND):
+                    self.assertTrue(osa.action == actions[auth_version])
```
Kami: To get a better assertion error message you should use `self.assertEqual` here and bellow.

libcloud/test/compute/test_openstack.py
```diff
@@ -120,6 +120,58 @@ def setUp(self):
...
+            "2.0": "/v2.0/tokens",
+            "2.0_apikey": "/v2.0/tokens",
+            "2.0_password": "/v2.0/tokens"
+        }
+
+        user_id = OPENSTACK_PARAMS[0]
+        key = OPENSTACK_PARAMS[1]
+
+        for (auth_version, mock_http_class) in tuples:
+            for _auth_url, status, action in auth_urls:
```
Kami: To make it more readable and easier to understand, I would rename it to:

``` python
for (_auth_url, should_append_default_path, expected_path) in auth_urls
```

Kami:
@charsyam Thanks, I added some more comments.

libcloud/test/compute/test_openstack.py
charsyam, line 168: @Kami this test just check of OpenStackAuthConnection's action value.
and auth_url is invalid. 
so it always fails except "http://192.168.0.100". so I capture exception.
and just check action is the same with our expectation.

Kami:
I've improved the commit message, made some small style changes and merged patch into trunk.

Next time please make sure the commit message actually describes what you did.

Thanks!

(this pull request can also be closed now)

Kami:
@openbaas / @charsyam You can close this pull request since those changes have been merged.

charsyam:
@Kami Thank you. I don't know why, but My account is somewhat weird. my main account is charsyam, but, My Pull Requests are always written by openbaas

PR #175, opened 2013-11-07
LIBCLOUD-434: list_records() return all records for a Route53 zone.

xofer:
The Route53 API returns a maximum of 100 records per request. This change makes multiple requests to get all records in the zone when results are truncated.


Kami:
Thanks, this looks like a good start, but it should be changed to match other drives and use a generator approach. This means implementing `iterate_zones` and `iterate_records` method.

`list_zones` and `list_records` doesn't need to be implemented in this case since it's already provided by a parent class.

You can find an example of how Zerigo driver does that here - https://github.com/apache/libcloud/blob/trunk/libcloud/dns/drivers/zerigo.py#L149

Kami:
@xofer ^^ let us know if you need any help with this.

xofer:
Thanks @Kami -- will give it a shot later today and let you know...

Kami:
@xofer hey, have you made any progress on this?

Let us know if you need any help.

xofer:
well, that's not what I wanted. Will open a new request.

xofer:
@Kami https://github.com/apache/libcloud/pull/187

PR #176, opened 2013-11-09
[#435] fix bug : when diskoffering doesn't exist

charsyam:
reason:
KT ucloud starts to support "no disk model"

patch:
 when diskofferingid is not in response. set it to 0.


libcloud/compute/drivers/ktucloud.py
```diff
@@ -55,8 +55,12 @@ def list_sizes(self, location=None):
         szs = self._sync_request('listAvailableProductTypes')
         sizes = []
         for sz in szs['producttypes']:
+            diskofferingid = 0
+            if 'diskofferingid' in sz:
+                diskofferingid = sz['diskofferingid']
```
Kami: `diskofferingid = sz.get('diskofferingid' , 0)` is nicer and more pythonic way to do that

Kami:
Also needs a test case for a size without `diskofferingid` attribute.

charsyam:
@Kami I added testcase and fixing lint issue :)

Kami:
@charsyam Patch has been merged into trunk, this pr can be closed now.

Thanks!

PR #178, opened 2013-11-13
[LIBCLOUD-428] Fix auth token caching and re-use in the OpenStack based drivers

Kami:
This branch fixes an issue described in LIBCLOUD-436 (https://issues.apache.org/jira/browse/LIBCLOUD-428).

As noted in the ticket description, the problem is that current version of code doesn't properly check if the token has expired before using it. This causes `InvalidCredsError` exception to be thrown if the token has expired. 

This is usually a non-issue for short running scripts, but if you are using a driver in a long running process, there is a bigger chance that the token will expire during the life-time of the process and as such, the exception will be thrown. 

While working on this bug, I also discovered an issue with `ex_force_auth_token` argument. Currently we support skipping the authentication process by passing a valid auth token to the driver constructor using `ex_force_auth_token` argument.

The problem is that we don't know about the token expiration time when this argument is used. Ss such, we can't invalidate the token when it's about to expire. My solution to this problem was documenting this behavior better in the documentation.

As you can see in the diff, I also needed to update the fixtures for tests to pass. The approach that I used is not ideal since it's basically a ticking time bomb, but I opted for a simple and less time-consuming solution at this point and hopefully we will figure a better one by 2031 :P

### DIFF
--- a/docs/compute/drivers/openstack.rst
+++ b/docs/compute/drivers/openstack.rst
@@ -78,6 +78,16 @@ component of the URL (``12345`` in the example bellow).
 4. Skipping normal authentication flow and hitting the API endpoint directly using the ``ex_force_auth_token`` argument
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+This is an advanced use cases which assumes you manage authentication and token
+retrieval yourself.
+
+If you use this argument, the driver won't hit authentication service and as
+such, won't be aware of the token expiration time.
+
+This means auth token will be considered valid for the whole life time of the
+driver instance and you will need to manually re-instantiate a driver with a new
+token before the currently used one is about to expire.
+
 .. literalinclude:: /examples/compute/openstack/force_auth_token.py
    :language: python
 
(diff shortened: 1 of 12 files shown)


Kami:
@alex, @briancurtin would appreciate if you two have a look at this one. Currently this issue is the only blocker for 0.14 release.

alex:
I didn't spot anythign in my review of this.

briancurtin:
Looks fine to me after a read-through, so as long as it works in practice for Michael, go for it.

I was talking to someone recently (@jnoller, I think) about writing up some long-running integration-style tests to catch this type of thing, since it has bitten us a bunch on the pyrax project as well. Another thing to add to the todo list!

Kami:
@briancurtin, @jnoller Yeah, having integration tests would definitely be nice.

When people originally started to work on tests, they had this idea in mind so they made all the credentials pluggable (`secres.py-dist`). The idea was that you could just put in real provider credentials and the tests would run against a live installation.

Sadly that's easier said than done. A lot of tests for edge cases rely on an account being in specific state which means they won't work against a live installation. Making tests work against a live installation, would at this point, require a substantial amount of work with changes to the existing tests.

One thing we could do is use a more gradual approach and start working on a separate integration test suite. This way we would have two independent test suites (unit & integration). With this approach we might trade long term overhead of maintaining two test suites for short term flexibility and speed, but it's probably worth it.

Kami:
OK, I added some more tests for `_populate_hosts_and_request_paths` method and went ahead and merged patch into trunk.

PR #179, opened 2013-11-20
GCE don't send field description in image if it doesn't exist or blank

gigimon:
Hello

In GCE we can set description for image, but this's not required and we set this field blank. But if this field blank GCE return response without this field. I fix this with tests


Kami:
Patch merged into trunk, this pull request can be closed now.

@gigimon thanks for the patch and @wrigri thanks for reviewing / cross-checking it. @gigimon this time I did it, but next time please make sure the commit message is a bit more descriptive.

Kami:
@gigimon  Can you please close this PR (changes have been merged)?

gigimon:
@Kami thx!

PR #183, opened 2013-12-02
LIBCLOUD-433: Make storage backend send content-type of "application/octet-stream" if none is supplied and none can be guessed.

micolous:
https://issues.apache.org/jira/browse/LIBCLOUD-433

### DIFF
--- a/libcloud/storage/base.py
+++ b/libcloud/storage/base.py
@@ -598,9 +598,9 @@ class StorageDriver(BaseDriver):
             content_type, _ = libcloud.utils.files.guess_file_mime_type(name)
 
             if not content_type:
-                raise AttributeError(
-                    'File content-type could not be guessed and' +
-                    ' no content_type value provided')
+                # Fallback to a content-type that will cause most browsers to
+                # download it again as a binary file.
+                content_type = 'application/octet-stream'
 
         file_size = None
 
--- a/libcloud/test/storage/test_atmos.py
+++ b/libcloud/test/storage/test_atmos.py
@@ -316,17 +316,14 @@ class AtmosTests(unittest.TestCase):
         file_path = os.path.abspath(__file__)
         container = Container(name='fbc', extra={}, driver=self)
         object_name = 'ftu'
-        try:
-            self.driver.upload_object(file_path=file_path, container=container,
-                                      object_name=object_name)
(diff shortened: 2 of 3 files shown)


Kami:
@micolous Thanks.

The change looks good overall, but the tests are failing (https://travis-ci.org/apache/libcloud/builds/14792563).

Please fix the failing tests and add a new one which tests that `application/octet-stream` is specified if we can't guess the content_type ourselves.

micolous:
Thanks, I'll have a look at this.

Travis helps as I am having difficulties trying to run the tests locally in
tox.

Kami:
@micolous What kind of problems do you have?

You don't have all the supported Python versions? If so, just running the tests locally with 2.6 / 2.7 should be a good start. Then you can leave Travis to run tests for other versions.

micolous:
@Kami, I've resolved some of the issues with `tox` now -- and yes, I don't have all of the supported Python versions.  I was having issues also trying to run the tests on `python2.5`, because I can no longer install the development packages for that version of Python (in order to pull all of it's dependencies).

I've resolved the `AtmosTests` failure, I'm a bit unsure where those hash errors in the Rackspace provider are coming from, they have strange results... I'll have a look at that.

micolous:
I've resolved the Rackspace test failures now, this seems to because of checking response hashes in it's Content-Type failure tests.  Because the behaviour is now different, I've updated this test.

Kami:
@micolous I've made some changes, squashed your commits and merged patch into trunk.

The changes I've made include:
- Added and exported a `DEFAULT_CONTENT_TYPE` constant (888b3d598ee0d67cea6ec899d1aa1ec330dd36c4)
- Added a test case for default content type being specified (888b3d598ee0d67cea6ec899d1aa1ec330dd36c4)
- Added a strict mode which defaults to `False`. If strict mode is used, old behavior is preserved and exception will be thrown if no content type is provided and none be detected. (ec90fcb8a32f5e4a0330ffa6d7b932b65bc82077)
- Updated upgrade notes (0e8c2f06e2b6c1ca245f86000471563d4aaeaa79)

Thanks!

Kami:
Also, this pull request can be closed now.

micolous:
Excellent, thanks!

PR #184, opened 2013-12-02
On GCE, Images no longer always have preferredKernel set.

wrigri:


libcloud/compute/drivers/gce.py
```diff
@@ -2469,7 +2469,7 @@ def _to_node_image(self, image):
         :rtype: :class:`NodeImage`
         """
         extra = {}
-        extra['preferredKernel'] = image['preferredKernel']
+        extra['preferredKernel'] = image.get('preferredKernel')
```
Kami: For consistency, please use `image.get('preferredKernel', None)` as the line bellow.

Otherwise LGTM, +1

wrigri:
@Kami OK, added default value to align with the line below.

Kami:
Squashed and merged into trunk.

Thanks!

PR #185, opened 2013-12-03
Specify public key on ex_create_keypair and  Successfully get public ip ...

mgogoulos:
...addresses for nodes for HP Cloud

https://issues.apache.org/jira/browse/LIBCLOUD-449

### DIFF
--- a/libcloud/compute/drivers/openstack.py
+++ b/libcloud/compute/drivers/openstack.py
@@ -39,6 +39,7 @@ from libcloud.common.openstack import OpenStackBaseConnection
 from libcloud.common.openstack import OpenStackDriverMixin
 from libcloud.common.types import MalformedResponseError, ProviderError
 from libcloud.compute.types import NodeState, Provider
+from libcloud.compute.base import is_private_subnet
 from libcloud.compute.base import NodeSize, NodeImage
 from libcloud.compute.base import NodeDriver, Node, NodeLocation, StorageVolume
 from libcloud.pricing import get_size_price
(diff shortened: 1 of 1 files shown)


libcloud/compute/drivers/openstack.py
```diff
@@ -39,6 +39,7 @@
 from libcloud.common.openstack import OpenStackDriverMixin
 from libcloud.common.types import MalformedResponseError, ProviderError
 from libcloud.compute.types import NodeState, Provider
+from libcloud.compute.base import is_private_subnet
```
Kami, line 42: This function was recently moved to `libcloud.utils.networking` and a new `is_public_subnet` function was also added there.

libcloud/compute/drivers/openstack.py
```diff
@@ -1730,7 +1731,7 @@ def ex_list_keypairs(self):
         return self._to_keypairs(
             self.connection.request('/os-keypairs').object)
 
-    def ex_create_keypair(self, name):
+    def ex_create_keypair(self, name, public_key=None):
```
Kami, line 1734: Please add a docstring for new `public_key` argument.

Kami:
Those changes also need tests, especially the IP address change.

Jc2k:
Why change ex_create_keypair? I think ex_import_keypair and/or ex_import_keypair_from_string do the same job already.

Kami:
@Jc2k Ah, that's actually a good point & catch. For consistency, `ex_create_keypair` should only take a `name` argument, right?

I do need to admit, that I find the existing API slightly "weird", but it's what we have at the moment and we should follow it for the consistency reasons.

In any case, if we want to change this and get rid of the "import" methods, we should do it now, before the 1.0 release.

mgogoulos:
closing, will open a new pull request for the second issue only

PR #186, opened 2013-12-04
service type has changed for cloudFilesCDN

jobelenus:
I am unsure if this change by CloudFiles is version specific. But I do know this is explicitly what is returned by version 2.0 (I don't have any 1.x versions). I happen to be using the IAD region as well.

### DIFF
--- a/libcloud/storage/drivers/cloudfiles.py
+++ b/libcloud/storage/drivers/cloudfiles.py
@@ -136,8 +136,8 @@ class CloudFilesConnection(OpenStackBaseConnection):
         if not ep:
             raise LibcloudError('Could not find specified endpoint')
 
-        if 'publicURL' in ep:
-            return ep['publicURL']
+        if 'internalURL' in ep:
+            return ep['internalURL']
         else:
             raise LibcloudError('Could not find specified endpoint')
 
--- a/libcloud/storage/drivers/cloudfiles.py
+++ b/libcloud/storage/drivers/cloudfiles.py
@@ -259,20 +259,21 @@ class CloudFilesStorageDriver(StorageDriver, OpenStackDriverMixin):
         raise LibcloudError('Unexpected status code: %s' % (response.status))
 
     def get_container_cdn_url(self, container):
-        container_name = container.name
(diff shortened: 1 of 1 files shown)


Kami:
@jobelenus Good catch and thanks for this pull request.

I have verified it by checking the service catalog for my account using auth 2.0 - https://gist.github.com/Kami/327b7530793987131461.

Overall the change looks good, but a bunch of tests fixtures need to be updated because now the tests are failing - https://travis-ci.org/apache/libcloud/builds/14930820 (I need to have a look why Travis stopped updating commit status for this project).

/cc @briancurtin @alex

jobelenus:
Yea, the "Could not find specified endpoint" was the error I would get without this change, and it was always a failure with the CDN endpoint. The non-CDN endpoint always worked. The service catalog entries are likely different across versions of Cloudfiles.

briancurtin:
This looks fine and accurate to me. Good catch, John!

I'm in training/meetings through tomorrow, but I can help with getting tests updated on Friday if needed.

Kami:
@jobelenus I've added the tests and merged those changes into trunk (8762175ea6b6ccd5f72d155012cda6eea8a1ee25, 6f3856458fc476dd26463e1352cdb72b2b97f4b0, 12a75c9fe87ae2236c19774f7cf71dd4aa7f2ae6) so this pull request can be closed now.

Thanks!

Kami:
@jobelenus Btw, this pull request can be closed now.

jobelenus:
Sorry about that.

jobelenus:
I see that the notion of public/internalURL did not make it into 0.14 ( [see commit](https://github.com/activefrequency/libcloud/commit/eb84616d894c928009e43fd5d4dbf0331e85cbc5) ). This changes gives significant speed improvements (upwards of a whole second) when your application server and cloudfiles are in the same region/datacenter.

Kami:
@jobelenus Yes, the reason for that is that I didn't even notice this change.

You should create a new pull request for this change. Then we can get it reviewed and merged into trunk.

PR #187, opened 2013-12-05
LIBCLOUD-434: route53 replace list_* with iterate_*

xofer:
The Route53 API returns a maximum of 100 records per request. This change implements iterate_zones and iterate_records methods.

### DIFF
--- a/libcloud/dns/drivers/route53.py
+++ b/libcloud/dns/drivers/route53.py
@@ -111,17 +111,11 @@ class Route53DNSDriver(DNSDriver):
         RecordType.TXT: 'TXT'
     }
 
-    def list_zones(self):
-        data = self.connection.request(API_ROOT + 'hostedzone').object
-        zones = self._to_zones(data=data)
-        return zones
+    def iterate_zones(self):
+        return self._get_more('zones')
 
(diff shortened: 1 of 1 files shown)


libcloud/dns/drivers/route53.py
```diff
@@ -308,3 +302,40 @@ def _to_record(self, elem, zone):
...
+        if last_key:
+            params['name'] = last_key
+        path = API_ROOT + 'hostedzone'
+        
+        if rtype == 'zones':
+            response = self.connection.request(path, params=params)
+            transform_func = self._to_zones
+        elif rtype == 'records':
+            zone = kwargs['zone']
+            path += '/{}/rrset'.format(zone.id)
```
Kami: Can you please use a standard string interpolation?

The reason for that is that we still support Python 2.5 where format sadly isn't available.

Kami:
Lint and tests seem to be failing https://travis-ci.org/apache/libcloud/builds/14970505. Not sure if all the test failures are related to format thing.

Kami:
@xofer Squashed all the commits and merged changes into trunk. Thanks!

This pull request can also be closed now.

PR #188, opened 2013-12-05
Issue LIBCLOUD-455: validate image id

virtuald:
https://issues.apache.org/jira/browse/LIBCLOUD-455

### DIFF
--- a/libcloud/compute/drivers/openstack.py
+++ b/libcloud/compute/drivers/openstack.py
@@ -1861,12 +1861,6 @@ class OpenStack_1_1_NodeDriver(OpenStackNodeDriver):
             else:
                 private_ips.extend(ips)
 
-        # sometimes the image is not set if openstack is in an error state
-        if api_node['image'] == u'':
-            image_id = None
-        else:
-            image_id = api_node['image']['id']
-
         return Node(
             id=api_node['id'],
             name=api_node['name'],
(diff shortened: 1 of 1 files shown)


libcloud/compute/drivers/openstack.py
```diff
@@ -1875,7 +1881,7 @@ def _to_node(self, api_node):
                 # Docs says "tenantId", but actual is "tenant_id". *sigh*
                 # Best handle both.
                 tenantId=api_node.get('tenant_id') or api_node['tenantId'],
-                imageId=api_node['image']['id'],
+                imageId=image_id,
```
Kami: More robust and Pythonic way would be to just do `imageId=api_node.get('image', {}).get('id', None)`

Kami:
It would also be good to use raw response data (https://gist.github.com/virtuald/1358cec79751925e3817) as a test fixture and add a test case for this specific scenario.

virtuald:
@Kami I'd add a test fixture, but it's not exactly clear how to do that, and I don't have the time to figure it out right now.

Kami:
@virtuald I've added a regression test case, modified the code to correctly handle the missing attribute and merged changes into trunk.

The approach I have originally suggested in my comment was wrong (sorry about that!). I thought that the `image` attribute value was an empty `dict`, but it was actually an empty string (nasty API).

This pull request can be closed now.

Thanks!

virtuald:
Thanks!

PR #189, opened 2013-12-07
Promote key pair management functionality to be part of the base compute API

Kami:
This pull request promotes key pair management functionality to be part of the base compute API as discussed on the mailing list.

It includes the following changes:
- Add new `KeyPair` class in `libcloud.compute.base` module
- Add new key pair management methods to `libcloud.compute.base.NodeDriver` class
- Update existing drivers which previously exposed this functionality through the extension methods (EC2, CloudStack, OpenStack) to use a new standard API
- For backward compatibility reasons and to ease the migration, existing extension methods will continue to work until a next major release (existing extension methods have been modified to call new methods which are now part of the standard API).
- Add some missing tests (while working on this I have noticed there were some missing or bad tests for the key pair management functionality)
- Update upgrade nodes section in the documentation
- Add a new section on how to use key pair management to the documentation

Only which currently exposes key pair management through extension methods and I haven't updated is NephoScale. The reason for that is that NephoScale offers two types of keys (password and ssh) and I haven't totally decided yet what I want to do with it.


Jc2k:
Perhaps it would be good to add get_key_pair? There could be a default implementation using list_key_pairs. It would raise KeyPairNotFound if there is no such key pair.

Also, a KeyPairAlreadyExists exception for the create and import API's would be good.

Kami:
@Jc2k Yeah, I was thinking about that as well. All the new APIs expose "get one" functionality (storage, dns) so it also makes sense to add a `get_key_pair` method here.

I just checked the API docs and all of the APIs (EC2, CloudStack, OpenStack) support retrieving information for a single key pair so we don't need to use a potentially expensive list_key_pairs.

Kami:
@Jc2k As far as the exception name goes, I was thinking about `ZoneDoesNotExist` since other APIs also use `DoesNotExist` suffix.

Jc2k:
I guess you meant `KeyPairDoesNotExist` - I was thinking of the DNS exceptions too, but misremembered. +1 for following what we do there.

Kami:
Yeah, sorry, I meant `KeyparDoesNotExist` :)

Kami:
@Jc2k `get_key_pair` and related changes implemented in c62c472b0887077ef3402ace51adb511e2b3cf60.

PR #190, opened 2013-12-11
Handle Rackspace pagination of DNS records.

thanatos:
Rackspace paginates DNS records returned from their API. In list_records, query
Rackspace multiple times, if needed, to get the full list of records.


briancurtin:
This looks fine and passes all of the tests - thanks!

@Kami is this alright to go in by itself, or should it wait for https://issues.apache.org/jira/browse/LIBCLOUD-453? I'm planning on working on that one soon, but I wouldn't want to hold up this submission over something else that may take some rounds of review/debate.

Kami:
@thanatos Yep, thanks for this pull request.

@briancurtin @thanatos The pull request looks good, but I would prefer if you would spent some more time on this and implement `iterate_*` methods.

You have already done the hard part and implementing `iterate_*` methods at this point mostly means shuffling some code around. You can find some examples of how to do that in a recent pull request - https://github.com/apache/libcloud/pull/175 & https://github.com/apache/libcloud/pull/187

thanatos:
@Kami Took a shot at implementing this with `iterate_records`. Let me know how it looks.

libcloud/dns/drivers/rackspace.py
```diff
@@ -154,13 +154,26 @@ def list_zones(self):
...
+                'limit': limit,
+                'offset': offset,
+            }
+            response = self.connection.request(
+                action='/domains/%s' % (zone.id), params=params).object
+            response_data = response['recordsList']['records']
+            for item in response_data:
+                record = self._to_record(data=item, zone=zone)
+                yield record
+            if len(response_data) < limit:
```
Kami: This would result in an unnecessary request (and probably an exception) if the last page has `limit` number of items, but there is no next page.

There are two ways to avoid that:
1. You can detect if there are more items if `link` item with `rel` `next` is present in the response. And if you are on the last page, this `next` link won't be present. (http://docs.rackspace.com/cdns/api/v1.0/cdns-devguide/content/pagination.html)
2. Save the number of all the retrieved records in a variable and stop when `yielded_records === response['totalEntries']`

Kami:
There is also exactly the same issue with `list_zones` so `iterate_zones` should be implemented in a same fashion as well.

Kami:
@alex, @briancurtin can one of you please help get this across the finish line?

briancurtin:
Would something like this do the trick alright? https://github.com/briancurtin/libcloud/commit/8db02b0d94c77ff5ef4a871fa15225042034cb95

Unit tests needed, but I just threw together an IRL test very quickly with 150 A records and got 'em all back.

thanatos:
@briancurtin: Thanks for taking the time to put that together! I added a `break` to terminate the loop that looks for the next link early if it finds it (thanatos@d1bc5c3).

(The pull request and the merge commit look a little funny. Let me know if I need to change something…)

briancurtin:
If/when @Kami is fine with the way this one was done, I will apply the same love to the other methods mentioned above. I just don't want to create more work by doing them before we've nailed the right implementation.

Kami:
@briancurtin @thanatos Yeah, looks good, but like I said, we should also do the same for `iterate_zones`.

thanatos:
@Kami Wrote an implementation for `iterate_zones`.

Kami:
@thanatos Can you please close this pull request and create a new one with your changes?

The reason is that this one is messed up and I can't merge it directly into trunk.

thanatos:
@Kami @briancurtin I've opened pulled request #230 with a cleaned up branch of the relevant changes in here.

PR #194, opened 2013-12-13
fixes ssh for servers with PubkeyAuthentication only

mgogoulos:
https://issues.apache.org/jira/browse/LIBCLOUD-461

### DIFF
--- a/libcloud/compute/ssh.py
+++ b/libcloud/compute/ssh.py
@@ -173,11 +173,8 @@ class ParamikoSSHClient(BaseSSHClient):
 
         if self.password:
             conninfo['password'] = self.password
-        elif self.key:
+        if self.key:
             conninfo['key_filename'] = self.key
-        else:
-            conninfo['allow_agent'] = True
-            conninfo['look_for_keys'] = True
 
         if self.timeout:
             conninfo['timeout'] = self.timeout


libcloud/compute/ssh.py
```diff
@@ -173,11 +173,8 @@ def connect(self):
 
         if self.password:
             conninfo['password'] = self.password
-        elif self.key:
+        if self.key:
             conninfo['key_filename'] = self.key
-        else:
-            conninfo['allow_agent'] = True
```
Kami, line 179: For explicitness sake and backward compatibility reasons I would prefer to still specify those two options.

Kami:
I needed paramiko docs and ticket description to see what happens if both arguments (multiple authentication methods) are specified.

It appears paramiko just tries different authentication methods in a pre-defined order of priority so this change looks good.

Kami:
Existing tests are failing btw - https://travis-ci.org/apache/libcloud/builds/15403675.

Those tests need to be fixed / updated. It would also be good to add a new test case which checks what happens when both arguments are specified.

Kami:
I went ahead fixed the failing tests and made the `look_for_keys` and `allow_agent` change described above and merged this patch into trunk.

Thanks!

mgogoulos:
Nice, thanks for adding the tests! Shame for me for postponing for such a long time!

PR #196, opened 2013-12-16
Add methods ex_add/ex_delete port_forwarding_rule to CloudstackNode (similar to ip_forwarding_rule)

gigimon:
Hi!

I have added port forwarding method to CloudstackNode object because class CloudstackPortForwardingRule has a method delete which doesn't work. And change call signature because node object was not in first place

And want ask about naming of port forwarding methods. We have ex_add_ip_forwarding_rule and ex_create_port_forwarding_rule maybe rename one of this?

### DIFF
--- a/libcloud/compute/drivers/cloudstack.py
+++ b/libcloud/compute/drivers/cloudstack.py
@@ -42,15 +42,35 @@ class CloudStackNode(Node):
         "Release a public IP that this node holds."
         return self.driver.ex_release_public_ip(self, address)
 
-    def ex_add_ip_forwarding_rule(self, address, protocol, start_port,
-                                  end_port=None):
+    def ex_create_ip_forwarding_rule(self, address, protocol,
+                                     start_port, end_port=None):
         "Add a NAT/firewall forwarding rule for a port or ports."
-        return self.driver.ex_add_ip_forwarding_rule(self, address, protocol,
-                                                     start_port, end_port)
+        return self.driver.ex_create_ip_forwarding_rule(self, address,
+                                                        protocol,
+                                                        start_port, end_port)
+
+    def ex_create_port_forwarding_rule(self, address,
+                                       private_port, public_port,
+                                       protocol,
+                                       public_end_port=None,
+                                       private_end_port=None,
+                                       openfirewall=True):
+        "Add a port forwarding rule for port or ports."
+        return self.driver.ex_create_port_forwarding_rule(self, address,
+                                                          private_port,
+                                                          public_port,
+                                                          protocol,
+                                                          public_end_port,
+                                                          private_end_port,
+                                                          openfirewall)
 
     def ex_delete_ip_forwarding_rule(self, rule):
-        "Delete a NAT/firewall rule."
+        "Delete a port forwarding rule."
         return self.driver.ex_delete_ip_forwarding_rule(self, rule)
+
+    def ex_delete_port_forwarding_rule(self, rule):
+        "Delete a NAT/firewall rule."
+        return self.driver.ex_delete_port_forwarding_rule(self, rule)
 
     def ex_start(self):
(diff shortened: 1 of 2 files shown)


sebgoa:
I am not sure what you mean by 'has a method delete which doesn't work' ? what exactly does not work ? you did not seem to modify that method.

Also what do you mean by "change call signature...' 

Ok to rename ex_add_ip_forwarding_rule to ex_create_ip_frowarding_rule, but we will have to keep track of the fact this will break backward compatibility.

gigimon:
ex_create_port_forwarding_rule was:

``` python
def ex_create_port_forwarding_rule(self, address, private_port,
                                       public_port, protocol, node,
                                       public_end_port=None,
                                       private_end_port=None,
                                       openfirewall=True):
```

Now it:

``` python
def ex_create_port_forwarding_rule(self, node, address,
                                       private_port, public_port,
                                       protocol,
                                       public_end_port=None,
                                       private_end_port=None,
                                       openfirewall=True):
```

Because in all methods argument 'node' in first place.

About method delete:
Class CloudStackPortForwardingRule has method delete, it call self.node.ex_delete_port_forwarding_rule(self), but CloudstackNode doesn't has a method ex_delete_port_forwarding_rule - I fix this.

Ok, I'm rename method and fix test.

Sorry for bad english :(

Kami:
@gigimon Can you please also add tests which cover those newly added methods?

gigimon:
@Kami ok, later today I add tests

gigimon:
@Kami I've added tests, please check, if all good I create one commit with all changess

libcloud/compute/drivers/cloudstack.py
```diff
@@ -42,16 +42,36 @@ def ex_release_public_ip(self, address):
         "Release a public IP that this node holds."
         return self.driver.ex_release_public_ip(self, address)
 
-    def ex_add_ip_forwarding_rule(self, address, protocol, start_port,
-                                  end_port=None):
+    def ex_create_ip_forwarding_rule(self, address, protocol,
```
Kami, line 45: I just went over the pull request again and I'm kinda confused now.

Why did you rename methods from add to create (`ex_add_ip_forwarding_rule` -> `ex_create_ip_forwarding_rule`)? Is it for consistency reasons or is it something else?

gigimon, line 45: Yes, it's for consistency reason. In pull request @runseb say me rename to create

Kami, line 45: OK, sounds good. Need to note this in the changelog though since it's a backward incompatible change.

Kami:
(I also plan open a corresponding JIRA ticket so it will be easier to track this issue)

@runseb While reviewing this pull request I noticed a couple of unrelated issues in the driver.

`ex_allocate_public_ip` and `ex_release_public_ip` method on the `CloudStackNode` class seems to be broken.

Both of those methods call methods on the driver class with invalid arguments. They pass `node` object to the driver method, but the driver method doesn't take a node object.

See:
- https://github.com/apache/libcloud/blob/trunk/libcloud/compute/drivers/cloudstack.py#L39 
- https://github.com/apache/libcloud/blob/trunk/libcloud/compute/drivers/cloudstack.py#L828

We have two options:
1. Fix those methods to actually do what the docstring says - allocate the IP **and bind it to the node**
2.  Remove those methods from the `CloudStackNode` class since in the current form, they don't belong there.

We also need better test coverage to detect issues like this automatically in the future.

Edit: I've opened a JIRA issue for it - https://issues.apache.org/jira/browse/LIBCLOUD-462. We can move this off topic debate there.

Kami:
@gigimon Rebased the commits and merged those changes into trunk. This PR can be closed now.

Thanks.

gigimon:
@Kami done

PR #197, opened 2013-12-16
LIBCLOUD-450: Update GCE to v1

wrigri:
Update GCE to v1.
The primary change for v1 from v1beta16 is that ephemeral disks have gone away and now all disks are persistent.  This means that create_node is now creating a disk first if an existing disk isn't supplied.
Also, add support for snapshots.

This is obviously a fairly large change, although most of the actual changed lines are for test fixtures.  Now that v1 has launched, the treadmill of releases with large changes should slow down to a more reasonable pace.


libcloud/common/google.py
```diff
@@ -82,6 +82,8 @@
     from Crypto.Hash import SHA256
     from Crypto.PublicKey import RSA
     from Crypto.Signature import PKCS1_v1_5
+    import Crypto.Random
+    Crypto.Random.atfork()
```
Kami, line 90: Hm, we aren't forking anywhere, but if this reseeds the random generator and adds more entropy to the pool, explicitly calling it here is not such a bad idea.

Also, no wonder people don't like pycrypto. This seems like a pretty awful API for reseeding the RNDG.

libcloud/compute/drivers/gce.py
```diff
@@ -941,6 +962,20 @@ def list_sizes(self, location=None):
                 list_sizes = [self._to_node_size(s) for s in response['items']]
         return list_sizes
 
+    def ex_list_snapshots(self):
```
Kami, line 790: Hm, this lists the snapshots for the persistent disks, right?

Compute part of the API does include some standard methods and classes for managing block storage volumes and snapshots. `list_volume_snapshots` is one of the supported methods.

The problem is that those methods refer to the block storage volumes which are not the same as instance storage (or persistent disks in this case), so I'm not yet sure what we should do here.

I need to have a better look and think some more about it. I just want to make sure this whole thing doesn't confuse the user.

libcloud/compute/drivers/gce.py
```diff
@@ -1447,27 +1510,76 @@ def ex_create_multiple_nodes(self, base_name, size, image, number,
...
-                    error = e.value
-                    code = e.code
-                if response['status'] == 'DONE':
-                    responses[i] = None
-                    name = '%s-%03d' % (base_name, i)
-                    if error:
-                        node_list[i] = GCEFailedNode(name, error, code)
+            time.sleep(2)
+            for i, status in enumerate(status_list):
+                # If disk does not yet exist, check on its status
```
Kami, line 1263: Is it possible to refactor some of the functionality outside of this loop into multiple smaller methods?

Currently it's really dense and hard to read / follow.

libcloud/compute/drivers/gce.py
```diff
@@ -1447,27 +1510,76 @@ def ex_create_multiple_nodes(self, base_name, size, image, number,
...
+                if not status['disk']:
+                    error = None
+                    try:
+                        response = self.connection.request(
+                            status['disk_response']['selfLink']).object
+                    except:
```
Kami: What exception is thrown here? Can we catch a more specific one?

Even if we cant, you should use `except Exception`.

libcloud/compute/drivers/gce.py
```diff
@@ -1447,27 +1510,76 @@ def ex_create_multiple_nodes(self, base_name, size, image, number,
...
-                except:
-                    e = self._catch_error(ignore_errors=ignore_errors)
-                    error = e.value
-                    code = e.code
-                if response['status'] == 'DONE':
-                    responses[i] = None
-                    name = '%s-%03d' % (base_name, i)
-                    if error:
-                        node_list[i] = GCEFailedNode(name, error, code)
+            time.sleep(2)
```
Kami: Is it possible to make this delay configurable?

libcloud/compute/drivers/gce.py
```diff
@@ -1557,18 +1672,82 @@ def create_volume(self, size, name, location=None, image=None,
...
         if not hasattr(location, 'name'):
             location = self.ex_get_zone(location)
         request = '/zones/%s/disks' % (location.name)
-        self.connection.async_request(request, method='POST',
-                                      data=volume_data,
-                                      params=params)
+
+        return request, volume_data, params
+
+    def create_volume(self, size, name, location=None, image=None,
```
Kami: Standard API looks like this:

``` python
def create_volume(self, size, name, location=None, snapshot=None):
```

If a user would use regular and not keyword arguments, this method would not work as expected.

To make it compatible with the base API, please change the signature to:

``` python
def create_volume(self, size, name, location=None, snapshot=None, image=None, use_existing=True):
```

libcloud/common/google.py
wrigri, line 90: I think the issue is that even though libcloud isn't forking anywhere, if someone else is using pycrypto and then forks to call libcloud, it can cause an error.  This is basically a safety call that just makes sure that libcloud is working from a fresh state.  (There may be a better way to do it, but this seemed to be the general suggestion when I was searching for the particular error seen by a coworker working on some different third-party code.)

libcloud/compute/drivers/gce.py
wrigri, line 790: I'm open for suggestions on this.  I saw the list_volume_snapshots as well, but those were attributes of the storage volumes.  For GCE, snapshots are global resources that may just happen to have been created from a specific disk.

libcloud/compute/drivers/gce.py
wrigri, line 1263: Yeah, I'm not 100% happy with it either, so I'll give it another pass and see if I can improve it.

wrigri:
@Kami Thanks for the review.  I replied to a few specific comments, but overall, I don't disagree with any of your feedback, so I'll work on some improvements.

Kami:
@wrigri Sounds good. Thanks.

@wrigri @erjohnso It would also be good to add some GCE specific docs with examples to Libcloud (https://github.com/apache/libcloud/tree/trunk/docs). Then you can also add a link to GCE salt cloud provisioner there :)

libcloud/common/google.py
Kami, line 90: OK, thanks for the clarification. That's kinda what I thought.

wrigri:
@Kami I think I addressed all of your feedback.  Please take a look and let me know if you have other feedback or suggestions.  Also, let me know if you have any thoughts on the list_snapshots thing.

libcloud/compute/drivers/gce.py
Kami, line 790: OK, in this case we should leave this extension method as it is.

One thing I'm wondering is - can you filter snapshots for a specific volume? If you can, it would make sense to also implement `list_volume_snapshots` from the base API.

libcloud/compute/drivers/gce.py
```diff
@@ -1982,22 +2299,31 @@ def ex_destroy_network(self, network):
...
         :rtype:   ``bool``
         """
         request = '/zones/%s/instances/%s' % (node.extra['zone'].name,
                                               node.name)
         self.connection.async_request(request, method='DELETE')
+        if destroy_boot_disk and node.extra['boot_disk']:
+            node.extra['boot_disk'].destroy()
         return True
 
     def ex_destroy_multiple_nodes(self, nodelist, ignore_errors=True,
```
Kami: For consistency please use `node_list` instead of `nodelist`.

Kami:
The patch is pretty big, but I think it mostly looks good.

Only thing I would ask you is if you could re-organize the methods a bit - this means moving "private" (methods prefixed with an underscore) after the "public" method to the bottom (https://libcloud.readthedocs.org/en/latest/development.html#function-and-method-ordering).

This will make the code and Sphinx documentation generated from the docstrings a bit easier to read and follow.

wrigri:
@Kami I reordered the methods in gce.py, and made that minor change to nodelist.

Also, I went ahead and added list_volume_snapshots.  It turns out it wasn't too hard to filter by the volume they were created from.  If the volume is deleted, the volume reference is removed from the snapshot, but that shouldn't affect list_volume_snapshots negatively.

Kami:
@wrigri Looks good, thanks.

Can you please sync your branch with the latest trunk and squash the commits? After you do this, I will go ahead and merge the patch.

Kami:
Hm, the pull request looks really messed up now.

I've tried manually checking out your branch and squashing the commits, but I also encountered conflicts. To save both of us some time & pain, I will make an exception this time and directly merge your branch into trunk.

Kami:
I've merged your branch into trunk - thanks!

Also, as we've talked in the past, it would now be great to add some GCE specific documentation to Libcloud docs.

Kami:
@wrigri Ugh, I just wanted to update the JIRA issue and I noticed you have uploaded a squashed patch there.

Sorry for that, I have totally missed this before.

PR #199, opened 2013-12-20
Issue LIBCLOUD-464: Add security group delete/destroy into the EC2 drive...

cderamus:
...r

### DIFF
--- a/libcloud/compute/drivers/ec2.py
+++ b/libcloud/compute/drivers/ec2.py
@@ -1277,7 +1277,7 @@ class BaseEC2NodeDriver(NodeDriver):
                   'GroupDescription': description}
         return self.connection.request(self.path, params=params).object
 
-    def ex_destroy_security_group_by_id(self, group_id):
+    def ex_delete_security_group_by_id(self, group_id):
         """
         Deletes a new Security Group using the group id.
 
@@ -1294,7 +1294,7 @@ class BaseEC2NodeDriver(NodeDriver):
 
         return element == 'true'
 
-    def ex_destroy_security_group_by_name(self, group_name):
+    def ex_delete_security_group_by_name(self, group_name):
         """
         Deletes a new Security Group using the group name.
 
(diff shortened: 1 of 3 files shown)


libcloud/compute/drivers/ec2.py
```diff
@@ -1277,6 +1277,31 @@ def ex_create_security_group(self, name, description):
                   'GroupDescription': description}
         return self.connection.request(self.path, params=params).object
 
+    def ex_destroy_security_group(self, group_id=None, group_name=None):
```
Kami: For consistency with other drivers and to make it easier to promote security group management methods to a top-level API in the near future, I would do the following:
1. Add two new methods - `ex_destroy_security_group_by_id(group_id)` and `ex_destroy_security_group_by_name(group_name)`
2. Add a wrapper method - `ex_destroy_security_group(name)` which calls `ex_destroy_security_group_by_name` underneath

As noted above, those methods will hopefully be promoted to a top-level API in the near-future and guaranteeing consistency and compatibility between drivers will be a bit easier then.

Edit: Fixed a typo.

cderamus:
Let me know if this is more in the vein of what you'd like buddy.

Kami:
Ugh, sorry, looks like I've missed one thing in my previous review.

For consistency with other drivers, can you please also rename the methods from "destroy" to "delete" (`ex_destroy_security_group` -> `ex_delete_security_group`, ...) ?

Besides that, the patch looks good.

cderamus:
Ah, nice catch on that. I try to stay consistent, but I clearly missed that one. I've updated for your review!

Kami:
Thanks!

While reviewing and merging the patch, I've noticed that you forgot to update `ex_delete_security_group` method and the tests didn't detect that because they had a typo in the name so they didn't actually run (`text` instead of `test`).

I've squashed your commits, fixed those two issues in 4df8dd2781d986f702adfd1392b569cad86a2b8e & adba6dee07066cb9ad745af37d98c7a564aba23c and merged your patch into trunk.

When I find some more time, I'll also have a look at how we can automatically detect and prevent issues like this from happening in the feature. It seems that for a start, we could do a simple integration with coverage.py and make sure that all the newly added test cases are actually executed.

cderamus:
Ah, that's what I get for updating code with egg nog in hand. Thanks for catching that!

PR #202, opened 2013-12-20
Issue: LIBCLOUD-466 Add ingress/egress VPC support to security group rul...

cderamus:
...es

### DIFF
--- a/libcloud/compute/drivers/ec2.py
+++ b/libcloud/compute/drivers/ec2.py
@@ -1318,6 +1318,282 @@ class BaseEC2NodeDriver(NodeDriver):
             if e.args[0].find('InvalidPermission.Duplicate') == -1:
                 raise e
 
+    def ex_authorize_security_group_ingress(self, id, from_port, to_port,
+                                            cidr_ips=None, group_pairs=None,
+                                            protocol='tcp'):
+        """
+        Edit a Security Group to allow specific ingress traffic using
+        CIDR blocks or either a group ID, group name or user ID (account).
+
+        :param      id: The id of the security group to edit
+        :type       id: ``str``
+
+        :param      from_port: The beginning of the port range to open
+        :type       from_port: ``int``
+
+        :param      to_port: The end of the port range to open
+        :type       to_port: ``int``
+
+        :param      cidr_ips: The list of ip ranges to allow traffic for.
+        :type       cidr_ips: ``list``
+
+        :param      group_pairs: User/group pairs to allow traffic for.
+        :type       group_pairs: ``dict``
+
+        :param      protocol: tcp/udp/icmp
+        :type       protocol: ``str``
+
+        :rtype: ``bool``
+        """
+
+        params = {'Action': 'AuthorizeSecurityGroupIngress',
+                  'GroupId': id,
+                  'IpPermissions.1.IpProtocol': protocol,
+                  'IpPermissions.1.FromPort':   from_port,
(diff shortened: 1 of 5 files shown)


libcloud/compute/drivers/ec2.py
```diff
@@ -1318,6 +1318,282 @@ def ex_authorize_security_group(self, name, from_port, to_port, cidr_ip,
...
+        params = {'Action': 'AuthorizeSecurityGroupIngress',
+                  'GroupId': id,
+                  'IpPermissions.1.IpProtocol': protocol,
+                  'IpPermissions.1.FromPort':   from_port,
+                  'IpPermissions.1.ToPort':     to_port}
+
+        if cidr_ips is not None:
+            i = 1
+            ip_ranges = {}
+            for cidr_ip in cidr_ips:
```
Kami: You could get rid of two lines if you use `enumerate` which is considered more "Pythonic".

So something like:

``` python
for index, cidr_ip in enumerate(cidr_ips, 1)
...
```

libcloud/compute/drivers/ec2.py
```diff
@@ -1318,6 +1318,282 @@ def ex_authorize_security_group(self, name, from_port, to_port, cidr_ip,
...
+        :param      from_port: The beginning of the port range to open
+        :type       from_port: ``int``
+
+        :param      to_port: The end of the port range to open
+        :type       to_port: ``int``
+
+        :param      cidr_ips: The list of ip ranges to allow traffic for.
+        :type       cidr_ips: ``list``
+
+        :param      group_pairs: User/group pairs to allow traffic for.
```
Kami: It would be nice to include an example of couple of possible values for this argument since it's not immediately obvious after just reading the docstrings.

Kami: Also, seems like the type should be `list` of `dict` and not `dict`

libcloud/compute/drivers/ec2.py
```diff
@@ -1318,6 +1318,282 @@ def ex_authorize_security_group(self, name, from_port, to_port, cidr_ip,
...
+        :rtype: ``bool``
+        """
+
+        params = {'Action': 'AuthorizeSecurityGroupIngress',
+                  'GroupId': id,
+                  'IpPermissions.1.IpProtocol': protocol,
+                  'IpPermissions.1.FromPort':   from_port,
+                  'IpPermissions.1.ToPort':     to_port}
+
+        if cidr_ips is not None:
```
Kami: There seems to be a lot of duplicated code between the methods.

To avoid that, it would be great if you could split duplicated code which populates the `params` dict into a separate method. You could call it `_get_common_security_group_params` or something like that.

Then inside those methods, you could do something like:

``` python
params = self._get_common_params...
params['Action'] = 'Action'
```

libcloud/compute/drivers/ec2.py
cderamus: Nice call. I will make that change momentarily.

libcloud/compute/drivers/ec2.py
cderamus: Noted!

Kami:
While merging and testing the patch, I have noticed a test failure under Python 2.5. I have forgot that `enumerate` function in 2.5 doesn't support `start` argument.

I've fixed that, squashed the commits and merged patch into trunk.

Thanks!

PR #203, opened 2013-12-20
Issue: LIBCLOUD-477 Add list/delete/create network (VPC) calls within EC...

cderamus:
...2

### DIFF
--- a/libcloud/compute/drivers/ec2.py
+++ b/libcloud/compute/drivers/ec2.py
@@ -503,6 +503,24 @@ class ExEC2AvailabilityZone(object):
                 % (self.name, self.zone_state, self.region_name))
 
 
+class EC2Network(object):
+    """
+    Represents information about a VPC (Virtual Private Cloud) network
+
+    Note: This class is EC2 specific.
+    """
+
+    def __init__(self, id, name, cidr_block, extra=None):
+        self.id = id
+        self.name = name
+        self.cidr_block = cidr_block
+        self.extra = extra or {}
+
+    def __repr__(self):
+        return (('<EC2Network: id=%s, name=%s')
+                % (self.id, self.name))
+
+
 class BaseEC2NodeDriver(NodeDriver):
     """
     Base Amazon EC2 node driver.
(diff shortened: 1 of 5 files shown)


libcloud/compute/drivers/ec2.py
```diff
@@ -1236,6 +1236,131 @@ def ex_describe_keypair(self, name):
...
+        """
+        Return all private virtual private cloud (VPC) networks
+
+        :return:    list of network dicts
+        :rtype:     ``list``
+        """
+        params = {'Action': 'DescribeVpcs'}
+
+        result = self.connection.request(self.path,
+                                         params=params.copy()).object
```
Kami: Is `.copy` call necessary here?

`request` method shouldn't mutate `params` so the copy call here should be unnecessary.

libcloud/compute/drivers/ec2.py
```diff
@@ -1236,6 +1236,131 @@ def ex_describe_keypair(self, name):
             'keyFingerprint': fingerprint
         }
 
+    def ex_list_networks(self):
+        """
+        Return all private virtual private cloud (VPC) networks
+
+        :return:    list of network dicts
+        :rtype:     ``list``
```
Kami: `list` of `dict`

libcloud/compute/drivers/ec2.py
```diff
@@ -1236,6 +1236,131 @@ def ex_describe_keypair(self, name):
...
+                               xpath='vpcSet/item',
+                               namespace=NAMESPACE):
+
+            # Get the network id
+            vpc_id = findtext(element=element,
+                              xpath='vpcId',
+                              namespace=NAMESPACE)
+
+            # Get tags
+            tags = dict((findtext(element=item,
```
Kami: Looks like a regular for loop would make this couple of lines a bit more readable.

libcloud/compute/drivers/ec2.py
```diff
@@ -1236,6 +1236,131 @@ def ex_describe_keypair(self, name):
...
+                    'state': findtext(element=result,
+                                      xpath='vpc/state',
+                                      namespace=NAMESPACE),
+                    'cidr_block': findtext(element=result,
+                                           xpath='vpc/cidrBlock',
+                                           namespace=NAMESPACE)}
+
+        # Attempt to tag our network if the name was provided
+        if name is not None:
+            # Build a resource object
```
Kami: If you define a new class which represents a Network, you could get rid of this code. 

In cases like this, defining a class is also generally better than using a dict. Dict enforces no structure which means it's fragile and it also makes it harder for the end user to know what to expect.

Then you can also modify `ex_list_networks` to return a list of instances of this class.

cderamus: No objection here. I'll go ahead and make that change now.

libcloud/compute/drivers/ec2.py
cderamus: Certainly isn't, just came over with a copy paste from a different call :).

libcloud/compute/drivers/ec2.py
```diff
@@ -1236,6 +1321,84 @@ def ex_describe_keypair(self, name):
...
+        :rtype:     ``dict``
+        """
+        params = {'Action': 'CreateVpc',
+                  'CidrBlock': cidr_block,
+                  'InstanceTenancy':  instance_tenancy}
+
+        result = self.connection.request(self.path, params=params).object
+
+        # Get our properties
+        response = {'vpc_id': findtext(element=result,
```
Kami: This method should now also return `EC2Network` instance.

Something like `network = self._to_network(elem=respose.object` should do (you might need to select a correct element first though).

You can then also pass this instance to `self.ex_create_tags` method.

Kami: I would also appreciate if you could synchronize your branch with trunk after you make those changes.

Otherwise I will have a bad time merging this patch since there will be a bunch of merge conflicts.

cderamus: Yup my fault on that. I'll make sure to synchronize my branch before submitting future PRs.

libcloud/compute/drivers/ec2.py
```diff
@@ -1408,6 +1491,70 @@ def ex_describe_keypair(self, name):
...
+                                            namespace=NAMESPACE))[0]
+
+        network = self._to_network(element)
+
+        if name is not None:
+            self.ex_create_tags(network, {'Name': name})
+
+        return network
+
+    def ex_delete_network(self, vpc_id):
```
Kami, line 1541: Now that we have `EC2Network` class, this method should take an instance of this class to be consistent with other methods (delete_key_pair, destroy_volume, etc.).

If you only have an ID, you can always constructor the `EC2Network` object manually and pass it to this method to avoid the `ex_list_networks` call.

libcloud/compute/drivers/ec2.py
Kami: No problem and thanks for all those PRs :)

Kami:
@cderamus I went ahead, made the "ex_delete_network" change mentioned above and merged changes into trunk.

I had some issues (conflicts) with applying the patch and squashing the commits so I just went ahead and directly merged your branch into trunk.

PR #205, opened 2013-12-22
Libcloud 469 add reserved instance support

cderamus:


libcloud/compute/drivers/ec2.py
```diff
@@ -559,6 +559,22 @@ def __repr__(self):
                 % (self.name, self.zone_state, self.region_name))
 
 
+class ExEC2ReservedNode(object):
```
Kami: You can drop the `Ex` prefix. I know one class in EC2 driver uses it, but that's mostly an artifact from the past.

Nowadays we only use ex prefix for methods and arguments.

Kami: This class could potentially inherit from `Node` and you can then just pass `None` for `public_ips` and `private_ips` argument, right?

Adding docstrings for attributes which are not already present on the Node class would also be good (just by looking at the constructor, I don't know what the `type` attribute represents).

libcloud/compute/drivers/ec2.py
```diff
@@ -608,6 +624,55 @@ def _get_terminate_boolean(self, element):
...
+            id=findtext(element=element,
+                        xpath='reservedInstancesId',
+                        namespace=NAMESPACE),
+            type=findtext(element=element,
+                          xpath='instanceType',
+                          namespace=NAMESPACE),
+            state=findattr(element=element,
+                           xpath="state",
+                           namespace=NAMESPACE),
+            driver=self.connection.driver,
```
Kami: `driver=self` is shorter :)

libcloud/compute/drivers/ec2.py
```diff
@@ -608,6 +624,55 @@ def _get_terminate_boolean(self, element):
...
+    def _to_reserved_node(self, element):
+        reservation = ExEC2ReservedNode(
+            id=findtext(element=element,
+                        xpath='reservedInstancesId',
+                        namespace=NAMESPACE),
+            type=findtext(element=element,
+                          xpath='instanceType',
+                          namespace=NAMESPACE),
+            state=findattr(element=element,
+                           xpath="state",
```
Kami: For consistency, please use single quotes around strings everywhere except for docstrings.

libcloud/compute/drivers/ec2.py
```diff
@@ -608,6 +624,55 @@ def _get_terminate_boolean(self, element):
...
+            extra={'availability': findattr(element=element,
+                                            xpath="availabilityZone",
+                                            namespace=NAMESPACE),
+                   'start': findattr(element=element,
+                                     xpath="start",
+                                     namespace=NAMESPACE),
+                   'duration': findattr(element=element,
+                                        xpath="duration",
+                                        namespace=NAMESPACE),
+                   'usage_price': findattr(element=element,
```
Kami: This attribute should probably be float, right?

libcloud/compute/drivers/ec2.py
```diff
@@ -608,6 +624,55 @@ def _get_terminate_boolean(self, element):
...
+                                           xpath="usagePrice",
+                                           namespace=NAMESPACE),
+                   'fixed_price': findattr(element=element,
+                                           xpath="fixedPrice",
+                                           namespace=NAMESPACE),
+                   'instance_count': findattr(element=element,
```
Kami: int? :)

libcloud/compute/drivers/ec2.py
```diff
@@ -608,6 +624,55 @@ def _get_terminate_boolean(self, element):
...
+                        xpath='reservedInstancesId',
+                        namespace=NAMESPACE),
+            type=findtext(element=element,
+                          xpath='instanceType',
+                          namespace=NAMESPACE),
+            state=findattr(element=element,
+                           xpath="state",
+                           namespace=NAMESPACE),
+            driver=self.connection.driver,
+            extra={'availability': findattr(element=element,
```
Kami: To get rid of a bunch of lines and making it easier to add a new attributes, you could use a for loop.

So, something like:

``` python
extra_attributes_map = {
    'availability': {
        'xpath:' 'availabilityZone',
        'type': str
    }
}

for attribute, values in extra_attributes_map.items():
   type_func = values['type']
   name = values['name']
   value = findattr(element=element, xpath=values['xpath'],
                               namespace=NAMESPACE)
   extra[attribute] = cast_func(value)
   ...
```

cderamus:
I went ahead and made the requested changes. To inherit from Node I also had to set the name argument to None as reserved instances do not have names or tags where we can pull the Name key/value pair.

Kami:
Thanks!

The PR looks mostly good, but I've made some minor changes after merging the patch:
1. I've modified `EC2ReservedNode` constructor to call parent constructor - 49e2f3bf55b9ef510e48c8c7174d7377f27e620a
2. I've removed unused arguments from the `EC2ReservedNode` constructor method - 49e2f3bf55b9ef510e48c8c7174d7377f27e620a
3. I've also passed `size` attribute to the `EC2ReservedNode` constructor. We can do this because `list_sizes` call is cheap (it doesn't result in HTTP request since sizes are stored locally in the module). - 11abec495e0125cf037e468b2dbed70d5254a055
4. I've moved test from `NimbusTests` to `EC2Tests` class since this functionality is EC2 specific. I imagine you just looked for `list_nodes` tests and you accidentally put `list_reserved_nodes` test above it. - 1655e850a71cca1af35ece6f795aa44fa5029958
5. I've renamed `list_reserved_nodes` method to `ex_list_reserved_nodes`. Sorry if I've confused you with "you can drop the ex prefix from the class name" comment. We still use `ex_` prefix, but only for method and argument names.

cderamus:
Ah, thanks for that and good call on the list_sizes.

PR #206, opened 2013-12-25
LIBCLOUD-470: VPC Elastic IP Support

cderamus:
I tried to make this change without impacting some of the old calls (eg: ex_describe_addresses) using keyword arguments. All of my tests using both the mock data and my own AWS account appear to work without issue for both EC2 and VPC elastic IP addresses. 

If you think it's better to create a class for network addresses such as EC2NetworkAddress I can certainly make that change. 

Lastly, Merry Christmas and enjoy the holiday!


Kami:
@cderamus Hey, I've noticed you closed the PR. I assume this is intentional? :)

And happy holidays to you as well :)

cderamus:
Yup there was an issue that I corrected shortly thereafter and I also assumed that you wouldn't want a PR with all of the other commits from prior issues/PRs. If that isn't a problem and is okay then I'll just re-open it now and you can review it at your convenience. 

What are you doing reading this on Christmas anyways? You should be relaxing!

Kami:
Yeah, that's correct - PR should be based directly off trunk and self-sustaining.

Merging pull requests can sometimes be relaxing. And I should ask you the same thing :)

PR #208, opened 2013-12-26
LIBCLOUD-470: Add VPC Elastic IP Support

cderamus:


libcloud/compute/drivers/ec2.py
```diff
@@ -961,6 +961,61 @@ def _to_network(self, element):
 
         return EC2Network(vpc_id, name, cidr_block, extra=extra)
 
+    def _to_addresses(self, response, only_allocated, all_properties):
```
Kami: Please add a docstring which describes what `only_allocated` and `all_properties` arguments do.

libcloud/compute/drivers/ec2.py
```diff
@@ -2119,113 +2174,166 @@ def _add_instance_filter(self, params, node):
...
+                            (standard/vpc)
+        :type       domain: ``str``
 
-        :return: String representation of allocated IP address
-        :rtype: ``str``
+        :return:    Dictionary of elastic IP properties
+        :rtype:     ``dict``
         """
-        params = {'Action': 'AllocateAddress'}
+        params = {'Action': 'AllocateAddress', 'Domain': domain}
```
Kami: Hm, documentation (http://docs.aws.amazon.com/AWSEC2/latest/APIReference/ApiReference-query-AllocateAddress.html) says that the only supported value for `Domain` query parameter is `vpc`, but you are also letting `standard` through.

Unless I'm missing something, you should only include `Domain` in the `params` if `domain` == `vpc`.

Edit: updated the comment.

libcloud/compute/drivers/ec2.py
```diff
@@ -2119,113 +2174,166 @@ def _add_instance_filter(self, params, node):
...
-        :return: True on success, False otherwise.
-        :rtype: ``bool``
+        :param      allocation_id: The allocation ID of the elastic IP
+                                   (VPC only)
+        :type       allocation_id: ``str``
+
+        :return:    A string representation of the association ID which is
+                    required for VPC disassociation. EC2/standard
+                    addresses return None
+        :rtype:     ``NoneType`` or ``str``
```
Kami: Should be `None` iirc

libcloud/compute/drivers/ec2.py
```diff
@@ -2119,113 +2174,166 @@ def _add_instance_filter(self, params, node):
...
+    def ex_associate_addresses(self, node, elastic_ip_address=None,
+                               allocation_id=None):
         """
         Note: This method has been deprecated in favor of
         the ex_associate_address_with_node method.
         """
-        return self.ex_associate_address_with_node(node=node,
-                                                   elastic_ip_address=
-                                                   elastic_ip_address)
+        if elastic_ip_address is not None:
```
Kami: I've noticed you use two different styles for the same check in the code.

Once you compare it to `None` (`is not None`) and the other time you just check if it's truthy (`if foo`). It would be good if you could use a consistent style.

I usually prefer to be explicit and use `is not None` check, but a valid value in this case is always truthy (0, empty list, None, etc. are not valid values for this argument) so you can get away with just checking if the argument is truthy.

libcloud/compute/drivers/ec2.py
```diff
@@ -2119,113 +2174,166 @@ def _add_instance_filter(self, params, node):
...
-    def ex_associate_addresses(self, node, elastic_ip_address):
+        response = self.connection.request(self.path, params=params).object
+        association_id = findtext(element=response,
+                                  xpath='associationId',
+                                  namespace=NAMESPACE)
+        return association_id
+
+    def ex_associate_addresses(self, node, elastic_ip_address=None,
+                               allocation_id=None):
         """
```
Kami, line 2785: This method won't work in the same way as `ex_associate_address_with_node`, because it only uses the first argument if user specifies both.

You should do:

``` python
return self.ex_associate_address_with_node(node=node,
                                                    elastic_ip_address=
                                                    elastic_ip_address,
                                                    allocation_id=allocation_id)
```

Or alternatively, if those arguments are mutually exclusive, you should throw.

libcloud/compute/drivers/ec2.py
cderamus: Noted. I almost always use is not None myself across all of my code and will try to remain consistent for this project.

libcloud/compute/drivers/ec2.py
cderamus, line 2785: They are indeed mutually exclusive and Amazon will throw an exception if you try to pass both as parameters. I'll go ahead and throw.

cderamus:
I went ahead and added the raise condition into ex_associate_address_with_node so it gets thrown should the user use that method or the deprecated ex_associate_address method.

libcloud/compute/drivers/ec2.py
```diff
@@ -2119,113 +2189,169 @@ def _add_instance_filter(self, params, node):
...
+                                        (VPC only)
+        :type       association_id: ``str``
+
+        :return:    True on success, False otherwise.
+        :rtype:     ``bool``
         """
         params = {'Action': 'DisassociateAddress'}
 
-        params.update({'PublicIp': elastic_ip_address})
+        if elastic_ip_address:
```
Kami: You should also throw here if both arguments are specified, right?

Please also add a test case for this.

cderamus: Right you are. I will take care of that later this morning.

Jc2k:
My reading of this is that `only_allocated` shows addresses that are allocated, but not associated. Have I misunderstood? If not, would `only_associated` be a better name?

Jc2k:
Also I wonder if allocation_id should be encapsulated within an ElasticIP class, so instead of passing elastic_ip_address AND allocation_id, you pass a single object. 

(This PR has caused me to look at the various elastic IP API's we have for other providers - bright box, cloudstack, openstack, haven't looked further yet - and i'm going to try and propose a 'standard' API for the feature. Hiding allocation_id would be needed to do that).

cderamus:
Yes, only associated would likely be a better name for this and certainly comes with less ambiguity.

cderamus:
Given that virtually all cloud vendors have some form of public/floating/elastic IP addresses I am all for creating a standard API for this with a class (eg: NodeAddress). I do wish Amazon wouldn't require the allocation ID for VPC related actions as it would certainly make this a bit easier.

cderamus:
Just checking in to see what direction you guys would like to go to support this functionality. If @Jc2k is indeed going to propose promoting this to the base API then I'll go ahead and close this out.

Kami:
@cderamus If a standard API promotion is going to take a while, I'm OK with merging this into trunk if you can update the code to use `ElasticIP` class as suggested by @Jc2k.

Jc2k:
+1. Hope to look at core API soon, but having this merged suits me.

Kami:
@cderamus Also, I don't want to rush you, but the whole 0.14.0 release has already been taking longer than usual and I would like to start a voting thread for it soon and it would be great if we could include this change in it.

cderamus:
@Kami I live for pressure! I'm planning on dedicating a few hours to the project this evening and will bump this to the top of my list. I should have no issues getting this shored up.

Kami:
@cderamus heh, great! :)

cderamus:
Sorry about the bad commit message on this one guys. I went ahead and added a new ElasticIp class into the driver and put all VPC specific attributes into the extra dict. As I was making the change I couldn't help but think that the all_properties boolean is sort of silly now that we're using a class.

cderamus:
I'll be submitting a follow-up with some additional changes momentarily.

cderamus:
@Jc2k I believe this is what you wanted. Leveraging the ElasticIp instance required a new keyword argument 'domain' which is only used for VPC actions. If anything other than vpc is passed in the class will throw. Thanks for the suggestions as this was definitely a better approach.

libcloud/compute/drivers/ec2.py
```diff
@@ -2582,113 +2675,152 @@ def _add_instance_filter(self, params, node):
...
+        :param      domain: The domain where the IP resides (vpc only)
+        :type       domain: ``str``
+
+        :return:    True on success, False otherwise.
+        :rtype:     ``bool``
         """
         params = {'Action': 'ReleaseAddress'}
 
-        params.update({'PublicIp': elastic_ip_address})
+        if domain is not None and domain is not 'vpc':
```
Kami: Second check won't work everywhere.

`is` compares identifies (e.g. if two variables point to the same objects)  which means you should only use it for immutable types (`True`, `False`, `None`, ...) otherwise you might get an unexpected result.

When you are comparing values you should use `==` or `!=` in this case.

This is also the reason why PyPy tests are currently failing.

libcloud/compute/drivers/ec2.py
```diff
@@ -920,10 +942,30 @@ def __init__(self, id, name, state, extra=None):
...
+class ElasticIp(object):
+    """
+    Represents information about an elastic IP adddress
+
+    Note: This class is used to support both EC2 and VPC IPs.
+          For VPC specific attributes are stored in the extra
+          dict to make promotion to the base API easier.
+    """
+
+    def __init__(self, ip, domain, instance_id, extra=None):
```
Kami, line 970: Can you please add some docstrings which document what each attribute represents.

libcloud/compute/drivers/ec2.py
```diff
@@ -1236,6 +1278,57 @@ def _to_network(self, element):
...
+        # Build our extra dict
+        extra = self._get_extra_dict(
+            element, RESOURCE_EXTRA_ATTRIBUTES_MAP['elastic_ip'])
+
+        # Return NoneType if only associated IPs are requested
+        if only_associated and not instance_id:
+            return None
+
+        # If all properties are not requested, only send back the IP
+        if not all_properties:
```
Kami: It looks like now that we use `ElasticIP` abstraction there is not much more sense in having `all_properties` arguments anymore so we can get rid of it.

libcloud/compute/drivers/ec2.py
cderamus: Aye, will make that change now. Thanks!

libcloud/compute/drivers/ec2.py
cderamus, line 970: Absolutely.

libcloud/compute/drivers/ec2.py
cderamus: Fantastic that's what I was thinking as well.

libcloud/compute/drivers/ec2.py
```diff
@@ -920,10 +942,42 @@ def __init__(self, id, name, state, extra=None):
         self.extra = extra or {}
 
     def __repr__(self):
-        return (('<EC2NetworkInterface: id=%s, name=%s')
+        return (('<EC2NetworkInterface: id=%s, name=%s>')
                 % (self.id, self.name))
 
 
+class ElasticIp(object):
```
Kami: Hm, I just checked the code and it seems like there is an even split between `Ip` and `IP`.

I prefer `IP` (`ElasticIP`) over `Ip` since it's an abbreviation for Internet Protocol. What do other ppl think?

``` bash
kami /w/lc/libcloud/libcloud (git:trunk)$ grep -r IP . | grep class
./test/loadbalancer/test_rackspace.py:class RackspaceLBWithVIPMockHttp(MockHttpTestCase):
./common/gandi.py:class IPAddress(BaseObject):
./compute/drivers/cloudstack.py:class CloudStackIPForwardingRule(object):

kami /w/lc/libcloud/libcloud (git:trunk)$ grep -r Ip . | grep class
./common/gogrid.py:class GoGridIpAddress(object):
./compute/drivers/openstack.py:class OpenStack_1_0_SharedIpGroup(object):
./compute/drivers/openstack.py:class OpenStack_1_0_NodeIpAddresses(object):
./compute/drivers/openstack.py:class OpenStack_1_1_FloatingIpPool(object):
./compute/drivers/openstack.py:class OpenStack_1_1_FloatingIpAddress(object):
```

cderamus:
+1 for IP

Kami:
@cderamus Can you please make this change, sync your branch and squash the commits?

I'm having problems (conflicts) doing this myself.

cderamus:
Sure I'll take care of that shortly.

cderamus:
@Kami I am also having issues when attempting to squash the commits. Want me to just close this PR and open up a new one with a clean fork? That may be easier at this point.

Kami:
@cderamus Yeah, if it makes it less painful than resolving the conflicts, then go for it.

PR #209, opened 2013-12-26
Issue LIBCLOUD-471: Add Get Console Output

cderamus:


libcloud/compute/drivers/ec2.py
```diff
@@ -2535,6 +2535,32 @@ def destroy_node(self, node):
...
+        :rtype: ``dict``
+        """
+        params = {
+            'Action': 'GetConsoleOutput',
+            'InstanceId': node.id
+        }
+
+        response = self.connection.request(self.path, params=params).object
+
+        return {'instance_id': findattr(element=response,
```
Kami: Hm, you already need to have the instance id to retrieve the output (`node.id`) so I'm not sure how value does it add to return a `dict` here.

Unless `timestamp` is really crucial, I would prefer the method to directly return the console output (`str`).

libcloud/compute/drivers/ec2.py
```diff
@@ -2535,6 +2535,32 @@ def destroy_node(self, node):
...
+                                        xpath='instanceId',
+                                        namespace=NAMESPACE),
+                'timestamp': findattr(element=response,
+                                      xpath='timestamp',
+                                      namespace=NAMESPACE),
+                'output': findattr(element=response,
```
Kami: To make it more user-friendly and useful, this function should also decode base64 encoded output and return a raw string.

``` python
base64.b64decode(b(base64_string))
```

Should do the trick.

libcloud/compute/drivers/ec2.py
cderamus: I would have a hard time arguing that it's crucial :). I actually just thought it would be a nicety to have available if someone wanted it. If you want me to make that change I certainly can do that.

libcloud/compute/drivers/ec2.py
cderamus: I originally had that in there, but didn't know if you would want to bring in additional imports into the driver. I agree it's a good idea to decode. I'll update that now.

libcloud/compute/drivers/ec2.py
Kami: After thinking about it some more, I actually think it would be useful to have access to `timestamp` here, so you should leave it as it is.

One use case I had in mind is determining if node is "stuck". If you have access to `timestamp` it makes it easier to do that. You can just periodically call this function and check if `timestamp` attribute has changed.

To make it more clear and useful, you should do two things:
- Update the docstring and document which keys the return `dict` will always contain
- Parse the timestamp and return a `datetime` object (`libcloud.utils.iso8601.parse_date` should do the trick)

cderamus:
Good to go!

Kami:
I've made two changes and merged patch into trunk:
1. Removed obsolete comment from the docstring and clarified which keys the returned dictionary contains. The docstring you have provided didn't help much because you would still need to check the code to see which keys the dictionary contains.
2. Fixed test failure on Python 3.3. `base64.b64decode` returns `bytes` in Python 3 and you need to call `.decode` on it to get a string back.

cderamus:
Great, thanks!

PR #210, opened 2013-12-26
LIBCLOUD 473 Add EBS Volume Attributes

cderamus:
https://issues.apache.org/jira/browse/LIBCLOUD-473

### DIFF
--- a/libcloud/compute/drivers/ec2.py
+++ b/libcloud/compute/drivers/ec2.py
@@ -759,13 +759,9 @@ class BaseEC2NodeDriver(NodeDriver):
 
         instance_id = findtext(element=element, xpath='instanceId',
                                namespace=NAMESPACE)
-        tags = dict((findtext(element=item, xpath='key', namespace=NAMESPACE),
-                     findtext(element=item, xpath='value',
-                              namespace=NAMESPACE))
-                    for item in findall(element=element,
-                                        xpath='tagSet/item',
-                                        namespace=NAMESPACE)
-                    )
+
+        # Get our tags
+        tags = self._get_resource_tags(element)
 
         name = tags.get('Name', instance_id)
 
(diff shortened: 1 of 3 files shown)


libcloud/compute/drivers/ec2.py
```diff
@@ -876,19 +876,89 @@ def _to_volume(self, element, name):
...
+
+            value = findtext(element=tag,
+                             xpath='value',
+                             namespace=NAMESPACE)
+
+            tags[key] = value
+
+        # Set our name if the Name key/value if available
+        # If we don't get anything back then use the volume id
+        name = tags.get('Name', volId)
```
Kami: This now shadows `name` argument which is passed to this function.

Maybe a better approach would be to use `name` argument  which is passed to the function (if available), otherwise fall back to retrieving a name from the tags (`name = name if name else tags.get('Name', volId)).

Kami: While you are at it, I would also do the following changes:
1. Modify `_to_volume` method signature and default `name` to `None`
2. Modify `self._to_volume(el,``)` call to omit name argument (empty string) all together
3. Add a docstring which explains that if not provided, name is obtained from the tags (if available)

libcloud/compute/drivers/ec2.py
```diff
@@ -876,19 +876,89 @@ def _to_volume(self, element, name):
...
                          namespace=NAMESPACE)
         size = findtext(element=element, xpath='size', namespace=NAMESPACE)
-        state = findtext(element=element, xpath='status', namespace=NAMESPACE)
-        create_time = findtext(element=element, xpath='createTime',
-                               namespace=NAMESPACE)
-        device = findtext(element=element, xpath='attachmentSet/item/device',
-                          namespace=NAMESPACE)
+
+        # Get our tag items
+        tag_items = findall(element=element,
```
Kami: There seem to be multiple places in the file where the tags are read into the dict and exactly the same code is repeated (`ex_describe_tags`, `_to_node`, `_to_network`, etc.)

It would be great if this functionality could be refactored in a separate function and re-used.

libcloud/compute/drivers/ec2.py
```diff
@@ -876,19 +876,89 @@ def _to_volume(self, element, name):
...
+
+        # Define and build our extra dictionary
+        extra = {}
+        for attribute, values in extra_attributes_map.items():
+            type_func = values['type']
+            value = findattr(element=element, xpath=values['xpath'],
+                             namespace=NAMESPACE)
+            if value is not None:
+                # Convert our create/attach time to ISO 8601
+                if attribute == 'create_time' or attribute == 'attach_time':
```
Kami: Maybe instead of `type`, you could call it a `format_func` / `cast_func`, specify `parse_date` as `format_func` for those two attributes and get rid of this if.

libcloud/compute/drivers/ec2.py
```diff
@@ -876,19 +876,89 @@ def _to_volume(self, element, name):
...
+            },
+            'delete': {
+                'xpath': 'attachmentSet/item/deleteOnTermination',
+                'type': str
+            }
+        }
+
+        # Define and build our extra dictionary
+        extra = {}
+        for attribute, values in extra_attributes_map.items():
```
Kami, line 934: This functionality now also appears to be repeated multiple times in this file.

It would be great if you could refactor this into a separate function (e.g. `_extract_extra_attributes` or something like that) which takes extraction definition (basically, `extra_attributes_map` argument) and returns a `dict` with extracted attributes.

libcloud/compute/drivers/ec2.py
cderamus: Not a problem, will do.

libcloud/compute/drivers/ec2.py
cderamus, line 934: I was thinking about doing that actually, but wanted to submit it as a separate issue for tracking purposes. Once this PR is closed I'll go ahead and create a JIRA issue for that and go through all calls where the extra_attributes_map exists if you're okay with that approach.

libcloud/compute/drivers/ec2.py
cderamus: Sounds good.

libcloud/compute/drivers/ec2.py
Kami, line 934: Yeah, I'm fine with that.

cderamus:
Assuming that you're okay with these changes I will create a new branch later today to convert all extra_attributes_map dictionaries from type -> cast_func to remain consistent and also will set up the new utility function that we previously discussed. Thanks!

Kami:
@cderamus Yeah, sounds good.

I've squashed the commits, updated the docstring and merged patch into trunk. Thanks!

PR #212, opened 2013-12-27
LIBCLOUD-477 Add Extra Utility Function

cderamus:
https://issues.apache.org/jira/browse/LIBCLOUD-477


Kami:
Thanks.

The patch looks good so I've merged in into trunk. I just made one little change (hope you don't mind).

I've renamed `cast_func` to `transform_func`. After giving it some more thought, I think `transform_func` is a better name because the function "transforms" the value.

cderamus:
Works for me!

PR #213, opened 2013-12-27
LIBCLOUD-474: Add Network Interface Support

cderamus:
https://issues.apache.org/jira/browse/LIBCLOUD-474


libcloud/compute/drivers/ec2.py
```diff
@@ -2435,6 +2595,134 @@ def ex_describe_addresses_for_node(self, node):
...
+        :return:    List of EC2NetworkInterface instances
+        :rtype:     ``list`` of :class `EC2NetworkInterface`
+        """
+        params = {'Action': 'DescribeNetworkInterfaces'}
+
+        return self._to_interfaces(
+            self.connection.request(self.path, params=params).object
+        )
+
+    def ex_create_network_interface(self, subnet_id, name=None,
```
Kami: Instead of `subnet_id`, this method should probably take `subnet` (instance of `EC2NetworkSubnet`).

libcloud/compute/drivers/ec2.py
```diff
@@ -2435,6 +2595,134 @@ def ex_describe_addresses_for_node(self, node):
...
+
+        interface = self._to_interface(element, name)
+
+        if name is not None:
+            tags = {'Name': name}
+            self.ex_create_tags(resource=interface, tags=tags)
+
+        return interface
+
+    def ex_delete_network_interface(self, network_interface_id):
```
Kami: Same here, this should take `network_interface` argument and `EC2NetworkInterface` instance.

libcloud/compute/drivers/ec2.py
```diff
@@ -2435,6 +2595,134 @@ def ex_describe_addresses_for_node(self, node):
...
+        params = {'Action': 'DeleteNetworkInterface',
+                  'NetworkInterfaceId': network_interface_id}
+
+        result = self.connection.request(self.path, params=params).object
+        element = findtext(element=result, xpath='return',
+                           namespace=NAMESPACE)
+
+        return element == 'true'
+
+    def ex_attach_network_interface(self, network_interface_id,
```
Kami: Same here, instance instead of id.

libcloud/compute/drivers/ec2.py
```diff
@@ -2435,6 +2595,134 @@ def ex_describe_addresses_for_node(self, node):
...
+        :param      network_interface_id: The ID of the interface
+        :type       network_interface_id: ``str``
+
+        :param      node: Node instance
+        :type       node: :class:`Node`
+
+        :param      device_index: The interface device index
+        :type       device_index: ``int``
+
+        :rtype:     ``str``
```
Kami, line 2696: Also add `:return` docstring which states that it returns an attachment id.

libcloud/compute/drivers/ec2.py
Kami: For explicitness sake and consistency with `ex_associate_address_with_node` and other methods, this method should probably be called `ex_attach_network_interface_to_node`.

libcloud/compute/drivers/ec2.py
```diff
@@ -2435,6 +2595,134 @@ def ex_describe_addresses_for_node(self, node):
...
+
+        return attachment_id
+
+    def ex_detach_network_interface(self, attachment_id, force=False):
+        """
+        Detatch a network interface from an instance.
+
+        :param      attachment_id: The attachment ID associated with the
+                                   interface
+        :type       attahcment_id: ``str``
```
Kami: Typo :)

libcloud/compute/drivers/ec2.py
```diff
@@ -612,6 +612,24 @@ def __repr__(self):
...
 
 
+class EC2NetworkInterface(object):
+    """
+    Represents information about a VPC network interface
+
+    Note: This class is EC2 specific.
+    """
+
+    def __init__(self, id, name, state, extra=None):
```
Kami, line 624: It would be great if you could document all the possible values for the `state` argument.

cderamus, line 624: Sure thing, I will add that in later this evening.

libcloud/compute/drivers/ec2.py
```diff
@@ -2435,6 +2595,136 @@ def ex_describe_addresses_for_node(self, node):
...
+        :param      node: EC2NetworkSubnet instance
+        :type       node: :class:`EC2NetworkSubnet`
+
+        :param      name:  Optional name of the interface
+        :type       name:  ``str``
+
+        :param      description:  Optional description of the network interface
+        :type       description:  ``str``
+
+        :param      private_ip_address:  Optional IP address assigned. If not
```
Kami: Looking at the docs (http://docs.aws.amazon.com/AWSEC2/latest/APIReference/ApiReference-query-CreateNetworkInterface.html), EC2 allows you to assign multiple private IP addresses to an interface, but only one primary address.

It would be good to clarify in the docstring that this is a primary private IP address.

libcloud/compute/drivers/ec2.py
```diff
@@ -2435,6 +2595,136 @@ def ex_describe_addresses_for_node(self, node):
...
+
+    def ex_detach_network_interface(self, attachment_id, force=False):
+        """
+        Detatch a network interface from an instance.
+
+        :param      attachment_id: The attachment ID associated with the
+                                   interface
+        :type       attachment_id: ``str``
+
+        :param      force: Forces the detachment.
```
Kami, line 2717: (just thinking out loud)

I wonder what this argument actually does. The documentation doesn't say anything besides "Specifies whether to force a detachment.".

I would assume it means it will also detach the interface even if it's currently in use.

libcloud/compute/drivers/ec2.py
```diff
@@ -2435,6 +2595,136 @@ def ex_describe_addresses_for_node(self, node):
...
+        :type       force: ``bool``
+
+        :rtype:     ``bool``
```
Kami, line 2721: Would be good to add

``` python
:return: ``True`` on successful detachment, ``False`` otherwise.
```

Kami:
Besides the couple of comments I've added it looks good, thanks!

libcloud/compute/drivers/ec2.py
cderamus, line 2717: It's rarely used, but there have been some edge cases where I have had to pass this over the wire. There's been the rare occasion where my instance was in a hung state and I was not able to detach the interface without performing a full server cycle. 

Again it's a rarity just like the volume force detachment parameter; however, I figured I would include it just on the off chance someone needed it :).

Kami, line 2717: Thanks for addition info / clarification.

And yeah, I see no problems with including it. I was just wondering what it does since EC2 docs were also very vague about it :)

Kami:
Squashed and merged into trunk. Thanks!

PR #215, opened 2013-12-29
LIBCLOUD 481: Add Image Properties

cderamus:
https://issues.apache.org/jira/browse/LIBCLOUD-481

While adding these properties I also noticed that the ex_imageids filter was broken. I've updated this and added a new test to validate that it works as expected. I also ran multiple live tests against my own Amazon account to ensure that it works as anticipated.


Kami:
Merged into trunk, thanks!

I would also appreciate if going forward, you could write a bit more descriptive commit messages. This means you should at the minimum include what have you changed / added / removed and in which driver.

As you have probably noticed, I reword most of the commits while merging the patch, so it's not a big of a deal, but if you could do it, it would save me some time and I could preserve the original commit message which would be ideal :)

cderamus:
Not a problem Tomaz and thanks for the assist.

PR #217, opened 2014-01-02
LIBCLOUD-486: Rename ex_destroy_image -> ex_delete_image

cderamus:
https://issues.apache.org/jira/browse/LIBCLOUD-486


Kami:
@cderamus Sorry for the delay.

For now, I would leave this as it is (`ex_destroy_image`). The reason for that is that there are some other drivers which also use this name. On top of that, renaming the method also breaks backward compatibility. Or if we decide to add a wrapper for the old method, this means we now need to support two methods until we fully deprecate it. None of this is ideal.

I know there is some confusion and no clear guidelines on when to use destroy and when to use delete. Most of the old APIs use destroy (e.g. `destroy_node`) and the new ones use delete.

I will try to find some time in the near future and publish some clear guidelines on the method naming.

cderamus:
Ah okay, my mistake for not going through the other drivers to validate that before proposing the change. 

No worries about the guidelines, it was only about five minutes of work :). Cheers.

PR #220, opened 2014-01-11
LIBCLOUD-470: Add VPC Elastic IP Support

cderamus:
This is a re-submission of LIBCLOUD-470 which adds full support for EC2/VPC Elastic IP addresses. Tests are included and everything checks out with my AWS developer account. Please let me know if you need anything else to get this merged into trunk.


Kami:
Merged into trunk, thanks!

PR #221, opened 2014-01-12
LIBCLOUD-493: Add additional data into the Node.extra dictionary

cderamus:
Note that this PR removes libcloud/test/compute/fixtures/ec2/describe_instances_with_tags.xml as it is no longer necessary. The tag tests are done within test_list_nodes now. Also note that I changed the name of a few select properties to match the libcloud standard. As an example clienttoken -> client_token.


cderamus:
One other note on this PR is that I removed the groups parameter in _to_nodes and now build the security groups list using _get_security_groups. This seemed like a cleaner implementation given that security groups also need to be built for network interfaces as well.

Kami:
@cderamus Yep, I noticed that and I like this approach more.

Changes look good so I've merged them into trunk. Thanks!

PR #225, opened 2014-01-17
Add support for creating nodes with Static or No external IPs.

wrigri:
This adds a new feature and fixes a bug.  The feature is that you can specify information about the external IP address on node creation.  The bug is that _to_node() currently fails if a node does not have an external IP address.  The bug fix is line 2955/2984 (The last diff), so even if we can't get the feature in quickly, I'd like to at least get the bug fix in.


Kami:
Merged into trunk, thanks!

PR #227, opened 2014-01-20
Issue LIBCLOUD-500 : support extra arguments to CloudStackNodeDriver.list_nodes()

gza:
After you can pass useful arguments like listall of filter by domain, VPC, ...

### DIFF
--- a/libcloud/compute/drivers/cloudstack.py
+++ b/libcloud/compute/drivers/cloudstack.py
@@ -354,13 +354,13 @@ class CloudStackNodeDriver(CloudStackDriverMixIn, NodeDriver):
 
         return locations
 
-    def list_nodes(self):
+    def list_nodes(self, **extra_args):
         """
         @inherits: :class:`NodeDriver.list_nodes`
         :rtype: ``list`` of :class:`CloudStackNode`
         """
-        vms = self._sync_request('listVirtualMachines')
-        addrs = self._sync_request('listPublicIpAddresses')
+        vms = self._sync_request('listVirtualMachines', **extra_args)
+        addrs = self._sync_request('listPublicIpAddresses', **extra_args)
 
         public_ips_map = {}
         for addr in addrs.get('publicipaddress', []):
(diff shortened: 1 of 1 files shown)


libcloud/compute/drivers/cloudstack.py
```diff
@@ -354,13 +354,13 @@ def list_locations(self):
 
         return locations
 
-    def list_nodes(self):
+    def list_nodes(self, **extra_args):
```
Kami: I really hate the kwargs abuse in the CloudStack driver. I know it's not your fault, you are probably just following the existing pattern.

In any case, we should do it right going forward. This means explicitly declaring all the supported arguments. Also keep in mind that all the extension arguments need to be prefixed with `ex_`.

gza: Ok, I'll do modifications...

gza: root need was to show all nodes even from other users if I have the right to do it.

But I have a question for default behavour.
Shoud list_nodes(), by default, show :
- all visible node (I do prefer)
  - no need for extra-arg, just pass listall=True to cloudstack call
- only my nodes (current behavour)
  - We shoud add an arg ex_listall = False (and user can force it to True)

Kami: 1 - all visible nodes. Filters in the extension arguments should be used to
reduce the resulting set not to expand it.

Kami: also, /cc @runseb

sebgoa: @gza @Kami we can define some extension arguments explicitly. Especially in this case ex_list_all which can be True of False has said. The entire doc is at http://cloudstack.apache.org/docs/api/apidocs-4.2/root_admin/listVirtualMachines.html default to false will list of nodes belonging to the caller, while True will list all nodes the caller is authorized to see.

so:

def list_nodes(self, ex_list_all=False)

I can work on a patch this week or @gza feel free to submit one

gza:
I prefer ex_listall to keep same words.

I did modifications for discussion but I can't write test until two weeks.
I am dad since few days and I don't have access to a cloudstack until I come back to work.

sebgoa:
Hi @gza Sorry for the super long delay here. Can you rebase your patch and let us know if you still want to merge it.

PR #228, opened 2014-01-21
Adding/Removing instances from targetpool requires a list.

wrigri:
This fixes a bug with attaching/detaching nodes from loadbalancers.
(Attach/Detach wasn't working)

This may have changed in the API at some point after I wrote the initial load balancing implementation, but I'm not sure when.  If so, I missed the change previously.


Kami:
Merged into trunk, thanks.

PR #229, opened 2014-01-24
Support for internal endpoints in cloudfiles

jobelenus:
Sorry for the ugliness of the commit history. I didn't do this cleanly from the start. The diff is clean, however.

Being able to use the internal network endpoint gives significant speed improvements (upwards of a whole second from my brief benchmarking) when your application server and cloudfiles are in the same region/datacenter.

RE: https://github.com/apache/libcloud/pull/186#issuecomment-33209925

### DIFF
--- a/libcloud/storage/drivers/cloudfiles.py
+++ b/libcloud/storage/drivers/cloudfiles.py
@@ -106,12 +106,14 @@ class CloudFilesConnection(OpenStackBaseConnection):
     auth_url = AUTH_URL
     _auth_version = '2.0'
 
-    def __init__(self, user_id, key, secure=True, **kwargs):
+    def __init__(self, user_id, key, secure=True,
+                 use_internal_url=False, **kwargs):
         super(CloudFilesConnection, self).__init__(user_id, key, secure=secure,
                                                    **kwargs)
         self.api_version = API_VERSION
         self.accept_format = 'application/json'
         self.cdn_request = False
+        self.endpoint_url = 'internalURL' if use_internal_url else 'publicURL'
 
     def get_endpoint(self):
         region = self._ex_force_service_region.upper()
@@ -136,8 +138,8 @@ class CloudFilesConnection(OpenStackBaseConnection):
         if not ep:
             raise LibcloudError('Could not find specified endpoint')
 
-        if 'publicURL' in ep:
-            return ep['publicURL']
+        if self.endpoint_url in ep:
+            return ep[self.endpoint_url]
         else:
             raise LibcloudError('Could not find specified endpoint')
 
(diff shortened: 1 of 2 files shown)


libcloud/storage/drivers/cloudfiles.py
```diff
@@ -112,6 +112,7 @@ def __init__(self, user_id, key, secure=True, **kwargs):
         self.api_version = API_VERSION
         self.accept_format = 'application/json'
         self.cdn_request = False
+        self.endpoint_url = 'internalURL' if 'use_internal_url' in kwargs else 'publicURL'
```
alex: Passing `use_internal_url=False` shouldn't result in using it.

Kami: Correct.

Please also explicitly declare `use_internal_url` argument in the method signature instead of "abusing" kwargs.

Kami:
Adding a test case for this new functionality would also be good.

jobelenus:
A test case would be interesting, since the internal endpoint is only accessible from within the datacenter network. Is the test runner within that network?

Kami:
@jobelenus Sorry, I didn't actually meant doing the whole integration test. Sadly this would be too painful.

I was just thinking of adding a simple unit test which tests that `get_endpoint()` method returns a correct value depending on the `use_internal_url` argument.

jobelenus:
If I've correctly understood how you're doing the Driver/Connection kwargs, this testcase should meet your requirements. I may be wrong though

Kami:
@jobelenus Yeah, it's close enough :)

Kami:
Can you please fix lint (https://travis-ci.org/apache/libcloud/builds/17566981) and squash all commits?

Once this is done, I'll go ahead and merge patch into trunk.

jobelenus:
Yea... that doesn't make sense from Travis. Over-indent then under-indent when I take away a tab?

Kami:
@jobelenus Do you happen to use tabs? If so, you should really use spaces.

Looking at the file, you are missing one whitespace on line 110.

jobelenus:
I do use tabs. I don't see a missing whitespace? my line 110 matches line 147 (which is untouched) in behavior.

Kami:
It should match with a first argument, but currently it's matching with an open parenthesis.

selection_028

jobelenus:
Ah. I didn't understand the Travis lint was expecting it to match to the first char of the first arg. Thank you.

Kami:
Np. Next time `autopep8` also might or might not (it might mess up the existing style) help :)

Kami:
Also, please ping me when the commits are squashed.

jobelenus:
Squashed up. Wow that was ugly. Shame on me for not being clean in the first place.

Kami:
Patch merged into trunk. This pull request can be closed now.

Thanks!

PR #230, opened 2014-01-24
Rackspace pagination

thanatos:
A set of commits to handle Rackspace's pagination of results in their DNS API in the Rackspace driver.

Originally in pull request #190, but moved here to clean it up.

### DIFF
--- a/libcloud/dns/drivers/rackspace.py
+++ b/libcloud/dns/drivers/rackspace.py
@@ -155,7 +155,7 @@ class RackspaceDNSDriver(DNSDriver, OpenStackDriverMixin):
         zones = self._to_zones(data=response.object['domains'])
         return zones
 
-    def list_records(self, zone):
+    def iterate_records(self, zone):
         self.connection.set_context({'resource': 'zone', 'id': zone.id})
         records = []
         offset = 0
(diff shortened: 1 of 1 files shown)


libcloud/dns/drivers/rackspace.py
```diff
@@ -396,6 +408,15 @@ def _ex_connection_class_kwargs(self):
         return kwargs
 
 
+def _rackspace_result_has_more(obj):
```
Kami: To be on the safe side I would also pass limit and number of items returned in the response.

Then you could also return `False` if `num_items` < `limit`.

thanatos: @Kami: do you mean the `totalEntries` key? (which I think is the total number, so we'd be checking if `total_entries <= offset + limit` so something like that?) (I'm using http://docs.rackspace.com/cdns/api/v1.0/cdns-devguide/content/list_domains.html as a reference.)

Kami: I was thinking of number of items, so `len(records_list['records'])` for records and `len(response['domains'])` for zones.

thanatos: @Kami: Oh, I see. Added a commit to that effect.

Kami:
@thanatos Looks good. Can you please squash the commits so I can merge patch into trunk?

thanatos:
@Kami: That's an odd request. Why not just merge the branch into `trunk`? If you really want a squashed commit (you'll lose the history?) there's one at thanatos@rackspace-pagination-squished. (1a8b287)

Kami:
@thanatos That's a standard workflow for our project. It keeps history cleaner.

Kami:
Patch merged into trunk. Thanks pull request can be closed now.

Thanks.

PR #231, opened 2014-01-29
when a request is made to a CDN endpoint there is no internalURL to use,...

jobelenus:
... we have to fall back to the publicURL. This was an oversight on my part with my first pull request  --https://github.com/apache/libcloud/pull/229

And this is the fix

### DIFF
--- a/libcloud/storage/drivers/cloudfiles.py
+++ b/libcloud/storage/drivers/cloudfiles.py
@@ -105,6 +105,8 @@ class CloudFilesConnection(OpenStackBaseConnection):
 
     auth_url = AUTH_URL
     _auth_version = '2.0'
+    INTERNAL_URL = 'internalURL'
+    PUBLIC_URL = 'publicURL'
 
     def __init__(self, user_id, key, secure=True,
                  use_internal_url=False, **kwargs):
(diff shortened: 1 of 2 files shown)


libcloud/storage/drivers/cloudfiles.py
```diff
@@ -105,6 +105,8 @@ class CloudFilesConnection(OpenStackBaseConnection):
 
     auth_url = AUTH_URL
     _auth_version = '2.0'
+    INTERNAL_URL = 'internalURL'
```
Kami: Those attributes need better name.

On top of that, if they are upper case (constants) they should probably be stored on module level and not on class level.

libcloud/storage/drivers/cloudfiles.py
```diff
@@ -113,7 +115,13 @@ def __init__(self, user_id, key, secure=True,
         self.api_version = API_VERSION
         self.accept_format = 'application/json'
         self.cdn_request = False
-        self.endpoint_url = 'internalURL' if use_internal_url else 'publicURL'
+        self.use_internal_url = use_internal_url
+
+    def _get_endpoint_url(self):
```
Kami: This method name and the variable name bellow is deceives. This method does not return an endpoint URL but a key which is used to look up the endpoint URL.

As such, the method and the variable need better names.

libcloud/storage/drivers/cloudfiles.py
jobelenus: I was thinking about making them module level, but they only apply to the connection class.

Kami:
Thanks for making those changes.

There were some lint issues which I have fixed and merged patch into trunk. Next time please run `tox -e lint` and make sure there are no lint issues.

jobelenus:
Ah thanks for that local command, I was always waiting for Travis to fail me :(

PR #232, opened 2014-01-30
Add 'name' argument to key pair importing examples

briancurtin:
The [key pair management](https://libcloud.readthedocs.org/en/latest/compute/key_pair_management.html) examples show two importing methods, from file and string, but leave out the `name` argument.


Kami:
Good catch, patched merged into trunk.

Thanks!

PR #233, opened 2014-01-30
Cache cdn container url for performance improvements

jobelenus:
preventing duplicate lookups on the same instance

### DIFF
--- a/libcloud/storage/drivers/cloudfiles.py
+++ b/libcloud/storage/drivers/cloudfiles.py
@@ -262,20 +262,22 @@ class CloudFilesStorageDriver(StorageDriver, OpenStackDriverMixin):
         raise LibcloudError('Unexpected status code: %s' % (response.status))
 
     def get_container_cdn_url(self, container):
-        container_name = container.name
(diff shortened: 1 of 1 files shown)


Kami:
I'm actually not too sure about this change. There were quite a lot of talks about caching in general in the past. Usually the consensus was that we want to avoid doing it in the Libcloud and leave it up to the user.

The reason for that is that doing caching correctly and in a robust manner is hard and has all kinds of consequences (invalidation, etc.).

On top of that, if we decided to do it we should be consistent and the same change for `get_object_cdn_url` and in all other drives.

Because of the complexities involved, I'd rather avoid doing it all together like I said above.

jobelenus:
How is the user supposed to cache the CDN's URL (which really ought not to change over the in-memory life of the instance of a single object here) and continue to use libcloud as an abstraction layer?

I was watching network traffic while using libcloud and saw 3 HTTP requests for one single "get_object" call. Two were about getting the container, and one was the object. If you're getting multiple objects, as through a storage layer, this is going to be very slow.

PR #234, opened 2014-01-31
fix docs for HP Cloud

mgogoulos:


Kami:
With the latest version (0.14.0) you can specify auth URL in two forms:
1. Full URL including a path (e.g. `https://192.168.1.101:5000/v2.0/tokens/`)
2. Just a base URL without a path (e.g. `https://192.168.1.101:5000`)

It looks like the problem in this case is that the existing URL specifies just a partial path. Is this indeed the case?

If it is, I would prefer to be explicit and specify a full URL with a path (first example).

mgogoulos:
This has been the case indeed

Kami:
Looks good.

Patch merged into trunk, thanks.

PR #235, opened 2014-01-31
OpenStack public ip addresses fix

mgogoulos:


mgogoulos:
References https://issues.apache.org/jira/browse/LIBCLOUD-503

libcloud/compute/drivers/openstack.py
```diff
@@ -1910,7 +1911,15 @@ def _to_node(self, api_node):
...
             else:
-                private_ips.extend(ips)
+                for ip in ips:
+                    #is_private_subnet does not check for ipv6
+                    try:
+                        if is_private_subnet(ip):
+                            private_ips.append(ip)
+                        else:
+                            public_ips.append(ip)
+                    except:
```
Kami, line 1998: It would be good to catch a more specific exception.

libcloud/compute/drivers/openstack.py
```diff
@@ -1910,7 +1911,15 @@ def _to_node(self, api_node):
             if label in public_networks_labels:
                 public_ips.extend(ips)
             else:
-                private_ips.extend(ips)
+                for ip in ips:
```
Kami, line 1991: It needs a test case which tests the following scenarios:
- Public IPv4
- Public IPv6
- Private IPv4
- Private IPv6

Kami:
@mgogoulos Bump^^

mgogoulos:
I did some changes and pushed the tests (they've already been in test_list_nodes inside). 
On my dev environment they pass, but Travis CI fails! Not sure why this happens, self.assertEqual('12.16.18.28', node.public_ips[0]) could easily be written as self.assertTrue('12.16.18.28' in node.public_ips) since the order of the ips has changed and seems to break the tests, what do you think?

Kami:
@mgogoulos Yeah, tests shouldn't rely on some arbitrary ordering, but in this case it seems like the order should always be the same? Or am I missing something?

Kami:
@mgogoulos In any case, I'm also fine with using `in` approach.

mgogoulos:
Fixed the tests, ready to merge as I see this issue is referenced on other tickets as well (https://issues.apache.org/jira/browse/LIBCLOUD-452)

Kami:
@mgogoulos Thanks.

Can you please sync this branch with trunk and squash all the commits? Currently, I can't apply patch cleanly onto trunk.

mgogoulos:
branch synced, commits were squashed and patch has been attached to https://issues.apache.org/jira/browse/LIBCLOUD-503

Kami:
Merged into trunk.

Thanks.

PR #236, opened 2014-01-31
If you know the name of the container, you shouldn't need to make a call...

jobelenus:
... to get a container obj/dict first, and then another for the cdn url.


Kami:
This change is unnecessary.

If you know the container name, you can manually instantiate the `Container` class and use that - https://libcloud.readthedocs.org/en/latest/other/working-with-oo-apis.html

PR #237, opened 2014-02-02
LIBCLOUD-504: Add support for records with multiple values such as MX/...

cderamus:
...NS records as well an additional fixture and test.

### DIFF
--- a/libcloud/dns/drivers/route53.py
+++ b/libcloud/dns/drivers/route53.py
@@ -289,12 +289,17 @@ class Route53DNSDriver(DNSDriver):
                                                     namespace=NAMESPACE))
         ttl = findtext(element=elem, xpath='TTL', namespace=NAMESPACE)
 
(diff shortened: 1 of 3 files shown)


Kami:
I was actually thinking about that in the past.

If we want to make it consistent across all the drivers, each value should be exposed as a separate record (that's how it currently works with other drivers).

For example:
- record 1: name=None, value=ASPMX.L.GOOGLE.COM., extra['priority']=1
- record 2:  name=None, value=ALT1.ASPMX.L.GOOGLE.COM., extra['priority']=5
- etc.

cderamus:
Gotcha, that makes sense. I'll update this shortly.

Kami:
I've merged changes to trunk, thanks.

While I was at it, I also made some changes to parse / handle SRV records better.

cderamus:
Ah great call, thanks!

PR #238, opened 2014-02-02
LIBCLOUD-505: Added four new calls to support pause/unpause and suspend/...

cderamus:
...resume calls. These calls are supported in the v1.1 of OpenStack and are useful for situations such as performing system maintenance.


libcloud/compute/drivers/openstack.py
```diff
@@ -2092,6 +2092,30 @@ def ex_get_metadata_for_node(self, node):
         """
         return node.extra['metadata']
 
+    def ex_pause_node(self, node):
+        uri = '/servers/%s/action' % (node.id)
+        data = {'pause': None}
```
Kami, line 2097: (I know it's not your fault, I'm more or less just complaining)

That's a weird API.

Kami:
In any case, the patch looks good so I've merged it into trunk.

Thanks!

libcloud/compute/drivers/openstack.py
cderamus, line 2097: I couldn't agree more :).

cderamus:
Thanks!

PR #239, opened 2014-02-02
LIBCLOUD-506: Added the resource extra attribute mapping and implemented...

cderamus:
... it for networks, nodes and volumes. I also went ahead and added a number of additional properties for each of the three resource types.

### DIFF
--- a/libcloud/compute/drivers/cloudstack.py
+++ b/libcloud/compute/drivers/cloudstack.py
@@ -37,155 +37,155 @@ Define the extra dictionary for specific resources
 RESOURCE_EXTRA_ATTRIBUTES_MAP = {
     'network': {
         'broadcast_domain_type': {
-            'xpath': 'broadcastdomaintype',
+            'key_name': 'broadcastdomaintype',
             'transform_func': str
         },
         'traffic_type': {
-            'xpath': 'traffictype',
+            'key_name': 'traffictype',
             'transform_func': str
         },
         'zone_name': {
-            'xpath': 'zonename',
+            'key_name': 'zonename',
             'transform_func': str
         },
         'network_offering_name': {
-            'xpath': 'networkofferingname',
+            'key_name': 'networkofferingname',
             'transform_func': str
         },
         'network_offeringdisplay_text': {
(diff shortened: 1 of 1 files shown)


libcloud/compute/drivers/cloudstack.py
```diff
@@ -600,14 +762,21 @@ def ex_list_networks(self):
         nets = res.get('network', [])
 
         networks = []
+        extra_map = RESOURCE_EXTRA_ATTRIBUTES_MAP['network']
         for net in nets:
+            extra = self._get_extra_dict(net, extra_map)
+
+            if 'tags' in net.keys():
```
Kami: `.keys()` is unnecessary, you can just do `if 'tags' in net`.

libcloud/compute/drivers/cloudstack.py
```diff
@@ -695,11 +864,18 @@ def list_volumes(self, node=None):
                                          method='GET')
 
         list_volumes = []
+        extra_map = RESOURCE_EXTRA_ATTRIBUTES_MAP['volume']
         for vol in volumes['volume']:
+            extra = self._get_extra_dict(vol, extra_map)
+
+            if 'tags' in vol.keys():
```
Kami: Same here.

libcloud/compute/drivers/cloudstack.py
```diff
@@ -31,6 +31,167 @@
 from libcloud.utils.networking import is_private_subnet
 
 
+"""
+Define the extra dictionary for specific resources
+"""
+RESOURCE_EXTRA_ATTRIBUTES_MAP = {
+    'network': {
+        'broadcast_domain_type': {
+            'xpath': 'broadcastdomaintype',
```
Kami: CloudStack API returns JSON so calling a key `xpath` is deceiving :)

Maybe we should call it `key_name` or something along those lines?

cderamus: Ah, nice call!

Kami:
Merged into trunk, thanks.

PR #240, opened 2014-02-02
LIBCLOUD-507: Add ex_limits call to return a dictionary or resource limi...

cderamus:
...ts associated with the account. Examples of resource limits would be the maximum number of instances, volumes, snapshots and networks.


Kami:
Looks good, patch merged into trunk.

Thanks.

PR #242, opened 2014-02-02
LIBCLOUD-509: Add extra dictionary to the NodeSize object within the bas...

cderamus:
...e compute API. This will allow additional properties specific to providers to be stored within the object. The GCE driver was updated and the addition of the # of CPUs was added into the CloudStack driver.

### DIFF
--- a/libcloud/compute/base.py
+++ b/libcloud/compute/base.py
@@ -276,28 +276,28 @@ class NodeSize(UuidMixin):
                  driver, extra=None):
         """
         :param id: Size ID.
-        :type id: ``str``
+        :type  id: ``str``
 
         :param name: Size name.
-        :type name: ``str``
+        :type  name: ``str``
 
         :param ram: Amount of memory (in MB) provided by this size.
-        :type ram: ``int``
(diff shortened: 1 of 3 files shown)


libcloud/compute/base.py
```diff
@@ -294,6 +295,10 @@ def __init__(self, id, name, ram, disk, bandwidth, price, driver):
 
         :param driver: Driver this image belongs to.
         :type driver: :class:`.NodeDriver`
+
+        :param extra: Optional provider specific attributes associated with
+                      this node.
```
Kami: node == node size :)

libcloud/compute/base.py
```diff
@@ -294,6 +295,10 @@ def __init__(self, id, name, ram, disk, bandwidth, price, driver):
 
         :param driver: Driver this image belongs to.
```
Kami: I know this is an old (probably very old typo), but it would be great if you can fix it (image -> size).

Kami:
Merged into trunk.

Thanks.

PR #243, opened 2014-02-03
LIBCLOUD-510: Allow creation of "private only addressing" nodes

mneagul:
This changes allow node creation on Eucalyptus instances that assign by default private only IP addresses.

Signed-off-by: Marian Neagul marian@info.uvt.ro

### DIFF
--- a/libcloud/compute/drivers/ec2.py
+++ b/libcloud/compute/drivers/ec2.py
@@ -1315,6 +1315,10 @@ class BaseEC2NodeDriver(NodeDriver):
 
         :keyword    ex_iamprofile: Name or ARN of IAM profile
         :type       ex_iamprofile: ``str``
+
+        :keyword    ex_addressingtype: Name of the addressing scheme that should be used
+        			for launching the node. It should be either ``public`` or ``private``.
+        :type       ex_addressingtype: ``str``
         """
         image = kwargs["image"]
         size = kwargs["size"]
@@ -1383,11 +1387,11 @@ class BaseEC2NodeDriver(NodeDriver):
             else:
                 params['IamInstanceProfile.Name'] = kwargs['ex_iamprofile']
 
-        if 'ex_addressingtype' in kwargs and kwargs['ex_addressingtype'] not in ["private", "public"]:
+        if 'ex_addressingtype' in kwargs and kwargs['ex_addressingtype'] not in ['private', 'public']:
             raise AttributeError('Unsupported addressing type')
 
         if 'ex_addressingtype' in kwargs:
-            params['AddressingType'] = kwargs["ex_addressingtype"]
+            params['AddressingType'] = kwargs['ex_addressingtype']
 
         object = self.connection.request(self.path, params=params).object
         nodes = self._to_nodes(object, 'instancesSet/item')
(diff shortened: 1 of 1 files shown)


libcloud/compute/drivers/ec2.py
```diff
@@ -1383,6 +1383,12 @@ def create_node(self, **kwargs):
             else:
                 params['IamInstanceProfile.Name'] = kwargs['ex_iamprofile']
 
+        if 'ex_addressingtype' in kwargs and kwargs['ex_addressingtype'] not in ["private", "public"]:
+            raise AttributeError('Unsupported addressing type')
+
+        if 'ex_addressingtype' in kwargs:
+            params['AddressingType'] = kwargs["ex_addressingtype"]
```
Kami: Is this functionality also supported by the EC2 or just Eucalyptus?

libcloud/compute/drivers/ec2.py
```diff
@@ -1383,6 +1383,12 @@ def create_node(self, **kwargs):
             else:
                 params['IamInstanceProfile.Name'] = kwargs['ex_iamprofile']
 
+        if 'ex_addressingtype' in kwargs and kwargs['ex_addressingtype'] not in ["private", "public"]:
```
Kami: For consistency, please single quotes around strings everywhere.

Kami: Please also update the docstring and explain what this argument does.

libcloud/compute/drivers/ec2.py
mneagul: Basically Eucalyptus. There is also some deprecated support in EC2.

Kami: The reason I'm asking is that because if it's only supported by Eucalyptus it should live in the Eucalyptus driver.

mneagul: I understand and agree. Considering this the changes should be moved to
 EucNodeDriver (which is currently mostly a stub).
This would require also changing the create_node method from
BaseEC2NodeDriver so we can update the params dictionary from EucNodeDriver.

Marian

mneagul: I will move the new functionality to EucNodeDriver and if everything is ok I will close the Pull request and reopen a new, clean one.

Kami: Yes, sadly moving this functionality to the Eucalyptus driver class will require quite a lot of refactoring.

The good thing is that this is a step in a right direction and it would be much appreciated if you can do it.

And opening a new pull request is OK.

mneagul:
Do you agree with the changes in commit 8874914 (omitting coding style :) ? This approach would allow the Eucalyptus and Nimbus drivers to pass additional params to the EC2 driver, while maintaining the same interface.

The only issue that I'm afraid of, is that it could be abused at some point.

Kami:
@mneagul Yeah, I hate some of our legacy code and `**kwargs` abuse which happens there.

I would actually do the following which is more robust and less error prone:
1. Explicitly define `create_node` method with all the arguments in the EC2 and Eucalyptus NodeDriver class
2. Add "private" `_create_node` method to the base class. This method is called by `create_node` in the EC2 and Eucalyptus node driver class. This method does all the work and can take `params` (or similar argument) which differs between EC2 and Eucalyptus driver.

Kami:
@mneagul Bump. Let me know if you need any help.

mneagul:
Hello,

Sorry for disappearing. I was cought with some deadline. I will close the pull request and reopen a new one this days.

Marijan

PR #244, opened 2014-02-04
LIBCLOUD-511: Added support for obtaining console output for OpenStack n...

cderamus:
...odes. New tests are also included.

### DIFF
--- a/libcloud/compute/drivers/openstack.py
+++ b/libcloud/compute/drivers/openstack.py
@@ -1644,7 +1644,6 @@ class OpenStack_1_1_NodeDriver(OpenStackNodeDriver):
         """
         resp = self.connection.request('/os-snapshots/%s' % snapshot.id,
                                        method='DELETE')
-
         return resp.status == httplib.NO_CONTENT
 
     def _to_security_group_rules(self, obj):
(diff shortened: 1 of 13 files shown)


libcloud/compute/drivers/openstack.py
```diff
@@ -1572,6 +1572,31 @@ def ex_delete_network(self, network):
                                        method='DELETE')
         return resp.status == httplib.ACCEPTED
 
+    def ex_get_console_output(self, node, length=None):
```
Kami, line 1583: Just curious, do you happen to know since which release this functionality is available?

I'm having troubles finding any references in documentation. I just found some links to the code which uses this functionality.

Kami, line 1583: Also, since there are no docs I don't know what happens if you pass `None` for `length`. Does it use some default value?

Kami, line 1583: I just tested it with Rackspace and that's what I get back:

```
# -------- begin 25657288:25633136 response ----------
HTTP/1.1 400 Bad Request
Content-Length: 88
Via: 1.1 Repose (Repose/2.12)
X-Compute-Request-Id: 1
Server: Jetty(8.0.y.z-SNAPSHOT)
Date: Tue, 04 Feb 2014 14:18:07 GMT
Content-Type: application/json; charset=UTF-8

{"badRequest": {"message": "There is no such action: os-getConsoleOutput", "code": 400}}
```

Did you test this with a live installation / provide? If so, which versions / providers did support this functionality?

cderamus, line 1583: It's offered in OpenStack v2: http://api.openstack.org/api-ref-compute-ext.html. See section Server console output (os-console-output). It is not supported within Rackspace from what I can tell. Passing None as length is fine and causes no issue from my testing.

Kami, line 1583: I see. OpenStack needs to do some SEO stuff, I had troubles finding this information on Google.

In any case, because Rackspace doesn't support it, we need to take this into account. The easiest (but definitely not the best and nicest) thing to do is to implement this method in the Rackspace driver and throw `NotImplement` error.

Kami, line 1583: Actually, just noticed you added this to the `OpenStack_1_0_NodeDriver` class.

A better approach would be to create a new `OpenStack_2_0_NodeDriver` class which inherits from ``OpenStack_1_1_NodeDriver` and add this method there.

Sorry for extra work, I missed some context the first time I reviewed this pull request.

cderamus, line 1583: I'm okay with the extra work, but do I need to go the whole distance and create new OpenStack_2_0_Response and OpenStack_2_0_Connection objects as well? Won't there also be a lot of changes in the test_openstack.py compute file as well?

Kami, line 1583: Honestly, for now, I'm OK with just using 1.1 classes.

If you want though, you can create new classes and just have them inherit from the 1.1 ones.

PR #245, opened 2014-02-04
LIBCLOUD-512: Added snapshot lifecycle support and fixture tests to supp...

cderamus:
...ort snapshot listing, creation and deletion.

### DIFF
--- a/libcloud/compute/drivers/openstack.py
+++ b/libcloud/compute/drivers/openstack.py
@@ -1644,7 +1644,6 @@ class OpenStack_1_1_NodeDriver(OpenStackNodeDriver):
         """
         resp = self.connection.request('/os-snapshots/%s' % snapshot.id,
                                        method='DELETE')
-
         return resp.status == httplib.NO_CONTENT
 
     def _to_security_group_rules(self, obj):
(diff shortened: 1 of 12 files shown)


libcloud/compute/drivers/openstack.py
```diff
@@ -1572,6 +1577,49 @@ def ex_delete_network(self, network):
...
+    def ex_delete_snapshot(self, snapshot):
+        """
+        Delete a VolumeSnapshot
+
+        :param      node: snapshot
+        :type       node: :class:`VolumeSnapshot`
+
+        :rtype:     ``bool``
+        """
+        return self.connection.request('/os-snapshots/%s' % snapshot.id,
```
Kami: You probably want to explicitly check the status code (e.g. `response.status == httplib.DELETED`) because IIRC, `success()` returns `True` for all the supported status codes/

libcloud/compute/drivers/openstack.py
```diff
@@ -1958,6 +2006,37 @@ def _to_volume(self, api_node):
             }
         )
 
+    def _to_snapshot(self, api_node):
+        if 'snapshot' in api_node:
+            api_node = api_node['snapshot']
+
+        # RackSpace vs. OpenStack
+        if 'displayName' in api_node:
+            return VolumeSnapshot(
```
Kami: It seems all the arguments except `extra` dictionary are the same.

To reduce duplication, I would just build `extra` dictionary depending on the provider.

E.g.

``` python
if rackspace:
    extra = ...
else:
  extra = ...

snapshot = VolumeSnapshot(..., extra=extra)
```

libcloud/test/compute/test_openstack.py
```diff
@@ -1862,6 +1879,24 @@ def _v1_1_slug_servers_12063_action(self, method, url, body, headers):
...
+            body = self.fixtures.load('_os_snapshots.json')
+        elif method == "POST":
+            body = self.fixtures.load('_os_snapshots_create.json')
+        else:
+            raise NotImplementedError()
+
+        return (httplib.OK, body, self.json_content_headers, httplib.responses[httplib.OK])
+
+    def _v1_1_slug_os_snapshots_3fbbcccf_d058_4502_8844_6feeffdf4cb5(self, method, url, body, headers):
+        if method == "DELETE":
```
Kami: I noticed this in other pull request as well.

It's a minor thing, but for consistency reasons, please use single quotes around strings everywhere (except docstrings).

Kami:
I went ahead and merged this pull request with my changes into trunk.

Thanks!

PR #246, opened 2014-02-06
Update CloudStack NodeStates for Stopped/Stopping

cderamus:


Kami:
We've discussed this on the IRC with @cderamus yesterday.

I think we should also map `stopping` and `expunging` to `pending` instead of `terminated`. The reason for that is that not all the operations are available until the node is fully stopped (`stopped` state).

This means that currently your code might fail if you wait until node.state == stopped because node might actually still be in stopping state.

Kami:
I've merged the changes mentioned above and merged this into trunk.

Thanks.

PR #248, opened 2014-02-08
LIBCLOUD-514: Add create/delete tag support into the CloudStack driver. ...

cderamus:
...New tests and fixtures are also included.

### DIFF
--- a/libcloud/compute/drivers/cloudstack.py
+++ b/libcloud/compute/drivers/cloudstack.py
@@ -1692,6 +1692,64 @@ class CloudStackNodeDriver(CloudStackDriverMixIn, NodeDriver):
 
         return limits
 
+    def ex_create_tags(self, resource_ids, resource_type, tags):
+        """
+        Create tags for a resource (Node/StorageVolume/etc).
+        A list of resource types can be found at http://goo.gl/6OKphH
+
+        :param resource: Resource IDs to be tagged
+        :type  resource: ``list`` of resource IDs
+
+        :param resource_type: Resource type (eg: UserVm)
+        :type  resource_type: ``str``
+
+        :param tags: A dictionary or other mapping of strings to strings,
+                     associating tag names with tag values.
+        :type  tags: ``dict``
+
+        :rtype: ``bool``
+        """
+        params = {'resourcetype': resource_type,
+                  'resourceids': ",".join(resource_ids)}
+
+        for i, key in enumerate(tags):
+            params['tags[%d].key' % i] = key
+            params['tags[%d].value' % i] = tags[key]
+
+        self._async_request(command='createTags',
+                            params=params,
+                            method='GET')
+        return True
+
+    def ex_delete_tags(self, resource_ids, resource_type, tags):
+        """
+        Delete tags from a resource.
+
+        :param resource: Resource IDs to be tagged
+        :type  resource: ``list`` of resource IDs
+
+        :param resource_type: Resource type (eg: UserVm)
+        :type  resource_type: ``str``
+
+        :param tags: A dictionary or other mapping of strings to strings,
+                     associating tag names with tag values.
+        :type tags: ``dict``
+
+        :rtype: ``bool``
+        """
+        params = {'resourceType': resource_type,
+                  'resourceIds': ",".join(resource_ids)}
+
+        for i, key in enumerate(tags):
+            params['tags[%s].key' % i] = key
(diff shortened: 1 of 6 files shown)


libcloud/compute/drivers/cloudstack.py
```diff
@@ -1692,6 +1692,64 @@ def ex_limits(self):
 
         return limits
 
+    def ex_create_tags(self, resource_ids, resource_type, tags):
```
Kami, line 1695: Since you need to specify `resource_type` I assume this means that all the resource IDs need to refer to the resource of the same type.

If this is indeed the case, it would be good to document it.

libcloud/compute/drivers/cloudstack.py
```diff
@@ -1692,6 +1692,64 @@ def ex_limits(self):
...
+        :type  resource_type: ``str``
+
+        :param tags: A dictionary or other mapping of strings to strings,
+                     associating tag names with tag values.
+        :type  tags: ``dict``
+
+        :rtype: ``bool``
+        """
+        params = {'resourcetype': resource_type,
+                  'resourceids': ",".join(resource_ids)}
```
Kami: Single quotes for consistency please.

libcloud/compute/drivers/cloudstack.py
```diff
@@ -1692,6 +1692,64 @@ def ex_limits(self):
...
+        :type  resource_type: ``str``
+
+        :param tags: A dictionary or other mapping of strings to strings,
+                     associating tag names with tag values.
+        :type tags: ``dict``
+
+        :rtype: ``bool``
+        """
+        params = {'resourceType': resource_type,
+                  'resourceIds': ",".join(resource_ids)}
```
Kami: Same here.

libcloud/compute/drivers/cloudstack.py
```diff
@@ -1692,6 +1692,64 @@ def ex_limits(self):
...
+        """
+        Delete tags from a resource.
+
+        :param resource: Resource IDs to be tagged
+        :type  resource: ``list`` of resource IDs
+
+        :param resource_type: Resource type (eg: UserVm)
+        :type  resource_type: ``str``
+
+        :param tags: A dictionary or other mapping of strings to strings,
```
Kami: I've checked the docs and it says that for the delete operation, only tag keys are mandatory.

Because of that, it would imo be better and feel more natural if user only needs to specify a list of tag keys to be deleted.

libcloud/test/compute/test_cloudstack.py
```diff
@@ -496,6 +496,18 @@ def test_ex_limits(self):
         self.assertEqual(limits['max_volumes'], 20)
         self.assertEqual(limits['max_snapshots'], 20)
 
+    def test_ex_create_tags(self):
+        node = self.driver.list_nodes()[0]
+        tags = {'Region': 'Candada'}
```
Kami: O Canada!

libcloud/compute/drivers/cloudstack.py
cderamus, line 1695: You got it.

libcloud/compute/drivers/cloudstack.py
cderamus: You are battling years of PHP/Python/JS muscle memory where I instinctively use double quotes. I'll try to be better about this :).

libcloud/compute/drivers/cloudstack.py
cderamus: Ah okay that works.

libcloud/test/compute/test_cloudstack.py
cderamus: :)

libcloud/test/compute/test_cloudstack.py
```diff
@@ -496,6 +496,18 @@ def test_ex_limits(self):
...
+        resp = self.driver.ex_create_tags([node.id], 'UserVm', tags)
+        self.assertTrue(resp)
+
+    def test_ex_delete_tags(self):
+        node = self.driver.list_nodes()[0]
+        tags = {'Region': 'Candada'}
+        resp = self.driver.ex_create_tags([node.id], 'UserVm', tags)
```
Kami: This should probably be `ex_delete_tags` and pass `tag_keys` :)

Kami:
Looks good. Changes merged into trunk.

Thanks.

PR #249, opened 2014-02-09
LIBCLOUD-516: Add list_size() support into Eucalyptus for cluster versio...

cderamus:
...ns >= 3.3.0. In this version Eucalyptus introduced the DescribeInstanceTypes call which is specific to EUCA and is not AWS compatible. This call requires a different XML namespace.


Kami:
Honestly, I don't think this approach is that bad.

libcloud/compute/drivers/ec2.py
```diff
@@ -3866,11 +3866,56 @@ def __init__(self, key, secret=None, secure=True, host=None,
         if path is None:
             path = '/services/Eucalyptus'
         self.path = path
+        self.EUCA_API_VERSION = '3.4.1'
```
Kami: To be consistent with AWS stuff, I would promote those variables to module level constants.

Kami: Now, this driver could also support standard `api_version` argument. This argument would default to `DEFAULT_EUCA_API_VERSION` (or smth).

cderamus: I'm not super familiar with Eucalyptus and am only using the community cloud at the moment. Is the API version something that can dynamically be pulled?

Kami: I have no idea. I know it can't be dynamically specified for AWS.

libcloud/compute/drivers/ec2.py
```diff
@@ -3855,22 +3861,72 @@ class EucNodeDriver(BaseEC2NodeDriver):
     connectionCls = EucConnection
 
     def __init__(self, key, secret=None, secure=True, host=None,
-                 path=None, port=None):
+                 path=None, port=None, api_version=None):
```
Kami, line 3864: Shouldn't `api_version` default to `DEFAULT_EUCA_API_VERSION`?

cderamus, line 3864: I suppose so :)

Kami:
I've made the api_version change and merged patch into trunk.

Thanks.

PR #250, opened 2014-02-10
Unpack the iterator before getting its length.

coderanger:
This matters for things like generators or network streams where it doesn't have a size normally.

### DIFF
--- a/libcloud/storage/drivers/dummy.py
+++ b/libcloud/storage/drivers/dummy.py
@@ -438,6 +438,8 @@ class DummyStorageDriver(StorageDriver):
         @inherits: :class:`StorageDriver.upload_object_via_stream`
         """
 
+        # Unpack the iterator
+        iterator = list(iterator)
         size = len(iterator)
         return self._add_object(container=container, object_name=object_name,
                                 size=size, extra=extra)


coderanger:
Having trouble making DummyFileObject actually behave like a file object. The dummy provider in general seems a bit unfinished, am I better off not using it?

Kami:
I need to have a more closer look, but IIRC, this class is only used in the dummy driver. There is also a `DummyIterator` class which is used in some tests.

As far as the completeness goes, it should be more or less complete, but it's a Dummy driver so there aren't really a tons of use cases for it. What kind of problems are you having exactly?

I believe that in the beginning, dummy drives were added for demonstration pruposed. Another use case I imagined a while back is building an online REPL where users can play with dummy drivers.

coderanger:
I'm trying to write unit tests for something that uses libcloud and I was hoping the dummy driver would make that relatively easy as an in-memory backend that otherwise gives mostly the same API.

Jc2k:
I've ended up subclassing the dummy drivers in one of my projects so that I can use them for tests too. I'm definitely a fan of having something in libcloud for this purpose. 

Here are a few more methods (they were to make my tests work, so might not be 100%):

```
    def iterate_container_objects(self, container):
        for obj in self._containers[container.name]['objects'].values():
            yield obj

    def list_container_objects(self, container):
        return list(self.iterate_container_objects(container))

    def upload_object_via_stream(self, iterator, container, object_name, extra=None):
        blocks = [block for block in iterator]
        size = sum(len(block) for block in blocks)
        o = self._add_object(
            container=container, object_name=object_name, size=size, extra=extra)
        o.blocks = blocks
        return o

    def download_object_as_stream(self, obj, chunk_size=None):
        return obj.blocks
```

coderanger:
It sounds like maybe I'm fixing this at the wrong level and DummyDriver should just be overhauled a bit to fit the bill of "simple in-memory backend", does that seem reasonable?

coderanger:
It is already most of the way there, just need to remove a bunch of usage of DummyFileObject in favor of returning real data.

Kami:
@coderanger Yeah, this sound good. And it would also be good to implement methods posted above by @Jc2k

PR #251, opened 2014-02-12
NodeState.STOPPED fix for VCloud driver

ashb:
This fixes up the vCloud drivers to correctly update the Node state to be
STOPPED rather than TERMINATED - a node can be restarted from this state


Kami:
@ashb Thanks.

For the completeness sake, it's worth noting that the reason this state was previously mapped to "TERMINATED" is because "STOPPED" state was added just recently.

Kami:
Merged into trunk.

Thanks.

PR #254, opened 2014-03-01
Fix a typo for an error message in the GCE driver.

fcuny:


Kami:
Merged into trunk.

Thanks.

PR #255, opened 2014-03-02
LIBCLOUD-525: Add VPC lifecycle support into the EC2 driver. This PR inc...

cderamus:
...ludes the ability to describe, create, delete, attach and detach VPC Internet gateways.

### DIFF
--- a/libcloud/compute/drivers/ec2.py
+++ b/libcloud/compute/drivers/ec2.py
@@ -1171,6 +1171,24 @@ class ElasticIP(object):
                 % (self.ip, self.domain, self.instance_id))
 
 
+class VPCInternetGateway(Node):
+    """
+    Class which stores information about VPC Internet Gateways.
+
+    Note: This class is VPC specific.
+    """
+
+    def __init__(self, id, name, vpc_id, state, driver, extra=None):
+        self.id = id
+        self.name = name
+        self.vpc_id = vpc_id
+        self.state = state
+        self.extra = extra or {}
+
+    def __repr__(self):
+        return (('<VPCInternetGateway: id=%s>') % (self.id))
+
+
 class BaseEC2NodeDriver(NodeDriver):
     """
     Base Amazon EC2 node driver.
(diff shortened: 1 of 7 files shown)


libcloud/compute/drivers/ec2.py
```diff
@@ -1171,6 +1171,24 @@ def __repr__(self):
                 % (self.ip, self.domain, self.instance_id))
 
 
+class VPCInternetGateway(Node):
```
Kami: Hm, this probably shouldn't inherit from `Node`, right?

libcloud/compute/drivers/ec2.py
```diff
@@ -3171,6 +3189,105 @@ def ex_find_or_import_keypair_by_key_material(self, pubkey):
...
+        result = self.connection.request(self.path, params=params).object
+        element = findtext(element=result,
+                           xpath='return',
+                           namespace=NAMESPACE)
+
+        return element == 'true'
+
+    def ex_detach_internet_gateway(self, gateway, network):
+        """
+        Detach a Internet gateway to a VPC
```
Kami: to a VPC = from a VPC :)

libcloud/compute/drivers/ec2.py
```diff
@@ -3557,6 +3674,40 @@ def _to_device_mapping(self, element):
...
+                          xpath='attachmentSet/item/vpcId',
+                          namespace=NAMESPACE)
+
+        state = findtext(element=element,
+                         xpath='attachmentSet/item/state',
+                         namespace=NAMESPACE)
+
+        # If there's no attachment state, let's
+        # set it to available
+        if state is None:
```
Kami: IIRC, `findtext` will return empty string (`''`) by default if the matching item is not found. This means we will never get into this branch.

You can pass `no_text_value=''` to `findtext` function, or even better, just do `if not state`.

libcloud/compute/drivers/ec2.py
cderamus: Correct.

Kami:
Merged into trunk.

Thanks.

PR #256, opened 2014-03-02
Add hs1.4xlarge/hs1.8xlarge into pricing.json/ec2 driver in regions where it's supported.

cderamus:
Adding in the hs1.4xlarge/hs1.8xlarge instance types into the driver and pricing data files.


Kami:
Looks good, but I can't cleanly apply patch onto trunk.

Can you please squash all the commits?

PR #258, opened 2014-03-10
Add method `ex_copy_image` to the GCE driver.

fcuny:
The `ex_copy_image` method allow to copy an image from a Google Cloud
Storage URL in the specified project.

I'm going to need this to implement this part in Ansible soon.

Thanks.

### DIFF
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -2346,6 +2346,42 @@ class GCENodeDriver(NodeDriver):
             return None
         return self._to_zone(response)
 
+    def ex_copy_image(self, name, url, description=None):
+        """
+        Copy an image to your image collection.
+
+        :param  name: The name of the image
+        :type   name: ``str``
+
+        :param  url: The URL to the image. The URL can start with `gs://`
+        :param  url: ``str``
+
+        :param  description: The description of the image
+        :type   description: ``str``
+
+        :return:  NodeImage object based on provided information or None if an
+                  image with that name is not found.
+        :rtype:   :class:`NodeImage` or ``None``
+        """
+
+        # the URL for an image can start with gs://
+        if url.startswith('gs://'):
+            url = url.replace('gs://', 'https://storage.googleapis.com/', 1)
+
+        image_data = {
+            'name': name,
+            'description': description,
+            'sourceType': 'RAW',
+            'rawDisk': {
+                'source': url,
+            },
+        }
+
+        request = '/global/images'
+        self.connection.async_request(request, method='POST',
+                                      data=image_data)
+        return self.ex_get_image(name)
+
     def _ex_connection_class_kwargs(self):
         return {'auth_type': self.auth_type,
                 'project': self.project}
--- a/libcloud/test/compute/fixtures/gce/global_images.json
+++ b/libcloud/test/compute/fixtures/gce/global_images.json
@@ -29,6 +29,21 @@
       "selfLink": "https://www.googleapis.com/compute/v1/projects/debian-cloud/global/images/centos-6-v20131118",
       "sourceType": "RAW",
       "status": "READY"
+    },
+    {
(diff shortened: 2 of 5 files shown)


libcloud/compute/drivers/gce.py
```diff
@@ -2346,6 +2346,42 @@ def ex_get_zone(self, name):
...
+        :type   description: ``str``
+
+        :return:  NodeImage object based on provided information or None if an
+                  image with that name is not found.
+        :rtype:   :class:`NodeImage` or ``None``
+        """
+
+        # the URL for an image can start with gs://
+        if url.startswith('gs://'):
+            url = ('http://storage.googleapis.com/' + url[len('gs://'):])
```
wrigri: The URL should be 'https:' not 'http:'

Also, the [len('gs://'):] construct seems overly complicated.  Maybe something like this instead?
url.replace('gs://', 'https://storage.googleapis.com/', 1)

(I can't imagine what a url would look like that had more than one 'gs://' in it, but added the '1' to the parameters just in case.)

libcloud/test/compute/fixtures/gce/global_images.json
```diff
@@ -29,7 +29,22 @@
...
+  "description": "CoreOS test image",
+  "id": "15196339658718959621",
+  "kind": "compute#image",
+  "name": "coreos",
+  "preferredKernel": "https://www.googleapis.com/compute/v1/projects/google/global/kernels/gce-v20130603",
+  "rawDisk": {
+    "containerType": "TAR",
+    "source": ""
+  },
+  "selfLink": "https://www.googleapis.com/compute/v1/projects/project_name/global/operations/operation-global_image_post",
```
wrigri: This link should probably be a link to the image, not to the operation.

libcloud/test/compute/test_gce.py
```diff
@@ -501,6 +501,14 @@ def test_ex_get_image(self):
         self.assertEqual(image.name, 'debian-6-squeeze-v20130926')
         self.assertTrue(image.extra['description'].startswith('Debian'))
 
+    def test_ex_copy_image(self):
+        name = 'coreos'
```
wrigri, line 505: Your names don't seem to match here.  The targetLink in the test fixtures references 'core-os' and this is 'coreos'.  The tests don't currently care about that level of detail, but it would be nice to have it consistent.

wrigri:
Looks good overall.  A couple of minor things, but otherwise it seems like a good addition.

libcloud/compute/drivers/gce.py
```diff
@@ -2346,6 +2346,42 @@ def ex_get_zone(self, name):
...
         return self._to_zone(response)
 
+    def ex_copy_image(self, name, url, description=None):
+        """
+        Copy an image to your image collection.
+
+        :param  name: The name of the image
+        :type   name: ``str``
+
+        :param  url: The URL to the image. The URL can starts with `gs://`
```
wrigri: s/starts/start/

fcuny:
Thanks @wrigri, I'll update this patch later today.

libcloud/compute/drivers/gce.py
```diff
@@ -2346,6 +2346,42 @@ def ex_get_zone(self, name):
...
+        :type   description: ``str``
+
+        :return:  NodeImage object based on provided information or None if an
+                  image with that name is not found.
+        :rtype:   :class:`NodeImage` or ``None``
+        """
+
+        # the URL for an image can start with gs://
+        if url.startswith('gs://'):
+            url.replace('gs://', 'https://storage.googleapis.com/', 1)
```
Kami: Python strings are immutable and `.replace` returns a new string. This means you should do:

``` python
url = url.replace('gs://', 'https://storage.googleapis.com/', 1)
```

Kami: Besides that, LGTM.

PR #259, opened 2014-03-11
Fix a path in the documentation for testing.

fcuny:


Kami:
@franckcuny Good catch, merged into trunk.

Thanks.

PR #260, opened 2014-03-12
Support deletion and deprecation of GCE's images.

fcuny:
Add a new class GCENodeImage to represent GCE's images, and support the
`delete` and `deprecated` functions.

Added a new extra field to the image object to specify the deprecation
status.

### DIFF
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -229,6 +229,37 @@ class GCEForwardingRule(UuidMixin):
             self.id, self.name, self.address)
 
 
+class GCENodeImage(NodeImage):
+    """A GCE Node Image class."""
+    def __init__(self, id, name, driver, extra=None):
+        super(GCENodeImage, self).__init__(id, name, driver, extra=extra)
+
+    def delete(self):
+        """
+        Delete this image
+
+        :return: True if successful
+        :rtype:  ``bool``
+        """
+        return self.driver.ex_delete_image(image=self)
+
+    def deprecate(self, replacement, state):
+        """
+        Deprecate this image
+
+        :param  replacement: Image to use as a replacement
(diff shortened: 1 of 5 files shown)


libcloud/compute/drivers/gce.py
```diff
@@ -229,6 +229,38 @@ def __repr__(self):
             self.id, self.name, self.address)
 
 
+class GCENodeImage(NodeImage):
+    """A GCE Node Image class."""
+    def __init__(self, id, name, driver, extra=None):
+        self.extra = extra
```
Kami: You already pass extra to the parent constructor which means you shouldn't need to set the attribute in the line above.

libcloud/compute/drivers/gce.py
```diff
@@ -1854,6 +1885,65 @@ def ex_destroy_address(self, address):
...
+
+        if not hasattr(replacement, 'name'):
+            replacement = self.ex_get_image(replacement)
+
+        if state is None:
+            state = 'DEPRECATED'
+
+        possible_states = ['DELETED', 'DEPRECATED', 'OBSOLETE']
+
+        if state not in possible_states:
```
Kami, line 1931: Should throw `ValueError` instead of returning `False`

libcloud/compute/drivers/gce.py
```diff
@@ -1854,6 +1885,65 @@ def ex_destroy_address(self, address):
...
+            return False
+
+        image_data = {
+            'state':       state,
```
Kami: IIRC, this will cause lint to fail (extra whitespace)

libcloud/compute/drivers/gce.py
fcuny, line 1931: Changed.

fcuny:
I'll push an update for the build failure.

Kami:
Merged into trunk.

Thanks.

PR #264, opened 2014-03-14
Add `ex_set_volume_auto_delete` to the GCE driver.

fcuny:
Sets the auto-delete flag for a volume attached to a node.

### DIFF
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -1838,6 +1838,33 @@ class GCENodeDriver(NodeDriver):
                                       data='ignored')
         return True
 
+    def ex_set_volume_auto_delete(self, volume, node, auto_delete=True):
+        """
+        Sets the auto-delete flag for a volume attached to a node.
+
+        :param  volume: Volume object to auto-delete
+        :type   volume: :class:`StorageVolume`
+
+        :param   ex_node: Node object to auto-delete volume from
+        :type    ex_node: :class:`Node`
+
+        :keyword auto_delete: Flag to set for the auto-delete value
+        :type    auto_delete: ``bool`` (default True)
+
+        :return:  True if successfull
+        :rtype:   ``bool``
+        """
+        request = '/zones/%s/instances/%s/setDiskAutoDelete' % (
+            node.extra['zone'].name, node.name
+        )
+        delete_params = {
+            'deviceName': volume,
+            'autoDelete': auto_delete,
+        }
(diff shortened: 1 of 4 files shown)


libcloud/compute/drivers/gce.py
```diff
@@ -1838,6 +1838,37 @@ def detach_volume(self, volume, ex_node=None):
...
+        :keyword ex_node: Node object to auto-delete volume from (required)
+        :type    ex_node: :class:`Node`
+
+        :keyword auto_delete: Flag to set for the auto-delete value
+        :type    auto_delete: ``bool`` (default True)
+
+        :return:  True if successfull
+        :rtype:   ``bool``
+        """
+        if not ex_node:
```
Kami: If the `ex_node` argument is required, you should simply not set the default value in the method signature and Python will throw if you don't pass this argument in.

Kami: Also, extension methods already have `ex_` prefix in the method name so you can remove it from the argument (`ex_node` -> `node`).

You only need to add `ex_` prefix to the argument if you add it to the standard method (e.g. `list_nodes`).

Kami:
Besides the naming thing, LGTM.

/cc @wrigri

fcuny:
Ha, interesting. For both suggestions, that's what I would normally have done. But the previous function declaration (`def detach_volume(self, volume, ex_node=None):`) was setting a `ex_node` to `None` by default (even if it's required). I was trying to copy the style :)

Would a second pull request to modify the `detach_function` signature be accepted ?

Thanks for the quick feedback, will update.

Kami:
@franckcuny Yeah, a second pull request would be much appreciated.

An sadly we currently only enforce things like this manually during the PR review so it's kinda easy to miss it. The good news is that we are working on an automatic system for checking that the driver and methods comply to the base API :)

wrigri:
Looks good to me

I'm a bit confused about the conversation regarding a second pull request for detach_volume.  The signature for detach volume is like it is because the base API doesn't allow for the node to be set as a positional argument.  So, even though volume.detach() won't actually work for GCE, adding a positional argument for node to detach_volume would cause volume.detach() to error out.

Kami:
Patch merged into trunk, the pull request can be closed now. Thanks.

And sorry for the delay. I thought I already merged this a while back.

Kami:
@franckcuny Can you please close this pull request (changes have already been merged a while back)?

Thanks

PR #265, opened 2014-03-16
Fixed the documented return type. This function doesn't return a value.

alex:


Kami:
Looks good. Merged into trunk.

Thanks.

(P.S. You have commit rights now so feel free to push such small changes / fixed with PR directly to trunk)

PR #266, opened 2014-03-20
allow creds to be passed in dict

bassdread:
Allow the GCE driver to be created by passing existing tokens.


wrigri:
Seems like a good change to me.  It would be great to have some documentation about how/when to use it, perhaps by adding a 4th example to this page: https://github.com/apache/libcloud/blob/trunk/docs/compute/drivers/gce.rst

bassdread:
@Kami @wrigri hi, I have cleaned up the code and added docs and example code. Could you have another look?

erjohnso:
@wrigri - can you take a look?

wrigri:
Sorry, I've been away from libcloud for a while.  This looks good to me, but I'd imagine that it will need a few updates to allow it to be patched in cleanly.

PR #267, opened 2014-03-20
Gce auto create disk

wrigri:
This adds support for the automatic creation and deletion of boot disks when creating/deleting instances.

It also deals with some inefficient code and should address most of the problems raised in LIBCLOUD-527

/cc @RussellSpitzer


RussellSpitzer:
I know i'm not a regular here but I think this is a good approach since the assumption for the _to\* methods is that they are simply converting responses to objects.

Kami:
@wrigri Can you please sync you branch with trunk and squash the commits since I have problems applying the patch cleanly.

Kami:
@wrigri bump^^

erjohnso:
@wrigri - I think this can be withdrawn?

wrigri:
Yes, I think it can be.  At the very least, if there are things in this PR that still need to be implemented, they should be rewritten based on the current state.

PR #271, opened 2014-03-30
Pep8 fixes

fcuny:
PEP8 1.5 introduced a bunch of new tests, and because of them the linter fails. Each of this patch can be applied individually to fix the appropriate errors. If all the patches are accepted, I'll be happy to squash them if it makes more sense.


Kami:
Merged, thanks.

PR #274, opened 2014-04-08
restore the connection request_path in case of failed request. 

mgogoulos:
This could happen if ex_project is specified and request gets ResourceNotFoundError, then request_path will be wrong for all other requests (eg on a script)

### DIFF
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -708,10 +708,10 @@ class GCENodeDriver(NodeDriver):
             try:
                 response = self.connection.request(request, method='GET').object
             except:
-                self.connection.request_path = save_request_path
-                raise
+                response = {}
             # Restore the connection request_path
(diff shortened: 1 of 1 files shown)


libcloud/compute/drivers/gce.py
```diff
@@ -705,7 +705,11 @@ def list_images(self, ex_project=None):
             new_request_path = save_request_path.replace(self.project,
                                                          ex_project)
             self.connection.request_path = new_request_path
-            response = self.connection.request(request, method='GET').object
+            try:
```
Kami, line 708: You could also use try / finally and and restore the request_path in finally block. This way you only need to do it once and don't need to re-raise exception.

mgogoulos, line 708: now list_images returns [] rather than ResourceNotFoundError . Do you think this is better?

PR #275, opened 2014-04-08
[LIBCLOUD-518] Automatically detect domain of Elastic IPs.

tomprince:


Kami:
(sorry for the delay)

Hm, I thought I commented on this PR, but now I can't find my comment. Anyways.
erjohnso
Can you please elaborate on this change? If domain is not required anymore, can we just get rid of this argument all together (backward incompatible)?

Kami:
@tomprince bump^^

tomprince:
> If domain is not required anymore, can we just get rid of this argument all together (backward incompatible)?

Sure, it can be removed. Most projects I work with give at least one release with a `DeprecationWarnging` before making incompatible changes. This way, they can upgrade without making any changes to the code, test, and then make changes to get rid of the warnings.

PR #276, opened 2014-04-10
Add a deprecated flag to a size for GCE nodes.

bassdread:
Currently sizes are flagged as deprecated in the Google GCE API

```
'deprecated': {'state': 'OBSOLETE', 'replacement': 'https://www.googleapis.com/compute/v1/projects/clever-span-499/zones/us-central1-a/machineTypes/n1-highmem-4'}
```

But there is no field to indicate this on the returned size.

The code below attaches a 'deprecated' value to 'extra' so it's possible to identify deprecated sizes.


Kami:
Sorry for the delay.

I've fixed the code (it was referencing undefined `machine_type` variable) and merged changes into trunk. Thanks.

PR #277, opened 2014-04-17
[LIBCLOUD-537] Added support for generic image management at Rackspace and EC2

ghost:
I was planning to take care of GCE in this first push as well. However, Google presents an interesting challenge with their persistent disk model and their use of the snapshot and image terms. We could use snapshots as a replacement for images. However, this would then limit our ability to use the default images provided by Google. It seems like we actually need a factory for evaluating both the images and snapshots in order to fold in GCE. I completed the _supported_methods_image_management.rst as well in case we want to include it in the documentation. I did submit an email regarding this effort to the dev mail list but I didn't hear anything back so I just went ahead and made the change.

### DIFF
--- a/libcloud/compute/base.py
+++ b/libcloud/compute/base.py
@@ -1069,7 +1069,7 @@ class NodeDriver(BaseDriver):
         raise NotImplementedError(
             'list_images not implemented for this driver')
 
-    def create_image(self, node, name, description=None, reboot=False):
+    def create_image(self, node, name, description=None):
         """
         Creates an image from a node object.
 
--- a/libcloud/compute/base.py
+++ b/libcloud/compute/base.py
@@ -645,19 +645,6 @@ class NodeDriver(BaseDriver):
         raise NotImplementedError(
             'list_nodes not implemented for this driver')
 
-    def list_images(self, location=None):
-        """
-        List images on a provider
-
-        :param location: The location at which to list images
-        :type location: :class:`.NodeLocation`
-
-        :return: list of node image objects
-        :rtype: ``list`` of :class:`.NodeImage`
-        """
-        raise NotImplementedError(
-            'list_images not implemented for this driver')
-
     def list_sizes(self, location=None):
         """
         List sizes on a provider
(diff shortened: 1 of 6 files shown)


ghost:
Fixed docstring errors

Kami:
Sorry for the delay, I will look some time in the next couple of days.

ghost:
No worries I'm just running off of my branch.

ghost:
Also noticing we need to add a generic state attribute to NodeImage in compute/base.py and have a statemap for NodeImage in compute/types.py. I'm primarily working to EC2/Rackspace so once we get an initial patch in place I can extend that to support a state on NodeImage for at least those drivers but specify it as None by default so it doesn't interfere with other drivers.

libcloud/compute/base.py
```diff
@@ -1066,6 +1053,95 @@ def destroy_volume_snapshot(self, snapshot):
...
+        :param node: Node to run the task on.
+        :type node: :class:`.Node`
+
+        :param name: name for new image.
+        :type name: ``str``
+
+        :param description: description for new image.
+        :type name: ``description``
+
+        :param reboot: description for new image.
```
Kami, line 1085: What does the reboot argument does?

Also, is this available across all the providers or is it provider specific?

libcloud/compute/base.py
```diff
@@ -1066,6 +1053,95 @@ def destroy_volume_snapshot(self, snapshot):
...
+        :param image_id: Node to run the task on.
+        :type image_id: ``str``
+
+        :rtype :class:`.NodeImage`:
+        :return: NodeImage instance on success.
+        """
+        raise NotImplementedError(
+            'get_image not implemented for this driver')
+
+    def copy_image(self, source_region, node_image, name, description=None):
```
Kami, line 1122: Probably a better argument order would be - `self, node_image, source_region, name, description`.

libcloud/compute/drivers/ec2.py
```diff
@@ -1376,6 +1377,22 @@ def list_images(self, location=None, ex_image_ids=None, ex_owner=None,
...
+        Get an image based on a image_id
+
+        :param image_id: Image identifier
+        :type image_id: ``str``
+
+        :return: A NodeImage object
+        :rtype: :class:`NodeImage`
+
+        """
+        images = self.list_images(ex_image_ids=[image_id, ])
```
Kami: This is a list and not a tuple so trailing comma is not needed.

Kami:
This pull request mostly looks good to be, but it would be nice if it can also get reviwed by a couple of other people - /cc @mahendra @Jc2k

libcloud/compute/base.py
ghost, line 1085: For Amazon it makes sure the instance reboots in order to guarantee file system integrity. I looked into Amazon, Rackspace, Google, and even Azure. It does appear that this is an Amazon specific argument and it is not needed as the default should be okay. I will remove it from the base and put back the original implementation in the ec2.py file.

libcloud/compute/base.py
ghost, line 1122: agreed I will adjust that

libcloud/compute/drivers/ec2.py
ghost: Okay I will fix these few things and make a commit here in a few minutes.

ghost:
Is there anything specific I need to do to trigger another Travis CI build?

Kami:
Nope, it should be triggered manually but it might take a while.

Kami:
Sorry for the delay, I've merged your changes into trunk.

I've noticed you manually created the supported methods rst file. Those files
are generated by `./contrib/generate_provider_feature_matrix_table.py` script which I updated (08d7d95dbe5e2d545b1d66a2744ae6adf4226a16) to also support generating this file for image management API.

PR #278, opened 2014-04-21
[LIBCLOUD-541] Add destination(target) instance tags support to GCE driv...

zerthimon:
Add firewall destination (target) instance tags support to GCE driver


wrigri:
This looks good to me.

Kami:
Fixed lint violations and merged changes into trunk. Thanks.

PR #280, opened 2014-04-22
The hostname should be fullyQualifiedDomainName instead of hostname

RoelVanNyen:
Otherwise we don`t know which machine this really is, you only see a part of the hostname.

### DIFF
--- a/libcloud/compute/drivers/softlayer.py
+++ b/libcloud/compute/drivers/softlayer.py
@@ -242,7 +242,7 @@ class SoftLayerNodeDriver(NodeDriver):
 
         return Node(
             id=host['id'],
-            name=host['hostname'],
+            name=host['fullyQualifiedDomainName'],
             state=state,
             public_ips=public_ips,
             private_ips=private_ips,
--- a/libcloud/test/compute/test_softlayer.py
+++ b/libcloud/test/compute/test_softlayer.py
@@ -43,7 +43,7 @@ class SoftLayerTests(unittest.TestCase):
     def test_list_nodes(self):
         nodes = self.driver.list_nodes()
         node = nodes[0]
-        self.assertEqual(node.name, 'libcloud-testing1')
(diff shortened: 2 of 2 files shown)


Kami:
@RoelVanNyen Can you please update the failing tests?

Thanks

RoelVanNyen:
Should be fixed now.

Thanks

Kami:
@RoelVanNyen Actually, sorry. I just quickly glanced over this the first time.

I think a better approach would be to store this value in the `extra['hostname']` attribute.

RoelVanNyen:
If you go to the softlayer panel and check the hosts you will see the names are equal to the fullyQualifiedDomainName and not the hostname, so it would be logical for me to use host['fullyQualifiedDomainName'] instead. Howhever if you really don`t want to do this then I would suggest putting the field into extra['fullyQualifiedDomainName'] instead of extra['hostname'].

What do you think?
Thanks

Kami:
@RoelVanNyen Ah, if the Softlayer control panel uses `fullyQualifiedDomainName` for the display name, then using that in Libcloud makes sense as well.

Kami:
Commits squashed and merged into trunk. Thanks.

RoelVanNyen:
Cool. However I don`t see your commit anywhere.
Where is it committed ?

Cheers,

phamnamlong:
Hi @RoelVanNyen, I find this fix quite problematic. Imagine the following scenario:

You create a node by: 

```
host = conn.create_node(name="first", image=image, size=size) 
```

Then, later you want to find this node by name:

```
def found_with_id(id):
  nodes = conn.list_nodes();
    for node in nodes:
      if node.name == id: # this is where the originally created name is not what it is now
        return node
  return None
```

Doesn't it break the generic purpose of libcloud where you will certainly have to "if conn is the SoftLayer" then check node.name == id + "example.com".

I suggest that the name is still the original created name while you will add extra attribute fullyQualifiedDomainName instead.

PR #283, opened 2014-04-28
[LIBCLOUD-544] GCE: fix adding node metadata at node creation

zerthimon:
Passing a simple metadata dict to create_node GCE driver method doesn't work.
GCE api requires the following structure for metadata:
metadata = {"items": [{"value": "val1", "key": "key1"}, {"value": "val2", "key": "key2"}]}

This fix converts simple dict to what the API wants.

### DIFF
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -1196,6 +1196,10 @@ class GCENodeDriver(NodeDriver):
                                               image=image,
                                               use_existing=use_existing_disk)
 
+        if ex_metadata is not None:
+            ex_metadata = {"items": [{"key": k, "value": v}
+                                     for k, v in ex_metadata.items()]}
+
         request, node_data = self._create_node_req(name, size, image,
                                                    location, ex_network,
                                                    ex_tags, ex_metadata,
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -1196,6 +1196,10 @@ class GCENodeDriver(NodeDriver):
                                               image=image,
                                               use_existing=use_existing_disk)
 
+        if ex_metadata != None:
+            ex_metadata = {"items": [{"key": k, "value": v}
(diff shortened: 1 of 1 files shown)


libcloud/compute/drivers/gce.py
```diff
@@ -1196,6 +1196,10 @@ def create_node(self, name, size, image, location=None,
                                               image=image,
                                               use_existing=use_existing_disk)
 
+        if ex_metadata != None:
```
Kami: `if ex_metadata is not None` is more ideomatic

erjohnso: Agreed

libcloud/compute/drivers/gce.py
```diff
@@ -1196,6 +1196,10 @@ def create_node(self, name, size, image, location=None,
...
+            ex_metadata = {"items": [{"key": k, "value": v}
+                                     for k, v in ex_metadata.items()]}
+
```
erjohnso, line 1202: Looks like a great fix.

Kami:
Sorry for the delay. This has now been merged into trunk. Thanks.

raphtheb:
This commit appears to have caused a regression with projects that sends a pre-structured metadata dictionnary. See https://github.com/saltstack/salt/issues/14985

I am looking into adding a verification clause and might open a new pull request when done.

erjohnso:
@raphtheb - sorry for that! Libcloud is used by both salt-cloud and ansible that I'm aware of. Adding better validation into libcloud would be most welcome!

raphtheb:
Created this pull request:

https://github.com/apache/libcloud/pull/349

It resolve the core issue, albeit in an arguably naive way. It will fail if someone sends an un-formatted dictionary containing an "items" entry, but i couldn't come up with a quick fix for that on a Friday afternoon.

raphtheb:
For the record, i squashed together my commits cleanly into a different pull request, apache/libcloud#353 .

PR #284, opened 2014-04-29
LIBCLOUD-545_Change create_image API call params from Reboot to NoReboot and negated set logic

ghost:
See excerpt from: http://docs.aws.amazon.com/AWSEC2/latest/APIReference/ApiReference-query-CreateImage.html

NoReboot
By default, this parameter is set to false, which means Amazon EC2 attempts to shut down the instance cleanly before image creation and then reboots the instance. When the parameter is set to true, Amazon EC2 doesn't shut down the instance before creating the image. When this option is used, file system integrity on the created image can't be guaranteed.

Type: Boolean

Default: false

Required: No


Kami:
Looks good. Merged into trunk. Thanks.

PR #285, opened 2014-04-29
[LIBCLOUD-536] Add support for the Outscale SAS cloud.

benoit-canet:
This commit introduce the Outscale SAS cloud an european cloud provider
featuring an EC2 compatible API.

Special care have been taken to leave room for Outscale INC an US based cloud
using the same infrastructure but governed by US laws.

Signed-off-by: Benoît Canet benoit@irqsave.net


libcloud/compute/drivers/ec2.py
```diff
@@ -4229,3 +4570,273 @@ def ex_create_tags(self, resource, tags):
...
+        """
+        supported = ['DisableApiTermination.Value', 'EbsOptimized',
+                     'GroupId.n', 'SourceDestCheck.Value', 'UserData.Value',
+                     'InstanceType.Value']
+        for attribute in attributes.keys():
+            if attribute not in supported:
+                message = "Attribute %s not implemented" % (attribute)
+                raise NotImplementedError(message)
+
+        return super(OutscaleNodeDriver, self).ex_modify_instance_attribute(
```
Kami, line 4690: Looks good, but for readability purposes please use keyword arguments everywhere.

libcloud/compute/drivers/ec2.py
```diff
@@ -4229,3 +4570,273 @@ def ex_create_tags(self, resource, tags):
...
+        """
+        Outscale does not support copying images.
+
+        Stub it.
+
+        @inherits: :class:`EC2NodeDriver.ex_copy_image`
+        """
+        pass
+
+    def ex_get_limits(self):
```
Kami, line 4734: This is more of a note for myself - seems like some of those methods should be moved from `BaseEC2NodeDriver` to `EC2NodeDriver`.

This way you don't need to include stub methods.

libcloud/compute/drivers/ec2.py
```diff
@@ -4229,3 +4570,273 @@ def ex_create_tags(self, resource, tags):
...
+        for instance_type in available_types:
+            attributes = OUTSCALE_INSTANCE_TYPES[instance_type]
+            attributes = copy.deepcopy(attributes)
+            price = self._get_size_price(size_id=instance_type)
+            attributes.update({'price': price})
+            sizes.append(NodeSize(driver=self, **attributes))
+        return sizes
+
+
+class OutscaleSasUSEastNodeDriver(OutscaleNodeDriver):
```
Kami: One class per region model has been replace with `region` argument so you can get rid of those region specific classes.

libcloud/compute/drivers/ec2.py
```diff
@@ -4229,3 +4570,273 @@ def ex_create_tags(self, resource, tags):
...
+
+    def ex_detach_network_interface(self, attachment_id, force=False):
+        """
+        Outscale does not support detaching a network interface
+
+        Stub it.
+
+        @inherits: :class:`EC2NodeDriver.ex_detach_network_interface`
+        """
+        return False
```
Kami: For methods which are not supported, you should raise `NotImplementedError('This method is not supported in the Outscale driver)`.

libcloud/compute/drivers/ec2.py
benoit-canet: Providers constants are different from region name.
I dont understand how compute/provider.py will know how to pass the correct region to the OutscaleNodeDriver.

Kami: Yes.

In the past we used a model where you needed 1 class per region. This approach isn't as scalable and nice to we deprecated it in favor of using a single constant and class which constructor takes `region` argument.

With this new approach you specify a region when instantiating a class so multiple classes and constants are not desired and needed anymore. For example:

``` python
cls = get_driver(Provider.OUTSCALE_SAS)
driver = cls('api key', 'api secret', region='us-east-1')
```

libcloud/compute/drivers/ec2.py
```diff
@@ -622,6 +629,340 @@
...
+}
+
+
+"""
+The function manipulating Outscale cloud regions will be overriden because
+Outscale instances types are in a separate dict so also declare Outscale cloud
+regions in another constant.
+"""
+OUTSCALE_REGION_DETAILS = {
+    'osc-sas-eu-west-3': {
```
Kami: For consistency with EC2 and other drivers, I would drop `osc-sas-` prefix from the region name and only use `eu-west-3` and so on.

libcloud/compute/drivers/ec2.py
benoit-canet, line 4690: I though about this but if I start passing disable_api_termination and friend as keywords the OutscaleNodeDriver will differs from the EC2NodeDriver and polymorphism wont work.
Are you sure it's not an issue ?

Kami, line 4690: Yeah, I don't think this should be an issue.

libcloud/compute/drivers/ec2.py
benoit-canet: Ok,
How can I pave the way for the future Outscale INC driver ?
Will adding another constant dictionary for it be ok ?

Kami: How is Outscale INC different from Outscale SAS?

libcloud/compute/drivers/ec2.py
benoit-canet: ok I'll drop the extra types.py constant so.

libcloud/compute/drivers/ec2.py
benoit-canet: Same codebase, and datacenters (regions) but differents corporate entities governed by differents laws (patriot act etc..)
Outscale SAS is a french company which target the european customers whereas Outscale SAS target US based customers.

benoit-canet: Also Outscale SAS prices are in euros and Outscale INC prices in $

Kami: That's what I thought, thanks for the clarification.

Since it's not just a different region it sounds like that Outscale INC should be a separate class. You can handle regions by adding another "regions" dict for Outscale INC.

libcloud/compute/drivers/ec2.py
```diff
@@ -4229,3 +4567,257 @@ def ex_create_tags(self, resource, tags):
...
+
+class OutscaleConnection(EC2Connection):
+    """
+    Connection class for Outscale
+    """
+
+    host = None
+
+
+class OutscaleNodeDriver(BaseEC2NodeDriver):
```
Kami, line 4580: This is an abstract class similar to `BaseEC2NodeDriver` and should never be instantiated directly, right?

If so, you should document that.

Kami:
Merged into trunk. Thanks.

PR #286, opened 2014-04-29
[LIBCLOUD-547] linode.py labels too long with name use linode id instead

earthgecko:
Using linode["id"] for labels and the labels step needed to be moved down to after the linode object exists so that the id can be used.


earthgecko:
Travis failed - closing

PR #287, opened 2014-04-29
[LIBCLOUD-547] linode.py labels too long with name use linode id instead

earthgecko:
Using linode["id"] for labels and the labels step needed to be moved down to after the linode object exists so that the id can be used.

Apologies about the old commit messages could not get rid of them rebased and all but the diff is sound.


Kami:
Sorry for the delay, I'll review the pull request now.

libcloud/compute/drivers/linode.py
```diff
@@ -320,6 +310,18 @@ def create_node(self, **kwargs):
             }
             self.connection.request(API_ROOT, params=params)
 
+        # Step 1d. Labels
+        # Use the linode id as the name can be up to 63 chars and the labels
+        # are limited to 48 chars
+        label = {
+            "lconfig": "[%s] Configuration Profile" % linode["id"],
```
Kami, line 317: This looks good, although if we wanted to be sure that labels are never longer we could truncate it in this case (`...[:48]).

Kami:
@earthgecko I wanted to apply your patch into trunk, but I'm getting conflicts.

Can you please sync you branch with upstream trunk and squash all the commits?

Kami:
I went ahead and manually merged the changes into trunk. Thanks.

PR #288, opened 2014-05-02
Added OpenStack_1_1_NodeDriver to the 'API Docs' section

TorstenS73:
This is supposed to fix https://issues.apache.org/jira/browse/LIBCLOUD-549.


Kami:
Looks good, merged into trunk. Thanks.

PR #289, opened 2014-05-05
LIBCLOUD-550_DisplayName

fifthecho:
Added support for "displayname" CloudStack deployment parameter and removed automatic setting of "name" parameter.

This will allow for deployment of instances without a hostname, but with a displayname, so that you can have multiple tenants using the same displayname with unique hostnames being generated from the instance UUID.

### DIFF
--- a/libcloud/compute/drivers/cloudstack.py
+++ b/libcloud/compute/drivers/cloudstack.py
@@ -700,6 +700,9 @@ class CloudStackNodeDriver(CloudStackDriverMixIn, NodeDriver):
                                         the node
         :type       ex_security_groups: ``list`` of ``str``
 
+        :keyword    ex_displayname: String containing instance display name
+        :type       ex_displayname: ``str``
+
         :rtype:     :class:`.CloudStackNode`
         """
 
@@ -727,9 +730,16 @@ class CloudStackNodeDriver(CloudStackDriverMixIn, NodeDriver):
         ex_key_name = kwargs.get('ex_keyname', None)
         ex_user_data = kwargs.get('ex_userdata', None)
         ex_security_groups = kwargs.get('ex_security_groups', None)
+        ex_displayname = kwargs.get('ex_displayname', None)
 
         if name:
             server_params['name'] = name
+
+        if name is None:
+            del server_params['name']
+
+        if ex_displayname:
+            server_params['displayname'] = ex_displayname
 
         if size:
             server_params['serviceofferingid'] = size.id


libcloud/compute/drivers/cloudstack.py
```diff
@@ -727,10 +730,17 @@ def _create_args_to_params(self, node, **kwargs):
         ex_key_name = kwargs.get('ex_keyname', None)
         ex_user_data = kwargs.get('ex_userdata', None)
         ex_security_groups = kwargs.get('ex_security_groups', None)
+        ex_displayname = kwargs.get('ex_displayname', None)
 
         if name:
             server_params['name'] = name
 
+        if name is None:
+            del server_params['name']
```
Kami, line 739: Can you even create a node if you don't specify a name?

(I know this doesn't directly relate to your code, but in case you can't, we can simplify the code)

fifthecho, line 739: Yes, if no name is specified, it will use the instance UUID as the name.

For example:
connection = cls(key=ACCESS_KEY, secret=SECRET_KEY, url=CLOUDSTACK_URL)
sizes = connection.list_sizes()
networks = connection.ex_list_networks()
securitygroups = connection.ex_list_security_groups()
images = connection.list_images()
locations = connection.list_locations()

SIZE_ID='e3bf3707-c04f-4cd1-a670-929b38c0fdfb'
LOCATION_ID='b1bb1d35-e57b-4156-a92e-d650b7c0d257'
KEY_NAME='jmoody_fifthecho'
IMAGE_ID='85d5a794-c3f4-49fd-afee-19132a850f79'
SECURITY_GROUPS=['default']
size = [s for s in sizes if s.id == SIZE_ID][0]
image = [i for i in images if i.id == IMAGE_ID][0]
location = [l for l in locations if l.id == LOCATION_ID][0]

node = connection.create_node(ex_displayname = 'test_py', ex_keyname = KEY_NAME, 
                              ex_security_groups = SECURITY_GROUPS, location = location, 
                              size = size, image = image)

Creates a node object:
<Node: uuid=7a27fabb99f7acfce0caac1ed6ab37eed1d15884, name=f73488e7-88dc-4e30-b6bb-72a1712dcee3, state=0, public_ips=[u'XXX.XXX.XXX.XXX'], private_ips=[], provider=CloudStack ...>

So, the node name is the instance UUID generated by CloudStack of "f73488e7-88dc-4e30-b6bb-72a1712dcee3"

Kami, line 739: OK, thanks for clarification.

Also, just noticed this line is unnecessary if we just initialize `server_params` to an empty dict which is a better approach anyway. I'll make this change and merge patch into trunk.

Kami:
Patch merged into trunk. Thanks.

PR #290, opened 2014-05-08
[LIBCLOUD-551] Fix Google DNS to use valid record split character

erjohnso:
@franckcuny - mind looking this over?

I ran into this because one of my testing domains was foo-bar.com.  When trying to look up a record with get_record(), I formed "foo-bar.com-A" and the get_records()'s split() call failed on unpacking the string.

Looking over other providers, it seems they've all settled on using ':'.  I also swapped the record type / name ordering for unpacking to make it more compatible with the other providers.

### DIFF
--- a/libcloud/dns/drivers/google.py
+++ b/libcloud/dns/drivers/google.py
@@ -330,7 +330,7 @@ class GoogleDNSDriver(DNSDriver):
         return records
 
     def _to_record(self, r, zone):
-        record_id = '%s-%s' % (r['name'], r['type'])
+        record_id = '%s:%s' % (r['type'], r['name'])
         return Record(id=record_id, name=r['name'],
                       type=r['type'], data=r, zone=zone,
                       driver=self, extra={})
--- a/libcloud/dns/drivers/google.py
+++ b/libcloud/dns/drivers/google.py
@@ -260,7 +260,7 @@ class GoogleDNSDriver(DNSDriver):
                 }
             ]
         }
-        request = '/managedZones/%s/changes' % (record.zone.id)
+        request = '/managedZones/%s/changes' % (record.zone)
         response = self.connection.request(request, method='POST',
                                            data=data)
         return response.success()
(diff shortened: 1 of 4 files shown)


Kami:
LGTM, +1

erjohnso:
@Kami - sorry to sneak another one in after your review.  This one-liner is actually an unrelated bug fix, but I was thinking it's so small it didn't merit it's own issue and PR.  But let me know if you prefer if this was split out.

Kami:
@erjohnso No problem.

Also, while we are at it, it would make sense to add `_get_record_id(self, record_type, record_name)` and use that instead of repeating the logic (it makes it easier to change it in the future and maybe even move it to the base driver).

libcloud/dns/drivers/google.py
```diff
@@ -260,7 +260,7 @@ def delete_record(self, record):
                 }
             ]
         }
-        request = '/managedZones/%s/changes' % (record.zone.id)
+        request = '/managedZones/%s/changes' % (record.zone)
```
Kami: Hm, how is this supposed to work? 

If you do it like that, it will use the string representation of the zone object (`str(zone)`) which you probably don't want. Unless the `zone` attribute is not an instance of `libcloud.dns.base.Zone` which means the driver is not following the base API.

erjohnso:
@Kami - Ok, you were correct that it wasn't matching the model, iiuc.  Google doesn't use numerical ID's in the API other than as informational output, so I think using the 'name' field as the 'id' is a much better fit between libcloud's model and Google's API.  But I did tuck the numerical 'id' into the 'extra' struct.

I was less confident in my ability to refactor base as you suggested.  Plus this is getting to be a big enough change that it's diverting more from the original goal (just swap out '-' with ':').

Kami:
@erjohnso Hm, looking at the API docs you should be able to use either (name or the id) in the zone case - `Identifies the managed zone addressed by this request. Can be the managed zone name or ID.`.

In any case, I'm fine with using `name`.

fcuny:
Oh, sorry about the inconsistency, I've overlooked that in the original patch.

LGTM too.

erjohnso:
Thanks @franckcuny and @Kami.  I've squashed the commit, generated the patch, and uploaded it to Jira issue https://issues.apache.org/jira/browse/LIBCLOUD-551

Kami:
Merged into trunk. Thanks.

PR #291, opened 2014-05-12
openstack: put floating ips in public_ips

bouttier:
This patch use OS-EXT-IPS:type introduced in Nova 2013.1 ("Grizzly") but stay compatible with older version.

This is useful in testing environment where floating ips can be private but we want see them as public.
I'm not sure that you want that merged but I think it could be nice so I propose ...

### DIFF
--- a/libcloud/compute/drivers/openstack.py
+++ b/libcloud/compute/drivers/openstack.py
@@ -2009,13 +2009,15 @@ class OpenStack_1_1_NodeDriver(OpenStackNodeDriver):
             if label in public_networks_labels:
                 public_ips.extend(ips)
             else:
-                for ip in ips:
-                    # is_private_subnet does not check for ipv6
+                for v in values:
+                    ip = v['addr']
+                    ip_type = v.get('OS-EXT-IPS:type')  # added in grizzly
                     try:
-                        if is_private_subnet(ip):
-                            private_ips.append(ip)
-                        else:
+                        # is_private_subnet does not check for ipv6
+                        if ip_type == 'floating' or not is_private_subnet(ip):
                             public_ips.append(ip)
+                        else:
+                            private_ips.append(ip)
                     except:
                         private_ips.append(ip)
 


Kami:
@bouttier Looks good, but can you please add some test cases for this change (`OS-EXT-IPS:type` not being present, `OS-EXT-IPS:type` being set to `floating`, `OS-EXT-IPS:type` being set to a value different than `floating`)?

libcloud/compute/drivers/openstack.py
```diff
@@ -2009,13 +2009,15 @@ def _to_node(self, api_node):
             if label in public_networks_labels:
                 public_ips.extend(ips)
             else:
-                for ip in ips:
-                    # is_private_subnet does not check for ipv6
+                for v in values:
+                    ip = v['addr']
+                    ip_type = v.get('OS-EXT-IPS:type')  # added in grizzly
```
Kami, line 2014: What are the other possible values for this attribute besides `floating`?

bouttier, line 2014: The other value is fixed (http://docs.openstack.org/api/openstack-compute/2/content/GET_os-ext-ips-v2_serverIPDetails_v2__tenant_id__servers__server_id__action_ext-os-ext-ips.html).

bouttier:
I'm not sure of how add test cases, theses values depends of openstack version (rackspace used in tests files apparently). I think `floating` and `fixed` cases are already tested by existing tests ; if `OS-EXT-IPS:type` is not set, I use `get` function so `ip_type = None != 'floating'`.

Kami:
@bouttier For testing we use response fixtures - you can find a bunch of examples in `libcloud/test/compute/`.

Kami:
@bouttier Actually, I just checked at your pull description and OpenStack docs again and I don't actually think we want this in trunk - why would you want to see private IP as public (it seems like this should be handled in the code which uses Libcloud and not in the Libcloud itself)?

Also, addresses should already be correctly split in the "private" and "public" attribute.

bouttier:
In my case, I had an issue with salt-cloud who want remove the public key  from .ssh/know_host on vm deletion. For that, salt-cloud search the first public ip (and I used an old version of libcloud that I fix in my way). I understand your point of view and I will search an other way in salt-cloud to select the right ip.

Kami:
@bouttier One thing we could do to make this easier for you it to store information about floating and fixed IP addresses in the Node `extra` attribute.

bouttier:
I haven't had time to find the best way to modify salt yet, but yes this can be useful, and not only for me apparently (#296).

PR #292, opened 2014-05-14
Add m3 instances for ec2 Oregon

RoelVanNyen:
I added the m3 instances, the flavors/prices were already there.


Kami:
Merged into trunk, thanks.

P.S. The reason prices are already there is because pricing information is automatically fetched and parsed from the AWS website by the`./contrib/` script. Sadly Amazon doesn't provide an API endpoint for retrieving sizes information so we can't use that approach for sizes.

PR #294, opened 2014-05-15
[LIBCLOUD-554] Add support for providing filters and VPC IDs to EC2: ex_list_networks()

zerthimon:
EC2 API allows VPC ID and filters to be provided to list only specific VPCs. 
http://docs.aws.amazon.com/AWSEC2/latest/APIReference/ApiReference-query-DescribeVpcs.html
Add support for specifying filters and vpc ids when listing networks (vpc's) with ec2 driver's method ex_list_networks()

### DIFF
--- a/libcloud/compute/drivers/ec2.py
+++ b/libcloud/compute/drivers/ec2.py
@@ -2256,11 +2256,14 @@ class BaseEC2NodeDriver(NodeDriver):
         Return a list of :class:`EC2Network` objects for the
         current region.
 
-        :param      network_ids: One or more VPC IDs
+        :param      network_ids: Return only networks matching the provided
+                                 network IDs. If not specified, a list of all
+                                 the networks in the corresponding region
+                                 is returned.
         :type       network_ids: ``list``
 
         :param      filters: The filters so that the response includes
-                             information for only certain VPCs
+                             information for only certain networks.
         :type       filters: ``dict``
 
         :rtype:     ``list`` of :class:`EC2Network`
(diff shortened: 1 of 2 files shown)


libcloud/compute/drivers/ec2.py
```diff
@@ -2251,15 +2251,42 @@ def ex_register_image(self, name, description=None, architecture=None,
...
+        :param      filters: The filters so that the response includes
+                             information for only certain VPCs
+        :type       filters: ``dict``
+
         :rtype:     ``list`` of :class:`EC2Network`
         """
         params = {'Action': 'DescribeVpcs'}
 
+        if network_ids is not None:
+            for network_idx, network_id in enumerate(network_ids, 1):
```
Kami: We still support Python 2.5 which doesn't support second argument (`start`) to `enumerate` function so you need to do it like this - https://github.com/zerthimon/libcloud/blob/LIBCLOUD-554_filters/libcloud/compute/drivers/ec2.py#L4307

libcloud/compute/drivers/ec2.py
```diff
@@ -2251,15 +2251,42 @@ def ex_register_image(self, name, description=None, architecture=None,
         )
         return image
 
-    def ex_list_networks(self):
+    def ex_list_networks(self, network_ids=None, filters=None):
         """
         Return a list of :class:`EC2Network` objects for the
         current region.
 
+        :param      network_ids: One or more VPC IDs
```
Kami: Better docstring would be in order here. For example, something along the lines of "Return only networks which match the provided network IDs. If not provided, a list of all the networks on your account is returned."

Kami:
Adding some tests for this new functionality would be good.

zerthimon:
Something is wrong with the tests:

Providing VPC IDs and filters returns correct number of matching objects  when working with EC2 API. In tests the calls return all objects in the fixture (ignores the filter)

Please advice!

Kami:
I went ahead, squashed the commits, fixed the tests and merged changes into trunk.

The problem in your case was that you asserted on the response, but you returned the same fixture which contains all the networks. I've changed the tests to assert that the correct query parameters are sent, because that's the thing we are actually interested in and want to test in this case.

PR #295, opened 2014-05-15
[LIBCLOUD-555] Added support for openstack compute availability zone

syndicut:
https://issues.apache.org/jira/browse/LIBCLOUD-555

### DIFF
--- a/libcloud/compute/drivers/openstack.py
+++ b/libcloud/compute/drivers/openstack.py
@@ -1318,6 +1318,9 @@ class OpenStack_1_1_NodeDriver(OpenStackNodeDriver):
                                                                  {}))
         }
 
+        if 'availability_zone' in kwargs:
+            server_params['availability_zone'] = kwargs['availability_zone']
+
         if 'ex_keyname' in kwargs:
             server_params['key_name'] = kwargs['ex_keyname']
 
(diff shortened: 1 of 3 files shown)


libcloud/compute/drivers/openstack.py
```diff
@@ -1318,6 +1318,9 @@ def _create_args_to_params(self, node, **kwargs):
                                                                  {}))
         }
 
+        if 'availability_zone' in kwargs:
+            server_params['availability_zone'] = kwargs['availability_zone']
```
Kami: This is an extension param so please prefix it with `ex_`:

``` python
 if 'ex_availability_zone' in kwargs:
    server_params['availability_zone'] = kwargs['ex_availability_zone']
```

Kami: Please also update `create_node` docstring and document this kwarg there - https://github.com/syndicut/libcloud/blob/LIBCLOUD-555_openstack_availability_zone/libcloud/compute/drivers/openstack.py#L1218

syndicut: Done!

Kami:
Merged into trunk.

Thanks.

PR #296, opened 2014-05-20
[LIBCLOUD-557] Updated list_nodes to categorize private_ips and public_i...

sasi5a1:
...ps


Kami:
This seems to be related to #291, right? (/cc @bouttier)

It seems the best approach would be to do something like described in my last comment - https://github.com/apache/libcloud/pull/291#issuecomment-43046177

Kami:
@sasi5a1 ^^ what do you think about my comment and the approach suggested there?

PR #297, opened 2014-05-21
LIBCLOUD-559_sec-groups EC2 new method to get security groups

zerthimon:
Current method of EC2 driver's listing security groups [ex_list_security_groups()] only returns security group names. This method doesn't provide any ability to filter security groups by names or IDs, nor does it list security group's rules with no way to know which rules are in place.

This patch adds new method of listing security groups: ex_get_security_groups() and new class: EC2SecurityGroup. The objects of new class have ingress_rules and egress_rules structures that reflect the EC2 security groups rules.

The filter dict building code was moved into a separate function.
This patch doesn't provide any unit tests (will be added if the concept is approved)

### DIFF
--- a/libcloud/compute/drivers/ec2.py
+++ b/libcloud/compute/drivers/ec2.py
@@ -64,6 +64,7 @@ __all__ = [
 
     'EC2NodeLocation',
     'EC2ReservedNode',
+    'EC2SecurityGroup',
     'EC2Network',
     'EC2NetworkSubnet',
     'EC2NetworkInterface',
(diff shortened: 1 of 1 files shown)


libcloud/compute/drivers/ec2.py
```diff
@@ -2527,6 +2547,50 @@ def ex_list_security_groups(self):
...
+        if group_names:
+            for name_idx, group_name in enumerate(group_names):
+                name_idx += 1 # We want 1-based indexes
+                name_key = 'GroupName.%s' % name_idx
+                params[name_key] = group_name
+
+        if filters:
+            params.update(self._build_filters(filters))
+
+        response = self.connection.request(self.path, params=params)
```
Kami, line 2588: You are actually performing the same request twice. You ignore `response` and perform it again a couple of lines bellow.

In any case, I will fix this while merging the patch.

Kami:
I've made some changes and fixes and merged this patch into trunk.

Hopefully, security group API promotion will be implemented soon and then we can deprecate and get rid of this method all together.

libcloud/compute/drivers/ec2.py
zerthimon, line 2588: Yes, thanks for pointing this out, it's clearly a bug.

zerthimon:
Thanks, Kami

PR #298, opened 2014-05-22
LIBCLOUD-560: Updated CloudStack driver to support new resource types in 4.3

cderamus:
4.3 added four new resource types to the listResourceLimits response. This update adds support for those types as well as support for ignoring unknown types.


Kami:
Merged into trunk.

Thanks.

PR #299, opened 2014-05-25
Support for Google-Containers

magnusart:
Recently Google have released a google VM Image that comes with docker preinstalled and supports docker provisioning via metadata.

See the documentation here: https://developers.google.com/compute/docs/containers

Toady in libcloud both debian and centos images are supported by matching against the VM name and then injecting the appropiate project name (`debian-cloud` and `centos-cloud`).

This adds support for `google-containers` as well.


Kami:
Looks good. Merged into trunk.

Thanks.

erjohnso:
@magnusart - thanks Magnus!  Sorry for the laggy LGTM. :)

PR #300, opened 2014-05-26
[LIBCLOUD-563] fix group id variable

zerthimon:
Fix variable name.

### DIFF
--- a/libcloud/compute/drivers/ec2.py
+++ b/libcloud/compute/drivers/ec2.py
@@ -4590,7 +4590,7 @@ class BaseEC2NodeDriver(NodeDriver):
 
         :rtype: ``dict``
         """
-        params = {'GroupId': id,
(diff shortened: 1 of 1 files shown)


Kami:
Yeah, I agree that it would be better to use a name which is not a built in function.

The problem though it that some base methods and classes use `id` for the name where changing it would break backward compatibility and making it painful for users to migrate / update their code.

Kami:
Btw, I just had a look and your fork and pull request seem out of that.

It looks like the version in trunk is already using `group_id` in the method signature - https://github.com/apache/libcloud/blob/trunk/libcloud/compute/drivers/ec2.py#L4584

zerthimon:
Yes, this was fixed in https://github.com/apache/libcloud/pull/301 which was merged before this one.
Closing.

PR #301, opened 2014-05-27
Fix various bugs and add "floating IP handler" to the OpenStack provider

hcs42:
This commit contains:
- Fixes of typos in docstrings and comments
- 7 various small bugfixes
- a small feature (add "floating IP handler" to the OpenStack provider)
- a small addition to the DUMMY provider

If I should break up this pull request into several pull requests and/or I should write Jira tickets, just let me know.

### DIFF
--- a/libcloud/compute/drivers/openstack.py
+++ b/libcloud/compute/drivers/openstack.py
@@ -2276,7 +2276,7 @@ class OpenStack_1_1_FloatingIpPool(object):
         """
         Get specified floating IP from the pool
 
-        :param      ip: floating IP to remove
+        :param      ip: floating IP to get
         :type       ip: ``str``
 
         :rtype: :class:`OpenStack_1_1_FloatingIpAddress`
--- a/libcloud/compute/drivers/openstack.py
+++ b/libcloud/compute/drivers/openstack.py
@@ -1758,7 +1758,7 @@ class OpenStack_1_1_NodeDriver(OpenStackNodeDriver):
         resp = self.connection.request('/os-security-groups/%s' %
                                        (security_group.id),
                                        method='DELETE')
-        return resp.status == httplib.NO_CONTENT
+        return resp.status in (httplib.NO_CONTENT, httplib.ACCEPTED)
 
     def ex_create_security_group_rule(self, security_group, ip_protocol,
                                       from_port, to_port, cidr=None,
--- a/libcloud/compute/drivers/openstack.py
+++ b/libcloud/compute/drivers/openstack.py
@@ -2168,6 +2168,62 @@ class OpenStack_1_1_NodeDriver(OpenStackNodeDriver):
         return self._to_floating_ip_pools(
             self.connection.request('/os-floating-ip-pools').object)
 
+    def _to_floating_ips(self, obj):
+        ip_elements = obj['floating_ips']
+        return [self._to_floating_ip(ip) for ip in ip_elements]
+
+    def _to_floating_ip(self, obj):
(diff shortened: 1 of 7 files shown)


libcloud/utils/publickey.py
```diff
@@ -16,6 +16,10 @@
 import base64
 import hashlib
 
+from Crypto.Util.py3compat import bchr
```
Kami: paramiko (and pycrypto) is an optional dependency, but `get_pubkey_openssh_fingerprint` should also work without those libraries so we can't rely on `bchr` function from `pycrypto`.

libcloud/utils/py3.py
```diff
@@ -93,6 +101,11 @@ def dictvalues(d):
 
     def tostring(node):
         return ET.tostring(node, encoding='unicode')
+
+    def hexadigits(s):
```
Kami, line 109: It would be nice to add a small test case for this.

libcloud/compute/drivers/openstack.py
```diff
@@ -2168,6 +2168,14 @@ def ex_list_floating_ip_pools(self):
         return self._to_floating_ip_pools(
             self.connection.request('/os-floating-ip-pools').object)
 
+    def ex_create_floating_ip_handler(self):
```
Kami: For consistency, please avoid handler approach and add method directly to the driver. I'm not totally sure about the name though since `ex_create_floating_ip` might be confusing since we already have `OpenStack_1_1_FloatingIpPool.create_floating_ip`.

libcloud/utils/publickey.py
hcs42: Is it OK if I add a `bchr` function to our `py3` module?

libcloud/utils/py3.py
hcs42, line 109: Will do.

libcloud/compute/drivers/openstack.py
hcs42: I did it this way so that I can easier reuse the code from `OpenStack_1_1_FloatingIpHandler`.

But that's not a lot of code, and I agree that it would be more convenient to use this functionality without having to create a handler. I will rewrite it and we will see how that looks like.

libcloud/utils/publickey.py
Kami: Yep

hcs42:
I updated three commits:
- "Fix publickey._to_md5_fingerprint for Python 3": added a unit test.
- "Fix publickey.get_pubkey_ssh2_fingerprint for Python 3": use our own py3.bchr function.
- "Add "floating IP" functions to the OpenStack provider": instead of using a handler, I added IP address handler methods directory to the driver class. I used the method name you mentioned, I don't find it's confusing.
  
  I had to add a "driver" parameter to the OpenStack_1_1_FloatingIpAddress class so that the object can use it to delete itself. I didn't want to use the "pool" object for this, I felt that that would be more confusing. Suggestions about better solutions are welcome.

Kami:
I've squashed the commits, fixed some lint issues and merged your changes into trunk.

Thanks!

Kami:
@hcs42 Btw, I know this has already been merged, but what do you think about this approach - https://issues.apache.org/jira/browse/LIBCLOUD-567?focusedCommentId=14012135&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-14012135 ?

hcs42:
Thanks! I posted my opinion in the JIRA ticket.

PR #302, opened 2014-05-30
[LIBCLOUD-568] Fixing cross service OAuth scopes for Google Compute Engine / DNS / Storage

erjohnso:
Prior to this fix, a user could call "get_driver()" for GCE but authorization was only allowed for the "compute" scope.  With the addition of the DNS module, users calling its "get_driver()" would only be authorized to the DNS service.

This change allows scopes to be set as keyword params in get_driver() (or via secrets.py) that get propagated down to the authorization connection class.  The default behavior is to grant authorization via scopes to all supported google services (compute, storage, dns).

It should be noted that the storage authorization via oauth scopes is bogus, but sets the stage for future enhancements we expect to contribute over the summer.

@wrigri, @franckcuny - A review would be much appreciated!

### DIFF
--- a/libcloud/common/google.py
+++ b/libcloud/common/google.py
@@ -506,10 +506,21 @@ class GoogleBaseConnection(ConnectionUserAndKey, PollingConnection):
                 auth_type = 'SA'
             else:
                 auth_type = 'IA'
-        if 'scope' in kwargs:
-            self.scope = kwargs['scope']
+
+        # Default scopes to read/write for compute, storage, and dns.  Can
+        # override this when calling get_driver() or setting in secrets.py
+        self.scope = None
+        if kwargs and type(kwargs) is dict:
+            self.scope = kwargs.get('scope', None)
             kwargs.pop('scope', None)
+        if not self.scope:
+            self.scope = [
+                'https://www.googleapis.com/auth/compute',
+                'https://www.googleapis.com/auth/devstorage.full_control',
+                'https://www.googleapis.com/auth/ndev.clouddns.readwrite',
+            ]
         self.token_info = self._get_token_info_from_file()
+
         if auth_type == 'SA':
             self.auth_conn = GoogleServiceAcctAuthConnection(
                 user_id, key, self.scope, **kwargs)
--- a/libcloud/common/google.py
+++ b/libcloud/common/google.py
@@ -253,7 +253,7 @@ class GoogleBaseAuthConnection(ConnectionUserAndKey):
     host = 'accounts.google.com'
     auth_path = '/o/oauth2/auth'
 
-    def __init__(self, user_id, key, scope,
+    def __init__(self, user_id, key, scopes=[],
                  redirect_uri='urn:ietf:wg:oauth:2.0:oob',
                  login_hint=None, **kwargs):
         """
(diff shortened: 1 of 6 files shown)


erjohnso:
Hmm... not sure what to do about those travis failures,

``` sh
WARNING:test command found but not installed in testenv
```

I don't think it has anything to do with my changes and if it does, I'm at a loss on how to interpret the error.

libcloud/common/google.py
```diff
@@ -506,10 +506,21 @@ def __init__(self, user_id, key, auth_type=None,
                 auth_type = 'SA'
             else:
                 auth_type = 'IA'
-        if 'scope' in kwargs:
-            self.scope = kwargs['scope']
+
+        # Default scopes to read/write for compute, storage, and dns.  Can
+        # override this when calling get_driver() or setting in secrets.py
+        self.scope = None
+        if kwargs and type(kwargs) is dict:
```
Kami: `type(kwargs) is dict` check shouldn't be necessary since, afaik, if you use *\* kwarg expansion it will always be a dict.

Kami: Actually, there is no need to use kwargs here (right?). If so, you should just explicitly put `scope=None` in the method signature. Even better would be to use `scopes` name since the argument value is a list.

erjohnso:
@Kami - I think this is good to go unless you want to hear from @wrigri or @franckcuny.

libcloud/common/google.py
```diff
@@ -506,16 +510,24 @@ def __init__(self, user_id, key, auth_type=None,
...
             else:
                 auth_type = 'IA'
-        if 'scope' in kwargs:
-            self.scope = kwargs['scope']
-            kwargs.pop('scope', None)
+
+        # Default scopes to read/write for compute, storage, and dns.  Can
+        # override this when calling get_driver() or setting in secrets.py
+        self.scopes = scopes
+        if self.scopes is None or not self.scopes:
```
wrigri: Is there a reason to specifically test for None here?  Wouldn't 'if not self.scopes:' work?

wrigri:
Looks good to me

libcloud/common/google.py
erjohnso: Thanks for looking it over @wrigri!

Old habits. Back when I used to do a lot python development (circa 2002), this was considered a performance tip to use "is" for singleton comparisons. Since we set None as the default value in the method signature, and this change assumes the user will likely never pass in a non-None value for 'scopes', this would have (in my old brain) better performance.

But, I suspect that's either a) no longer a consideration 12 years later, and b) if there is a performance benefit, it's likely masked out by HTTP calls to remote services anyway. ;-)

As I said, just an "old habit". :)

wrigri: So you made me curious, and as it turns out, it looks like 'if not x:' is actually faster than 'if x is None:' even when x is an empty list.

Using Python 2.7.3 on Linux:

> > > timeit.timeit('if a is None: pass', setup='a=None', number=100000000)
> > > 2.322221040725708
> > > timeit.timeit('if not a: pass', setup='a=None', number=100000000)
> > > 1.8063910007476807
> > > timeit.timeit('if not a: pass', setup='a=[]', number=100000000)
> > > 2.0614678859710693

erjohnso:
@Kami - I think this is all set now. :)

Kami:
Thanks.

I've merged it into trunk with a minor fix - I've changed the default value for `scopes` attribute from `[]` to `None` and initialized empty list inside the function.

The problem is that default values for arguments are created at the time of the function definition. This means that if you would mutate `scope` argument inside the function, you would actually be mutating the default parameter value which is almost never something you would want.

erjohnso:
Thank you @Kami, nice catch!

PR #303, opened 2014-06-01
[LIBCLOUD-569] Removing the overloaded constructor from the rimuhosting ...

amastracci:
...base class.

Rimu uses gzip'ed responses but the overloaded constructor was not calling the decompress function resuuulting in a MalformedResponseError. All functionality in overloaded constructor already existed in the base clase.

### DIFF
--- a/libcloud/compute/drivers/rimuhosting.py
+++ b/libcloud/compute/drivers/rimuhosting.py
@@ -43,16 +43,9 @@ class RimuHostingException(Exception):
 
 
 class RimuHostingResponse(JsonResponse):
-    def __init__(self, response, connection):
-        self.body = response.read()
-        self.status = response.status
-        self.headers = dict(response.getheaders())
-        self.error = response.reason
-        self.connection = connection
-
-        if self.success():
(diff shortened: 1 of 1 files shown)


Kami:
Looking at the differences between base and overloaded constructor - base constructor also lower cases the header keys, but it seems like RimuHosting driver never accesses `response.headers` properly so the driver shouldn't be affected by this.

Also, I don't have RimuHosting account so I can't test those changes myself. Have you actually tested those changes with the RimuHosting API and can confirm that they are working?

Thanks

amastracci:
Yes, I was able to successfully bring up a Rimu compute node with/without libcloud debugging. Stepped through the request/response in the debugger and the driver was correctly handling the response.

Kami:
@amastracci I noticed you closed this pull request. Was this intentional? Or can I proceed and merge it?

amastracci:
Pilot error. Yes go ahead and merge it up.

Kami:
Merged into trunk.

Thanks.

PR #308, opened 2014-06-03
Removing ex_ from the front of vps_parameters to avoid overwrites of ser...

doberloh:
The base rimu compute class has options in the create function that can set custom memory and disk values for a VPS. The problem is in the logic that looks to see if vps_parameters is in the data structure sent in the create api call. In the first block it looks to see if vps_parameters exists which is the correct naming. In the second one it looks for ex_vps_parameters, which will never exist. And then it re initializes vps_parameters which wipes the values set for memory in the previous block. This change fixes that.

### DIFF
--- a/libcloud/compute/drivers/rimuhosting.py
+++ b/libcloud/compute/drivers/rimuhosting.py
@@ -307,7 +307,7 @@ class RimuHostingNodeDriver(NodeDriver):
             data['vps_parameters']['memory_mb'] = kwargs['ex_memory_mb']
 
         if 'ex_disk_space_mb' in kwargs:
-            if 'ex_vps_parameters' not in data:
+            if 'vps_parameters' not in data:
                 data['vps_parameters'] = {}
             data['vps_parameters']['disk_space_mb'] = \
                 kwargs['ex_disk_space_mb']


Kami:
Fix looks good. While we are at it, this is also a good opportunity to improve the existing code and remove some unnecessary duplication.

We could save a couple of lines by, pre-defining `vps_parameters` to an empty dict (unless RimuHosting API doesn't like if you send them an empty dict):

``` python
data = {
            'instantiation_options': {
                'domain_name': name, 'distro': image.id
            },
            'vps_parameters': {},
            'pricing_plan_code': size.id,
        }
...
```

This way we can get rid of duplicated `'vps_parameters' not in data` check.

It would also be good to add some tests to make sure that correct parameters are sent to the provider (not a blocker for this pull request).

doberloh:
I did not test it with an empty vps_parameters. I wont have time today to try it due to work but I could try it out in the coming days. I can also add some checks to make sure that proper params are sent to the function.

Kami:
I went ahead, made vps_parameters changes I've suggested and merged pull request into trunk.

I've modified it to not send empty 'vps_parameters' attribute (just in case) so it should behave exactly the same as it did previously.

Thanks.

doberloh:
That sounds great. Once we ramp up here on developing our internal scripts I'm sure I can help make this driver a bit better. 

Thanks

PR #309, opened 2014-06-04
[LIBCLOUD-573] Update 'CA_CERTS_PATH' to look for updated MacOS X Homebrew location

paparomeo:
https://issues.apache.org/jira/browse/LIBCLOUD-573

### DIFF
--- a/libcloud/security.py
+++ b/libcloud/security.py
@@ -42,7 +42,10 @@ CA_CERTS_PATH = [
     # macports: curl-ca-bundle
     '/opt/local/share/curl/curl-ca-bundle.crt',
 
-    # homebrew: curl-ca-bundle
+    # homebrew: openssl
+    '/usr/local/etc/openssl/cert.pem',
+
+    # homebrew: curl-ca-bundle (backward compatibility)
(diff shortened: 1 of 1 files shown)


libcloud/security.py
```diff
@@ -42,8 +42,8 @@
     # macports: curl-ca-bundle
     '/opt/local/share/curl/curl-ca-bundle.crt',
 
-    # homebrew: curl-ca-bundle
-    '/usr/local/opt/curl-ca-bundle/share/ca-bundle.crt',
+    # homebrew: openssl
+    '/usr/local/etc/openssl/cert.pem',
```
Kami, line 46: Libcloud tries the paths in the way they are defined and the first bundle which is found is used.

It seems like `/usr/local/etc/openssl/cert.pem` is preferred since it's generated from a key chain, but we should probably still keep path to curl-ca-bundle path there for backward compatibility reasons (not all users might have installed openssl formula, etc.).

We can simply decrease the priority of curl-ca-bundle and move it after the openssl bundle path.

paparomeo:
@Kami: agreed. Pull request updated.

Kami:
Merged into trunk.

Thanks!

PR #310, opened 2014-06-06
Fix floating IP initialization

hcs42:
The OpenStack_1_1_FloatingIpAddress constructor was called incorrectly: "self" was passed as "pool" instead of "driver". (The bug was introduced in my earlier commit.)


Kami:
Sorry for the delay. I've merged the changes into trunk. Thanks.

On a related note -

There were a couple of other spots which also didn't use keyword arguments and specify a `driver` argument, so I also updated those. This is also a good example of why always using keyword arguments is a good idea - it doesn't just make code more readable / clear, but it also prevents issues like this from happening :)

hcs42:
> Sorry for the delay.

No worries :)

> On a related note -

I totally agree.

PR #311, opened 2014-06-09
[LIBCLOUD-576] SWIFT: removed superfulous call to upper() on service_region

zestrada:
As stated in the jira ticket, OpenStack uses mixed-case regions, but rackspace only has uppercase regions. It seems that as a holdover from supporting cloudfiles all openstack regions were converted to upper, causing endpoint_get to fail.


Kami:
Good catch. I'll merge it into trunk tomorrow. 

Thanks.

Kami:
Merged into trunk.

Thanks.

PR #314, opened 2014-06-10
[LIBCLOUD-540] [WIP] Copy object

rgbkrk:
This is the beginning of a patch for https://issues.apache.org/jira/browse/LIBCLOUD-540. It's been sitting on my machine gathering dust, so I'm submitting the WIP it to make sure I finish it off.

### DIFF
--- a/libcloud/storage/base.py
+++ b/libcloud/storage/base.py
@@ -584,6 +584,29 @@ class StorageDriver(BaseDriver):
 
         return True
 
+    def copy_object(self, src_container_name, src_object_name,
+                    dst_container_name, dst_object_name):
+        """
+        Copy object from one container to another (or the same container with a
+        new name).
+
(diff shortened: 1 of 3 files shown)


libcloud/storage/base.py
```diff
@@ -584,6 +584,29 @@ def _save_object(self, response, obj, destination_path,
 
         return True
 
+    def copy_object(self, src_container_name, src_object_name,
+                    dst_container_name, dst_object_name):
+        """
+        Copy object from one container to another (or the same container with a
+        new name).
+
+        :param src_container_name: Source container name
```
Kami, line 593: To be consistent with existing APIs, this should probably take a container object, so `src_container` and `dst_container`, right?

This way it's also more flexible if a provider uses id instead of a name as container's unique identifier.

PR #316, opened 2014-06-11
Cloudstack Driver - Create/Delete Network functionality + tests

boul:
Functions added on the cloudstack driver:

ex_list_network_offering
ex_create_network
ex_delete_network

New Object:
CloudStackNetworkOffering

And some additions to the CloudStackNetwork dictionary
netmask
gatway
vpcid


libcloud/compute/drivers/cloudstack.py
```diff
@@ -453,6 +465,30 @@ def __repr__(self):
                    self.networkofferingid, self.zoneid, self.driver.name))
 
 
+class CloudStackNetworkOffering(object):
+    """
+    Class representing a CloudStack Network Offering.
+    """
+
+    def __init__(self, name, displaytext, guestiptype, id, serviceofferingid,
```
Kami: For consistency with existing code and other drivers, please use `_` to separate words in the argument names - display_text, guest_ip_type, service_offerring_id, for_vpc, etc.

libcloud/compute/drivers/cloudstack.py
```diff
@@ -880,6 +916,122 @@ def ex_list_networks(self):
...
+                            netoffer['displaytext'],
+                            netoffer['guestiptype'],
+                            netoffer['id'],
+                            netoffer['serviceofferingid'],
+                            netoffer['forvpc'],
+                            self))
+
+        return networkofferings
+
+    def ex_create_network(self, displaytext, name, networkoffering,
```
Kami: Same here, please use underscore in the argument names.

libcloud/compute/drivers/cloudstack.py
```diff
@@ -880,6 +916,122 @@ def ex_list_networks(self):
...
+                                    name,
+                                    networkoffering.id,
+                                    result['id'],
+                                    location.id,
+                                    self,
+                                    extra=extra)
+
+        return network
+
+    def ex_delete_network(self, network, forced=None):
```
Kami: For consistency with existing code base, please call the argument `force` instead of `forced`.

libcloud/compute/drivers/cloudstack.py
Kami: Instead of using kwargs, please explicitly declare `gateway`, `netmask` and `vpc_id` argument in the method signature.

Using kwargs in cases like this makes programmatic introspection and many other things hard / impossible - https://libcloud.readthedocs.org/en/latest/development.html#don-t-abuse-kwargs

boul:
Thx for the quick feedback, see 50e6c75 with the suggested changes for consistancy reasons.

libcloud/compute/drivers/cloudstack.py
sebgoa: Could you also add deleteNetworkOffering and createNetworkOffering ? Since you are defining the CloudStackNetworkOffering class, you could add the other methods

boul: createNetworkOffering and deleteNetworkOffering are only-available to ROOT-admins. I might add this at a later stage. Especially createNetworkOffering could be tricky based on the different network providers. I'd rather put some focus on CRUD VPC's.

Kami: Yeah, and it's also worth noting that compute drivers should only implement user / non-admin functionality.

We need to find a better and unified solution (maybe a separate driver / class) for handling admin actions, but this requires a bit more thought and work.

sebgoa: On Jun 11, 2014, at 3:56 PM, Tomaz Muraus notifications@github.com wrote:

> In libcloud/compute/drivers/cloudstack.py:
> 
> > -        networkofferings = []
> >   +
> > -        for netoffer in netoffers:
> >   +
> > -            networkofferings.append(CloudStackNetworkOffering(
> > -                            netoffer['name'],
> > -                            netoffer['displaytext'],
> > -                            netoffer['guestiptype'],
> > -                            netoffer['id'],
> > -                            netoffer['serviceofferingid'],
> > -                            netoffer['forvpc'],
> > -                            self))
> >   +
> > -        return networkofferings
> >   +
> > -    def ex_create_network(self, displaytext, name, networkoffering,
> 
> Yeah, and it's also worth noting that compute drivers should only implement user / non-admin functionality.
> 
> We need to find a better and unified solution (maybe a separate driver / class) for handling admin actions, but this requires a bit more thought and work.

Yes agreed, good point by Roland.

libcloud/compute/drivers/cloudstack.py
```diff
@@ -880,6 +917,122 @@ def ex_list_networks(self):
...
+                                    netoffer['guestiptype'],
+                                    netoffer['id'],
+                                    netoffer['serviceofferingid'],
+                                    netoffer['forvpc'],
+                                    self))
+
+        return networkofferings
+
+    def ex_create_network(self, displaytext, name, networkoffering,
+                          location, **kwargs):
```
sebgoa: You still need to fix the kwargs issue. We prefer to explicitly list the arguments that you need: gateway, net mask, vpc_id

boul: Fixed it. Cheers!

sebgoa:
Can you squash your commits ? that would make a clean PR. thx

Kami:
I've squashed the commits, renamed some variables to use underscores and merged patch into trunk. Thanks.

boul:
Thx just squashed before noticing you already did. Thx!

Kind Regards,
Roeland Kuipers

http://www.schubergphilis.com

Sent from a mobile device.

sebgoa:
@Kami you were not that nice with me when I submitted pr :) …maybe once

Kami:
@runseb heh, I can't always easily squash the commits - if the branch is not in sync with trunk, trying to apply the patch results in conflicts and resolving this is quite painful (although, I do it every now and then).

PR #317, opened 2014-06-19
Fix ex_unpause_node in openstack

porduna:
ex_unpause_node in the OpenStack driver was calling the 'pause' action instead of the 'unpause' action.


Kami:
Good catch, merged into trunk. Thanks.

PR #318, opened 2014-06-20
[LIBCLOUD-579] Allow nodes to be launched into a specific subnet by addi...

zerthimon:
...ng subnet parameter to create_node() method of ec2 driver


Kami:
Merged into trunk. Thanks.

PR #319, opened 2014-06-20
use existing _pathlist() function to convert key and an array of values ...

zerthimon:
...into AWS query param format instead of doing this in every function


Kami:
Merged into trunk. Thanks.

PR #320, opened 2014-06-20
[LIBCLOUD-580] node filters for ec2 driver

zerthimon:
Extend EC2 driver list_nodes() method with ability to provide filters to list only specific nodes.

### DIFF
--- a/libcloud/compute/drivers/ec2.py
+++ b/libcloud/compute/drivers/ec2.py
@@ -1875,7 +1875,7 @@ class BaseEC2NodeDriver(NodeDriver):
         'terminated': NodeState.TERMINATED
     }
 
-    def list_nodes(self, ex_node_ids=None, filters=None):
+    def list_nodes(self, ex_node_ids=None, ex_filters=None):
         """
         List all nodes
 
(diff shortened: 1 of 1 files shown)


libcloud/compute/drivers/ec2.py
```diff
@@ -1875,7 +1875,7 @@ class BaseEC2NodeDriver(NodeDriver):
         'terminated': NodeState.TERMINATED
     }
 
-    def list_nodes(self, ex_node_ids=None):
+    def list_nodes(self, ex_node_ids=None, filters=None):
```
Kami: Since it's an extension argument, it should be prefixed with ex - `ex_filters`.

Also, ex_node_ids and filters arguments are kinda overlapping (you could get the same result when using filters), but yeah, we should probably still leave it there for backward compatibility purposes.

zerthimon:
Should parameter filters be renamed to ex_filters in other methods too (ex_list_networks, ex_get_security_groups, ex_list_route_tables and others in pending pull requests), for consistency ?

zerthimon:
@Kami anything else to be done here ?

Kami:
@zerthimon Nope. I'll go ahead and merge it shortly. Thanks.

Kami:
Merged into trunk. Thanks.

PR #321, opened 2014-06-22
Ssh key management for softlayer

Itxaka:
This pull requests add ssh key management for the SoftLayer driver.

Add methods for:
list_key_pairs()
get_key_pair()
import_key_pair_from_string()
delete_key_pair()
create_key_pair()

It also adds a new property to create node "ex_keyname" which is the name of the key to be associated to the new node.

Added tests for all methods and modified the test_create_node() test to use a sshkey as well.


libcloud/compute/drivers/softlayer.py
```diff
@@ -188,6 +194,47 @@ def _get_object_mask(self, service, mask):
             return {}
 
 
+class KeyPair(object):
```
Kami: Key pair management is now part of the base compute API so you should use the existing `libcloud.compute.base.KeyPair` class.

Itxaka:
Fixed the Keypar stuff that @Kami brougth to my attention. Im guessing that the way of storing extra stuff for the Keypairs is to use the extra field, so that's where I stored the ID of the key :D

Itxaka:
Any update if this is gonna be merged @Kami ?

Kami:
@Itxaka Really sorry for the delay.

I will review it again now and try to get it merged asap.

libcloud/compute/drivers/softlayer.py
```diff
@@ -411,6 +422,82 @@ def create_node(self, **kwargs):
 
         return self._to_node(raw_node)
 
+    def _to_key_pairs(self, elems):
```
Kami: Minor style thing - for consistency, please move "private" methods to the end of the class after all the "public" ones.

libcloud/compute/drivers/softlayer.py
```diff
@@ -411,6 +422,82 @@ def create_node(self, **kwargs):
...
+    def _to_key_pair(self, elem):
+        key_pair = KeyPair(name=elem['label'],
+                           public_key=elem['key'],
+                           fingerprint=elem['fingerprint'],
+                           private_key=elem.get('private', None),
+                           driver=self,
+                           extra={'id': elem['id']})
+        return key_pair
+
+    def _key_name_to_id(self, key):
```
Kami: Argument should probably be called `name` since it represents a key name and not a key object.

libcloud/compute/drivers/softlayer.py
```diff
@@ -411,6 +422,82 @@ def create_node(self, **kwargs):
...
+    def list_key_pairs(self):
+        result = self.connection.request(
+            'SoftLayer_Account', 'getSshKeys'
+        ).object
+        elems = [x for x in result]
+        key_pairs = self._to_key_pairs(elems=elems)
+        return key_pairs
+
+    def get_key_pair(self, name):
+        key_id = self._key_name_to_id(name)
```
Kami: Minor thing - for consistency and explicitness sake, please use keyword arguments (`self._key_name_to_id(name=name)`).

libcloud/compute/drivers/softlayer.py
```diff
@@ -411,6 +422,82 @@ def create_node(self, **kwargs):
...
+        ).object
+        return self._to_key_pair(result)
+
+    # TODO: Check this with the libcloud guys,
+    # can we create new dependencies?
+    def create_key_pair(self, name):
+        if crypto is False:
+            raise NotImplementedError("create_key_pair needs"
+                                      " the pycrypto library")
+        key = RSA.generate(2048)
```
Kami: We should probably also allow user to specify key size and default to `4096` bits just to be on the safe side.

For example:

``` python
def create_key_pair(self, name, ex_size=4096)
```

libcloud/compute/drivers/softlayer.py
```diff
@@ -411,6 +422,82 @@ def create_node(self, **kwargs):
...
+        result = self.connection.request(
+            'SoftLayer_Security_Ssh_Key', 'getObject', id=key_id
+        ).object
+        return self._to_key_pair(result)
+
+    # TODO: Check this with the libcloud guys,
+    # can we create new dependencies?
+    def create_key_pair(self, name):
+        if crypto is False:
+            raise NotImplementedError("create_key_pair needs"
```
Kami: Minor style thing - for consistency, please use single quotes around strings.

Kami:
There are some minor style things which need to be addressed, but besides that it looks good.

I will get it merged into trunk as soon as those comments are addressed.

Thanks!

Itxaka:
@Kami thanks for the comments, will try to fix them as soon as possible.

Regarding the sshkey stuff, I still don't know if getting the local generation for ssh keys before uploading its an ok thing and could be implemented across libcloud.

Thing is, some(all?) of the cloud providers allow you to import an ssh key, but for that we need a method to generate one, upload the public part and return to the user the keypair object (public and private) so I believe it would be good if it could be used across all drivers as a new generic method. For automation purpouses it makes too much sense (we are doing it that way internally, on node creation we generate a new key from scratch for each machine, but as libcloud does not support it we are doing it outside of it)

It requires a new dependecy thougth, pycrypto.

Cheers!

Itxaka:
@Kami Everything modified to confirm to the proper guidelines and all tests passing :)

Kami:
@Itxaka Thanks.

I'm having troubles applying this patch on top of trunk (I would squash the commits myself). Can you please sync your branch with latest trunk and squash all the commits?

In the worst case, if you are also having troubles, just sync your branch with trunk, do git diff --no-prefix trunk > t.diff, create new branch, git apply t.diff and commit all the changes under a single commit.

Itxaka:
Having issues, will recreate the branch and a new pull request.

PR #322, opened 2014-06-23
[LIBCLOUD-581] fix setting Name tags and name attributes when creating s...

zerthimon:
...ome objects

### DIFF
--- a/libcloud/compute/drivers/ec2.py
+++ b/libcloud/compute/drivers/ec2.py
@@ -3535,7 +3535,7 @@ class BaseEC2NodeDriver(NodeDriver):
 
         interface = self._to_interface(element, name)
 
-        if name and self.ex_create_tags(interface, {'Name': name})
+        if name and self.ex_create_tags(interface, {'Name': name}):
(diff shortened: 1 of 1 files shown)


libcloud/compute/drivers/ec2.py
```diff
@@ -2193,7 +2197,10 @@ def create_volume(self, size, name, location=None, snapshot=None,
         volume = self._to_volume(
             self.connection.request(self.path, params=params).object,
             name=name)
-        self.ex_create_tags(volume, {'Name': name})
+
+        if self.ex_create_tags(volume, {'Name': name}):
+            volume.extra['tags']['Name'] = name
```
Kami, line 2202: Have you tested this change?

Looking at the code, it doesn't seem like `_to_volume` adds `tags` attribute to the `extra` dictionary anywhere.

Kami, line 2202: Edit - never mind.

Just noticed your change a couple of lines bellow.

Kami:
Merged into trunk. Thanks.

PR #323, opened 2014-06-23
use _build_filters function instead of specifying request params manuall...

zerthimon:
...y

### DIFF
--- a/libcloud/compute/drivers/ec2.py
+++ b/libcloud/compute/drivers/ec2.py
@@ -3157,7 +3157,7 @@ class BaseEC2NodeDriver(NodeDriver):
 
         filters = {'region-name': self.region_name}
         if only_available:
-                  filters['state'] = 'available'
+            filters['state'] = 'available'
 
         params.update(self._build_filters(filters))
 
@@ -3196,9 +3196,10 @@ class BaseEC2NodeDriver(NodeDriver):
         """
         params = {'Action': 'DescribeTags'}
 
-        filters = {'resource-id': resource.id,
-                   'resource-type': 'instance'
-                  }
+        filters = {
+            'resource-id': resource.id,
+            'resource-type': 'instance'
+        }
 
         params.update(self._build_filters(filters))
 
--- a/libcloud/compute/drivers/ec2.py
+++ b/libcloud/compute/drivers/ec2.py
@@ -1998,10 +1998,9 @@ class BaseEC2NodeDriver(NodeDriver):
             'Action': 'DescribeVolumes',
         }
         if node:
-            params.update({
-                'Filter.1.Name': 'attachment.instance-id',
-                'Filter.1.Value': node.id,
-            })
+            filters = {'attachment.instance-id': node.id}
+            params.update(self._build_filters(filters))
+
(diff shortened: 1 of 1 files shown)


libcloud/compute/drivers/ec2.py
```diff
@@ -1998,10 +1998,9 @@ def list_volumes(self, node=None):
             'Action': 'DescribeVolumes',
         }
         if node:
-            params.update({
-                'Filter.1.Name': 'attachment.instance-id',
-                'Filter.1.Value': node.id,
-            })
+            filters = {'attachment.instance-id': node.id}
```
Kami, line 2001: Hm, with this approach, value attribute key will be set to `Filter.1.Value.1`, previously it was set to `Filter.1.Value`.

Are you sure this will still work? We could modify `_build_filters` to remove `.1` suffix if filter has only one value, but might break an existing code and needs to be tested.

libcloud/compute/drivers/ec2.py
```diff
@@ -3156,12 +3155,11 @@ def ex_list_availability_zones(self, only_available=True):
         """
         params = {'Action': 'DescribeAvailabilityZones'}
 
+        filters = {'region-name': self.region_name}
         if only_available:
-            params.update({'Filter.0.Name': 'state'})
-            params.update({'Filter.0.Value.0': 'available'})
```
Kami, line 3161: `_build_parameters` use 1-based indexing, but previously it used 0-based indexing. In any case, I think 1-based indexing is something we should be using anyway in this case, so this change should be fine.

libcloud/compute/drivers/ec2.py
zerthimon, line 2001: According to the EC2 API reference, the way filters are built is common for all resources.

<quote>
Filter.n.Name
The name of a filter. For more information about supported filter names, see the Supported Filters section.
</quote>

<quote>
Filter.n.Value.m
A value for the filter. For more information about supported values for each filter, see the Supported Filters section.
</quote>

from: http://docs.aws.amazon.com/AWSEC2/latest/APIReference/ApiReference-query-DescribeVolumes.html

So its weird that it even worked before, or did it work at all ?
This looks like a bug.

libcloud/compute/drivers/ec2.py
zerthimon, line 3161: Yep, probably 0-based indexing also works (?) with EC2 API. However, 1-based indexing is mentioned in all examples in EC2 API reference:

<quote>
https://ec2.amazonaws.com/?Action=DescribeVolumes
&Filter.1.Name=attachment.instance-id
&Filter.1.Value.1=i-1a2b3c4d
&Filter.2.Name=attachment.delete-on-termination
&Filter.2.Value.1=true
&AUTHPARAMS
</quote>

That's why I used 1-based indexes in _build_filters() function in the first place.

libcloud/compute/drivers/ec2.py
Kami, line 2001: Yeah, I was wondering the same thing. In any case, I think your changes are good.

Kami:
Merged into trunk. Thanks.

PR #324, opened 2014-06-24
use _get_boolean() function instead of parsing api response in every met...

zerthimon:
...hod


Kami:
Merged into trunk. Thanks.

PR #327, opened 2014-06-26
Add network_association_id to ex_list_public_ips and CloudstackAddress obj...

boul:
...ect

ex_list_networks should read ex_list_public_ips

With the network_association_id it is possible to relate and already acquired public ip to a network.
Without this information it is impossible to find out to which network a public ip belongs.


sebgoa:
Merged thanks for the patch. Make sure to extend the test cases to check for this parameter.

PR #329, opened 2014-06-27
[LIBCLOUD-588] Add ex_node_ids and ex_filters to list_volumes()

zerthimon:
Current list_volumes() has a "node" parameter, which builds a specific filter. I added a generic filtering while leaving "node" parameter for backward compatibility.
The "node's" attachment.instance-id parameter's value overrides the ex_filters one.

### DIFF
--- a/libcloud/compute/drivers/ec2.py
+++ b/libcloud/compute/drivers/ec2.py
@@ -2010,18 +2010,45 @@ class BaseEC2NodeDriver(NodeDriver):
                     )
         return locations
 
-    def list_volumes(self, node=None):
-        params = {
-            'Action': 'DescribeVolumes',
-        }
-        if node:
-            filters = {'attachment.instance-id': node.id}
-            params.update(self._build_filters(filters))
+    def list_volumes(self, node=None, ex_volume_ids=None, ex_filters=None):
+        """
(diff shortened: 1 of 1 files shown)


libcloud/compute/drivers/ec2.py
```diff
@@ -2010,18 +2010,45 @@ def list_locations(self):
...
+        if ex_volume_ids:
+            params.update(self._pathlist('VolumeId', ex_volume_ids))
+
+        if ex_filters and node:
+            ex_filters.update({'attachment.instance-id': node.id})
+        elif node:
+            ex_filters = {'attachment.instance-id': node.id}
+
+        if ex_filters:
+            params.update(self._build_filters(ex_filters))
```
allardhoeve, line 2045: You could write this as:

``` python
def list_volumes(self, node=None, ex_volume_ids=None, ex_filters=None):
    ex_filters = ex_filters or {}

    if ex_volume_ids:
         params.update(self._pathlist('VolumeId', ex_volume_ids))

    if node:
        ex_filters = {'attachment'...}

    if ex_filters:
        params.update(....)
```

allardhoeve:
Looks good to me.

PR #330, opened 2014-06-28
[LIBCLOUD-356] Add config_drive to support user_data

stickystyle:


libcloud/compute/drivers/openstack.py
```diff
@@ -1341,6 +1341,7 @@ def _create_args_to_params(self, node, **kwargs):
         if 'ex_userdata' in kwargs:
             server_params['user_data'] = base64.b64encode(
                 b(kwargs['ex_userdata'])).decode('ascii')
+            server_params['config_drive'] = True
```
Kami: I'm not actually sure if we want to always send `config_drive` attribute when `user_data` is provided.

I believe older versions and some installations of OpenStack support `user_data` without `config_driver`.

If that is in fact the case, we should also add `ex_config_drive` argument to the method (and this argument only makes sense if `ex_userdata` is provided as well).

/cc @briancurtin @alex

stickystyle: That seems pretty reasonable, I've honestly not worked with many older versions of OpenStack. I'll amend the pull request based on your suggestions.

stickystyle:
Going on your suggestion and looking through the convention in other drivers @Kami, would it be the correct style to raise a ValueError if the user set ex_config_drive to True and failed to pass in ex_userdata ?

Kami:
@stickystyle I think that's reasonable, although in cases like this, we usually leave validation like this to be performed by the server / API.

This way, we don't need to update the code if the business logic changes and passing one value without another becomes OK (which is unlikely in this case, but still).

libcloud/compute/drivers/openstack.py
Kami: Alright, let me know when you update the pull request and I'll get those changes merged into trunk.

stickystyle:
@Kami I've updated the pull request to have config_drive as a kwarg, and didn't include the ValueError exception as I agree with you that leaving it up to the API is probably a more maintainable solution from the client side.

Let me know if there are any other changes that need to be made, thanks!

libcloud/compute/drivers/openstack.py
```diff
@@ -1342,6 +1347,9 @@ def _create_args_to_params(self, node, **kwargs):
             server_params['user_data'] = base64.b64encode(
                 b(kwargs['ex_userdata'])).decode('ascii')
 
+        if 'ex_config_drive' in kwargs:
```
Kami, line 1350: Probably be better to do something like the code bellow because user could potentially pass in `ex_config_drive=False` to the method and we would still send `True`.

``` python
if 'ex_config_drive' in kwargs:
    server_params['config_drive'] = kwargs['ex_config_drive']
```

Kami:
Really sorry for the long delay (we don't get notifications when you push commits so I didn't notice that you updated the PR).

In any case, this change has finally been merged now. Thanks again.

PR #332, opened 2014-06-30
LIBCLOUD-590 - Reduce redundant API calls of CloudStack compute driver's...

atsaki:
CloudStack compute driver's list_nodes method executes too many API calls and it slows down this method.
The patch reduces redundant API calls and speed up list_nodes method.

https://issues.apache.org/jira/browse/LIBCLOUD-590

### DIFF
--- a/libcloud/compute/drivers/cloudstack.py
+++ b/libcloud/compute/drivers/cloudstack.py
@@ -1364,11 +1364,7 @@ class CloudStackNodeDriver(CloudStackDriverMixIn, NodeDriver):
         vms = self._sync_request('listVirtualMachines', params=args)
         addrs = self._sync_request('listPublicIpAddresses', params=args)
         port_forwarding_rules = self._sync_request('listPortForwardingRules')
-
-        try:
-            ip_forwarding_rules = self._sync_request('listIpForwardingRules')
-        except:
-            ip_forwarding_rules = {}
+        ip_forwarding_rules = self._sync_request('listIpForwardingRules')
 
         public_ips_map = {}
         for addr in addrs.get('publicipaddress', []):
--- a/libcloud/compute/drivers/cloudstack.py
+++ b/libcloud/compute/drivers/cloudstack.py
@@ -1363,6 +1363,12 @@ class CloudStackNodeDriver(CloudStackDriverMixIn, NodeDriver):
             args['projectid'] = project.id
         vms = self._sync_request('listVirtualMachines', params=args)
         addrs = self._sync_request('listPublicIpAddresses', params=args)
+        port_forwarding_rules = self._sync_request('listPortForwardingRules')
+
+        try:
+            ip_forwarding_rules = self._sync_request('listIpForwardingRules')
+        except:
+            ip_forwarding_rules = {}
 
         public_ips_map = {}
         for addr in addrs.get('publicipaddress', []):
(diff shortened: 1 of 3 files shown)


sebgoa:
thanks for your patches, I am a bit busy this coming days, but I will try to comment/merge as soon as possible.

libcloud/compute/drivers/cloudstack.py
```diff
@@ -654,6 +654,12 @@ def list_nodes(self, project=None):
             args['projectid'] = project.id
         vms = self._sync_request('listVirtualMachines', params=args)
         addrs = self._sync_request('listPublicIpAddresses', params=args)
+        port_forwarding_rules = self._sync_request('listPortForwardingRules')
+
+        try:
+            ip_forwarding_rules = self._sync_request('listIpForwardingRules')
+        except:
```
allardhoeve: This looks to broad to me. What does it catch? Why would you catch all the things and not fail here?

libcloud/compute/drivers/cloudstack.py
```diff
@@ -671,14 +677,14 @@ def list_nodes(self, project=None):
             public_ips = list(public_ips)
             node = self._to_node(data=vm, public_ips=public_ips)
 
-            addresses = public_ips_map.get(vm['id'], {}).items()
-            addresses = [CloudStackAddress(node, v, k) for k, v in addresses]
+            addresses = public_ips_map.get(str(vm['id']), {}).items()
```
allardhoeve: Why change this to a string?

libcloud/compute/drivers/cloudstack.py
```diff
@@ -654,6 +654,12 @@ def list_nodes(self, project=None):
             args['projectid'] = project.id
         vms = self._sync_request('listVirtualMachines', params=args)
         addrs = self._sync_request('listPublicIpAddresses', params=args)
+        port_forwarding_rules = self._sync_request('listPortForwardingRules')
```
allardhoeve, line 1366: You could dispense with a lot of the `.get(key, [])` complexity if you'd use a defaultdict here.

``` python
port_forwarding_rules = defaultdict(dict, **self._sync_request('listPortForwardingRules)
```

While a little more complex to set up here, this will make the rest of the code much easier to read.

allardhoeve:
This is a complex function that can be broken down into smaller bits to make it much more legible. That is not the scope of the PR though.

I _think_ this is fine, but we could actually use some more tests here. The lack of tests makes me hesitant to merge it. I would be much more comfortable if the `test_list_nodes` method contained at least an entry that has the correct IPs. I would like to see these added to the tests with their expected contents:
1. node.extra['ip_addresses']
2. node.extra['ip_forwarding_rules']
3. node.extra['port_forwarding_rules']

sebgoa:
@atsaki do you plan to work on this ? @allardhoeve made some good comments, could you answer ?thanks

allardhoeve:
My comments still stand ofcourse.

atsaki:
I'm working on this. Please wait for a few days.

atsaki:
I updated the code to follow the comments.
Can you check them?

allardhoeve:
Thanks.

This is better than before.

I have no access to CloudStack, so I can't test this and I also don't feel qualified to plus one it.

sebgoa:
@atsaki this was closed by mistake as we are trying to close some old PRs. Can you reopen again, I will merge this week.

atsaki:
@runseb I reopened https://github.com/apache/libcloud/pull/641.

PR #334, opened 2014-07-02
[LIBCLOUD-592] Create a generic method for listing AWS EIPs

zerthimon:
Currently there are three methods in ec2 driver for listing EIPs:
They all satisfy different use cases:
ex_describe_all_addresses() - list all EIPs (vpc and ec2) optionally list only associated EIPs.
ex_describe_addresses() - list only EIPs for the specific nodes.
ex_describe_addresses_for_node() - list only EIPs for specific node.

This PR creates a generic method ex_list_addresses() to give users more flexibility while listing EIPs, by implementing all the functionality from this API call: http://docs.aws.amazon.com/AWSEC2/latest/APIReference/ApiReference-query-DescribeAddresses.html

### DIFF
--- a/libcloud/compute/drivers/ec2.py
+++ b/libcloud/compute/drivers/ec2.py
@@ -3322,6 +3322,42 @@ class BaseEC2NodeDriver(NodeDriver):
         response = self.connection.request(self.path, params=params).object
         return self._get_boolean(response)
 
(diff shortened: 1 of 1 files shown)


zerthimon:
Similarly like with other EC2 resources which only have one generic "describe/list" method, should other EIP listing methods be deprecated ?

PR #335, opened 2014-07-02
allow specifying standard as a domain in ex_associate_address_with_node

zerthimon:
AWS uses "domain" field to distinguish between EIPs allocated for Classic EC2 ('standard') and EIPs allocated for VPC (value 'vpc')

domain
    Indicates whether this Elastic IP address is for use with instances in EC2-Classic (standard) or instances in a VPC (vpc).
    Type: xsd:string
    Valid values: standard | vpc

http://docs.aws.amazon.com/AWSEC2/latest/APIReference/ApiReference-ItemType-DescribeAddressesResponseItemType.html

The method ex_allocate_address() also uses 'standard' and 'vpc' as allowed parameters.

This patch adds 'standard' as a value to domain parameter.
None is left for backward compatibility.

### DIFF
--- a/libcloud/compute/drivers/ec2.py
+++ b/libcloud/compute/drivers/ec2.py
@@ -3352,7 +3352,7 @@ class BaseEC2NodeDriver(NodeDriver):
         :param      elastic_ip: Elastic IP instance
         :type       elastic_ip: :class:`ElasticIP`
 
-        :param      domain: The domain where the IP resides (vpc only)
+        :param      domain: The domain where the IP resides (standard/vpc)
         :type       domain: ``str``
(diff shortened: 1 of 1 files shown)


zerthimon:
Something went wrong with the travis tests. Is there a way to tell it to recheck ?

allardhoeve:
Yes, you can push an empty commit to the branch by using `git commit --allow-empty`. Ugly trick, I know, but it can be done. Make sure to merge trunk into this branch, there were loads of CI fixes lately.

allardhoeve:
Looks good to me, but get the tests to work first, please.

PR #337, opened 2014-07-06
[LIBCLOUD-594] Add firewall rule support for CloudStack driver

atsaki:
CloudStack driver's doesn't support firewall rule now.
This update add following methods.
- ex_list_firewall_rules
- ex_create_firewall_rule
- ex_delete_firewall_rule

https://issues.apache.org/jira/browse/LIBCLOUD-594


libcloud/compute/drivers/cloudstack.py
```diff
@@ -367,6 +367,60 @@ def __eq__(self, other):
...
+    """
+
+    def __init__(self, id, address, cidr_list, protocol,
+                 icmp_code=None, icmp_type=None,
+                 start_port=None, end_port=None):
+
+        """
+        A Firewall rule.
+
+        @note: This is a non-standard extension API, and only works for EC2.
```
sebgoa: ..only works for CloudStack ? might want to edit that line

atsaki: Thanks. I fixed it.

PR #343, opened 2014-07-30
[LIBCLOUD-601] Add check windows os in is_valid_ip_address to run inet_aton

ashangit:


libcloud/utils/networking.py
```diff
@@ -73,7 +74,10 @@ def is_valid_ip_address(address, family=socket.AF_INET):
     :return: ``bool`` True if the provided address is valid.
     """
     try:
-        socket.inet_pton(family, address)
+        if (platform.system() == 'Windows'):
+            socket.inet_aton(address)
```
Kami, line 78: I guess there is still an edge case with IPv6 which `inet_aton` doesn't support, but `inet_pton` does.

Maybe we should throw if running on Windows and user specifies `family=socket.AF_INET6`.

ashangit, line 78: In fact there is an issue with IPv6 when using inet_aton.
So we could throw an exception or use an external library dnspython wihich provide inet_aton for ipv4 and ipv6 and for windows also.
Let me know if adding such library is something acceptable or I will just add the throw.

Kami:
Sorry for the delay. I've added a check for validating IPv6 address on Windows (e105433e941262e03eb3bb922db47dabdd8efa75) and merged changes into trunk.

Thanks.

PR #345, opened 2014-08-05
[W.I.P.] [RFC] New base libcloud.networking API

Kami:
Note: This branch is still very much W.I.P. so please wait with the comments. I've only opened it so it's easier to link to the whole diff.

You also shouldn't based any of your changes of this branch since I'm constantly updating it and doing force pushes.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,0 +1,34 @@
+[MASTER]
+# Add <file or directory> to the black list. It should be a base name, not a
+# path. You may set this option multiple times.
+ignore=
+
+
+# Pickle collected data for later comparisons.
+persistent=yes
+
+# List of plugins (as comma separated values of python modules names) to load,
+# usually to register additional checkers.
+load-plugins=
+
+
+[MESSAGES CONTROL]
+disable=redefined-builtin,too-many-arguments,too-few-public-methods,missing-docstring,invalid-name,abstract-method,no-self-use
+
+# checks for
+# * unused variables / imports
+# * undefined variables
+# * redefinition of variable from builtins or from an outer scope
+# * use of variable before assigment
+#
+[VARIABLES]
+
+# Tells wether we should check for unused import in __init__ files.
+init-import=no
+
+# A regular expression matching names used for dummy variables (i.e. not used).
+dummy-variables-rgx=_|dummy
+
+# List of additional names supposed to be defined in builtins. Remember that
+# you should avoid to define new builtins when possible.
+additional-builtins=
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -44,6 +44,15 @@ Compute
   Reported by Chris DeRamus.
   [Tomaz Muraus]
 
+Networking
+~~~~~~~~~~
+
+- Add initial implementation of a base networking API with a reference drivers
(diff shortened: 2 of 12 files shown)


carletes:
A minor suggestion: What about `NetworkingDriver.attach_floating_ip()` (and `NetworkingDriver.detach_floating_ip()`) instead of `attach_floating_ip_to_node()`?

For symmetry with `NodeDriver.attache_volume()` and `NodeDriver.detach_volume()`, mainly --- and since it's quite clear that we're attaching the floating IP to a node, I would say.

Kami:
@carletes In OpenStack Quantum networking, there is a concept of a `Port` (http://docs.openstack.org/api/openstack-network/2.0/content/Overview-d1e71.html#Concepts-d1e369) and floating IP can be attached to a port. Usually a node is then attached to a port, but not necessary.

Those are utility / convenience methods and at the moment, they are only implemented for OpenStack Nova networking where it only involves a single API call, but in the future, I also plan to implement it in the Quantum driver where it will involve multiple API calls (create floating ip, attach it to the port, attach node to the port).

There are multiple reasons why I want to include "to_node" / "from_node" in the name:
1. Being explicit - in this case method is available on the `NetworkingDriver` and not `NodeDriver` (like in the volume case), so calling it just `attach_floating_ip` would be kinda ambiguous.
2. Reducing confusion between a port and a node / instance

carletes:
@Kami Thanks for the answer regarding the ports!

By the way: I'm using https://pypi.python.org/pypi/ipaddr for address operations in my `libcloud-vagrant` package and I'm very happy with it. It could simplify a bit your code in `OpenStackSubnet.AllocationPool`, I think

Kami:
@carletes Yeah, that's slightly off topic, but we currently only really support IPv4. The reason for that is that Python's stdlib doesn't provide a cross-platform approach for handling IPv4.

A while back I was researching existing packages for handling that, but I think there were some Python version compatibility issues (we support 2.5 - 3.x so the package needs to support the same version as well).

Kami:
@carletes Also, in the past, our policy was to avoid 3rd party dependencies because of the following reasons;
- Python version compatibility issues
- Quality / lack of tests
- Python packaging ecosystem was a bit of a mess (setuptools, distutils, etc.). This got better over the years though.

Python's packaging ecosystem is not as good as Node.js one for example, but I still think we should revise this policy. I think it is fine to add a dependency if it meets those things:
- It's compatible with all the Python versions we support
- It has extensive test coverage
- It's an utility package

hutson:
@Kami is there any chance this change could salvage work from [LIBCLOUD-122](https://issues.apache.org/jira/browse/LIBCLOUD-122)?

Kami:
@hbetts Will have a look when I get a chance.

Overall, I need to spend some more polishing and finishing up this branch (add EC2 driver and tests). After that, we can hopefully get more discussion going and get it merged into trunk in some reasonable time frame.

tonybaloney:
@Kami I would consider adding security groups to this specification. all providers support some sort of sec grouping policy.

http://developer.openstack.org/api-ref-networking-v2-ext.html#listSecGroups

Kami:
Old and out of date PR. Closing.

PR #348, opened 2014-08-15
[LIBCLOUD-544] Fix the regression caused by the original implementation ...

raphtheb:
...of the bugfix. We now perform a simple sanity check to know whether the metadata dictionary has already been formatted to GCE's specs.


raphtheb:
Cancelling this, found one line i should have deleted before submitting

PR #349, opened 2014-08-15
Libcloud 544 gce metadata fix

raphtheb:
This is a slightly improved version of the fix i submitted minutes ago which had some trace-generating code in it. Removed my own silly tests.


libcloud/compute/drivers/gce.py
```diff
@@ -1211,9 +1211,16 @@ def create_node(self, name, size, image, location=None,
                                               use_existing=use_existing_disk,
                                               ex_disk_type=ex_disk_type)
 
-        if ex_metadata is not None:
-            ex_metadata = {"items": [{"key": k, "value": v}
-                                     for k, v in ex_metadata.items()]}
+        if not ex_metadata or not isinstance(ex_metadata, dict):
+            # Should output a relevant error message if non-dict is received?
+            ex_metadata = None
```
Kami, line 1215: We should throw `ValueError` if user passes in an invalid value (not a dictionary) instead of silently ignoring it.

libcloud/compute/drivers/gce.py
```diff
@@ -1211,9 +1211,16 @@ def create_node(self, name, size, image, location=None,
...
+        else:
+            items = []
+            for k, v in ex_metadata.items():
+                items.append({'key': k, 'value': v})
+            if 'items' not in ex_metadata:
```
Kami, line 1219: I think we need a better solution of handling this - we should probably throw if `ex_metadata` dictionary already contains a key named `items` instead of just overwriting this key.

Kami, line 1219: Actually, why can't we do something like this:

``` python
metadata = {'items': []}
for key, value in ex_metadata.items():
    metadata['items'].append({'key': key, 'value': value})
```

This way, `ex_metadata` argument can still contain key named `items`...

raphtheb, line 1219: I've already added the ValueError, that's good and i have it staged locally.

I do believe we can't do the aforementioned code as that would re-introduce the issue i'm attempting to fix.This is the output of that code that gets sent to GCE while running it:

```
{'items': [{'value': [{'value': '1', 'key': 'items'}, {'value': 'two', 'key': '2'}, {'value': 'testbox', 'key': 'salt-cloud-profile'}], 'key': 'items'}]}
```

This is what salt sends to libcloud:

```
{'items': [{'value': '1', 'key': 'items'}, {'value': 'two', 'key': '2'}, {'value': 'testbox', 'key': 'salt-cloud-profile'}]}
```

I'll have a closer, fresher look on Monday and find a better way to handle this. 

Cheers and have a great weekend!

erjohnso:
@raphtheb - thanks for this.  I've just tested it with SaltStack and this seems to work just fine.  I'm guessing this is good to merge, but if I know @Kami, he'll ask you to squash your commits before merging. :)

raphtheb:
Thanks for havign a look, @erjohnso 

For clarity, i am opening a new pull-request (https://github.com/apache/libcloud/pull/353) with only one clean, squashed commit from an alternate branch. The end diff is the same, but will allow for a cleaner history. This one can safely be ignored/destroyed/rejected.

PR #351, opened 2014-08-25
[LIBCLOUD-427] update linode plans mapping

bsdlp:


Kami:
Merged into trunk. Thanks.

P.S. Nice chicken in (BBQ?) sauce :P

bsdlp:
the best duck poutine i've ever had in montreal during pycon 2014 :smile:

Kami:
Ah, I was totally off.

Anyway, sounds nice. I had a duck confit in a sour-cherry sauce a month or so ago at a local restaurant here in Ljubljana. It was really, really good, the only downside was that the portion was very small and it was very expensive.

PR #353, opened 2014-08-29
[LIBCLOUD-544] Resolve a bug introduced by the original fix of the issue...

raphtheb:
This is meant to replace github.com/apache/libcloud/pull/349
With one clean commit.

1- Add separated verifications for "None" and malformed metadata.
   A clear error will be raised upon malformed metadata.
   If undefined or already None, we assume None.
2- Provide comments explaining the odd GCE dictionary format.
3- Perform check on metadata dictionary structure. We expect one "items" key
   and a tuple of arbitrary values.
4- Prefix the keys/values provided by a simple dictionary by "items" if not
   found, enforcing point 3's structure.

### DIFF
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -1211,9 +1211,23 @@ class GCENodeDriver(NodeDriver):
                                               use_existing=use_existing_disk,
                                               ex_disk_type=ex_disk_type)
 
-        if ex_metadata is not None:
-            ex_metadata = {"items": [{"key": k, "value": v}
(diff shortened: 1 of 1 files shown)


erjohnso:
:+1: I tested this under the old #349 PR and this looks good to merge to me.

Kami:
Looks good. Thanks.

I've added tests and went ahead and merged patch into trunk.

anglerud:
It would be great if this change could be added for ex_create_multiple_nodes' ex_metadata parameter as well.

PR #354, opened 2014-08-31
Adds ssh key management to Softlayer compute driver

Itxaka:
Adds the following methods for the Softlayer driver:
- list_key_pairs()
- get_key_pair()
- import_key_pair_from_string()
- delete_key_pair()
- create_key_pair() (only if Pycripto is installed)

Adds tests for all new methods.
Modifies the test_create_node() to use an ssh key.
Adds a new property on the Softlayer node "ex_keyname"which is the name of the key to be associated to the new node.


Itxaka:
@Kami Here it is :)

Kami:
Merged into trunk. Thanks.

PR #355, opened 2014-09-03
[LIBCLOUD-609] Added image_id and size_id fields when calling DigitalOce...

sherifzain:
https://issues.apache.org/jira/browse/LIBCLOUD-609


Kami:
Merged into trunk. Thanks.

PR #356, opened 2014-09-09
[LIBCLOUD-612] Customize HTTP vendor prefix for Google Storage

sfriesel:
https://issues.apache.org/jira/browse/LIBCLOUD-612

Google storage uses/expects custom headers like meta data
and storage class to be prefixed with "x-goog" instead of
"x-amz". This enables use of object tagging in the Google
Storage provider.

### DIFF
--- a/libcloud/storage/drivers/google_storage.py
+++ b/libcloud/storage/drivers/google_storage.py
@@ -134,3 +134,4 @@ class GoogleStorageDriver(BaseS3StorageDriver):
     namespace = NAMESPACE
     supports_chunked_encoding = False
     supports_s3_multipart_upload = False
+    http_vendor_prefix = 'x-goog'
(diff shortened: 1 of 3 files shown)


sfriesel:
Unittests passed on 2.7 and I successfully uploaded an object with metadata and retrieved them again.

```
>>> bucket.upload_object("testfile", 'testfiles/foobar', extra={'meta_data':{'mytag': 'myvalue'}})
>>> bucket.get_object('testfiles/foobar').meta_data
{'mytag': 'myvalue'}
```

sfriesel:
@Kami Is anything missing? I could provide the whole debug log of the interaction or maybe some credentials for testing if needed

Kami:
@sfriesel I'll have a look asap.

Kami:
I need to spend some more time testing this myself. The reason for this is that we are using S3 compatible API which requires you to send `x-amzn` headers.

From your pull request, it seems like you can use `x-goog` headers even if you use the S3 compatible API, but I need to confirm this myself.

Kami:
I just tested your changes and verified it works correctly.

While testing the changes I also noticed meta_data is only returned when retrieving information for a single object (`get_object`), but not when listing objects (`list_container_objects`).

I will update documentation and add a note about this known limitation.

Kami:
Merged into trunk. Thanks!

PR #364, opened 2014-09-23
Cloudstackvpc

jeroendekorte:


boul:
This commit contains extra functionality for the cloudstack driver with respect to VPC support.
Credits go to my colleague: Jeroen de Korte.

vpc: list create delete
egress_fw_rules: list create delete
vpcofferings: list
- necessary tests

sebgoa:
This should get closed automatically, but in case it does not, please close it.
https://github.com/apache/libcloud/commit/d608085e3da79240b2b56709ef6adb58bd9c9dab

thanks for the patch

PR #365, opened 2014-09-25
[LIBCLOUD-618] Make openstack create_volume return a StorageVolume

gertjanol:
The create_volume call in the OpenStack driver returns a bool to signify whether the call succeeded. This call should return a StorageVolume object, as other drivers already do (ec2, cloudstack, gce, etc).


Kami:
Good catch.

Merged into trunk. Thanks.

gertjanol:
Thanks!

PR #366, opened 2014-09-25
[LIBCLOUD-619] Add extra data from openstack volume to object

gertjanol:
OpenStack Volumes have a state, but libcloud has no way to retrieve it. This PR adds all available data from the Openstack volume to the `extra`-dict on the `StorageVolume`.


Kami:
I modified the code to be more robust and use `.get` since I'm not sure if those attributes are always present (e.g. different OpenStack versions) and merged your changes into trunk.

Thanks.

gertjanol:
Thanks!

PR #368, opened 2014-10-03
CLOUDSTACK: option to start VM in a STOPPED state

boul:
extra option on create_node so a vm can be created in a STOPPED state, by default Cloudstack starts VM's after creation.
Plus tests.


sebgoa:
looks good to me, thanks for the tests.

PR #369, opened 2014-10-03
CLOUDSTACK: ex_list_nics, ex_attach_nic_to_vm, ex_detach_nic_from_vm + test...

boul:
Functionality to list NICS and to add/remove them too.
And some extra attribute maps with respect to nics.
Including unit tests
Some cosmetic (indent) change slipped in this commit as well.

### DIFF
--- a/libcloud/compute/drivers/cloudstack.py
+++ b/libcloud/compute/drivers/cloudstack.py
@@ -181,6 +181,10 @@ RESOURCE_EXTRA_ATTRIBUTES_MAP = {
         'project_id': {
             'key_name': 'projectid',
             'transform_func': str
+        },
+        'nics:': {
+            'key_name': 'nic',
+            'transform_func': list
         }
     },
     'volume': {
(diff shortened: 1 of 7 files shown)


boul:
@runseb let me know if you want to get rid of those indent changes which slipped through, most probably because I switched to pycharm vs pydev.
When do you expect a new release with this functionality added?

sebgoa:
Hi @boul yes it would be nice to revert those indent changes. Next release will probably be in couple weeks

boul:
Reverted the slipped indents

libcloud/compute/drivers/cloudstack.py
```diff
@@ -2712,6 +2751,88 @@ def ex_list_os_types(self):
...
+                gateway=item['gateway'],
+                ip_address=item['ipaddress'],
+                is_default=item['isdefault'],
+                mac_address=item['macaddress'],
+                driver=self,
+                extra=extra))
+
+        return nics
+
+    def ex_add_nic_to_node(self, node, network, ip_address=None):
```
Kami: Maybe better and more consistent names would be attach and detach (attach_nic_to_node, detach_nic_from_node).

boul: good suggestion, would you prefer to name the method attach_nic_to_node or ex_attach_nic_to_node, since it's a specific extension?

Kami: Yeah, sorry, those should include `ex_` prefix since they are extension methods.

boul:
Allright, methods have been renamed (attach/detach vs add/remove)

boul:
@kami @runseb: Is there still something we need to fix before it can be merged into trunk? thx

PR #370, opened 2014-10-09
[LIBCLOUD-622] Support for config_drive in the Openstack driver

rnirmal:
Adds a 'ex_config_drive' option to create_node and ex_rebuild_node
for the Openstack driver.

### DIFF
--- a/libcloud/compute/drivers/openstack.py
+++ b/libcloud/compute/drivers/openstack.py
@@ -1175,6 +1175,10 @@ class OpenStack_1_1_NodeDriver(OpenStackNodeDriver):
                                     Can be either ``AUTO`` or ``MANUAL``.
         :type       ex_disk_config: ``str``
 
+        :keyword    ex_config_drive: If True enables metadata injection in a
+                                     server through a configuration drive.
+        :type       ex_config_drive: ``bool``
+
         :keyword    ex_admin_pass: The root password for the node
         :type       ex_admin_pass: ``str``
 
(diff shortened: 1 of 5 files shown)


libcloud/test/compute/fixtures/openstack_v1.1/_servers_26f7fbee_8ce1_4c28_887a_bfe8e4bb10fe.json
```diff
@@ -40,7 +40,7 @@
                 }
             ]
         },
-        "config_drive": "",
+        "config_drive": "True",
```
Kami, line 43: Hm, so a value is a string and not a boolean?

libcloud/compute/drivers/openstack.py
```diff
@@ -1358,6 +1366,11 @@ def ex_rebuild(self, node, image, **kwargs):
                                     Can be either ``AUTO`` or ``MANUAL``.
         :type       ex_disk_config: ``str``
 
+        :keyword    ex_config_drive: Enables metadata injection in a server
```
Kami: Even if the OpenStack API takes a string and not a boolean, Libcloud API should take a boolean and this should get converted to the OpenStack format inside the method.

Kami:
Thanks.

There is a minor issue with a string / boolean in a public API, but besides that it looks good.

Kami:
It would also be nice if you could add some documentation on how to use `ex_config_drive` and `ex_userdata` :)

libcloud/test/compute/fixtures/openstack_v1.1/_servers_26f7fbee_8ce1_4c28_887a_bfe8e4bb10fe.json
rnirmal, line 43: Ah gotcha, yeah I was just mimicking what the api had, I'll get that changed to a boolean.

rnirmal:
The value is still a string on the API side, but I've set it up as a boolean on the client side. I've also added an example to show how to use it.

Kami:
Merged into trunk. Thanks.

PR #371, opened 2014-10-09
CLOUDSTACK: Implemented VPC network acls and tests

boul:
VPC Support was already added in a previous commit, this commit also introduces support for ACLList on VPC's. Including unit tests. Thanks @jeroendekorte


libcloud/compute/drivers/cloudstack.py
```diff
@@ -671,11 +772,10 @@ def __init__(self, display_text, name, vpc_offering_id, id, zone_id, cidr,
         self.extra = extra or {}
 
     def __repr__(self):
-        return (('<CloudStackVPC: id=%s, displaytext=%s, name=%s, '
-                 'vpc_offering_id=%s, zoneid=%s, cidr=%s, driver%s>')
-                % (self.id, self.displaytext, self.name,
-                   self.vpc_offering_id, self.zoneid, self.cidr,
-                   self.driver.name))
+        return (('<CloudStackVPC: id=%s, name=%s, '
```
sebgoa: Why did you remove the zone and display text ?

libcloud/compute/drivers/cloudstack.py
```diff
@@ -671,10 +772,11 @@ def __init__(self, display_text, name, vpc_offering_id, id, zone_id, cidr,
         self.extra = extra or {}
 
     def __repr__(self):
-        return (('<CloudStackVPC: id=%s, displaytext=%s, name=%s, '
-                 'vpc_offering_id=%s, zoneid=%s, cidr=%s, driver%s>')
-                % (self.id, self.displaytext, self.name,
-                   self.vpc_offering_id, self.zoneid, self.cidr,
```
sebgoa, line 677: @boul  @jeroendekorte Why do you change the order of the variables in the class ? does it matter ? also the driver value is really the name of the driver. I haven't checked the other classes but we could return the full driver object.

libcloud/compute/drivers/cloudstack.py
```diff
@@ -592,6 +604,95 @@ def __eq__(self, other):
...
+        :type       description: ``str``
+
+        :rtype: :class:`CloudStackNetworkACLList`
+        """
+
+        self.id = acl_id
+        self.name = name
+        self.vpc_id = vpc_id
+        self.description = description
+
```
sebgoa, line 639: @boul @jeroendekorte so this class does not return anything ?

boul, line 639: @jeroendekorte can you take a look at the comments from @runseb. Cheers!.

libcloud/compute/drivers/cloudstack.py
jeroendekorte, line 677: @runseb I've checked the code for the driver comment and I always pass the full driver object as a parameter.

libcloud/compute/drivers/cloudstack.py
jeroendekorte, line 639: @runseb Correct it does not return anything, is this something we should do? I use this class only as a data structure. Please correct me if I'm wrong.

boul:
@runseb I fixed the _repr_ ordering as you suggested, to match the object order. I kept driver.name as this seems common in the project, see EC2NodeLocation for example. And i'm not sure what will break if  I change this.

class EC2NodeLocation(NodeLocation):
    def **init**(self, id, name, country, driver, availability_zone):
        super(EC2NodeLocation, self).**init**(id, name, country, driver)
        self.availability_zone = availability_zone

```
def __repr__(self):
    return (('<EC2NodeLocation: id=%s, name=%s, country=%s, '
             'availability_zone=%s driver=%s>')
            % (self.id, self.name, self.country,
               self.availability_zone, self.driver.name))
```

libcloud/compute/drivers/cloudstack.py
boul, line 677: fixed ordering see latest commit

PR #373, opened 2014-10-11
adds ex_security_group_ids to ec2 in order to be able to launch nodes with security groups on a VPC

Itxaka:
Due to amazon having a different param for the security groups on an instance launched on a VPC and libcloud already supports VPCs (ex_subnet) this is the piece is missing to support the security groups on a VPC node
- adds a ex_security_group_ids kwarg which is a list of security group ids
- checks for the existence of ex_subnet if ex_security_group_ids is called, as it depends on it to make sense
- adds a test to see if the params are being constructed correctly

EC2 API page: http://docs.aws.amazon.com/AWSEC2/latest/APIReference/ApiReference-query-RunInstances.html

```
SecurityGroupId.n
One or more security group IDs. You can create a security group using CreateSecurityGroup.
Type: String
Default: Amazon EC2 uses the default security group.
Required: No
```

```
SecurityGroup.n
[EC2-Classic, default VPC] One or more security group names. For a nondefault VPC, you must use SecurityGroupId.n.
Type: String
Default: Amazon EC2 uses the default security group.
Required: No
```

### DIFF
--- a/libcloud/compute/drivers/ec2.py
+++ b/libcloud/compute/drivers/ec2.py
@@ -2135,6 +2135,10 @@ class BaseEC2NodeDriver(NodeDriver):
                                         assign to the node.
         :type       ex_security_groups:   ``list``
 
+        :keyword    ex_security_group_ids: A list of ids of security groups to
+                                        assign to the node.[for VPC nodes only]
+        :type       ex_security_group_ids:   ``list``
+
         :keyword    ex_metadata: Key/Value metadata to associate with a node
         :type       ex_metadata: ``dict``
 
@@ -2186,6 +2190,20 @@ class BaseEC2NodeDriver(NodeDriver):
             for sig in range(len(security_groups)):
                 params['SecurityGroup.%d' % (sig + 1,)] =\
                     security_groups[sig]
+
+        if 'ex_security_group_ids' in kwargs and not 'ex_subnet' in kwargs:
+            raise ValueError('You can only supply ex_security_group_id'
+                             ' combinated with ex_subnet')
+
+        security_group_id = kwargs.get('ex_security_group_ids', None)
+
+        if security_group_id:
+            if not isinstance(security_group_id, (tuple, list)):
+                security_group_id = [security_group_id]
+
+            for sig in range(len(security_group_id)):
+                params['SecurityGroupId.%d' % (sig + 1,)] =\
+                    security_group_id[sig]
 
         if 'location' in kwargs:
             availability_zone = getattr(kwargs['location'],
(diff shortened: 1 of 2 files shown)


Itxaka:
@Kami :) :shipit:

libcloud/compute/drivers/ec2.py
```diff
@@ -2187,6 +2191,20 @@ def create_node(self, **kwargs):
                 params['SecurityGroup.%d' % (sig + 1,)] =\
                     security_groups[sig]
 
+        if 'ex_security_group_ids' in kwargs and 'ex_subnet' not in kwargs:
+            raise ValueError('You can only supply ex_security_group_id'
+                             ' combinated with ex_subnet')
+
+        security_group_id = kwargs.get('ex_security_group_ids', None)
```
Kami: This variable stores a list so it should be called `security_group_ids`.

libcloud/test/compute/test_ec2.py
```diff
@@ -871,6 +872,22 @@ def test_create_node_ex_security_groups(self):
                           ex_securitygroup=security_groups,
                           ex_security_groups=security_groups)
 
+    def test_create_node_ex_security_group_id(self):
```
Kami: Also needs a test case for edge case (when user specifies `ex_security_group_ids`, but not `ex_subnet`).

libcloud/compute/drivers/ec2.py
```diff
@@ -2135,6 +2135,10 @@ def create_node(self, **kwargs):
                                         assign to the node.
         :type       ex_security_groups:   ``list``
 
+        :keyword    ex_security_group_ids: A list of ids of security groups to
```
Kami, line 2138: Edit: Ignore this comment - pr description already addresses it.

It's just unfortunate that existing argument is called `ex_security_groups` instead of `ex_security_group_names`.

libcloud/test/compute/test_ec2.py
Itxaka: I tried but could not catch the Exception with self.assertRaises, it actually raised the Exception. Will look into this.

Kami: That's probably because you were actually calling the method yourself and not just passing a callable to `assertRaises` (you should only pass callable + arguments to assertRaises and let assetRaises call the callable).

For example:

``` python
self.assertRaises(ValueError, myfunc(blah, foo=bar))
```

Not good because you are calling the function yourself.

For example:

``` python
self.assertRaises(ValueError, myfunc, blah, foo=bar)
```

Good - you are letting assertRaises call the function and intercept the exceptions.

Itxaka:
@Kami ?

Kami:
@Itxaka Sorry for the delay.

I will have a look later today and hopefully also get it merged then.

Kami:
LGTM. Patch has been merged into trunk.

Thanks.

PR #374, opened 2014-10-16
Use backports.ssl_match_hostname, instead of having our own logic

alex:


Kami:
Changes look good to me.

I need to some more testing and if everything works OK, I will go ahead and merge it into trunk as soon as 0.16.0 release is out.

PR #377, opened 2014-10-20
CLOUDSTACK: Minor fix for start-vm option

boul:
Is not None vs is not False, as suggested.


Kami:
Merged into trunk. Thanks.

PR #379, opened 2014-10-27
GCE: fix GCE internal auth

erjohnso:
@Kami - Very sorry, but I think with the merge trouble we were having on LIBCLOUD-625[1], I must have goofed on one of conflicts.  This minor change fixes LIBCLOUD-625.

[1] https://github.com/apache/libcloud/commit/20d977075117f05a0d8cd8ceb91c4dfcd93a7766

### DIFF
--- a/libcloud/common/google.py
+++ b/libcloud/common/google.py
@@ -111,7 +111,7 @@ def _get_gce_metadata(path=''):
     try:
         url = "http://metadata/computeMetadata/v1/" + path.lstrip('/')
         headers = {'Metadata-Flavor': 'Google'}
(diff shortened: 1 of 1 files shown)


Kami:
@erjohnso No problem and it was probably my fault - I've added a `method` argument to the `get_response_object` function shortly after telling you that you should use this method, so you probably haven't noticed my change.

If nothing else, this is also a good example why defaulting to keyword arguments is usually a good idea :)

Anyway, this has been merged into master. Thanks.

erjohnso:
Thanks @Kami !

PR #380, opened 2014-10-31
Correcting parameter names

gomezgoiri:
The section "I get Could not find specified endpoint error" contained incorrect parameter names: "ex_service__" instead of "ex_force_service__".


Kami:
Good catch. Patch merged into trunk - thanks!

PR #382, opened 2014-11-07
CLOUDSTACK: Add option to expunge vm on destroy

boul:


libcloud/compute/drivers/cloudstack.py
```diff
@@ -1034,15 +1034,27 @@ def _create_args_to_params(self, node, **kwargs):
...
         return server_params
 
-    def destroy_node(self, node):
+    def destroy_node(self, node, ex_expunge=None):
         """
         @inherits: :class:`NodeDriver.reboot_node`
         :type node: :class:`CloudStackNode`
 
+        :keyword    ex_expunge: If true is passed, the vm is expunged
+                                immediately. False by default.
```
sebgoa, line 1043: might be better to set it to False by default then, and check if it's True before passing it.

Kami, line 1043: Yeah, the docstring even says it defaults to `False`.

boul:
Check, changed the default to be False and only passing the arg when True as Cloudstack defaults to False so there is no need to send False to Cloudstack, this way it should also work with older CS versions which do not expose this feature.

libcloud/compute/drivers/cloudstack.py
```diff
@@ -1034,15 +1034,27 @@ def _create_args_to_params(self, node, **kwargs):
...
+        :type       ex_expunge: ``bool``
+
         :rtype: ``bool``
         """
+
+        args = {
+            'id': node.id,
+        }
+
+        if ex_expunge is not False:
```
sebgoa: even better with:

if ex_expunge:

boul: Less is more :) Done.

PR #385, opened 2014-11-10
[LIBCLOUD-632] Added images from ubuntu-os-cloud image project

dagi3d:
https://issues.apache.org/jira/browse/LIBCLOUD-632

### DIFF
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -2548,7 +2548,8 @@ class GCENodeDriver(NodeDriver):
                 image = self._match_images('coreos-cloud', partial_name)
             elif partial_name.startswith('opensuse'):
                 image = self._match_images('opensuse-cloud', partial_name)
-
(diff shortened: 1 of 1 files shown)


Kami:
LGTM

/cc @erjohnso

erjohnso:
LGTM too.  @Kami - any chance we can get this merged for 0.16.0?

Kami:
@erjohnso Sadly it's too late for 0.16.0 since the voting thread has already been started.

If it's a critical fix, we can just start a voting thread for 0.16.1 release once the 0.16.0 is out.

Kami:
Merged into trunk. Thanks.

erjohnso:
@Kami - no worries. I wouldn't call it 'critical', but I do suspect users are anxious to begin using Ubuntu on GCE.  I'll look into a patch to allow users to specify the image project so we don't have to keep hard-coding in image projects each time we come out with a new distro.

Thanks for the quick merge!

PR #386, opened 2014-11-10
Added us-east-2 and us-east-3 regions

ZuluPro:
Added new regions to constant VALID_REGIONS.

This list was created in 2012, I made an update.


Kami:
Merged into trunk. Thanks.

PR #387, opened 2014-11-11
Made optional Google token file writing

ZuluPro:
I had an issue where I can't write the token file in home directory.

File reading is made as optional by a try/except IOError.
So I made the same with file writing by change GoogleBaseConnection._write_token_info_to_file() with try/except IOError.

I use this patch in production without trouble.

### DIFF
--- a/libcloud/common/google.py
+++ b/libcloud/common/google.py
@@ -698,10 +698,14 @@ class GoogleBaseConnection(ConnectionUserAndKey, PollingConnection):
         """
         Write token_info to credential file.
         """
-        filename = os.path.realpath(os.path.expanduser(self.credential_file))
-        data = json.dumps(self.token_info)
-        with open(filename, 'w') as f:
-            f.write(data)
+        try:
+            filename = os.path.realpath(os.path.expanduser(
+                self.credential_file))
+            data = json.dumps(self.token_info)
+            with open(filename, 'w') as f:
+                f.write(data)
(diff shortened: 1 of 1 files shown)


Kami:
/cc @erjohnso 

I would personally prefer to allow user to specify if they want the tokens to be cached on the filesystem (e.g. `ex_cache_token=True` argument in the class constructor). I think that's better than quietly swallowing exceptions.

erjohnso:
This could also cause you to trip over API limits and temporarily lock out your access depending on the auth mechanism you're using. Caching/re-using tokens will prevent you from exceeding API limits. So, I'd be concerned that not caching tokens and silently swallowing errors could lead to a very hard to debug situation. See here [https://developers.google.com/accounts/docs/OAuth2]

I haven't tested this, but I suspect you could get the same effect if you chose to use `/dev/null` as your credential file.

I don't think we should go with this approach.  My first suggestion is to dig into why you're seeing IOErrors and fix that issue (NFS mount, bad file permissions, use a different credential file location, etc). If you really want to bypass using a credential file, give the `/dev//null` trick a try.

erjohnso:
@ZuluPro - did you get a chance to dig into root cause of the `IOError`, or try to see if using `/dev/null` is an acceptable workaround?

ZuluPro:
Yes, I think, thx for your time ! I close the request.

PR #388, opened 2014-11-12
Add t2 nodes to us-west-1

ghost:
for some reason these were left out but supported now in the region.


Kami:
Fixed broken tests and merged into trunk. Thanks.

PR #393, opened 2014-11-18
[google compute] adding TargetInstance

erjohnso:
Adding another Google Compute Engine API resource type, TargetInstance. This resource can be used with GCE's ForwardingRules (part of load-balancer) to direct traffic to a specific node (aka instance).

Public docs are: https://cloud.google.com/compute/docs/reference/latest/targetInstances


erjohnso:
/cc @verb 

Also, global IP addresses was just merged in #394

verb:
Thanks for the heads up.  I will rewrite the support for global
forwardingRules to be consistent with the region='global' style introduced
in #394

PR #394, opened 2014-11-19
[google] global IP addresses

erjohnso:
I should've really added this as part of #390 but didn't catch it until after it had been merged.

Google Compute Engine typically has region-based IP addresses, but with the introduction of HTTP load-balancing, a `global` static IP addresses have been introduced.  This PR is a small change to allow the `GCEAddress` class to allow for 'global' to be specified as the region.

New tests added to ensure both the original 'region-only' and new 'global' addresses are supported.

### DIFF
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -97,8 +97,9 @@ class GCEAddress(UuidMixin):
         return self.driver.ex_destroy_address(address=self)
 
     def __repr__(self):
-        return '<GCEAddress id="%s" name="%s" address="%s">' % (
-            self.id, self.name, self.address)
+        return '<GCEAddress id="%s" name="%s" address="%s" region="%s">' % (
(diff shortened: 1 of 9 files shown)


libcloud/compute/drivers/gce.py
```diff
@@ -97,8 +97,9 @@ def destroy(self):
         return self.driver.ex_destroy_address(address=self)
 
     def __repr__(self):
-        return '<GCEAddress id="%s" name="%s" address="%s">' % (
-            self.id, self.name, self.address)
+        return '<GCEAddress id="%s" name="%s" address="%s" region="%s">' % (
+            self.id, self.name, self.address,
+            (hasattr(self.region, "name") and self.region.name or self.region))
```
Kami, line 102: That's fine for now, but next time, for readability purposes, I would prefer to declare a separate `region` variable in the method and use that.

Kami:
Merged into trunk. Thanks.

PR #395, opened 2014-11-20
[google compute] improved Images coverage

erjohnso:
Updates to improve the `GCENodeImage` coverage of the GCE API.  Specifically, this update has:
- Added object attributes to match the upstream Image resource (e.g. `diskSizeGb`, `sourceType`, etc)
- Moving toward deprecating older attributes such as `preferredKernel`. This has long been removed from the GCE API, but there is a chance very old customer images still retain this value.
- Add support for setting timestamps for deprecating images for when they would will be automatically marked as `DELETED`, `OBSOLETE`, and `DEPRECATED`.
- Also adding a nice-to-have feature for allowing users to specify a search-list of projects versus just the user's project and hard-coded list of supported image projects.

No breaking changes introduced and tests have been enhanced for these changes.

### DIFF
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -243,7 +243,8 @@ class GCENodeImage(NodeImage):
         """
         return self.driver.ex_delete_image(image=self)
 
-    def deprecate(self, replacement, state):
+    def deprecate(self, replacement, state, deprecated=None, obsolete=None,
+                  deleted=None):
         """
         Deprecate this image
 
@@ -254,10 +255,20 @@ class GCENodeImage(NodeImage):
                        \'DELETED\', \'DEPRECATED\' or \'OBSOLETE\'.
         :type   state: ``str``
 
+        :param  deprecated: RFC3339 timestamp to mark DEPRECATED
+        :type   deprecated: ``str`` or ``None``
+
+        :param  obsolete: RFC3339 timestamp to mark OBSOLETE
+        :type   obsolete: ``str`` or ``None``
+
+        :param  deleted: RFC3339 timestamp to mark DELETED
+        :type   deleted: ``str`` or ``None``
+
         :return: True if successful
         :rtype:  ``bool``
         """
-        return self.driver.ex_deprecate_image(self, replacement, state)
+        return self.driver.ex_deprecate_image(self, replacement, state,
+                                              deprecated, obsolete, deleted)
 
 
 class GCENetwork(UuidMixin):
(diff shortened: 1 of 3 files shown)


libcloud/compute/drivers/gce.py
```diff
@@ -2538,10 +2587,11 @@ def ex_get_image(self, partial_name):
         if partial_name.startswith('https://'):
             response = self.connection.request(partial_name, method='GET')
             return self._to_node_image(response.object)
-        image = self._match_images(None, partial_name)
+        image = self._match_images(ex_project_list, partial_name)
         if not image:
             if (partial_name.startswith('debian') or
-                    partial_name.startswith('backports')):
+                    partial_name.startswith('backports') or
+                    partial_name.startswith('nvme-backports')):
```
sebgoa, line 2594: @erjohnso just checking that this is not a typo. nvme-backports ? not name-backports ?

libcloud/compute/drivers/gce.py
```diff
@@ -2218,6 +2246,27 @@ def ex_deprecate_image(self, image, replacement, state=None):
             'replacement': replacement.extra['selfLink'],
         }
 
+        if deprecated is not None:
+            try:
+                _ = timestamp_to_datetime(deprecated)    # NOQA
```
sebgoa, line 2251: @Kami quick ping on this , what's our convention for comments at end of a line ?

libcloud/compute/drivers/gce.py
erjohnso, line 2594: @runseb - thanks for checking!  Yes, `nvme` is correct.

libcloud/compute/drivers/gce.py
erjohnso, line 2251: If it's not clear, I'm ensuring that the string can be properly converted but I don't actually want to use the generated timestamp.  The flake8 tests don't like unused variables so I was suppressing it's check on these lines.

Kami, line 2251: (sorry for a late response)

We follow pep8 which says 2 spaces before in-line comment (e.g. `bar = 42  # foo`)

libcloud/compute/drivers/gce.py
```diff
@@ -2218,6 +2246,27 @@ def ex_deprecate_image(self, image, replacement, state=None):
             'replacement': replacement.extra['selfLink'],
         }
 
+        if deprecated is not None:
```
Kami, line 2249: Sorry for a late comment. Number of duplicated lines of code could be reduced by doing something like that:

``` python
for attribute, value in [('deprecated', deprecated), ...]:
   is value is None:
    continue

   try:
    timestamp_to_datetime(value)
   except:
    raise ValueError('deprecated must be an RFC3339 timestamp')

   image_data[name] = value
```

Anyway, that's just a suggestion and not a priority :)

erjohnso, line 2249: I like it! Adding it to the next PR.

PR #396, opened 2014-11-20
[google compute] improve MachineType (size) coverage of GCE API

erjohnso:
Minor improvements to GCE Machine Types (sizes in libcloud parlance) to match GCE API.
- added missing resource attributes to `GCENodeSize`
- hiding some parameters from `extra` since they've been removed (namely `imageSpaceGb`)
- minor change to the disk.size value based on presence/absence of attributes

Updated test fixture and test.

### DIFF
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -3508,11 +3508,27 @@ class GCENodeDriver(NodeDriver):
         :rtype: :class:`GCENodeSize`
         """
         extra = {}
+        disk_size = 10
         extra['selfLink'] = machine_type.get('selfLink')
+        extra['creationTimestamp'] = machine_type.get('creationTimestamp')
         extra['zone'] = self.ex_get_zone(machine_type['zone'])
         extra['description'] = machine_type.get('description')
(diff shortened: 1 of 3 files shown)


libcloud/compute/drivers/gce.py
```diff
@@ -3508,11 +3508,27 @@ def _to_node_size(self, machine_type):
         :rtype: :class:`GCENodeSize`
         """
         extra = {}
+        disk_size = 10
```
sebgoa, line 3511: @erjohnso so we cannot change the disk_size through an argument ?

erjohnso, line 3511: @runseb - no, you can't change parameters on a GCE MachineType. These are analogous to AWS "sizes" like x2.large, c2.medium (or whatever they are).  They are static descriptions of the "size" of VM.

Google has made a change to the API since this code was originally contributed and now the most appropriate way to report a MachineType's disk size is to use `maximumPersistentDisksSizeGb` over the now rarely seen `imageSpaceGb` attribute.

sebgoa, line 3511: ok cool, I have not used GCE in couple months :)

PR #399, opened 2014-11-24
[google compute] improved API coverage for Projects

erjohnso:
This PR adds support for:
- Setting common instance metadata. Common instance (or project-wide) metadata is typically used for defining a `startup-script` and setting `sshKeys`.
- Setting a usage export bucket. The usage export feature allows storing compute engine resource usage to Google Cloud Storage.

For setting common instance metadata, there is a safety check (boolean flag) to that must be set to forcefully overwrite or remove `sshKeys`.

### DIFF
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -3808,6 +3808,16 @@ class GCENodeDriver(NodeDriver):
         extra['selfLink'] = snapshot.get('selfLink')
         extra['creationTimestamp'] = snapshot.get('creationTimestamp')
         extra['sourceDisk'] = snapshot.get('sourceDisk')
+        if 'description' in snapshot:
+            extra['description'] = snapshot['description']
+        if 'sourceDiskId' in snapshot:
+            extra['sourceDiskId'] = snapshot['sourceDiskId']
+        if 'storageBytes' in snapshot:
+            extra['storageBytes'] = snapshot['storageBytes']
+        if 'storageBytesStatus' in snapshot:
+            extra['storageBytesStatus'] = snapshot['storageBytesStatus']
+        if 'licenses' in snapshot:
+            extra['licenses'] = snapshot['licenses']
 
         return GCESnapshot(id=snapshot['id'], name=snapshot['name'],
                            size=snapshot['diskSizeGb'],
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -1361,7 +1361,7 @@ class GCENodeDriver(NodeDriver):
 
         return self.ex_get_image(name)
 
-    def ex_create_network(self, name, cidr):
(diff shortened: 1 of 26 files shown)


libcloud/compute/drivers/gce.py
```diff
@@ -304,6 +304,54 @@ def __init__(self, id, name, metadata, quotas, driver, extra=None):
         self.extra = extra
         UuidMixin.__init__(self)
 
+    def set_common_instance_metadata(self, metadata=None, force=False):
```
sebgoa, line 319: @erjohnso Why are you defining these and then the extension methods ex_ ? @Kami what do you think ? more of a convention thing than anything else.

libcloud/test/compute/fixtures/gce/setUsageExportBucket_post.json
```diff
@@ -0,0 +1,14 @@
+{
```
sebgoa, line 1: @erjohnso thanks for contributing tests, that's great. @Kami do you see an issue with the fixture file name using camel case. A detail really, but since we try to be consistent, maybe _ would be more pythonic even in file names ?

libcloud/compute/drivers/gce.py
Kami, line 319: Those methods are located on a class which is already not a part of a standard Libcloud API so adding `ex_` prefix here is unnecessary and redundant. And other methods are located on the NodeDriver class which means they need to have `ex_` prefix.

As far as having methods on both - node driver and project class - it's something we do for the user convenience.

libcloud/test/compute/fixtures/gce/setUsageExportBucket_post.json
erjohnso, line 1: I named it to match the upstream API's method name, but I'm fine with changing it if you prefer.

sebgoa:
Can you rebase, this does not apply cleanly anymore

erjohnso:
@runseb - my git foo is (I hope) slowly improving. I _think_ I've rebased successfully, and here were my steps:

``` sh
$ cd ~erjohnso/libcloud
$ git remote add upstream https://github.com/apache/libcloud
$ git checkout trunk
$ git pull upstream trunk
$ git checkout GCE_Projects
$ git merge trunk
$ git rebase
$ git push -f
```

Tox tests are still passing on my local dev environment, so hopefully Travis agrees.

erjohnso:
@runseb - Travis seems happy (phew!). Let me know if I should do anything else here.

sebgoa:
@erjohnso right :) so that did not really work, you pulled all the previous patches that we made into this pr. @Kami is great with git, let's ask him. I have to run

erjohnso:
Well, yeah. I guess I should've gone the other way to test a merge into trunk and fixed those issues... :(

Closing this in favor of #409 which was just branched from trunk.

PR #400, opened 2014-11-25
[google compute] adding Routes

erjohnso:
This PR adds support for Google Compute Engine's Routes.  This covers adding, listing, deleting, and fetching routes (all standard operations).  The Routes API is here[1] and the GCE docs for routes are here[2]

Includes tests and new fixtures.

[1] https://cloud.google.com/compute/docs/reference/latest/routes
[2] https://cloud.google.com/compute/docs/networking#routing

### DIFF
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -3837,6 +3837,16 @@ class GCENodeDriver(NodeDriver):
         extra['selfLink'] = snapshot.get('selfLink')
         extra['creationTimestamp'] = snapshot.get('creationTimestamp')
         extra['sourceDisk'] = snapshot.get('sourceDisk')
+        if 'description' in snapshot:
+            extra['description'] = snapshot['description']
+        if 'sourceDiskId' in snapshot:
+            extra['sourceDiskId'] = snapshot['sourceDiskId']
+        if 'storageBytes' in snapshot:
+            extra['storageBytes'] = snapshot['storageBytes']
+        if 'storageBytesStatus' in snapshot:
+            extra['storageBytesStatus'] = snapshot['storageBytesStatus']
+        if 'licenses' in snapshot:
+            extra['licenses'] = snapshot['licenses']
 
         return GCESnapshot(id=snapshot['id'], name=snapshot['name'],
                            size=snapshot['diskSizeGb'],
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -296,6 +296,34 @@ class GCENetwork(UuidMixin):
             self.id, self.name, self.cidr)
 
 
+class GCERoute(UuidMixin):
+    """A GCE Route object class."""
+    def __init__(self, id, name, dest_range, priority, network="default",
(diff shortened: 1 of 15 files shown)


libcloud/compute/drivers/gce.py
```diff
@@ -1254,6 +1296,70 @@ def ex_create_image(self, name, volume, description=None,
...
+            route_data['nextHopIp'] = next_hop
+        else:
+            node = self.ex_get_node(next_hop)
+            route_data['nextHopInstance'] = node.extra['selfLink']
+
+        request = '/global/routes'
+        self.connection.async_request(request, method='POST',
+                                      data=route_data)
+
+        return self.ex_get_route(name)
```
sebgoa, line 1361: @erjohnso just trying to give you a little feedback to avoid a blanket commit :) what happens if the ex_get_route fails, like if the route has not been created yet, since it uses an async request.

libcloud/compute/drivers/gce.py
```diff
@@ -2333,6 +2439,20 @@ def ex_destroy_forwarding_rule(self, forwarding_rule):
...
+
+        :param  route: Route object to destroy
+        :type   route: :class:`GCERoute`
+
+        :return:  True if successful
+        :rtype:   ``bool``
+        """
+        request = '/global/routes/%s' % (route.name)
+        self.connection.async_request(request, method='DELETE')
+        return True
```
sebgoa, line 2454: @erjohnso so it always succeeds ?

libcloud/test/compute/fixtures/gce/n-create-route
```diff
@@ -0,0 +1,13 @@
+{
```
sebgoa: @erjohnso silly comment, but any reason to switch to hyphens in file name here and not underscore ?

libcloud/compute/drivers/gce.py
erjohnso, line 1361: Hi @runseb - thanks for checking!

The GCE driver sub-classes (indirectly) the PollingConnection class and it's async method is actually blocking [https://github.com/apache/libcloud/blob/trunk/libcloud/common/base.py#L818]. For the GCE CRUD requests, we return an "operation" and that gets polled until the operation's status is DONE [https://github.com/apache/libcloud/blob/trunk/libcloud/common/google.py#L706].

This is also a common pattern in the rest of the GCE driver.

libcloud/compute/drivers/gce.py
erjohnso, line 2454: Hmm.. always is pretty strong statement, but let's say "a lot more often than not". ;)  There isn't much that can go wrong with the request but if something does go wrong, it would surface in a traceback and provide the user a useful error message.

libcloud/test/compute/fixtures/gce/n-create-route
erjohnso: My bad! I didn't mean to submit this file - will whack it now.  Nice catch!

erjohnso:
Messed up on syncing to trunk, closing in favor of #410

PR #402, opened 2014-11-25
[google compute] add region attr to zone

erjohnso:
One-liner fix to add missing `region` to the `GCEZone` object.


Kami:
@erjohnso Can you please manually close this ticket - I accidentally closed #403 instead of this one via commit message.

PR #403, opened 2014-11-27
[LIBCLOUD-638] Can't set CORS headers via container.upload_object_via_stream()

pzrq:
https://issues.apache.org/jira/browse/LIBCLOUD-638

### DIFF
--- a/libcloud/test/storage/test_cloudfiles.py
+++ b/libcloud/test/storage/test_cloudfiles.py
@@ -666,6 +666,10 @@ class CloudFilesTests(unittest.TestCase):
 
         try:
             self.driver.upload_object_via_stream(
+                # We never reach the Python 3 only bytes vs int error
+                # currently at libcloud/utils/py3.py:89
+                #     raise TypeError("Invalid argument %r for b()" % (s,))
+                # because I raise a NotImplementedError.
                 iterator=iter(b'blob data like an image or video'),
                 container=container,
                 object_name="test_object",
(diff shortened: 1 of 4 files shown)


libcloud/test/storage/test_cloudfiles.py
```diff
@@ -635,6 +635,51 @@ def test__upload_object_part(self):
...
+        }
+        expected_headers.update(cors_headers)
+
+        def intercept_request(request_path,
+                              method=None, data=None,
+                              headers=None, raw=True):
+
+            # What we're actually testing
+            # ... would prefer assertDictEqual but for Python <=2.6 support
+            self.assertEqual(expected_headers, headers)
```
Kami, line 660: You can replace `import unittest` with from `libcloud.test import unittest`.

This way it will use stdlib `unittest` library with Python >= 2.7 and unittest2 with Python < 2.7 which means you can then use assertDictEqual.

libcloud/test/storage/test_cloudfiles.py
```diff
@@ -635,6 +635,51 @@ def test__upload_object_part(self):
...
+                # We never reach the Python 3 only bytes vs int error
+                # currently at libcloud/utils/py3.py:89
+                #     raise TypeError("Invalid argument %r for b()" % (s,))
+                # because I raise a NotImplementedError.
+                iterator=iter(b'blob data like an image or video'),
+                container=container,
+                object_name="test_object",
+                headers=cors_headers,
+            )
+        except NotImplementedError:
```
Kami, line 678: We should add `else` and throw if `NotImplementedError` is not thrown since this means we never got into `intercept_request`, right?

Kami:
Besides the two comments I've mentioned, LGTM, +1.

Also, thanks for clearing / fixing up the docstrings.

Kami:
Oops, I accidentally closed this ticket instead of #402 via the commit message.

@pzrq Can you please re-open it or create a new PR if you can't re-open it? Thanks.

PR #405, opened 2014-11-28
Fixes the c3.X instances to have the proper disk size

Itxaka:
See subject

### DIFF
--- a/libcloud/compute/drivers/ec2.py
+++ b/libcloud/compute/drivers/ec2.py
@@ -219,35 +219,35 @@ INSTANCE_TYPES = {
         'id': 'c3.large',
         'name': 'Compute Optimized Large Instance',
         'ram': 3750,
-        'disk': 16,
+        'disk': 32,  # x2
         'bandwidth': None
     },
     'c3.xlarge': {
         'id': 'c3.xlarge',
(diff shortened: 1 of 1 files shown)


libcloud/compute/drivers/ec2.py
```diff
@@ -219,35 +219,35 @@
         'id': 'c3.large',
         'name': 'Compute Optimized Large Instance',
         'ram': 3750,
-        'disk': 16,
+        'disk': 32,  # x2
```
sebgoa, line 222: I think you can remove all of the # x2 comments

Itxaka, line 222: Its just so its easier to know that its actually 2 different disks. There is a couple of instances in the file which have the same comment, probably so people dont get confused like I was at first :D

sebgoa, line 222: so it's 2 disks of 16 that make 32, or two disks of 32 ?

Itxaka, line 222: 2 total disks, same as it was here https://github.com/Itxaka/libcloud/blob/fix_aws_c3_instances_disks/libcloud/compute/drivers/ec2.py#L336

Itxaka:
@kami ?

PR #407, opened 2014-12-01
Added S3 EU Central region (Frankfurt)

jarus:


Kami:
Thanks for the patch, but just adding constant is not enough in this case.

Adding support for the Frankfurt region among other things also requires implementing support for AWS Signature version - https://issues.apache.org/jira/browse/LIBCLOUD-626

gertjanol:
Hi @Kami, we would like to use libcloud to communicate with the EC2-service in Frankfurt and are running into the missing support for SignatureVersion 4. I'm thinking about implementing it for libcloud. Are all Amazon services that libcloud supports compatible with v4, or does libcloud need to keep support for v2? Do you have an idea of other problems that we might run into?

Kami:
@gertjanol We should keep support for signature v2 for now (it shouldn't be too hard to support both versions) and to begin with, only use v4 for EC2 Frankfurt region.

And yeah, implementing the support for signature version 4 would be great. I believe the EC2 documentation also contains some examples of how to generate the signature with Python - http://docs.aws.amazon.com/general/latest/gr/sigv4-signed-request-examples.html

PR #411, opened 2014-12-04
False.lower() doesn't work.

tomprince:


Kami:
Good catch - patch has been merged into trunk.

Thanks.

PR #412, opened 2014-12-04
fix ex_disk_auto_delete for create_node()

Ygrex:
create_node() has a parameter ex_disk_auto_delete that is True by default, but autoDelete attribute is never set for boot disks


erjohnso:
Thanks @Ygrex!

PR #416, opened 2014-12-16
[LIBCLOUD-647] - in newCCI changing the value of sshKeys to a list conta...

doberloh:
...ining a dictionary with format {'id':<numcerical_key_value>}

### DIFF
--- a/libcloud/compute/drivers/softlayer.py
+++ b/libcloud/compute/drivers/softlayer.py
@@ -410,7 +410,11 @@ class SoftLayerNodeDriver(NodeDriver):
             newCCI['datacenter'] = {'name': datacenter}
 
         if 'ex_keyname' in kwargs:
-            newCCI['sshKeys'] = [self._key_name_to_id(kwargs['ex_keyname'])]
+            newCCI['sshKeys'] = [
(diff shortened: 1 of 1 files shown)


Kami:
Can you please provide more background and reference to the documentation which explains how the payload of `sshKeys` attribute should look like?

Thanks

doberloh:
So I found this issue while trying to use libcloud v0.16.0. I suspected there was an issue in the libcloud code when I was getting errors back from the softlayer API while sending ex_keyname='keyname' into the create_node function. I opened a ticket with them and they sent me back a short code snippet using their python library here: https://github.com/softlayer/softlayer-python/tree/master/SoftLayer. In it they sent sshKeys as a list of one dictionary in the format of the change I have made. You can see where they handle ssh keys in their python library at line 350 of the _generate_create_dict method: https://github.com/softlayer/softlayer-python/blob/master/SoftLayer/managers/vs.py. It is the same code that I have modified the libcloud code to be. Hopefully that was enough info :)

Kami:
Thanks for the clarification.

Looks good, merged into trunk. Thanks.

PR #418, opened 2014-12-18
[AWS] placement group

ardling:
creating/deleteing/listing placement groups with tests.
Plese give feedback if something wrong.

### DIFF
--- a/libcloud/compute/drivers/ec2.py
+++ b/libcloud/compute/drivers/ec2.py
@@ -2183,8 +2183,8 @@ class BaseEC2NodeDriver(NodeDriver):
         :keyword    ex_subnet: The subnet to launch the instance into.
         :type       ex_subnet: :class:`.EC2Subnet`
 
-        :keyword    ex_placement_group: The placement group to launch
-                                        the instance into.
+        :keyword    ex_placement_group: The name of the placement group to
+                                        launch the instance into.
         :type       ex_placement_group: ``str``
         """
         image = kwargs["image"]
@@ -2618,7 +2618,7 @@ class BaseEC2NodeDriver(NodeDriver):
         """Creates new Placement Group
 
         :param name: Name for new placement Group
-        :type: ``str``
+        :type name: ``str``
 
         :rtype: ``bool``
         """
@@ -2632,7 +2632,7 @@ class BaseEC2NodeDriver(NodeDriver):
         """Deletes Placement Group
 
         :param name: Placement Group name
-        :type: ``str``
+        :type name: ``str``
 
         :rtype: ``bool``
         """
@@ -2645,14 +2645,14 @@ class BaseEC2NodeDriver(NodeDriver):
         """List Placement Groups
 
         :param names: Placement Group names
-        :type: ``list`` of ``str``
+        :type names: ``list`` of ``str``
 
         :rtype: ``list`` of :class:`.EC2PlacementGroup`
         """
         params = {'Action': 'DescribePlacementGroups'}
 
-        for n, name in enumerate(names, 1):
-            params['GroupName.%s' % n] = name
+        for index, name in enumerate(names):
+            params['GroupName.%s' % index + 1] = name
 
         response = self.connection.request(self.path, params=params).object
         return self._to_placement_groups(response)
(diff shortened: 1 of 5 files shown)


libcloud/compute/drivers/ec2.py
```diff
@@ -2591,6 +2614,49 @@ def delete_image(self, image):
         response = self.connection.request(self.path, params=params).object
         return self._get_boolean(response)
 
+    def ex_create_placement_group(self, name):
+        """Creates new Placement Group
+
+        :param name: Name for new placement Group
+        :type: ``str``
```
Kami: Should be `:type name:`

libcloud/compute/drivers/ec2.py
```diff
@@ -2591,6 +2614,49 @@ def delete_image(self, image):
...
+                  'Strategy': 'cluster',
+                  'GroupName': name}
+        response = self.connection.request(self.path, params=params).object
+        return self._get_boolean(response)
+
+    def ex_delete_placement_group(self, name):
+        """Deletes Placement Group
+
+        :param name: Placement Group name
+        :type: ``str``
```
Kami: Should be `:type name:`

libcloud/compute/drivers/ec2.py
```diff
@@ -2591,6 +2614,49 @@ def delete_image(self, image):
...
+        params = {'Action': 'DeletePlacementGroup',
+                  'GroupName': name}
+        response = self.connection.request(self.path, params=params).object
+        return self._get_boolean(response)
+
+    def ex_list_placement_groups(self, names=()):
+        """List Placement Groups
+
+        :param names: Placement Group names
+        :type: ``list`` of ``str``
```
Kami: Should be `:type names`

libcloud/compute/drivers/ec2.py
```diff
@@ -2591,6 +2614,49 @@ def delete_image(self, image):
...
+        :type: ``str``
+
+        :rtype: ``bool``
+        """
+        params = {'Action': 'DeletePlacementGroup',
+                  'GroupName': name}
+        response = self.connection.request(self.path, params=params).object
+        return self._get_boolean(response)
+
+    def ex_list_placement_groups(self, names=()):
```
Kami, line 2644: You should never use a mutable default argument - http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments

You can do `names=None` and inside the function do `names = names or []`

libcloud/compute/drivers/ec2.py
```diff
@@ -2591,6 +2614,49 @@ def delete_image(self, image):
...
+        """List Placement Groups
+
+        :param names: Placement Group names
+        :type: ``list`` of ``str``
+
+        :rtype: ``list`` of :class:`.EC2PlacementGroup`
+        """
+        params = {'Action': 'DescribePlacementGroups'}
+
+        for n, name in enumerate(names, 1):
```
Kami: We still support Python 2.5 which doesn't support "step" argument to `enumerate` function so please remove it. Default step is 1 so it should still work fine.

Kami: For consistency, please also use `index` instead of `n` for the variable name.

libcloud/compute/drivers/ec2.py
```diff
@@ -2166,6 +2182,10 @@ def create_node(self, **kwargs):
 
         :keyword    ex_subnet: The subnet to launch the instance into.
         :type       ex_subnet: :class:`.EC2Subnet`
+
+        :keyword    ex_placement_group: The placement group to launch
```
Kami: To make it more obvious and explicit, let's change it to `The name of the placement group to launch the instance into.`.

Kami:
Thanks!

I've added some comments in-line. Overall it looks good, there are just some minor styling and compatibility issues which need to be addressed.

libcloud/compute/drivers/ec2.py
ardling, line 2644: I know about mutable default arguments, but () is tuple and tuples are immutable. Do I realy must fix it?

Kami:
Sorry for the delay - I didn't see you pushed a new commit (we don't get notifications for that).

Next time, feel free to ping us when you address the comments.

Anyway, I've merged patch into trunk. Thanks!

PR #421, opened 2014-12-24
@Added JoyentNodeDriver.ex_get_node()

ZuluPro:
Joyent didn't have a method for get only one Node.

Because I doesn't want to make a for loop for find a node, I added this method.
Error raising is better too with that, a real Joyent's error message is now raised.

### DIFF
--- a/libcloud/compute/drivers/joyent.py
+++ b/libcloud/compute/drivers/joyent.py
@@ -205,7 +205,6 @@ class JoyentNodeDriver(NodeDriver):
         result = self.connection.request('/my/machines/%s' % (node.id),
                                          data=data, method='POST')
         return result.status == httplib.ACCEPTED
-        
 
     def ex_get_node(self, node_id):
         """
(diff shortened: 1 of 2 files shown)


libcloud/compute/drivers/joyent.py
```diff
@@ -205,6 +205,20 @@ def ex_start_node(self, node):
...
+        """
+        Return a Node object based on a node ID.
+
+        :param  node_id: ID of the node
+        :type   node_id: ``str``
+
+        :return:  A Node object for the node
+        :rtype:   :class:`Node`
+        """
+        result = self.connection.request('/my/machines/%s' % node.id)
```
Kami: Please use parenthesis around format string arguments - `% (node.id)`

Kami:
Thanks.

Overall it looks good, but we need a test case for it and lint failure needs to be fixed - https://travis-ci.org/apache/libcloud/builds/45025149

ZuluPro:
I use my_machines_single.json for mock a single machine.

Kami:
Merged into trunk. Thanks.

PR #422, opened 2014-12-30
[google compute] adding all list params to all list functions

erjohnso:
This PR adds support to all `list_XXX()` and `ex_list_XXX()` driver methods using the parameters provided by GCE API as described in this example, https://cloud.google.com/compute/docs/reference/latest/diskTypes/list.
- Users can now provide an `ex_filter` to restrict search results to filter matches such as nodes that match `name eq "test.*"`.
- Users can now step through large resource sets `ex_max_results` for each method call until no more results are returned. In GCE terms, this allows for "pagination" and uses a `pageToken` query parameter to step through the resource set. Users can 'reset' the step pointer by specifying `ex_reset_page_token`, e.g.
  
  ``` python
  subset = gce_driver.ex_list_snapshots(ex_max_results=2):
  while subset:
      # do something with this subset of snapshots...
      subset = gce_driver.ex_list_snapshots(ex_max_results=2)
  gce_driver.ex_list_snapshots(ex_reset_page_token=True)
  ```

### DIFF
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -770,6 +770,7 @@ class GCENodeDriver(NodeDriver):
             raise ValueError('Project name must be specified using '
                              '"project" keyword.')
 
+        self.page_token_cache = {}
         self.auth_type = auth_type
         self.project = project
         self.scopes = scopes
@@ -801,7 +802,8 @@ class GCENodeDriver(NodeDriver):
         else:
             self.region = None
 
-    def ex_list_disktypes(self, zone=None):
+    def ex_list_disktypes(self, zone=None, ex_filter=None, ex_max_results=None,
+                          ex_reset_page_token=False):
         """
         Return a list of DiskTypes for a zone or all.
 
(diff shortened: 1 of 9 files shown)


verb:
I think I'd rather see something like a paged list iterator that worked something like:

```
for sublist in list_iterator(ex_list_XXX, max_results=X):
  # do something with this subset of XXX
```

This approach seems a little more pythonic, and it removes the ambiguity surrounding what happens when you call a series of ex_list_XXX list function with different arguments without resetting the page token (e.g. ex_list_XXX(region='A', max_results=2), ex_list_XXX(region='B')) by providing scope for the pageToken.

Additionally, it removes a lot of the boilerplate from list functions.

The first implementation that came to mind looks something like:

```
def ex_paged_list(self, list_fn, max_results, **kwargs):
    params = {'maxResults': max_results}

    def paged_request(request, method):
        response = self.connection.request(request, method=method,
                                           params=params)
        if 'nextPageToken' in response.object:
            params['pageToken'] = response.object['nextPageToken']
        else:
            params.pop('pageToken', None)

        return response

    more_results = True
    while more_results:
        yield list_fn(request_handler=paged_request, **kwargs)
        more_results = 'pageToken' in params
```

```
>>> for u in driver.ex_paged_list(list_fn=driver.ex_list_urlmaps, max_results=2):
...   pprint(u)
... 
[<GCEUrlMap id="15808585643830800853" name="cli-map">,
 <GCEUrlMap id="10438265192591835206" name="lc-map">]
[<GCEUrlMap id="4266107551250249032" name="web-map">]
```

Filter functions could be implemented the same way and could probably be chained (not implemented above) something like:

```
for X in ex_paged_list(ex_filtered_list(ex_list_urlmaps, 'FILTER'), max_results=2):
  # stuff
```

The downside of this particular implementation is adding a strange 'request_handler' parameter to the list_methods so that the self.connection.request method could be overridden.  It looks like pagination is implicit in the API, though, so maybe we could implement it in the GCEConnection request handler somehow.

What do you think?

libcloud/compute/drivers/gce.py
```diff
@@ -801,7 +802,8 @@ def __init__(self, user_id, key=None, datacenter=None, project=None,
         else:
             self.region = None
 
-    def ex_list_disktypes(self, zone=None):
+    def ex_list_disktypes(self, zone=None, ex_filter=None, ex_max_results=None,
+                          ex_reset_page_token=False):
```
Kami, line 806: If the method already has an `ex_` prefix, the argument shouldn't have it.

The arguments only need to have `ex_` prefix if the arguments are non-standard and used with a standard method (e.g. list_nodes, etc.).

erjohnso:
@verb - I like your idea much better! I'd be totally fine with closing out this PR if you want to take this over. :)

@Kami - thanks, that makes sense. I was partly on auto-pilot with the `ex_` convention because `filter` is a reserved word and happens to match the GCE API. But we can use a different param name such as filter_match or some such.

verb:
@erjohnso SGTM.  I'll polish it up and create a PR.  You may want to hang on to your git history until then, though. :)

erjohnso:
@verb - great! Yeah, I'll keep my branch but close the PR.  Thanks for this!

PR #423, opened 2014-12-30
[LIBCLOUD-602] improve gce.image_list()

erjohnso:
Improves the GCE driver's `list_images()` method to actually return what the user would expect; a list of vendor supported operating system images including what they have on their own project.  The default behavior now returns on non-deprectated images, but a full list can be returned using the `ex_include_deprecated` param.

Although this changes the behavior of `list_images()`, it more closely matches the behavior of GCE's command-line utility `gcloud compute` that is part of the Cloud SDK package[1]. It is also more likely what the user wants to see anyway.

[1] https://cloud.google.com/sdk/

### DIFF
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -730,6 +730,19 @@ class GCENodeDriver(NodeDriver):
         "userinfo-email": "userinfo.email"
     }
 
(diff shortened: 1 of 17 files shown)


ZuluPro:
+1, better than search manually images from every projects.

By default, driver.list_images actually returns an empty list.

PR #427, opened 2015-01-06
LIBCLOUD-605 GCE HTTP-LB support

verb:
Adds support for many HTTP LB primitives to the GCE NodeDriver.  Reviewed by @erjohnso.

PR #429, opened 2015-01-07
[LIBCLOUD-605] Support HTTP-LB in GCE NodeDriver

verb:
This change adds coverage for GCE's HTTP-LB features via ex_\* functions
in the GCE NodeDriver.

Also, this updates the existing ex_create_forwarding_rule to use a
more generic 'target' parameter.  Previously this was 'targetpool', but
the GCE API has changed this to 'target' because it can be a targetPool,
targetInstance or targetHttpProxy.  This update is backwards compatible.

@erjohnso This replaces PR #427


erjohnso:
Thanks @verb!

PR #430, opened 2015-01-09
[LIBCLOUD-649] Add new methods for LinodeNodeDriver

wirkijowski:
Unit tests included
Jira ticket: https://issues.apache.org/jira/browse/LIBCLOUD-649

### DIFF
--- a/libcloud/compute/drivers/linode.py
+++ b/libcloud/compute/drivers/linode.py
@@ -46,7 +46,7 @@ from libcloud.common.linode import (API_ROOT, LinodeException,
 from libcloud.compute.types import Provider, NodeState
 from libcloud.compute.base import NodeDriver, NodeSize, Node, NodeLocation
 from libcloud.compute.base import NodeAuthPassword, NodeAuthSSHKey
-from libcloud.compute.base import NodeImage
+from libcloud.compute.base import NodeImage, StorageVolume
 
 
 class LinodeNodeDriver(NodeDriver):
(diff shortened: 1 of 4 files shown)


Kami:
Thanks.

I quickly glanced over the changes and they look good to me.

I'm just wondering, if we can make "ex_create_volume" and "ex_list_volume" methods some how comply with the standard API. Looking at the Linode API docs, LinodeId argument is required so sadly this might not be possible.

libcloud/compute/drivers/linode.py
```diff
@@ -480,6 +484,137 @@ def linode_set_datacenter(self, dc):
...
+
+    def ex_list_volumes(self, node, disk_id=None):
+        """
+        List existing disk volumes for for given Linode.
+
+        :keyword    node: Node to list disk volumes for. (required)
+        :type       node: :class:`Node`
+
+        :keyword    disk_id: Id for specific disk volume. (optional)
+        :keyword     disk_id: ``int``
```
Kami: Should be `:type disk_id:`

wirkijowski:
In order to comply with standart libcloud API, there could be create_volume() method. There are two ways it could be done. 
The first one is create_volume() could return StorageVolume instance which would be created for real after attach_volume(). But this would need that programmer should remeber that they operate on nonexistent volume until it's attached. 
The second way is to use create_volume() with additional ex_ params. But this params need to be required.

ex_list_volumes() could be rewritten to list_volumes() with optional ex_ params, definig node or disk_id.

Kami:
Sorry for the delay.

It sounds like we should just stay with extension methods for now since none of the approaches makes it fully compliant with the base API, in fact, it might just make it more confusing to the end user (it's named as the base API, but it doesn't behave as such).

Kami:
Squashed the commits and merged changes into trunk. Thanks.

wirkijowski:
Thanks:)

PR #431, opened 2015-01-10
LIBCLOUD-650: ex_create_multi_value_record only returns one record

Itxaka:
Fixes that bug and also adds tests for that function.

Bug discovery and solution by @jvrplmlmn


Itxaka:
@Kami ?

Kami:
Sorry for the delay (recovering after surgery).

Good catch and merged into trunk.

PR #433, opened 2015-01-11
Libcloud 652 abq 3 driver

lochiiconnectivity:
[LIBCLOUD-652] Abiquo 2.x is no longer supported by the company, I believe all providers are running 3.x which has major changes to its API. I have updated the Abiquo compute driver.

### DIFF
--- a/libcloud/common/abiquo.py
+++ b/libcloud/common/abiquo.py
@@ -125,10 +125,16 @@ class AbiquoResponse(XmlResponse):
             raise InvalidCredsError(driver=self.connection.driver)
         elif self.status == httplib.FORBIDDEN:
             raise ForbiddenError(self.connection.driver)
+	elif self.status == httplib.NOT_ACCEPTABLE:
+             raise LibcloudError('Not Acceptable')
         else:
-            errors = self.parse_body().findall('error')
-            # Most of the exceptions only have one error
-            raise LibcloudError(errors[0].findtext('message'))
+	    parsebody = self.parse_body()
+            if parsebody is not None and hasattr(parsebody, 'findall'):
+             errors = self.parse_body().findall('error')
+             # Most of the exceptions only have one error
+             raise LibcloudError(errors[0].findtext('message'))
+	    else:
+             raise LibcloudError('unparseable body')
 
     def success(self):
         """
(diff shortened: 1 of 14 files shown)


Kami:
@lochiiconnectivity Thanks.

If all the providers are indeed running API v3, then removing support for v2 is fine. Otherwise, we should support both versions (we do a similar thing with CloudSigma and some other drivers) - can you please verify that?

libcloud/common/abiquo.py
```diff
@@ -125,10 +125,16 @@ def parse_error(self):
...
-            errors = self.parse_body().findall('error')
-            # Most of the exceptions only have one error
-            raise LibcloudError(errors[0].findtext('message'))
+            parsebody = self.parse_body()
+            if parsebody is not None and hasattr(parsebody, 'findall'):
+                errors = self.parse_body().findall('error')
+                # Most of the exceptions only have one error
+                raise LibcloudError(errors[0].findtext('message'))
+            else:
+                raise LibcloudError('unparseable body')
```
Kami: Probably better to at least return raw body in the exception?

Otherwise it will be hard to debug it without using LIBCLOUD_DEBUG.

libcloud/compute/drivers/abiquo.py
```diff
@@ -236,37 +251,38 @@ def ex_populate_cache(self):
...
-            loc = self._to_location(e_vdc, dc_dict[dc_link], self)
-
-            # Save into cache the link to the itself because we will need
-            # it in the future, but we save here to don't extend the class
-            # :class:`NodeLocation`.
-            # So here we have the dict: :class:`NodeLocation` ->
-            # link_datacenter
-            self.connection.cache['locations'][loc] = get_href(e_vdc, 'edit')
+            loc = get_href(e_vdc, 'location')
+            if (loc is not None):
```
Kami: Minor style thing - parenthesis are unnecessary.

lochiiconnectivity:
For the last two points, I've pushed 0298c71 which should address these.
with regards to the support for Abiquo 2.x , I will obtain a covering statement from Abiquo shortly, and I'd like to point out that the driver did not support the last versions of 2.x, 2.3, 2.4 and 2.6, these made significant changes to the API making the trunk version incompatible, (for example making JSON the default media type, see note in http://wiki.abiquo.com/display/ABI26/API+Changelog+v2.6).

Kami:
@lochiiconnectivity Yeah, I noticed that in the API v3 change log. Didn't know this back-ward incompatible change was already introduced in v2.6.

Let us know when you hear back from Abiquo.

Kami:
@lochiiconnectivity Any updates from Abiquo?

lochiiconnectivity:
Having spoken to Abiquo, they've confirmed 2.x is not officially supported and they've looked across their customer base and can't see anything that would cause an issue if this support (for 2.x) was removed.

Kami:
Thanks. Merged into trunk.

PR #436, opened 2015-01-20
[LIBCLOUD-656] adding the tok02 region for softlayer compute.

doberloh:
This is a simple one liner that adds the tok02 region to the softlayer compute driver. When the driver was originally created the region did not exist, which would explain why it wasnt present. Let me know if you need any more info.


Kami:
Merged into trunk. Thanks!

PR #437, opened 2015-01-23
Fixed boolean comparison of a string

quilombo:
Fixed corresponding unittest by using assertTrue instead of assertEqual

### DIFF
--- a/libcloud/compute/drivers/openstack.py
+++ b/libcloud/compute/drivers/openstack.py
@@ -1989,7 +1989,7 @@ class OpenStack_1_1_NodeDriver(OpenStackNodeDriver):
         # state
         image = api_node.get('image', None)
         image_id = image.get('id', None) if image else None
-        config_drive = api_node.get("config_drive", "false")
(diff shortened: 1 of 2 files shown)


libcloud/compute/drivers/openstack.py
```diff
@@ -1989,11 +1989,7 @@ def _to_node(self, api_node):
         # state
         image = api_node.get('image', None)
         image_id = image.get('id', None) if image else None
-
-        if api_node.get("config_drive", "false").lower() == "true":
-            config_drive = True
-        else:
-            config_drive = False
+        config_drive = api_node.get("config_drive", "false")
```
Kami: Good catch, but the value should still be a boolean.

If the API now returns a boolean, we can just do `config_drive = api_node.get("config_drive", False)`

quilombo: yes, will apply your suggestion.
thanks a lot

Kami:
Merged into trunk. Thanks.

PR #441, opened 2015-01-28
[google compute] raise notfound on images that are not found

erjohnso:
If a user tried to delete an image by name (string) and the image was not found, the code would fail with a `NoneType` does not contain `name` attribute error.

This fixes the case where if an image can't be found by string name, it will raise a `ResourceNotFoundError`.

/cc @tanpeter - if you think this will work and is good, I can merge it in.

### DIFF
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -3810,7 +3810,7 @@ n
 
         :return:  GCENodeImage object based on provided information or None if
                   an image with that name is not found.
-        :rtype:   :class:`GCENodeImage` or ``None``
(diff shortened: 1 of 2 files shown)


tanpeter:
This change seems like a right thing to do and make it more consistent with other methods. My only concern is that the moment users upgrade their libcloud library their code might break due to uncaught exception.

erjohnso:
Thanks for the feedback @tanpeter. I'd expect that users were already checking for `None` if the image wasn't found, so I'd expect their fix to adjust to the exception to be pretty low (but still surprising).

When we roll a new release, we can flag this change to warn users in the announce email and release notes.

PR #443, opened 2015-02-01
Port DigitalOcean driver to API v2

andrewsomething:
DigitalOcean's API v2 has been out for sometime now, and v1 will soon be heading towards deprecation. It would be great for the next release to move to the new version.

https://developers.digitalocean.com/v2/

Filed on Jira as https://issues.apache.org/jira/browse/LIBCLOUD-660


Kami:
Great, thanks.

Looking at the API docs though, the page says API v2 is still in beta and the things might change. Is there any ETA when API will be out of beta? 

Should we wait with switching to API v2 until it comes out of beta (or is it more like a gmail beta :)?)

andrewsomething:
It is starting to have that Gmail feeling. =) My understanding is that the beta tag will be dropped "soon," and an EOL date will be announced for v1 at that time.

If you want to hold off till then, that's understandable. Though it's been fairly stable for sometime already. The changelog is here: https://developers.digitalocean.com/changelog/

libcloud/compute/drivers/digitalocean.py
```diff
@@ -116,30 +120,93 @@ def create_node(self, name, size, image, location, ex_ssh_key_ids=None,
...
+                                      params=params, method='POST')
+        return res.status == httplib.CREATED
+
+    def ex_power_on_node(self, node):
+        params = {'type': 'power_on'}
+        res = self.connection.request('/v2/droplets/%s/actions' % (node.id),
+                                      params=params, method='POST')
+        return res.status == httplib.CREATED
 
     def ex_list_ssh_keys(self):
```
sebgoa: SSH key pair management is now part of the base API. So you should have the list_key_pairs(), get_key_pair(), create_key_pair(), delete_key_pair(). Also use the KeyPair() object:
    from libcloud.compute.base import KeyPair

sebgoa:
@Kami @andrewsomething maybe we can create a V2 driver class , like in the openstack driver.
that way both can be supported until v1 is really deprecated and v2 is stable.
Squashing commits would also be nice.

Kami:
Yeah, I'm fine with supporting both versions for now.

There is a good example of how to do that at https://github.com/apache/libcloud/blob/trunk/libcloud/compute/drivers/cloudsigma.py

andrewsomething:
Alright, moved to using the SSH key pair management from the base API. Might not get around to attempting at supporting both versions until the weekend.

Thanks for the feedback!

mgogoulos:
Hi all, we have supported API v2 since September at https://github.com/mistio/libcloud/blob/mistio/libcloud/compute/drivers/digitalocean.py it's a shame we haven't found the time to contribute it here to save some time from others doing so :(  

The v2 version is greatly improved and  DO pushes against using v2 and not v1 (if I remember well they even plan to discontinue support for v1 after some time). We wanted to use both of them so in our approach we have renamed DigitalOceanNodeDriver/DigitalOceanConnection as DigitalOceanFirstGenNodeDriver/DigitalOceanFirstGenConnection and DigitalOceanNodeDriver/DigitalOceanConnection is now for v2

andrewsomething:
So it ended up being a little longer than I would have liked to get around to this... Commits have all been squashed, and this should support both v1 and v2. 

The test failure on Travis seems completely unrelated:

```
pkg_resources.VersionConflict: (pep8 1.6.2 (/home/travis/build/apache/libcloud/.tox/lint/lib/python2.7/site-packages), Requirement.parse('pep8<1.6,>=1.5.7'))
ERROR: InvocationError: '/home/travis/build/apache/libcloud/.tox/lint/bin/flake8 --ignore=E402 --exclude=test libcloud/'
```

Kami:
Yeah, it's an issue with a conflicting dependency (same issue as here - https://github.com/StackStorm/st2/pull/1238). I will push a fix to trunk shortly.

Kami:
I've pushed a fix for a conflicting dependency issue to trunk - 733e7abfd3404014a7a078294821cd820ac07e64 .

If you rebase trunk on top of your PR, the build should pass. Also, thanks for the changes. I will look into them asap.

Kami:
Thanks and sorry for the delay.

I will go over the changes today and test them and if everything looks ok, merge them into trunk.

Kami:
Made some minor changes and fixes (dc88ea99731541a368f3ae692ddb3527ddf3d14b, 24f6f3ae966366e4f0bfb113f8b986284e8bf0ae) and added some very basic docs (4f330e79941e82968e190f68ff2e34168f0f861f) and merged changes into trunk.

Thanks!

andrewsomething:
Great! Thanks so much. If I've got the Jira-foo right, I should be subscribed to new issues with the digitalocean label.

andrewsomething:
Great! Thanks so much. If I've got the Jira-foo right, I should be subscribed to new issues with the digitalocean label.

Kami:
Great :)

PR #445, opened 2015-02-03
[google compute] fixing node status

erjohnso:
Per comment thread on #442, making a small change to GCE driver's node states.  Also added a safeguard to catch any future new states Google introduces that might not have a corresponding map entry.

### DIFF
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -790,13 +790,23 @@ class GCENodeDriver(NodeDriver):
     type = Provider.GCE
     website = 'https://cloud.google.com/'
 
+    # Google Compute Engine node states are mapped to Libcloud node states
+    # per the following dict. GCE does not have an actual 'stopped' state
+    # but instead uses a 'terminated' state to indicate the node exists
+    # but is not running. In order to better match libcloud, GCE maps this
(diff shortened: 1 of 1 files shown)


libcloud/compute/drivers/gce.py
```diff
@@ -794,9 +794,9 @@ class GCENodeDriver(NodeDriver):
         "PROVISIONING": NodeState.PENDING,
         "STAGING": NodeState.PENDING,
         "RUNNING": NodeState.RUNNING,
-        "STOPPED": NodeState.TERMINATED,
-        "STOPPING": NodeState.TERMINATED,
-        "TERMINATED": NodeState.TERMINATED
+        "STOPPING": NodeState.PENDING,
+        "TERMINATED": NodeState.STOPPED,
```
Kami, line 808: LGTM, please just add a comment here explaining why we are mapping terminated to pending (e.g. gce has no stopped status and node being stopped is represented by terminated status).

erjohnso:
Thanks @Kami - comments added.  I'll assume when travis gives the green light that I'm free to merge.

Kami:
Yep.

Actually, looks like travis might be stuck so if tests pass locally feel free to go ahead and merge it.

PR #447, opened 2015-02-09
list images does not break in case of invalid OS image types

mgogoulos:
Make sure that list_images will not break if one of the IMAGE_PROJECTS keys  (OS image types) changes in future. Right now if you run list_images with an invalid project name, the connection.request_path is not  restored after the exception is run, so all requests that follow will be invalid. 

```
from libcloud.compute.providers import get_driver; from libcloud.compute.types import Provider; driver = get_driver('gce')

c = driver('email', 'key_path', project='name')

c.list_images(ex_project='centos-cloudX')
InvalidRequestError

c.list_nodes()
InvalidRequestError
```

### DIFF
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -1285,9 +1285,13 @@ class GCENodeDriver(NodeDriver):
         image_list = self.ex_list_project_images(ex_project=None,
                                                  ex_include_deprecated=dep)
         for img_proj in list(self.IMAGE_PROJECTS.keys()):
-            image_list.extend(
-                self.ex_list_project_images(ex_project=img_proj,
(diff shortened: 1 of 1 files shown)


erjohnso:
Thanks @mgogoulos - this looks great.  @Kami just cut 0.17.0, so I don't think it will get added to the release, but definitely will merge when travis catches up.

Kami:
@erjohnso If you think this is a big issue, you can also vote -1 and we can cancel the vote and roll out a new release.

Or we can wait and release 0.17.1 soon after 0.17.0 is out.

erjohnso:
I think it's a great fix and hardens the driver, but I don't think we need to hold off on 0.17.0.  Adding it the next release is fine.

Thanks again @mgogoulos!

PR #448, opened 2015-02-10
improve GCE create_node, make sure ex_get_disktype function will not fail

mgogoulos:
create_node supports passing ex_disk_type as either str or GCEDiskType. If you pass an str, or if you don't pass anything -most usual case I believe- it gets an str value of 'pd-standard' since it is a kwarg,  then the function will fail because self.ex_get_disktype is called without the location, and line 
      request = '/zones/%s/diskTypes/%s' % (zone.name, name)

will break with a 'NoneType' object has no attribute 'name'

This can easily be fixed with this commit (also applies to ex_create_multiple_nodes). 

Then nodes with ssh key deployed are easily created:

```
from libcloud.compute.providers import get_driver; from libcloud.compute.types import Provider; driver = get_driver('gce')
conn = driver('email', 'key', project='project')
node = conn.create_node(
                name=machine_name, # str
                image=image, # NodeImage
                size=size, # NodeSize
                location=location, # NodeLocation
                ex_metadata=metadata # {'sshKeys': 'user:%s' % public_ssh_rsa_key}
            )
```

### DIFF
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -2160,7 +2160,7 @@ n
         if image and not hasattr(image, 'name'):
             image = self.ex_get_image(image)
         if not hasattr(ex_disk_type, 'name'):
-            ex_disk_type = self.ex_get_disktype(ex_disk_type)
+            ex_disk_type = self.ex_get_disktype(ex_disk_type, zone=location)
 
         # Use disks[].initializeParams to auto-create the boot disk
         if not ex_disks_gce_struct and not ex_boot_disk:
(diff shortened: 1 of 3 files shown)


erjohnso:
Looks good @mgogoulos, but tests are not passing. Do you think you can take a look at travis and get them fixed up?  For example, https://travis-ci.org/apache/libcloud/jobs/50234015#L5396.

mgogoulos:
I've added the missing fixture for this, it should complete fine!

mgogoulos:
Forgot to include the new fixtures file..

erjohnso:
Great, thank you very much @mgogoulos!  One last request, could you please rebase and squash into a single commit?  See, https://libcloud.readthedocs.org/en/latest/development.html#squash-the-commits-and-generate-the-patch. We don't need the patch, but see the two `Notes` in that section for rebasing/squashing if you've never done it.

mgogoulos:
Should be fine now, after tests complete!

erjohnso:
Thank you very much @mgogoulos - merging now.

PR #453, opened 2015-02-13
[google compute] s/assertTrue/assertEqual/ as needed

mbrukman:
cc: @erjohnso for review.


erjohnso:
Wow, that was embarrassing.  Thankfully, they still all passed after fixing the tests.  Thanks Misha, working on merging now.

PR #454, opened 2015-02-13
[compute] s/self.assertTrue(True)/pass/

mbrukman:
The idiomatic "noop" for a Python block is `pass`, including unit tests.


allardhoeve:
This is a good change

PR #455, opened 2015-02-15
Vultr ssh key support

jcastillo2nd:
Added in support for vultr to get the ssh keys list similar to the DigitalOcean driver method ex_list_ssh_keys().

Also updated the create_node() method to accept a key id for creating vultr nodes with those keys.

### DIFF
--- a/libcloud/compute/drivers/vultr.py
+++ b/libcloud/compute/drivers/vultr.py
@@ -39,6 +39,17 @@ class VultrResponse(JsonResponse):
             raise LibcloudError(self.body)
 
 
+class SSHKey(object):
+    def __init__(self, id, name, pub_key):
+        self.id = id
+        self.name = name
+        self.pub_key = pub_key
+
+    def __repr__(self):
+        return (('<SSHKey: id=%s, name=%s, pub_key=%s>') %
+                (self.id, self.name, self.pub_key))
+
+
 class VultrConnection(ConnectionKey):
     """
     Connection class for the Vultr driver.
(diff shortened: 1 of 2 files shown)


Kami:
Thanks, I'll add some comments in-line.

libcloud/compute/drivers/vultr.py
```diff
@@ -119,6 +133,15 @@ def create_node(self, name, size, image, location):
 
         return created_node
 
+    def ex_list_ssh_keys(self):
```
Kami, line 136: Any particular reason why this is implemented as an extension method and not as a standard `list_key_pairs` method? (https://github.com/apache/libcloud/blob/trunk/libcloud/compute/base.py#L1160)

Looking at the code, it looks like you should be able to implement it using `list_key_pairs` method which is part of the standard API (use name for the id attribute, you can leave fingerprint as None or calculate it from the public key).

Kami, line 136: Also, looking at the docs at http://search.cpan.org/~eskaaren/WebService-Vultr-0.1/lib/WebService/Vultr.pm, it looks like vultr also returns public key so you can also set that attribute.

```
Example Response: { "541b4960f23bd": { "SSHKEYID": "541b4960f23bd", "date_created": null, "name": "test", "ssh_key": "ssh-rsa AA... test@example.com" } }
```

jcastillo2nd:
None actually. Was going off of the DigitalOcean API implementation, but it looks like they both can be addressed to implement list_key_pairs.

The more I spend time in the libcloud base classes the more I will get familiar with it. Still pretty new to this library, but I am using it for a couple of projects.

From a curl, it certainly does. The digitalocean v1 API doesn't provide the public key however. Is that why it was setup as an extension and not the standard method?

Kami:
@jcastillo2nd No problem at all and thanks for your contribution.

As far as DigitalOcean driver goes - I believe those extension methods were added to the DigitalOcean driver before standard API for key pair management was added to Libcloud. The new DigitalOcean API v2 now also returns a public key so those extension methods can be replaced with standard API methods.

jcastillo2nd:
I've previously re-forked the repo, and have this setup in another branch on my current fork. Closing this PR and opening a new one with the previous points addressed. New PR https://github.com/apache/libcloud/pull/530

PR #460, opened 2015-02-17
Fix HostVirtual list_records method when no records exist yet

vanclevstik:
`list_records` incorrectly returned `ZoneDoesNotExistError` when there were no records present.
Fixed that with additional checking if API call return "No Records" error and return empty list instead of exception.


Kami:
Fixed minor lint and version compatibility issue (968425c62da91b271b04975c5fa5b062f88594f6) and merged changes into trunk. Thanks.

PR #463, opened 2015-02-17
fixes create node for Host Virtual plus small enhancements

mgogoulos:
- return an empty list on list_nodes in case no servers are found, instead of an Exception
- pass args explicitly in create_node rather than using kwargs
- fix URL for build server, new url is /cloud/server/build/stub_node_id rather than /cloud/server/build that does not work currently
- adds a complete example in create_node for node creation with ssh key deployed


mgogoulos:
Closing this, as we are working with Dinesh Bhoopathy of HostVirtual for a more complete PR

PR #467, opened 2015-02-24
[LIBCLOUD-672] [LIBCLOUD-673] Create volume from snapshot on EC2 and OS

allardhoeve:
Both platforms support creating a volume from snapshots but do not implement it.

Please see individual commit messages for reasoning on why the changes are as they are.

### DIFF
--- a/libcloud/compute/base.py
+++ b/libcloud/compute/base.py
@@ -997,9 +997,9 @@ class NodeDriver(BaseDriver):
                                (optional)
         :type location: :class:`.NodeLocation`
 
-        :param snapshot:  Name of snapshot from which to create the new
-                               volume.  (optional)
-        :type snapshot:  ``str``
+        :param snapshot:  Snapshot from which to create the new
+                          volume.  (optional)
+        :type snapshot: :class:`.VolumeSnapshot`
 
         :return: The newly created volume.
         :rtype: :class:`StorageVolume`
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,12 @@ Changes with Apache Libcloud in development
 Compute
 ~~~~~~~
 
+- Add support for creating volumes based on snapshots to EC2 and OS drivers.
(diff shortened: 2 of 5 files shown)


allardhoeve:
@Kami, what do you think?

Kami:
Sorry for the delay, very busy with day job recently, will look into it...

libcloud/compute/base.py
```diff
@@ -997,9 +997,9 @@ def create_volume(self, size, name, location=None, snapshot=None):
                                (optional)
         :type location: :class:`.NodeLocation`
 
-        :param snapshot:  Name of snapshot from which to create the new
-                               volume.  (optional)
-        :type snapshot:  ``str``
+        :param snapshot:  Snapshot from which to create the new
```
Kami, line 1000: This is a breaking change - can you please document it in docs/upgrade_notes.rst?

Kami:
Besides the upgrade notes entry, the change looks good to me.

allardhoeve:
I notice there have been no upgrade notes since 1.16.0? So no notes for 1.17?

Kami:
Yeah, there were no special notes for 0.17. Just add a section for "in development" now.

allardhoeve:
Will document, but this is actually not a breaking change:

The base driver defines a signature that is not used by any of the drivers implementing `create_volume`.
1. `gandi` does not use the snapshot parameter.
2. `gce` uses the parameter, but accepts either a string or an object or None, so this breaks nothing.
3. `ibm` does not use the snapshot parameter.
4. `opennebula` does not use the snapshot parameter.
5. `profitbricks` breaks the contract entirely.

In this PR both `ec2` and `openstack` implement the signature like this. This matches `gce` partly, so all drivers using the snapshot functionality implement it using this signature.

allardhoeve:
Documented.

allardhoeve:
@Kami, done

Kami:
Merged, thanks.

PR #469, opened 2015-02-26
Route53 creating/updating records fixes

vanclevstik:
-  Before it was not possible to set an A record that points directly to domain zone name. Fixed it with correctly parsing empty name.
-  MX records priority was not handled correctly.

### DIFF
--- a/libcloud/dns/drivers/route53.py
+++ b/libcloud/dns/drivers/route53.py
@@ -372,6 +372,8 @@ class Route53DNSDriver(DNSDriver):
 
             rrecs = ET.SubElement(rrs, 'ResourceRecords')
             rrec = ET.SubElement(rrecs, 'ResourceRecord')
+            if 'priority' in extra:
+                data = '{0} {1}'.format(extra['priority'], data)
             ET.SubElement(rrec, 'Value').text = data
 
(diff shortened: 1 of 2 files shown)


libcloud/dns/drivers/route53.py
```diff
@@ -315,7 +315,7 @@ def _update_multi_value_record(self, record, name=None, type=None,
 
         rrs = ET.SubElement(change, 'ResourceRecordSet')
         ET.SubElement(rrs, 'Name').text = record.name + '.' + \
-            record.zone.domain
+            record.zone.domain if record.name else record.zone.domain
```
Kami, line 318: This is slightly confused / hard to read - would be easier if we used an actual if - I will make this change when merging the PR.

Kami:
Made the change (55ebfd199efa3ee19b5ea6a30a67cfc8bce9b76f) and merged code into trunk. Thanks.

libcloud/dns/drivers/route53.py
vanclevstik, line 318: Cool! Yeah I agree usual if makes it a bit more explicit.

vanclevstik:
Thanks!

PR #470, opened 2015-02-28
Use native match_hostname functionality from ssl module in Python >= 3.2

Kami:
Python >= 3.2 provides `match_hostname` functionality in the `ssl` module starting from Python 3.2.

With this change we now use native functionality on those versions and only require and use `backports.ssl_match_hostname` on Python < 3.2.


allardhoeve:
:+1:

PR #472, opened 2015-03-02
[LIBCLOUD-670] Improvements to HostVirtual driver

dineshbhoopathy:
- methods to handle 'Packages' concept [basically, it is a reserved compute instance with pre-allocated resources (RAM,Disk,IPs etc) in any specified location]
- changed create_node to use the extra packages method
- fix website url in docs
- fix method ordering
- fix kwargs abuse in create_node 
- fix docstrings
- add test cases for extra methods
- fix the fixtures
- fix default location


docs/compute/_supported_methods_block_storage.rst
```diff
@@ -100,7 +100,7 @@ Provider                              list volumes create volume destroy volume
 .. _`Gandi`: http://www.gandi.net/
```
Kami, line 100: Just FYI - those files are generated automatically using `./contrib/generate_provider_feature_matrix_table.py`.

Kami:
Changes look good to me. I went ahead, squashed the commits and merged them into trunk. Thanks.

PR #473, opened 2015-03-03
Add VolumeSnapshot.created attribute

allardhoeve:
One very important aspect of snapshots is when they are made, especially when used as historical backups. The field in VolumeSnapshot.extra that is used to represent this data varies between OS, EC2, GCE and Rackspace. Also, EC2 boasts a `datetime` value in `extra`, while OpenStack and derivatives use a plain string.

This PR adds a `created` attribute to VolumeSnapshot, that is a parsed date. This unified all implementations.

### DIFF
--- a/libcloud/compute/base.py
+++ b/libcloud/compute/base.py
@@ -545,12 +545,6 @@ class VolumeSnapshot(object):
     """
     A base VolumeSnapshot class to derive from.
     """
-    id = None
-    driver = None
-    size = None
-    extra = None
-    created = None
-
     def __init__(self, id, driver, size=None, extra=None, created=None):
         """
         VolumeSnapshot constructor.
--- a/libcloud/compute/base.py
+++ b/libcloud/compute/base.py
@@ -545,23 +545,38 @@ class VolumeSnapshot(object):
     """
     A base VolumeSnapshot class to derive from.
     """
-    def __init__(self, id, driver, size=None, extra=None):
+    id = None
+    driver = None
+    size = None
+    extra = None
+    created = None
+
+    def __init__(self, id, driver, size=None, extra=None, created=None):
         """
         VolumeSnapshot constructor.
 
         :param      id: Snapshot ID.
         :type       id: ``str``
 
+        :param      driver: The driver that represents a connection to the
+                            provider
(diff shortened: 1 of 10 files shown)


libcloud/compute/base.py
```diff
@@ -545,23 +545,38 @@ class VolumeSnapshot(object):
     """
     A base VolumeSnapshot class to derive from.
     """
-    def __init__(self, id, driver, size=None, extra=None):
+    id = None
+    driver = None
+    size = None
+    extra = None
+    created = None
```
allardhoeve: Added to aid introspection in all sorts of IDEs and ipython.

allardhoeve:
@Kami

allardhoeve:
- CloudFrames implements a CloudFramesSnapshot, which has nothing to do with a VolumeSnapshot :cry:

libcloud/test/compute/test_cloudstack.py
```diff
@@ -39,7 +40,7 @@
 from libcloud.test.file_fixtures import ComputeFileFixtures
 
 
-class CloudStackCommonTestCase(TestCaseMixin):
+class CloudStackCommonTestCase(unittest.TestCase, TestCaseMixin):
```
allardhoeve: To aid introspection in PyCharm.

libcloud/compute/base.py
Kami: The behavior is not exactly the same. Previously those variables were only instance variables, now they are also class variables which could have undesired consequences in some cases (https://stackoverflow.com/questions/207000/python-difference-between-class-and-instance-attributes).

I would prefer to keep it as it was. As far as introspection goes -I need to dig into it, but at least as far as helping with types go - you can help IDE's by adding type asserts and docstring annotation (we already do the latter in most places).

Kami:
Besides the class variables change, this PR looks good to me.

libcloud/compute/base.py
allardhoeve: I did not know this. My %(deity)s, how often I've used this incorrectly.
Thank you.

allardhoeve:
Removed instance variables and merged trunk into the branch.

PR #474, opened 2015-03-03
Allow testing using nosetests

allardhoeve:
Nosetests tried to execute all imported `TestCaseMixin`s. This patch lets it skip all `TestCaseMixin` base classes that are imported.

Yes, we test using tox, but nosetests is a superior test runner in that it let's you run a subset of the tests. Also, peoples preferences will vary.

Tox also allows a subset to be run by passing arguments on to `setup.py`, but our `setup.py` does not support the `-s` flag.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,2 +1,5 @@
 [wheel]
 universal = 1
+
+[nosetests]
+exclude=TestCaseMixin


Kami:
@allardhoeve Unrelated to this PR - have you received my email sent from tomaz at apache org to allardhoeve at gmail?

Kami:
As far as this PR goes - looks good, thanks.

```
...
Ran 5207 tests in 30.707s

OK
```

PR #476, opened 2015-03-05
Add state attribute to StorageVolume

allardhoeve:
1. Currently, volume state is defined by the driver in a key in `extra`. This key varies between drivers. Also, the value varies between drivers.
2. However, it is a commonly accessed attribute that should be unified across drivers, just as `Node` state is unified in `NodeState`.
3. This PR does exactly that.

Some caveats:
1. Except for OS and EC2, all other drivers return StorageVolume objects with state `UNKNOWN`.
   1. I have no access to GCE.
   2. I could find no exhaustive list of GCE disk states.
   3. The same holds for all the other drivers.
   4. I do not think this is a problem, because the instance variable does not currently exist at all :wink: 

UPDATE: GCE list is here: https://cloud.google.com/compute/docs/reference/latest/disks#resource

```
Acceptable values are:
"CREATING": The disk is being created.
"FAILED": The disk failed to be created.
"READY": The disk is ready to use.
"RESTORING": The disk is being restored from either a snapshot or an image.
```

### DIFF
--- a/libcloud/compute/drivers/ec2.py
+++ b/libcloud/compute/drivers/ec2.py
@@ -2422,7 +2422,7 @@ class BaseEC2NodeDriver(NodeDriver):
         :param      volume: Instance of ``StorageVolume``
         :type       volume: ``StorageVolume``
 
-        :param      name: Name of snapshot
+        :param      name: Name of snapshot (optional)
         :type       name: ``str``
 
         :rtype: :class:`VolumeSnapshot`
@@ -3445,7 +3445,8 @@ class BaseEC2NodeDriver(NodeDriver):
         Create tags for a resource (Node or StorageVolume).
 
         :param resource: Resource to be tagged
-        :type resource: :class:`Node` or :class:`StorageVolume`
+        :type resource: :class:`Node` or :class:`StorageVolume` or
+                        :class:`VolumeSnapshot`
 
         :param tags: A dictionary or other mapping of strings to strings,
                      associating tag names with tag values.
--- a/libcloud/compute/drivers/ec2.py
+++ b/libcloud/compute/drivers/ec2.py
@@ -41,7 +41,8 @@ from libcloud.compute.providers import Provider
 from libcloud.compute.base import Node, NodeDriver, NodeLocation, NodeSize
 from libcloud.compute.base import NodeImage, StorageVolume, VolumeSnapshot
 from libcloud.compute.base import KeyPair
-from libcloud.compute.types import NodeState, KeyPairDoesNotExistError
+from libcloud.compute.types import NodeState, KeyPairDoesNotExistError, \
+    StorageVolumeState
 
 __all__ = [
     'API_VERSION',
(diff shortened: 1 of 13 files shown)


gertjanol:
[EC2 Volume states](http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_Volume.html): `creating | available | in-use | deleting | deleted | error`
[OpenStack Volume states](http://developer.openstack.org/api-ref-blockstorage-v2.html#volumes-v2): `creating | available | attaching | in-use | deleting | error | error_deleting | backing-up | restoring-backup | error_restoring | error_extending`

allardhoeve:
Where did you find this?

allardhoeve:
Fixed unknown node states, thanks @gertjanol

allardhoeve:
@Kami, does this need a proposal @ dev?

libcloud/compute/drivers/ec2.py
```diff
@@ -4683,6 +4695,14 @@ def _to_volume(self, element, name=None):
         volId = findtext(element=element, xpath='volumeId',
                          namespace=NAMESPACE)
         size = findtext(element=element, xpath='size', namespace=NAMESPACE)
+        raw_state = findtext(element=element, xpath='status',
+                             namespace=NAMESPACE)
+
+        try:
+            state = self.VOLUME_STATE_MAP[raw_state]
```
Kami: Minor style thing - doing `self.VOLUME_STATE_MAP.get(raw_state, StorageVolumeState.UNKNOWN)` saves a couple of lines of code. On top of that, code also won't break and require an immediate update if a provider adds a new state (unlikely, but still).

libcloud/compute/types.py
```diff
@@ -221,6 +221,18 @@ def fromstring(cls, value):
         return getattr(cls, value.upper(), None)
 
 
+class StorageVolumeState(object):
+    UNKNOWN = None
```
Kami: Should probably also use a number for `unknown` - that's what other ENUM classes to and this way we can differentiate between "no state" and "state not recognized by libcloud".

libcloud/compute/drivers/ec2.py
allardhoeve: Done.

libcloud/compute/types.py
allardhoeve: And done.

Kami:
Need to update the changelog.

Besides that, LGTM, +1

allardhoeve:
thanks!

allardhoeve:
Documented the changelog. Will wait for the build, then merge.

Kami:
Looks like the build is stuck :/

allardhoeve:
Travis is very, very slow lately

Kami:
Yeah, build just seems to be stuck. Probably best to just merge since since we don't have admin permissions and we can't restart the build (we could push a "dummy" commit but that's annoying...).

PR #482, opened 2015-03-10
LIBCLOUD-676: Allow describing AMI tags

jlk:
Per the referenced Jira - I can't think of a reason for why ex_describe_tags() shouldn't be able to be used for other resources besides instances and storage, so removed the filter.

### DIFF
--- a/libcloud/compute/drivers/ec2.py
+++ b/libcloud/compute/drivers/ec2.py
@@ -3404,10 +3404,11 @@ class BaseEC2NodeDriver(NodeDriver):
 
     def ex_describe_tags(self, resource):
         """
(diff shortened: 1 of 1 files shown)


allardhoeve:
Yeah, that description was faulty anyway. Looks good to me.

Kami:
Yeah, looks good to me as well.

I've also checked the EC2 docs and `resource-type` filter is indeed optional (http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeTags.html).

Kami:
Fixed lint issue and merged patch into trunk. Thanks.

PR #484, opened 2015-03-10
[LIBCLOUD-678] Enable to pass arguments to ex_list_portforwarding_rules ...

atsaki:
CloudStack driver's ex_list_port_forwarding_rules isn't accept argument now. 

This patch enables to pass arguments to ex_list_port_forwarding_rules to get filtered result.


Kami:
Merged into trunk. Thanks.

PR #486, opened 2015-03-13
Consistent logic for dependencies resolution

cgtx:
Different versions of Python have different dependencies.  These differences
are parsed in various ways.  This pull request establishes a consistent method
for setting dependencies per version.

In addition, the code from `backports.ssl_match_hostname` has been merged into
the standard library in Python 2.7.9.  The condition for adding that dependency
now takes into account that fact.

https://www.python.org/downloads/release/python-279/

> The entirety of Python 3.4's ssl module has been backported for Python 2.7.9.
> See PEP 466 for justification.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,6 +33,16 @@ from libcloud.utils.dist import get_packages, get_data_files
 
 libcloud.utils.misc.SHOW_DEPRECATION_WARNING = False
 
+# Different versions of python have different requirements.  We can't use
+# libcloud.utils.py3 here because it relies on backports dependency being
+# installed / available
+PY2 = sys.version_info.major == 2
+PY3 = sys.version_info.major == 3
+pre_PY25 = PY2 and sys.version_info < (2, 5)
+pre_PY26 = PY2 and sys.version_info < (2, 6)
+pre_PY279 = PY2 and sys.version_info < (2, 7, 9)
+pre_PY32 = PY3 and sys.version_info < (3, 2)
+post_PY27 = PY2 and sys.version_info >= (2, 7)
 
 HTML_VIEWSOURCE_BASE = 'https://svn.apache.org/viewvc/libcloud/trunk'
 PROJECT_BASE_DIR = 'http://libcloud.apache.org'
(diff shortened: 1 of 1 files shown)


Kami:
I intentionally avoided Python 2.7.9 check to avoid additional complexity. Additional complexity mostly results because we can't use re-use libcloud.utils.py3 here because of the import issues.

In any case, now that you have re-factored the code a bit I'm fine with additional check. Thanks.

setup.py
```diff
@@ -231,10 +235,10 @@ def run(self):
 forbid_publish()
 
 install_requires = []
-if pre_python26:
+if pre_PY26:
     install_requires.extend(['ssl', 'simplejson'])
 
-if sys.version_info < (3, 2):
+if pre_PY279 or pre_PY32:
```
Kami: I was confused for a second since I thought `pre_PY32` means any version before 3.2 so this check wouldn't work, but it really means any Python 3 version which is smaller than 3.2.

Should probably rename the variables to make this more obvious.

cgtx:
@Kami Fair point, but I don't have a clue what variable name to use.  What do you suggest?  The best I can come up with is something like `PY2_pre_279`.

Kami:
@ctx Yeah, I'm fine that name.

cgtx:
@Kami Variables have been renamed.  I also inverted post_PY27 into PY2_pre_27.  This makes the unittest2_required check a little more sane.

setup.py
```diff
@@ -33,6 +33,16 @@
 
 libcloud.utils.misc.SHOW_DEPRECATION_WARNING = False
 
+# Different versions of python have different requirements.  We can't use
+# libcloud.utils.py3 here because it relies on backports dependency being
+# installed / available
+PY2 = sys.version_info.major == 2
```
Kami, line 39: This won't work with Python version <= 2.6 where `sys.version_info` is a tuple. I will fix it when merging the patch.

Kami:
Merged into trunk. Thanks.

setup.py
Kami, line 39: Fix in 88e3c822b3ed7534f085c37d8beb1c230a828ea2

PR #487, opened 2015-03-17
Add Onapp IaaS platform support

MatthiasWiesnerCC:


Kami:
Cool, thanks. Will have a look asap.

prepare_tests.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
Kami: Can you please add Apache 2.0 license headers to those non-code related files and move them to `contrib/`?

Kami:
Can you please add some documentation?

See `docs/compute/drivers/vsphere.rst` for an example.

libcloud/compute/drivers/onapp.py
```diff
@@ -0,0 +1,842 @@
...
+                                data=json.dumps(data),
+                                headers={
+                                    "Content-type": "application/json"},
+                                method="PUT")
+
+    #
+    # Methods to be implemented
+    #
+    def list_sizes(self, location=None):
+        pass
```
Kami: Instead of having those methods here, either add `raise NotImplementedError` or remove them - base NodeDriver class already throws this error by default if the method is not overridden in a sub class.

This way it's possible to programatically check and know that those methods are unimplemented. Python returns implicit `None` by default so doing that means it's hard / impossible to programatically check if those methods are implemented.

libcloud/compute/drivers/onapp.py
```diff
@@ -0,0 +1,842 @@
...
+
+    @staticmethod
+    def _create_args_to_params(**kwargs):
+        """
+        Extract server params from keyword args to create a VS
+
+        :param   kwargs: keyword args
+        :return: ``dict``
+        """
+        server_params = dict()
```
Kami: Looks like the size of this method could be reduced drastically by using a for loop (we follow a similar pattern in some other drivers) :)

MatthiasWiesnerCC:
- move `run_test.sh` to `contrib`
- added onapp doc, with example
- removed unused methods
- shortened helper method

PR #489, opened 2015-03-31
[LIBCLOUD-683] add optional extra fields to Openstack Node responses

jimbobhickville:
Several fields utilized by Rackspace Public Cloud were
not being returned.  A few of them are from Openstack
core, and some are from extensions used by Rackspace.


Kami:
Merged into trunk. Thanks.

PR #491, opened 2015-03-31
Add support for GCE list paging and filtering

verb:
GCE will return a maximum of 500 resources in a single list. This change
adds an iterator that allows filtering and/or paging of list results.

### DIFF
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -64,7 +64,28 @@ class GCEResponse(GoogleResponse):
 
 
 class GCEConnection(GoogleBaseConnection):
-    """Connection class for the GCE driver."""
+    """
+    Connection class for the GCE driver.
+
+    GCEConnection extends :class:`google.GoogleBaseConnection` for 2 reasons:
+      1. modify request_path for GCE URI.
+      2. Implement gce_params functionality described below.
+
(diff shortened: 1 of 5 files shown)


Kami:
Will dig in ASAP.

On a related note - we already have an establish pattern for handling paginated collections which should probably also work here. See https://github.com/apache/libcloud/blob/trunk/libcloud/dns/drivers/zerigo.py#L149 for an example.

erjohnso:
Ah, good point. Reminded me that the Google DNS driver also uses generators: https://github.com/apache/libcloud/blob/trunk/libcloud/dns/drivers/google.py

verb:
Ah, didn't see that. It looks like this would be the first time iterate_\* is used in the compute driver.

It would be pretty easy to overlay the iterate_\* functions on top of GCEList if my underlying approach is still sound. This would keep the filter functionality available in ex_list().

Alternatively we could establish a pattern for filter, too.

verb:
friendly ping and just checking in to make sure nothing is blocked on me.

Kami:
@erjohnso can you please review this one and take it across the finish line?

Kami:
@verb Sorry for the delay. I have quickly glanced over it and it looks good to me, but I haven't done any testing.

erjohnso:
Sorry for the delay, @verb!  Thanks!!

PR #496, opened 2015-04-02
Support diskSizeGb and local SSD when creating an instance on GCE

phildav:
GCE has options to specify the size of boot disk when creating an instance, as well as adding a local SSD to the instance.
This pull request adds 2 new options in create_node:
- ex_boot_disk_size_gb to specify the size of the boot disk to create. If not set the default GCE was is used.
- ex_with_local_ssd to attach a local SSD and specify which interface to use.

### DIFF
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -2216,11 +2216,12 @@ n
                 ex_disks_gce_struct[0]['initializeParams']['diskSizeGb'] = \
                     ex_boot_disk_size_gb
             if ex_with_local_ssd:
-                disk_type_local_ssd = self.ex_get_disktype('local-ssd', zone=location)
+                disk_type_local_ssd = self.ex_get_disktype('local-ssd',
+                                                           zone=location)
                 ex_disks_gce_struct.append(
                     {
                         "type": "SCRATCH",
-                        "initializeParams":{
+                        "initializeParams": {
                             "diskType": disk_type_local_ssd.extra['selfLink']
                         },
                         "autoDelete": True,
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -2049,7 +2049,7 @@ class GCENodeDriver(NodeDriver):
     def create_node(self, name, size, image, location=None,
                     ex_network='default', ex_tags=None, ex_metadata=None,
                     ex_boot_disk=None, use_existing_disk=True,
-                    ex_boot_disk_size_gb=None, ex_with_local_ssd=False,
+                    ex_boot_disk_size_gb=None, ex_with_local_ssd=None,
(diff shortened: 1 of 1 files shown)


erjohnso:
Hi @phildav - thanks for the contribution!

As you can see, `create_node` is becoming quite cluttered and tricky to keep updated with GCE API changes and ensure backwards compatibility.  Thats why, in a recent change, I punted on providing new arguments for things like local SSD and instead just allowed for a generic JSON/python argument `ex_disks_gce_struct`.  The idea obviously is to let the programmer specify anything here that the upstream GCE API accepts for disks.

I can understand philosophically in wanting new params for local ssd and disk size, but it's not strictly necessary, correct?

Also, if we go with this pull-request, I'd like to see tests (and rigorous ones) that ensure backwards compatibility.

phildav:
Hi @erjohnso , thank you for the quick reply.
The reason why I added this into libcloud is that I'm using salt cloud, and it's not aware of `ex_disks_gce_struct`. Building this structure is not totally obvious since it require to know what gce expects. Although it does not belong to the list of features in common with other clouds, my feeling is that it make sense to have this in libcloud instead of its multiple users.
Let me know which direction you want to go with this PR: write specific tests, rework disk options or close it and wait for a better idea...

erjohnso:
Hi @phildav 

My preference would be to see `salt-cloud` extended to support a `local-ssd` type,

https://github.com/saltstack/salt/blob/develop/salt/cloud/clouds/gce.py#L2047

which I think would be cleaner from a Salt user's perspective (rather than requiring them to specify a new attribute/parameter).

But, this would then require and update to the GCE driver to properly format the `ex_disks_gce_struct` and adding that to `**kwargs` before creating the new VM,

https://github.com/saltstack/salt/blob/develop/salt/cloud/clouds/gce.py#L2047

I know it's a bit more work inside salt-cloud, but I do think it's cleaner that way.

wdyt?

phildav:
Sure it can be done in salt-cloud. But then every client of libcloud who wants to support these options has to know
- how to build the `ex_disks_gce_struct` properly
- resolve diskType and image links with `ex_get_disktype` and `ex_get_image`
- how to build the request to get local SSD to work (took me a few tries).

For what it's worth (I'm not a long time user of libcloud), since it's not trivial to build this `ex_disks_gce_struct` structure then I would put that code in common.
I agree that `create_node` may not be the best place if that piece of code gets large, so maybe we can add a helper that creates `ex_disks_gce_struct` properly in libcloud so it's available for every user ?

erjohnso:
@phildav - sorry for the lag.

I'm OK moving forward with your approach to extend libcloud's `create()` call extended to include local-ssd, but would like to see some stringent tests included.  When it comes time to roll this up into `salt-cloud`, I'd still suggest exposing it as `local-ssd` to the user as I suggested versus setting a new param.

PR #498, opened 2015-04-03
[LIBCLOUD-686] Address validation fails with CPython 2.7.9 on Windows.

SamuelMarks:


SamuelMarks:
I'm uncertain if my use of check on `force_ipv4` is valid. When IPv4 isn't forced, that doesn't mean it's not allowed; right?

Kami:
Thanks for the PR.

There is already a similar PR which tries to address that - #343. We should figure out how to consolidate those two and get it merged.

I personally think approach in #343 is better start since it handles this difference in the function where this problem can arise.

What do you think?

Kami:
Sorry for the delay. I've added a check for validating IPv6 address on Windows (e105433e941262e03eb3bb922db47dabdd8efa75) and merged changes into trunk.

Thanks.

PR #502, opened 2015-04-09
[LIBCLOUD-691] Add Onapp IaaS platform support

MatthiasWiesnerCC:
Corresponding issue:
https://issues.apache.org/jira/browse/LIBCLOUD-691


Kami:
Great, thanks.

I will review it and add the comments inline.

libcloud/compute/drivers/onapp.py
```diff
@@ -0,0 +1,540 @@
...
+    """
+    Base OnApp node driver.
+    """
+
+    connectionCls = OnAppConnection
+    type = Provider.ONAPP
+    name = 'OnApp'
+    website = 'http://onapp.com/'
+
+    def create_node(self, label, memory, cpus, cpu_shares, hostname,
```
Kami: This method is part of the standard API so all the arguments which are not part of standard API (looks like all of them :)) should have an `ex_` prefix.

Kami:
@MatthiasWiesnerCC bump^^

TooAngel:
I took over the onapp integration on our side. I prefixed the non standard parameter with `ex_`. Also simplified the implementation to have less onapp specific classes and renamed the parameters which can be mapped to libcloud standard parameters.

TooAngel:
@Kami bump^^ :-)

Kami:
Sorry for the delay.

The changes looked OK (except a weird test failure - see below), so I went ahead and merged it into trunk. Thanks.

There was a just a weird test failure which only happened on every Python version except 2.7.

``` ======================================================================
FAIL: test_create_node (__main__.OnAppNodeTestCase)
  File "libcloud/test/compute/test_onapp.py", line 67, in test_create_node
    self.assertEqual('onapp-new-fred', node.name)
AssertionError: 'onapp-new-fred' != <MagicMock name='mock.object.__getitem__().__getitem__()' id='139971643315264'>
```

It turned out the failure was related to using `is` to compare string values in the test (b3ca86bf89563c8f99d9abb2611da8324d1fbcad). The `is` operator compares the object pointers and not the values so it should only be used to check for values such as `True` and `None`.

The test passing on 2.7 was just a pure luck :)

libcloud/test/compute/test_onapp.py
```diff
@@ -0,0 +1,111 @@
...
+from libcloud.test.file_fixtures import ComputeFileFixtures
+from libcloud.compute.drivers.onapp import OnAppNodeDriver
+from libcloud.test import LibcloudTestCase
+from libcloud.test.secrets import ONAPP_PARAMS
+from libcloud.compute.base import Node
+
+
+class OnAppNodeTestCase(LibcloudTestCase):
+    def setUp(self):
+        def _request(*args, **kwargs):
```
Kami, line 16: Also, just a quick heads up for the future - we have a framework in place for mocking HTTP responses (pretty much every driver tests use it).

So in the future, please use follow the approach other tests use.

TooAngel:
Thank you for merging.

I also changed the tests to the libcloud test framework (https://github.com/apache/libcloud/commit/fec2964059f2bf7339c81875ee20304953feb4d5). Should I open a separate issue for that, send the pull request (https://github.com/apache/libcloud/pull/560), ...?

PR #504, opened 2015-04-13
[LIBCLOUD-692] Run tests in a docker container

MatthiasWiesnerCC:
Corresponding issue:
https://issues.apache.org/jira/browse/LIBCLOUD-692


tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py25,py26,py27,pypy,py32,py33,py34,lint
```
Kami, line 2: We still support 2.5 so please add this back and you can then explicitly specify which environments to use in the Dockerfile - `CMD tox -e py26,py27,pypy,py32,py33,py34,lint`.

Kami:
Thanks - the change looks good to me.

Can you please also add some documentation on how to use docker to run the tests to `docs/testing.rst`.

tox.ini
Kami, line 2: Or ideally, we would fine a base Docker image (e.g. older version of Ubuntu) where we can easily also install Python 2.5 or Python 2.5 PPA which we can use with a newer version of Ubuntu.

Kami:
Added some docs (ccc53c5202f5aae7fd0e5ad3bf64d26ed3a4681a) and merged changes into trunk.

Thanks.

PR #508, opened 2015-04-23
Add a 'break' in  iterate_container_objects()

ghost:
I think the 'while' circulation(line. 730) in function iterate_container_objects() is a endless loop. We just need to execute the 'for' sentence on line 747 once, so I add a 'break' on line 750.

### DIFF
--- a/libcloud/storage/drivers/cloudfiles.py
+++ b/libcloud/storage/drivers/cloudfiles.py
@@ -747,6 +747,7 @@ class CloudFilesStorageDriver(StorageDriver, OpenStackDriverMixin):
                 for obj in objects:
                     yield obj
                 params['marker'] = obj.name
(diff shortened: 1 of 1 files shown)


Kami:
Thanks.

It appears the code is actually working as intended - in case there are multiple pages, we don't want to break out of the loop. We just want to update the `marker` query string parameter which will result in us retrieving results from the next page and eventually when we are on the last page either one of the following termination conditions should be reached:
- No content status code is returned.
- There are no more objects on the last page.

Having said that, there might actually be a bug in how we handle the last page. I will look into trying to reproduce that case (multiple pages, last page has >= 1 objects).

It would also help if you can gist the output of running your code which results in an infinite loop using `LIBCLOUD_DEBUG`.

PR #509, opened 2015-04-23
Reset file's current position in read_in_chunks()

ghost:
If a cloud storage driver's parameter  supports_chunked_encoding is False,  the function read_in_chunks will be called twice during the upload_object_via_stream progress. 

**read_in_chunks()  be called in _upload_object() at first (line. 629  libcloud.storage.base)**

``` ruby
……
if iterator:
            if self.supports_chunked_encoding:
                headers['Transfer-Encoding'] = 'chunked'
                upload_func_kwargs['chunked'] = True
            else:
                # Chunked transfer encoding is not supported. Need to buffer
                # all the data in memory so we can determine file size.
                iterator = libcloud.utils.files.read_in_chunks(
                    iterator=iterator)
                data = libcloud.utils.files.exhaust_iterator(iterator=iterator)

                file_size = len(data)
                upload_func_kwargs['data'] = data
        else:
            file_size = os.path.getsize(file_path)
            upload_func_kwargs['chunked'] = False
……
```

**Then, read_in_chunks()  be called in _stream_data() (line. 734  libcloud.storage.base)**

``` ruby
……
generator = libcloud.utils.files.read_in_chunks(iterator, chunk_size)

bytes_transferred = 0
        try:
            chunk = next(generator)
        except StopIteration:
……
```

But in the second time, the current position of the file to be uploaded is the end, so the StopIteration exception will be raised and upload a empty file. So, I make this change in read_in_chunks.

### DIFF
--- a/libcloud/utils/files.py
+++ b/libcloud/utils/files.py
@@ -58,6 +58,9 @@ def read_in_chunks(iterator, chunk_size=None, fill_size=False,
     """
     chunk_size = chunk_size or CHUNK_SIZE
 
+    if isinstance(iterator, file) and iterator.tell():
+        iterator.seek(0, OS.SEEK_SET)
+        
     if isinstance(iterator, (file, httplib.HTTPResponse)):
         get_data = iterator.read
         args = (chunk_size, )


Kami:
Thanks.

Hm, I just had a look at the code and it seems like we should never end up in a scenario when `_stream_data` is called with `data` argument which would result in iterator being exhausted twice. In short, your code should be unnecessary (it seems to just be working around a deeper bug / problem somewhere else).

In case `data` is provided, `_upload_data` method should be called and not `_stream_data`.

Can you please also provide a specific piece of code which reproduces this scenario? Thanks.

PR #511, opened 2015-05-04
use get instead of pop, for region details dict

mgogoulos:


Kami:
Can you please provide more context and background on this change - why is it needed? what breaks with it? regression test? etc.

Thanks :)

libcloud/compute/drivers/ec2.py
```diff
@@ -5659,7 +5659,7 @@ def __init__(self, key, secret=None, secure=True, host=None, port=None,
         self.region_name = region
         self.api_name = details['api_name']
         self.country = details['country']
-        self.signature_version = details.pop('signature_version',
+        self.signature_version = details.get('signature_version',
```
Kami, line 5662: I think this won't work because the parent class constructor method doesn't take `signature_version` argument. That's also why `pop` is used in the first place, but I could also be wrong.

mgogoulos:
This fix was required for me in order for ec2 Frankfurt to work (uses signature_version 4). I noticed that under circumstances of running connections in parallel it would use signature_version 2 (and thus fail to authenticate), and it has worked without problems with this small patch (also in other regions as well). Unfortunately I cannot be explicit on how to reproduce the bug though

Kami:
@mgogoulos Alright, thanks.

I thought those issues shouldn't happen anymore since we now use `_ex_connection_kwargs` instead of class variables, but there might still be a bug hiding somewhere in that code so I will dig in and try to reproduce the issue and find the root cause.

Kami:
@mgogoulos Looks like #523 which fixes this has been merged.

I originally missed a bit of context here since I didn't expand the diff - we pop item from `REGION_DETAILS` which is a global constant and we didn't create a new copy first which means global constant was modified which would cause things to break.

Thanks again.

libcloud/compute/drivers/ec2.py
jimbobhickville, line 5662: It doesn't appear to be passing it along to the parent constructor at all.  Doing `.pop()` is modifying the global `REGION_DETAILS` for the duration of the process isn't it?

mgogoulos, line 5662: this was opened as another PR and merged - https://github.com/apache/libcloud/pull/523 - forgot to close it!

PR #513, opened 2015-05-08
LIBCLOUD-700 Add support for setting the region

tonybaloney:
when connecting to a Dimension Data cloud, added all public endpoints.

### DIFF
--- a/libcloud/compute/drivers/dimensiondata.py
+++ b/libcloud/compute/drivers/dimensiondata.py
@@ -140,6 +140,18 @@ class DimensionDataConnection(ConnectionUserAndKey):
 
     allow_insecure = False
 
+    def __init__(self, user_id, key, secure=True, host=None, port=None,
+                 url=None, timeout=None, proxy_url=None, **conn_kwargs):
+        super(DimensionDataConnection, self).__init__(
+            user_id=user_id,
+            key=key,
+            secure=secure,
+            host=host, port=port,
+            url=url, timeout=timeout,
+            proxy_url=proxy_url)
+        if conn_kwargs['region'] is not None:
+            self.host = conn_kwargs['region'].get('host')
+
     def add_default_headers(self, headers):
         headers['Authorization'] = \
             ('Basic %s' % b64encode(b('%s:%s' % (self.user_id,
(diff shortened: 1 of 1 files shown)


libcloud/compute/drivers/dimensiondata.py
```diff
@@ -203,12 +243,27 @@ class DimensionDataNodeDriver(NodeDriver):
...
+        if region not in API_ENDPOINTS:
+            raise ValueError('Invalid region: %s' % (region))
+
+        self.selected_region = API_ENDPOINTS[region]
+
+        super(DimensionDataNodeDriver, self).__init__(key=key, secret=secret,
+                                                      secure=secure, host=host,
+                                                      port=port,
+                                                      region=region, **kwargs)
+        self.connection.host = self.selected_region.get('host')
```
Kami: I believe this won't work correctly when user instantiates multiple drivers with a different region and runs those inside the same process.

You should use `_ex_connection_kwargs` approach. This way, `host` argument will be passed to the Connection class constructor when instantiating it instead of modifying a class level variable. Please see EC2 and some other drivers for an example.

tonybaloney:
I based it on the joyent driver, which follows a similar pattern (so of note that one may suffer the same issue!).

Running DimensionData and Rackspace side by side does give results for the DimensionData driver, then 
    pprint(driver.list_nodes())
NameError: name 'driver' is not defined

Will follow the EC2 pattern instead and resubmit changes

tonybaloney:
Updated and tested against multiple regions in the same process by: 

DimensionData = get_driver(Provider.DIMENSIONDATA)

drivers = [DimensionData('anthony','****_',region='dd-na'),
           DimensionData('anthony','**_**',region='dd-au')]

for driver in drivers:
     nodes = driver.list_nodes()
     print(len(nodes))

Gives 14 and 154 as expected.

Kami:
@tonybaloney Thanks, I will have a look.

Also, to clarify it, your example wouldn't uncover the bug. You would need to do something like this:

``` python
driver1 = cls(region='region-1')
driver2 = cls(region='region-2')

pprint(driver1.list_nodes())
pprint(driver2.list_nodes())
pprint(driver1.list_nodes())  # If there is a bug, this line would result in region from driver2 also being used here
```

tonybaloney:
print(len(driver1.list_nodes()))
print(len(driver2.list_nodes()))
print(len(driver1.list_nodes()))

gives 5, 154 and 5, so its not giving that bug

Kami:
I made a minor change (use normal dictionary access instead of get - we should throw if dictionary doesn't contain "host" item since this most likely indicates a programmer error) and merged changes into trunk.

Thanks.

PR #514, opened 2015-05-13
[LIBCLOUD-707] Fixed exception when parsing CloudSigma static IPs

chrisob:
Fixes issue with CloudSigma compute driver: https://issues.apache.org/jira/browse/LIBCLOUD-707


Kami:
Merged into trunk. Thanks.

PR #515, opened 2015-05-13
[LIBCLOUD-360] [LIBCLOUD-709] Retry on rate limit and connection timeout.

arpithkp:
[Retry Logic](https://issues.apache.org/jira/browse/LIBCLOUD-360)
[Rate Limit Logic](https://issues.apache.org/jira/browse/LIBCLOUD-709)

### DIFF
--- a/libcloud/common/abiquo.py
+++ b/libcloud/common/abiquo.py
@@ -180,11 +180,14 @@ class AbiquoConnection(ConnectionUserAndKey, PollingConnection):
     responseCls = AbiquoResponse
 
     def __init__(self, user_id, key, secure=True, host=None, port=None,
-                 url=None, timeout=None):
+                 url=None, timeout=None,
+                 retry_delay=None, backoff=None):
         super(AbiquoConnection, self).__init__(user_id=user_id, key=key,
                                                secure=secure,
                                                host=host, port=port,
-                                               url=url, timeout=timeout)
+                                               url=url, timeout=timeout,
+                                               retry_delay=retry_delay,
+                                               backoff=backoff)
 
         # This attribute stores data cached across multiple request
         self.cache = {}
--- a/libcloud/common/base.py
+++ b/libcloud/common/base.py
@@ -44,10 +44,11 @@ from libcloud.utils.py3 import StringIO
 from libcloud.utils.py3 import u
 from libcloud.utils.py3 import b
 
-from libcloud.utils.misc import lowercase_keys
+from libcloud.utils.misc import lowercase_keys, retry
 from libcloud.utils.compression import decompress_data
-from libcloud.common.types import LibcloudError, MalformedResponseError
 
+from libcloud.common.exceptions import exception_from_message
+from libcloud.common.types import LibcloudError, MalformedResponseError
 from libcloud.httplib_ssl import LibcloudHTTPConnection
 from libcloud.httplib_ssl import LibcloudHTTPSConnection
 
@@ -114,7 +115,9 @@ class Response(object):
             self.body = b(self.body).decode('utf-8')
 
         if not self.success():
-            raise Exception(self.parse_error())
+            raise exception_from_message(code=self.status,
+                                         message=self.parse_error(),
+                                         headers=self.headers)
 
         self.object = self.parse_body()
 
@@ -460,11 +463,13 @@ class Connection(object):
     driver = None
     action = None
     cache_busting = False
+    backoff = None
+    retry_delay = None
 
     allow_insecure = True
 
     def __init__(self, secure=True, host=None, port=None, url=None,
-                 timeout=None, proxy_url=None):
+                 timeout=None, proxy_url=None, retry_delay=None, backoff=None):
         self.secure = secure and 1 or 0
         self.ua = []
         self.context = {}
@@ -492,9 +497,11 @@ class Connection(object):
             (self.host, self.port, self.secure,
              self.request_path) = self._tuple_from_url(url)
 
-        if timeout:
-            self.timeout = timeout
-
+        if timeout is None:
+            timeout = self.__class__.timeout
+        self.timeout = timeout
+        self.retry_delay = retry_delay
+        self.backoff = backoff
         self.proxy_url = proxy_url
 
     def set_http_proxy(self, proxy_url):
@@ -675,6 +682,9 @@ class Connection(object):
         else:
             headers = copy.copy(headers)
 
+        retry_enabled = os.environ.get('LIBCLOUD_RETRY_FAILED_HTTP_REQUESTS',
+                                       False)
+
         action = self.morph_action_hook(action)
         self.action = action
         self.method = method
@@ -738,8 +748,17 @@ class Connection(object):
 
                 self.connection.endheaders()
             else:
-                self.connection.request(method=method, url=url, body=data,
-                                        headers=headers)
+                if retry_enabled:
+                    retry_request = retry(timeout=self.timeout,
+                                          retry_delay=self.retry_delay,
+                                          backoff=self.backoff)
+                    retry_request(self.connection.request)(method=method,
+                                                           url=url,
+                                                           body=data,
+                                                           headers=headers)
+                else:
+                    self.connection.request(method=method, url=url, body=data,
+                                            headers=headers)
         except ssl.SSLError:
             e = sys.exc_info()[1]
             self.reset_context()
@@ -946,7 +965,7 @@ class ConnectionKey(Connection):
     Base connection class which accepts a single ``key`` argument.
     """
     def __init__(self, key, secure=True, host=None, port=None, url=None,
-                 timeout=None, proxy_url=None):
+                 timeout=None, proxy_url=None, backoff=None, retry_delay=None):
         """
         Initialize `user_id` and `key`; set `secure` to an ``int`` based on
         passed value.
@@ -954,7 +973,9 @@ class ConnectionKey(Connection):
         super(ConnectionKey, self).__init__(secure=secure, host=host,
                                             port=port, url=url,
                                             timeout=timeout,
-                                            proxy_url=proxy_url)
+                                            proxy_url=proxy_url,
+                                            backoff=backoff,
+                                            retry_delay=retry_delay)
         self.key = key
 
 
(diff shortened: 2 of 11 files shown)


Kami:
Thanks.

I will have a look asap.

arpithkp:
@Kami : could you please take a look at this.

Kami:
I think having some kind of retry mechanism in Libcloud is reasonable, but I haven't had the time to look at the implementation yet.

In the past, we were against that and allowed users to built their own retry mechanism on top, but this was a less of an issue back in the past where 1 method call usually mapped to a single API call, but these days, a lot of methods result into multiple API calls.

In general, it should probably also be opt-in to start with (or maybe start with retry only for "GET" requests) since some operations are not idempotent and may return an error but still complete which could potentially result in additional resource being created.

Kami:
As far as the implementation goes - I will try to have a look ASAP, but unlikely that this will happen earlier than this weekend since I'm traveling and mostly offline in the next couple of days.

arpithkp:
@Kami any update on this

jimbobhickville:
I helped @arpithkp work through this issue and make sure that it was as backwards-compatible as possible.  Any code that is looking at the raw exception messages should still get the same string back, and new code can take advantage of having the response information as part of the exception, to check for specific codes.  

We're happy to rework things if the approach isn't to the liking of the core libcloud folks, but this is something we really need badly to improve the reliability of our product, so if we could get some feedback ASAP, that would rock.

Kami:
@arpithkp @jimbobhickville Sorry for the delay. Busy with day job and haven't managed to get to it yet.

I will try to look into it today.

libcloud/common/abiquo.py
```diff
@@ -180,11 +180,12 @@ class AbiquoConnection(ConnectionUserAndKey, PollingConnection):
     responseCls = AbiquoResponse
 
     def __init__(self, user_id, key, secure=True, host=None, port=None,
-                 url=None, timeout=None):
+                 url=None, timeout=None, **kwargs):
```
Kami: Would it be possible to explicitly declare all the possible arguments (https://libcloud.readthedocs.org/en/latest/development.html#don-t-abuse-kwargs)?

Would make it a lot more clear and obvious on what is going on (aka what arguments are expected).

libcloud/common/base.py
```diff
@@ -460,11 +463,13 @@ class Connection(object):
...
     action = None
     cache_busting = False
+    backoff = None
+    retry_delay = None
 
     allow_insecure = True
 
     def __init__(self, secure=True, host=None, port=None, url=None,
-                 timeout=None, proxy_url=None):
+                 timeout=None, proxy_url=None, retry_delay=None, backoff=None):
```
Kami, line 472: Please add a docstring with a short explanation for the new arguments.

libcloud/common/base.py
```diff
@@ -495,6 +500,12 @@ def __init__(self, secure=True, host=None, port=None, url=None,
         if timeout:
             self.timeout = timeout
 
+        if retry_delay:
+            self.retry_delay = retry_delay
```
Kami: I think it's better if we always set the instance variable - if user doesn't provide a value it will simply be `None`.

libcloud/common/exception.py
```diff
@@ -0,0 +1,72 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
```
Kami: Minor thing - can you please call the module `exceptions.py` (plural) so it's consistent with other files (types, etc.)?

libcloud/utils/misc.py
```diff
@@ -282,3 +292,28 @@ def __repr__(self):
 
     def __str__(self):
         return str(self.__repr__())
+
+
+def retry(timeout, retry_delay, backoff, func, *args, **kwargs):
+    """
+    Retry decorator helps to handle common exception.
+    """
+    timeout = DEFAULT_TIMEOUT if timeout is None else timeout
```
Kami: Instead of doing that, you should simply use default function argument values:

``` python
retry(timeout=DEFAULT_TIMEOUT, retry_delay=DEFAULT_SLEEP, backoff=DEFAULT_BACKOFF, func, *args, **kwargs):
```

libcloud/utils/misc.py
```diff
@@ -282,3 +292,28 @@ def __repr__(self):
...
+    bsleep = DEFAULT_SLEEP if retry_delay is None else retry_delay
+    backoff = DEFAULT_BACKCOFF if backoff is None else backoff
+    end = datetime.now() + timedelta(seconds=timeout)
+    exc_info = None
+    while datetime.now() < end:
+        try:
+            result = func(*args, **kwargs)
+            return result
+        except RateLimit as exc:
```
Kami: Since the function lives in the `misc` module and it's already "kinda" generic it would make sense if user could also pass in an exception classes (or maybe a callable) upon which we should retry.

For example, an argument called `retry_exceptions=(RateLimit, socket.error,)` or something along those lines.

libcloud/utils/misc.py
Kami: Can you please also add a docstring which explain what each argument does.

Kami: I think it would also make sense and make it nicer if we make this function a decorator so it could be used like that:

``` python
request_func_with_retry = retry(func, timeout, retry_delay, ...)(original function args and kwargs)
request_func_with_retry()
```

Kami:
I've added some comments - from a high level perspective, the change looks good to me, but there are some minor code and style things which should be addressed.

In addition to that, I would like to introduce a new module-level variable and an environment variable with which user can turn the retry mechanism on. For backward compatibility reasons, I would prefer to keep this feature off by default for the upcoming release. After a while when this feature gets more testing, we can turn it on by default in a future release.

For example, something along the lines of

``` python
RETRY_FAILED_HTTP_REQUESTS = os.environ.get('LIBCLOUD_RETRY_FAILED_HTTP_REQUESTS`, False)
```

Then user can turn it on by setting the module level variable / environment variable:

``` python
LIBCLOUD_RETRY_HTTP_REQUESTS=True python my_script.py
```

Or

``` python
import libcloud
libcloud.LIBCLOUD_RETRY_HTTP_REQUESTS=True
```

libcloud/utils/misc.py
jimbobhickville: We originally implemented it as a decorator, but it made testing difficult because the timing values were all set at compile-time and you couldn't override them to make it re-run in a tighter loop in the tests, causing the test time to blow up considerably.  Is there a good solution for that problem?  The only solution I'm aware of is to make sure you set the default values before you load any code that uses the decorator, which is hard to guarantee.

libcloud/common/base.py
arpithkp, line 472: added docstring for the retry method.

Kami:
@arpithkp next time please ping me (include my nickname in the comment) when you push the changes since I don't receive notifications for new commits / comments.

libcloud/utils/misc.py
Kami: @jimbobhickville That's correct - you would either need to set the default value before the import or you need to explicitly specify the timeout argument.

Kami:
@arpithkp Also, when the decorator and the environment variable / module level variable change is done, please also make sure to correctly squash / rebase the commits so you are preserved as the commit author.

Currently, Eric is noted as the commit author which I assume it not accurate :)

arpithkp:
@Kami Now I've everything in place, I was waiting to complete my work before notifying you. Please let me know if anything still needs to be addressed.

libcloud/common/base.py
```diff
@@ -738,8 +748,15 @@ def request(self, action, params=None, data=None, headers=None,
 
                 self.connection.endheaders()
             else:
-                self.connection.request(method=method, url=url, body=data,
-                                        headers=headers)
+                if retry_enabled:
+                    retry(self.connection.request, self.timeout,
```
jimbobhickville: I think what @Kami was trying to say, and I finally get it, is that if retry is a decorator, you can just use it inline here to create a new method that has the retry logic applied:

```
retrying_request = retry(self.connection.request, self.timeout, self.retry_delay, self.backoff)
retrying_request(method=method, url=url, body=data, headers=headers)
```

This will let you still override the values at runtime, but also allow retry to be used as a decorator elsewhere if people want to.  Does that make sense?

jimbobhickville: The argument order might be wrong, or you could just make everything but the decorated method be a kwarg instead, which is probably easier to read anyway.

Kami: Yep, what @jimbobhickville said.

Kami:
@arpithkp Thanks.

I will have a look today (and also hopefully get it merged if everything looks OK :)).

arpithkp:
@Kami If there is nothing to address, can you merge this soon please . It helps us to move forward with our internal code changes which is blocked from this PR.

libcloud/test/compute/test_retry_limit.py
```diff
@@ -0,0 +1,75 @@
...
+    ('content-type', 'application/json')]
+SIMPLE_RESPONSE_STATUS = ('HTTP/1.1', 429, 'CONFLICT')
+
+
+class RateLimitTestCase(unittest.TestCase):
+
+    def _raise_socket_error(self):
+        raise socket.gaierror('')
+
+    def test_retry_connection(self):
```
Kami, line 38: I was just merging the PR and cleaning some minor style things and noticed that this method doesn't actually test the retry mechanism.

You mock the `request` method which means you are testing the mock and not the actual request method which performs a retry.

Can you please work on some actual tests for the retry mechanism inside `request` method?

libcloud/test/compute/test_retry_limit.py
```diff
@@ -0,0 +1,75 @@
...
+        con = Connection(timeout=1, retry_delay=0.1)
+        con.connection = Mock()
+        connect_method = 'libcloud.common.base.Connection.request'
```
Kami, line 41: Also, either the variable name is not named correctly or you are mocking the wrong method.

Even if you mocked `libcloud.common.base.Connection.connect` method this test would still not be testing the retry mechanism because off two things:
1. We now only retry when environment variable is set
2. Retry happens inside the `request` method after the `connect` method is called

Kami:
@arpithkp I've made some style changes / improvements (6ffffa338bb624f2e8d3f79a5c38d960667af1c1, 5f870dbfb30deb32970010b81e89b29fcff9daaa, c86d45b930cee84b895cbf706f99ae34b9e161ac) and merged what we have so far into trunk. 

Can you please work on some actual tests for the retry inside the request method (see my comment above) and documentation which explains how retry works and how to enable it in a new PR?

It would be great if you can do that as soon as possible because I will wait on those changes before rolling out a new release.

Thanks.

cryptickp:
Thanks for making those changes and merging this. I'll work on the unit tests.

libcloud/test/compute/test_retry_limit.py
cryptickp, line 38: @Kami : Yes this test doesn't make any sense so I'll remove this in next PR. Also I noticed this [test](https://github.com/apache/libcloud/blob/trunk/libcloud/test/test_connection.py#L285) actually testing retry logic along with two other test below it, I'm thinking no other test are required to test it. Let me know your thoughts.

ioggstream:
@cryptickp should we close this ticket on the agile board? It's still open there ;)

arpith-kp:
@ioggstream please close it.

PR #516, opened 2015-05-13
List VPCs project support

syed:
Adding support to list VPCs which are in a project.

### DIFF
--- a/libcloud/compute/drivers/cloudstack.py
+++ b/libcloud/compute/drivers/cloudstack.py
@@ -1810,8 +1810,6 @@ class CloudStackNodeDriver(CloudStackDriverMixIn, NodeDriver):
         if project is not None:
             args['projectid'] = project.id
 
-
-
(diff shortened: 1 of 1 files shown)


libcloud/compute/drivers/cloudstack.py
```diff
@@ -1798,14 +1798,20 @@ def ex_list_vpc_offerings(self):
 
         return vpcofferings
 
-    def ex_list_vpcs(self):
+    def ex_list_vpcs(self, project=None):
         """
         List the available VPCs
```
Kami, line 1845: Please add a docstring which indicates what `project` argument does.

Kami, line 1845: Besides that, LGTM.

Kami:
Sorry for the delay.

I don't get notified when you push new commits so I didn't notice you addressed the comment.

In any case, patch has been merged into trunk now. Thanks.

PR #517, opened 2015-05-13
[LIBCLOUD-710] Updated CloudSigma's NODE_STATE_MAP to use new-ish Nod…

chrisob:
…eStates

Fixes issue: https://issues.apache.org/jira/browse/LIBCLOUD-710

### DIFF
--- a/libcloud/compute/drivers/cloudsigma.py
+++ b/libcloud/compute/drivers/cloudsigma.py
@@ -984,10 +984,10 @@ class CloudSigma_2_0_NodeDriver(CloudSigmaNodeDriver):
     NODE_STATE_MAP = {
         'starting': NodeState.PENDING,
         'stopping': NodeState.PENDING,
-        'unavailable': NodeState.PENDING,
+        'unavailable': NodeState.ERROR,
         'running': NodeState.RUNNING,
         'stopped': NodeState.STOPPED,
-        'paused': NodeState.STOPPED
+        'paused': NodeState.PAUSED
     }
 
     def __init__(self, key, secret, secure=True, host=None, port=None,


libcloud/compute/drivers/cloudsigma.py
```diff
@@ -984,10 +984,10 @@ class CloudSigma_2_0_NodeDriver(CloudSigmaNodeDriver):
     NODE_STATE_MAP = {
         'starting': NodeState.PENDING,
         'stopping': NodeState.PENDING,
-        'unavailable': NodeState.PENDING,
```
Kami, line 987: @vpetersson Is mapping "unavailable" to "error" indeed more correct than mapping it to "pending" - aka can node come out of the "unavailable" state?

Kami:
Thanks.

Will wait for @vpetersson's response about the new mapping for the "unavailable" state. The pause one looks good to me though.

vpetersson:
@Kami 'unavailable' is a tricky one. In most cases, it's just a temporary error, but it's really a catch-all error when we're able to get the proper status from the host. Perhaps @gcbirzan can chime in with a more detailed explanation.

chrisob:
From what I understand, unavailable is indeed usually a temporary status,
but while a server is in state unavailable, no API actions can be taken on
it (API returns a 500 error when attempting to interact with an unavailable
sever), thus why I figured ERROR would be the best state map for it.
Hopefully Cristian has something to say about this though :)

Just curious, does NodeState.ERROR usually imply an unrecoverable error?

Kami:
@chrisob Yeah, ERROR usually implies an end state which is not recoverable.

I don't think we have any detailed docs on the states right now, but I will make sure they get added.

Kami:
This has been merged a while ago.

Can you please close the PR?

PR #519, opened 2015-05-13
[LIBCLOUD-711] Fixed occasional CRC check failure when decompressing …

chrisob:
…large responses

fixes issue: https://issues.apache.org/jira/browse/LIBCLOUD-711

### DIFF
--- a/libcloud/utils/compression.py
+++ b/libcloud/utils/compression.py
@@ -14,7 +14,6 @@
 # limitations under the License.
 
 import zlib
-import gzip
 
 from libcloud.utils.py3 import PY3
 from libcloud.utils.py3 import StringIO
@@ -36,7 +35,9 @@ def decompress_data(compression_type, data):
         else:
             cls = StringIO
 
-        return gzip.GzipFile(fileobj=cls(data)).read()
+        decomp = zlib.decompressobj(16+zlib.MAX_WBITS)
+        return decomp.decompress(data)
+
     else:
         raise Exception('Invalid or onsupported compression type: %s' %
                         (compression_type))
--- a/libcloud/utils/compression.py
+++ b/libcloud/utils/compression.py
@@ -15,9 +15,6 @@
 
 import zlib
 
-from libcloud.utils.py3 import PY3
-from libcloud.utils.py3 import StringIO
-
 
 __all__ = [
     'decompress_data'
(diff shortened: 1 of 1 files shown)


Kami:
Thanks for the report.

Looking at the issue it seems more like a problem with the response getting corrupted and not an issue with the actual library we use to decompress the response - that is why thee CRC exists in the first place - to make sure the data is intact and not corrupted.

If it was an issue with a library, it would fail all of the time, not just some times. It could be a some kind of network issue on your or CloudSigma's side, a loadbalancer at CloudSigma corrupting the response or something else in the middle causing this issue to happen.

Only thing I find weird is that using zlib doesn't cause this issue to happen. Python zlib library should still include a CRC.

In any case, I will dig in and also try to reproduce the issue myself.

/cc @vpetersson

chrisob:
I would agree with you, but we've been polling CloudSigma's API with gzip
request/response headers (specifically the /servers/detail/?limit=0 action)
every few minutes for a while now using curl's decompression library, and
haven't had any issues.  Also, while debugging this issue, the compressed
response data is complete when decompressed with command line gzip (but
it's missing the EOF?).

Granted, it is strange that there's no complaints about it with other
drivers. If you need any other info to help troubleshoot, let me know.  In
the meantime I'll do some testing in other environments on the off chance
that something's off with mine.

Kami:
@chrisob Alright, thanks.

I will also apply try to reproduce the issue you have encountered and test your patch.

Kami:
I did a bit more research and it looks like the problem could be related to the server using "deflate" compression and incorrectly advertising it as "gzip" to the client via content-encoding header (see http://stackoverflow.com/a/9856879).

If that's the case, it's really an issue with the server and not issue in Libcloud. We could add try / catch and fall back to zlib if gzip throws checksum error exceptions, but this seems hacky.

In any case, I will also test out your changes.

Kami:
Also, this might be relevant - https://stackoverflow.com/questions/3122145/zlib-error-error-3-while-decompressing-incorrect-header-check/22310760#comment42261099_22310760

In any case, we already check for zlib / deflate content-encoding header and use zlib to decompress data if this header is returned. Also using zlib when gzip header is returned seems wrong since it's really a server misconfiguration / issue and we shouldn't silently accept it (although, it does look like a lot of clients do that :/)

pka:
This patch solves the problem for me (libcloud 0.20.1, Python 2.7.6 )!

vpetersson:
(@Kami just a side-note: I'm no longer with CloudSigma :)

kamilchm:
Works for me too.

kamilchm:
@Kami why can't we do it like requests do https://github.com/kennethreitz/requests/commit/6d6252aa9fc4bb38d7f68073b41092c31ddb146b#commitcomment-19841258 ?
[Python gzip module](https://docs.python.org/2/library/gzip.html) states that data compression is provided by the zlib module.
I'm not an experrt on decompression topic, I just want to use `libcloud` to manage cloudsigma and it's unusable right now :(

PR #522, opened 2015-05-13
[google compute] LIBCLOUD-696: improve node image detection

erjohnso:
Google removed 'image' from the instance resource a while ago, but sometimes set a sourceImage on a persistent disk. This fix will consult the sourceImage attribute on the instance's (node's) boot_disk to try and determine the image.  There is still no guarantee this will be set (e.g. if a persistent disk was created from a snapshot, this will still be None), but this is an improvement.

Fixes https://issues.apache.org/jira/browse/LIBCLOUD-696


erjohnso:
Flaky SoftLayer test. My tox tests passed locally, so I merged.

PR #523, opened 2015-05-15
[LIBCLOUD-712] Use get() to take signature version

mvschaik:
The pop() call will remove the field from the REGION_DETAILS dictionary,
causing subsequent calls to use the default value.

Fixes https://issues.apache.org/jira/browse/LIBCLOUD-712


allardhoeve:
Good catch. Will merge.

Kami:
This also fixes #511.

@allardhoeve Please make sure to give credit to both people when updating changelog - thanks :)

PR #525, opened 2015-05-20
Added Cloudwatt doc with example

ZuluPro:
Added:
- Basic doc
- Examples
- Logo

Fixed volume attachment in commit 88e90d8. Default value was 'auto' for device but it doesn't work with Cloudwatt. I've put None for compatibly with auto-finding.


Kami:
Merged into trunk. Thanks.

PR #527, opened 2015-05-22
[LIBCLOUD-703] packet bare metal cloud provider integration

crunchywelch:
I humbly submit this PR which adds support for our bare metal cloud hosting platform, Packet. Let me know if this looks ok, or if it requires additional work. The JIRA issue is here:

https://issues.apache.org/jira/browse/LIBCLOUD-703

Thanks!


crunchywelch:
I have faxed in my ICLA today as well. Thanks!

Kami:
Great, thanks.

I will try to have a look ASAP (hopefully over the weekend).

libcloud/compute/drivers/packet.py
```diff
@@ -0,0 +1,254 @@
...
+        res = self.connection.request('/devices/%s/actions' % (node.id),
+                                      params=params, method='POST')
+        return res.status == httplib.OK
+
+    def destroy_node(self, node):
+        res = self.connection.request('/devices/%s' % (node.id),
+                                      method='DELETE')
+        return res.status == httplib.OK
+
+    def list_ssh_keys(self):
```
Kami: Looks like you should be able to implement ssh key management functionality using key pair management methods which are part of the libcloud standard API - https://github.com/apache/libcloud/blob/trunk/libcloud/compute/base.py#L1183

This also means getting rid of `SSHKey` class in favor of `libcloud.compute.base.KeyPair`.

In your case, you can use store id in the `name` attribute and description in the `extra` dict.

libcloud/compute/drivers/packet.py
```diff
@@ -0,0 +1,254 @@
...
+    host = PACKET_ENDPOINT
+    responseCls = PacketResponse
+
+    def add_default_headers(self, headers):
+        """
+        Add headers that are necessary for every request
+        """
+        headers['Content-Type'] = 'application/json'
+        headers['X-Auth-Token'] = self.user_id
+        headers['X-Consumer-Token'] = \
```
Kami, line 64: I use this is a testing token and should be `self.key`? :)

libcloud/compute/drivers/packet.py
```diff
@@ -0,0 +1,254 @@
...
+        extra_keys = ['created_at', 'updated_at',
+                      'userdata', 'billing_cycle', 'locked']
+        if 'state' in data:
+            state = self.NODE_STATE_MAP.get(data['state'], NodeState.UNKNOWN)
+        else:
+            state = NodeState.UNKNOWN
+
+        public_ips = []
+        private_ips = []
+        if 'ip_addresses' in data and data['ip_addresses'] is not None:
```
Kami, line 189: This could be made a bit more readable if you refactor it in a new method - e.g. `_get_ips_for_node` or similar.

docs/compute/_supported_methods_block_storage.rst
```diff
@@ -55,6 +55,7 @@ Provider                              list volumes create volume destroy volume
 `Opsource`_                           no           no            no             no            no            no             no             
 `Outscale INC`_                       yes          yes           yes            yes           yes           yes            yes            
 `Outscale SAS`_                       yes          yes           yes            yes           yes           yes            yes            
+`Packet`_                             no           no            no             no            no            no             no
```
Kami, line 58: Just for a future reference - in case you updated those files manually, they are generated automatically using `./contrib/generate_provider_feature_matrix_table.py` script.

libcloud/compute/drivers/packet.py
```diff
@@ -0,0 +1,254 @@
...
+                      'powering_on': NodeState.REBOOTING,
+                      'powering_off': NodeState.REBOOTING,
+                      'rebooting': NodeState.REBOOTING,
+                      'inactive': NodeState.STOPPED,
+                      'deleted': NodeState.TERMINATED,
+                      'deprovisioning': NodeState.TERMINATED,
+                      'failed': NodeState.ERROR,
+                      'active': NodeState.RUNNING}
+
+    def list_nodes(self, location):
```
Kami: For consistency with the base API, `location` argument should be an instance of `NodeLocation` and not a string.

crunchywelch:
thanks @Kami, will make some adjustments!

libcloud/compute/drivers/packet.py
crunchywelch: so, this is not really a location, it is a project id which is a little unique to how we structure things on the platform. I'll rename the string to not cause confusion here, I hope that works

libcloud/compute/drivers/packet.py
crunchywelch, line 189: A+, will do!

libcloud/compute/drivers/packet.py
crunchywelch: cool, works for me

libcloud/compute/drivers/packet.py
crunchywelch, line 64: heh, fair enough, will fix

docs/compute/_supported_methods_block_storage.rst
crunchywelch, line 58: cool, gtk thanks

crunchywelch:
Ok, I think I got all those notes, lmk if you see anything else I should adjust!

Cheers,

libcloud/compute/drivers/packet.py
Kami, line 64: I actually meant the `X-Consumer-Token`. Or is this some kind of static token which is shared among all users?

crunchywelch, line 64: Yeah, exactly, we issue a consumer token for any platform or library that is integrated with us so we can tell what application requests are originating from. If calls are coming in with only an auth key, they will work, but be rate limited. This token in particular is the one we issued for libcloud. It doesn't need to be secured, people could technically abuse it or remove it if they wanted to, but then we'd rely on the auth token to relate those requests to a particular account and deal with it at that level. We have one for our mobile apps, docker machine, rancher, devo.ps, etc...

Does that make sense? Do you see an issue with it?

Kami, line 64: Ah, I see, that makes sense. Thanks for the clarification.

P.S. We also try to be a good citizen and indicate that the request is made by Libcloud by sending a right user-agent header (e.g. `User-Agent: libcloud/0.17.1-dev (Rackspace Cloud (Next Gen))` :)).

Kami:
I tries to test the driver using a dummy token and looks like an API returns 404 on invalid token. Something like 401 would be better and then Libcloud would also correctly throw `InvalidCreds` exception.

Or maybe exception is related to the project not existing? In any case, 401 would be more appropriate (and hopefully token validation happens before project validation anyway) :)

``` bash
# -------- begin 140127380731600 request ----------
curl -i -X GET -H 'Host: api.packet.net' -H 'X-LC-Request-ID: 140127380731600' -H 'Accept-Encoding: gzip,deflate' -H 'X-Consumer-Token: kcrhMn7hwG8Ceo2hAhGFa2qpxLBvVHxEjS9ue8iqmsNkeeB2iQgMq4dNc1893pYu' -H 'X-Auth-Token: accesskeyid' -H 'Content-Type: application/json' -H 'User-Agent: libcloud/0.17.1-dev (Packet) ' --compress 'https://api.packet.net:443/projects/aaaaa/devices?include=plan'
# -------- begin 140127380731600:140127380758680 response ----------
HTTP/1.1 404 Not Found
Status: 404 Not Found
X-Request-Id: ec216d58-04b5-432b-92ad-60efc377e9fd
X-Xss-Protection: 1; mode=block
X-Content-Type-Options: nosniff
Transfer-Encoding: chunked
Server: nginx/1.6.2
X-Runtime: 0.005357
Cache-Control: no-cache
Date: Sun, 24 May 2015 14:49:02 GMT
X-Frame-Options: SAMEORIGIN
Content-Type: application/json; charset=utf-8

15
{"error":"Not found"}
0

# -------- end 140127380731600:140127380758680 response ----------
```

Kami:
I made a minor change (`projectid`, `project_id` -> `ex_project_id` since project id argument is not part of the standard API) and merged driver into trunk.

Thanks.

crunchywelch:
great, thanks ! Do I need to provide the patch on the jira issue next? Looking at step 11 here:

http://libcloud.readthedocs.org/en/latest/development.html

Kami:
No, it's fine, I already merged it using a patch from Github :)

crunchywelch:
Sweet, thanks man! And thank you for being a maintainer!

On Sun, May 24, 2015 at 5:26 PM, Tomaz Muraus notifications@github.com
wrote:

> No, it's fine, I already merged it using a patch from Github :)
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/apache/libcloud/pull/527#issuecomment-105063113.

## 

Aaron Welch

## SVP of Product

212.933.9785 x301
welch@packet.net

PR #531, opened 2015-05-28
get_container can parse a 200 status code

wpjunior:


Kami:
Thanks - can you please provide more details on why this change is needed?

In general, it seems like a pretty harmless change, but API should always return the same status code on success so accepting multiple codes could mask some kind of issue.

PR #534, opened 2015-06-04
Gandi Compute Driver improvement

ZuluPro:
- Added sphinx documentation with example of node creation
- Added docstring for all public methods
- Added SSH keypair management: list, get, import, delete
- Change authentification initialisation in create_node: User must choose between login/password, SSH keys or both
- Added ex_get_node and ex_get_volume
- Added tests for all commit

### DIFF
--- a/libcloud/compute/drivers/gandi.py
+++ b/libcloud/compute/drivers/gandi.py
@@ -215,12 +215,15 @@ class GandiNodeDriver(BaseGandiDriver, NodeDriver):
         :keyword    inet_family: version of ip to use, default 4 (optional)
         :type       inet_family: ``int``
 
+        :keyword    keypairs: IDs of keypairs or Keypairs object
+        :type       keypairs: ``int`` or :class:`.KeyPair`
+
         :rtype: :class:`Node`
         """
 
-        if kwargs.get('login') is None or kwargs.get('password') is None:
-            raise GandiException(
-                1020, 'login and password must be defined for node creation')
+        if not kwargs.get('login') and not kwargs.get('keypairs'):
+            raise GandiException(1020, "Login and password or ssh keypair "
+                                 "must be defined for node creation")
 
         location = kwargs.get('location')
         if location and isinstance(location, NodeLocation):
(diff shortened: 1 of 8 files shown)


libcloud/compute/drivers/gandi.py
```diff
@@ -214,12 +260,15 @@ def create_node(self, **kwargs):
...
+        :keyword    keypairs: IDs of keypairs or Keypairs object
+        :type       keypairs: list of ``int`` or :class:`.KeyPair`
+
         :rtype: :class:`Node`
         """
 
-        if kwargs.get('login') is None or kwargs.get('password') is None:
-            raise GandiException(
-                1020, 'login and password must be defined for node creation')
+        if not kwargs.get('login') and not kwargs.get('keypairs'):
```
Kami, line 269: We need to fix this to use `auth` argument, but that's fine for now.

libcloud/compute/drivers/gandi.py
```diff
@@ -617,3 +759,68 @@ def ex_update_disk(self, disk, new_size=None, new_name=None):
...
+        """
+        List registered key pairs.
+
+        :return:   A list of key par objects.
+        :rtype:   ``list`` of :class:`libcloud.compute.base.KeyPair`
+        """
+        kps = self.connection.request('hosting.ssh.list').object
+        return self._to_key_pairs(kps)
+
+    def get_key_pair(self, key_id):
```
Kami: For consistency with the base API, the first argument should be called name.

Kami:
There are some minor style issues (method ordering), besides that it looks good to me.

ZuluPro:
I changed `key_id` for name. I also change the function behavior for list and filter instead of directly get object.

For next commits, how must I raise Error for an unfound KeyPair ? Is there a hierarchy of Exceptions to use ?

Kami:
@ZuluPro Great.

There is `libcloud.compute.types.KeyPairDoesNotExistError`.

Kami:
Changes merged into trunk. Thanks.

PR #536, opened 2015-06-15
Remove duplicate "Storage" in title

djoume:


Kami:
The title contained word "storage" twice since the service name is" Google Storage" and it's part of the Libcloud storage API.

In any case, It's indeed a bit redundant and I'm fine with that change so I merged it intro trunk. Thanks.

Kami:
Can you please also close this PR - I forgot to reference it in the commit message.

PR #537, opened 2015-06-25
LIBCLOUD-717 Implement v2 API request speed optimization and other ch…

jcastillo2nd:
…anges
- Increased pagination to API maximum
- Added create_node parameter to support droplet features (backups, ipv6, user_data, etc )
- Reorganized method groups for style
- Converted all list_\* to paginated request

In large result sets, the API request was returning the default 25 results per request, making it very slow in large sets.

https://issues.apache.org/jira/browse/LIBCLOUD-717

I have some doubts about the ex_create_attr, but this was done to support selecting features. The prior parameter for ssh key selection/assignment was kept, but updated to throw a deprecation warning.

I've also added an example of how the create_node is called with the options.

I'm not sure about the rest documentation of the ex_create_attr dictionary values, but I look foward to the review.

### DIFF
--- a/libcloud/common/digitalocean.py
+++ b/libcloud/common/digitalocean.py
@@ -16,6 +16,7 @@
 """
 Common settings and connection objects for DigitalOcean Cloud
 """
+import warnings
 
 from libcloud.utils.py3 import httplib, parse_qs, urlparse
 
@@ -116,7 +117,7 @@ class DigitalOcean_v2_Connection(ConnectionKey):
         This method adds ``per_page`` to the request to reduce the total
         number of paginated requests to the API.
         """
-        params['per_page'] = '200'
+        params['per_page'] = self.driver.ex_per_page
         return params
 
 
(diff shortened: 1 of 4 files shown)


libcloud/common/digitalocean.py
```diff
@@ -109,6 +109,16 @@ def add_default_headers(self, headers):
...
         return headers
 
+    def add_default_params(self, params):
+        """
+        Add parameters that are necessary for every request
+
+        This method adds ``per_page`` to the request to reduce the total
+        number of paginated requests to the API.
+        """
+        params['per_page'] = '200'
```
Kami: It would be nice if user could also specify / override this value - maybe by passing an argument to the driver constructor?

jcastillo2nd: Yes, that may work with the driver constructor. I will review this later today to see how to implement this, as the add_default_params is in the connection class.

I don't know if the connection class ever gets a reference from the driver, though that may actually be passed in as `self`. If that's the case, I could just use a base driver attribute assigned by a keyword parameter in the driver that defaults to 200. I will have to test that later, though if there are any cleaner suggestions, I'm all ears.

jcastillo2nd:
@Kami The ex_per_page kwarg has been added to the v2 base driver, and this is in turn picked up by the add_default_params in the v2 connection class. Let me know if you see anything else, and any feedback on using the warnings.warn for v1 initialization regarding deprecation. Nov. 1st 2015 is the deadline for the v1 support on the platform, and after that.

Kami:
Sorry for the delay.

I reviewed the changes and merged them into master (with a minor fix - not defaulting to an empty dict since it's a mutable type and using top-level argument for user-data - https://github.com/apache/libcloud/pull/559).

In the future, it would make it easier if you create a separate branch for your changes :)

PR #543, opened 2015-07-07
[LIBCLOUD-720] GCE job timeout modification.

gabik:
Google jobs timeout hardcoded to be 180 seconds.
There is no access to this property from the Node class.
Added timeout argument to the _init_ of GCENodeDriver.

https://issues.apache.org/jira/browse/LIBCLOUD-720

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -659,6 +659,10 @@ Compute
   (GITHUB-371)
   [Jeroen de Korte]
 
+- Allow setting job timeout in GENodeDriver.
+  (LIBCLOUD-720)
+  [Gabi Kazav]
+
 Storage
 ~~~~~~~
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -659,7 +659,7 @@ Compute
   (GITHUB-371)
   [Jeroen de Korte]
 
-- Allow setting job timeout in GENodeDriver.
+- Allow setting job timeout in GCENodeDriver.
   (LIBCLOUD-720)
   [Gabi Kazav]
 
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -1037,7 +1037,8 @@ class GCENodeDriver(NodeDriver):
         :type     timeout: ``int``
         """
         if timeout is not None:
-            self.connectionCls.timeout=timeout
+            if timeout > 0:
+                self.connectionCls.timeout = timeout
 
         if not project:
             raise ValueError('Project name must be specified using '
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -997,7 +997,8 @@ class GCENodeDriver(NodeDriver):
     }
 
     def __init__(self, user_id, key=None, datacenter=None, project=None,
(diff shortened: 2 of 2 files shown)


erjohnso:
I'm not so sure this is strictly necessary and I don't see that any of the other providers have exposed `timeout` in their driver `__init__()` methods.  I did a bit of testing, and you could do something like this instead of the proposed change:

``` python
start_time = datetime.datetime.now()
print "=> Compute demo/test start time: %s' % str(start_time)
gce = get_driver(Provider.GCE)('','', project='graphite-demos', auth_type='GCE')
print "=> default timeout:", gce.connection.timeout
gce.connection.timeout = 4
print "=> new timeout:", gce.connection.timeout
# Get project info and print name
project = gce.ex_get_project()
print "=> Project: %s' % project.name
node = gce.create_node("foo", "n1-standard-1", "debian-7", "us-central1-f")
```

And when I execute my demo script, I see that the timeout was honored.

``` sh
=> Compute demo/test start time: 2015-07-14 16:06:10.399974
=> default timeout: 180
=> new timeout: 4
=> Project: graphite-demos
  File "d.py", line 640, in <module>
    main_compute()
  File "d.py", line 233, in main_compute
    node = gce.create_node(name, "n1-standard-1", "debian-7", "us-central1-f")
  File "../libcloud/compute/drivers/gce.py", line 2393, in create_node
    self.connection.async_request(request, method='POST', data=node_data)
  File "../libcloud/common/base.py", line 902, in async_request
    (self.timeout))
libcloud.common.types.LibcloudError: <LibcloudError in None 'Job did not complete in 4 seconds'>
```

Note that you can't pass in the `timeout` argument when instantiating the driver, so you'll need to override its internal `connection` object's `timeout` parameter the way I did in the example.

@Kami - any other comments / tips / pointers?

gabik:
We have several projects using AWS and GCE, some of the projects need longer timeout and I dont see reason why not adding setter to them.

I agree about the other drivers, so I think I should move the setter to be more global.
Maybe setter in the Connection class will be better idea.
I am not sure changing the interval property is the best idea..  
The main idea is checking the timeout is positive integer.

so for example I can add the following method:

```
gce.connection.set_timeout(4)
```

instead

```
gce.connection.timeout = 4
```

What do you guys think?

erjohnso:
Sorry for the laggy response, @gabik. That sounds fine by me, but I usually only work on the GCE parts, so I'd defer to @Kami or someone else that knows the guts of libcloud better than I do.

jimbobhickville:
Yeah, timeout is already automatically passed along in the slurped `**kwargs`:

https://github.com/apache/libcloud/blob/trunk/libcloud/common/base.py#L1099

But it's the per request timeout, not the whole polling loop timeout.  We should probably disambiguate them and add a new kwarg for `PollingConnection` that gets passed along, IMO.

jimbobhickville:
I was thinking more about this.  I think we should rename `timeout` on `PollingConnection` to be `async_timeout` or something, and then let it be passed along in the `__init__` as an argument.  This would require modifying a few classes that inherit from `PollingConnection` but having the same property (`timeout`) mean two things is confusing.

PR #545, opened 2015-07-13
Add eu-central-1 to EC2 drivers

JamesGuthrie:
Signed-off-by: James Guthrie james@prodigi.ch

### DIFF
--- a/libcloud/compute/drivers/ec2.py
+++ b/libcloud/compute/drivers/ec2.py
@@ -5679,12 +5679,20 @@ class IdempotentParamError(LibcloudError):
         return repr(self.value)
 
 
-class EC2EUNodeDriver(EC2NodeDriver):
+class EC2EUWestNodeDriver(EC2NodeDriver):
     """
     Driver class for EC2 in the Western Europe Region.
     """
     name = 'Amazon EC2 (eu-west-1)'
     _region = 'eu-west-1'
+
+
+class EC2EUCentralNodeDriver(EC2NodeDriver):
+    """
+    Driver class for EC2 in the Central Europe Region.
+    """
+    name = 'Amazon EC2 (eu-central-1)'
+    _region = 'eu-central-1'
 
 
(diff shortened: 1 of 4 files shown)


ralphbean:
Why did this get closed?

JamesGuthrie:
Because it's outdated. Unfortunately the documentation hasn't been updated to reflect the change in the API though, and I only realised after making the PR that the documentation is outdated and my PR redundant.

The change is (mentioned [here](https://libcloud.readthedocs.org/en/latest/changelog.html#changes-with-apache-libcloud-0-14-0-beta3)) is:

> Deprecate the following EC2 provider constants: EC2_US_EAST, EC2_EU, EC2_EU_WEST, EC2_AP_SOUTHEAST, EC2_AP_NORTHEAST, EC2_US_WEST_OREGON, EC2_SA_EAST, EC2_SA_EAST and replace it with a new EC2 constant. Driver referenced by this new constant now takes a “region” argument which dictates to which region to connect.

So the following example (lifted from the docs):

```
cls = get_driver(Provider.EC2_US_WEST)
driver = cls('temporary access key', 'temporary secret key', token='temporary session token')
```

becomes:

```
cls = get_driver(Provider.EC2)
driver = cls("auth key", "auth secret", token="token", region="us-west-1")
```

So the driver type is no longer used to determine the location which the driver connects to. Instead, if you specify a valid region, that is the region it connects to. Valid regions are: 
- eu-west-1
- sa-east-1
- ap-southeast-2
- us-west-1
- us-east-1
- ap-southeast-1
- ap-northeast-1
- us-gov-west-1
- eu-central-1
- us-west-2

JamesGuthrie:
I've opened #606 to fix the docs.

ralphbean:
Thanks @JamesGuthrie.  I wandered down the same path you did!

JamesGuthrie:
No problem! Hopefully you will be the last to do so.

PR #547, opened 2015-07-13
[LIBCLOUD-721] work around http proxy for raw requests

nferch:


Kami:
Good catch.

The change looks good to me, but can you please also add a test case for it?

nferch:
Added, PTAL.

Kami:
Sorry for the delay.

I haven't had the time to look into it yet, but I will do it now.

Kami:
I tried to reproduce the issue using a local HTTP proxy and CloudFiles driver (uploading an object), but I couldn't.

It might only be an issue with S3. I will also try your patch with CloudFiles (to make sure it doesn't break anything) and then try to replicate the issue with S3 driver.

Kami:
@nferch 

I also tried to reproduce the issue with S3, but it works just fine. It could be related to the proxy software you are using. It worked fine for me using `tinyproxy`.

Which http proxy software do you use?

Kami:
@nferch 

I just tried your patch. It works fine with S3, but it breaks with CloudFiles. I get the following error back:

``` python
# -------- begin 140015939937808:140015939925632 response ----------
HTTP/1.1 404 Not Found
Date: Sun, 02 Aug 2015 13:28:41 GMT
Content-Length: 70
Content-Type: text/html; charset=UTF-8
X-Trans-Id: txb96027bae3c1459ab33ad-0055be1b09ord1

<html><h1>Not Found</h1><p>The resource could not be found.</p></html>
# -------- end 140015939937808:140015939925632 response ----------

  File "example_storage.py", line 39, in <module>
    object_name='test4'))
  File "/w/lc/libcloud/libcloud/storage/base.py", line 152, in upload_object
    file_path, self, object_name, extra=extra, **kwargs)
  File "/w/lc/libcloud/libcloud/storage/drivers/cloudfiles.py", line 430, in upload_object
    verify_hash=verify_hash, headers=headers)
  File "/w/lc/libcloud/libcloud/storage/drivers/cloudfiles.py", line 790, in _put_object
    driver=self)
libcloud.common.types.LibcloudError: <LibcloudError in <libcloud.storage.drivers.cloudfiles.CloudFilesStorageDriver object at 0x7f5800609e10> "Server didn't return etag">
```

It looks like the original issue is related to a specific proxy software. I would be OK with merging the patch if it fixes the issue for you and doesn't break anything else, but your patch breaks CloudFiles driver + HTTP proxy for me (it might also break other driver, I didn't have the time to test it all).

It would be good to know which proxy software you use (maybe you can tweak the configuration). If that's not possible, an alternative would be to add a flag (`proxy_raw_requests_use_full_url` or whatever) which user can use to toggle this mode on. It should be disabled by default though since it breaks it for me when using tiny proxy (as described above).

nferch:
@Kami The proxy software is Squid, but I don't have a lot of visibility to configuration that would induce this error.

Are you able to see what the full HTTP request was?

PR #548, opened 2015-07-17
[LIBCLOUD-723] Added cloud specific parameter to base compute wait_un…

davidwilson2038:
Fixed LIBCLOUD-723 bug with Azure driver where the wait_until_running method is broken because it calls self.list_nodes() without any options (list_nodes is defined as requiring an ex_cloud_service parameter). I added an optional ex_cloud_opts parameter to the wait_until_running method to make it work and also allow users of other drivers to specify additional configuration parameters if their list_nodes function allows.

### DIFF
--- a/libcloud/compute/base.py
+++ b/libcloud/compute/base.py
@@ -1260,8 +1260,9 @@ class NodeDriver(BaseDriver):
         raise NotImplementedError(
             'delete_key_pair not implemented for this driver')
 
-    def wait_until_running(self, nodes, wait_period=3, timeout=600,
-                           ssh_interface='public_ips', force_ipv4=True):
+    def wait_until_running(self, nodes, ex_cloud_opts=None, wait_period=3,
+                           timeout=600, ssh_interface='public_ips',
+                           force_ipv4=True):
         """
         Block until the provided nodes are considered running.
 
@@ -1270,6 +1271,9 @@ class NodeDriver(BaseDriver):
 
         :param nodes: List of nodes to wait for.
         :type nodes: ``list`` of :class:`.Node`
+
+        :param ex_cloud_opts: Cloud specific options for list_nodes method.
+        :type ex_cloud_opts: ``str``
 
         :param wait_period: How many seconds to wait between each loop
                             iteration. (default is 3)
@@ -1316,7 +1320,7 @@ class NodeDriver(BaseDriver):
         uuids = set([node.uuid for node in nodes])
 
         while time.time() < end:
-            all_nodes = self.list_nodes()
+            all_nodes = self.list_nodes(ex_cloud_opts)
             matching_nodes = list([node for node in all_nodes
                                    if node.uuid in uuids])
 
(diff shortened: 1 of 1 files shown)


Kami:
@davidwilson2038 Good catch and thanks.

libcloud/compute/base.py
```diff
@@ -1260,8 +1260,9 @@ def delete_key_pair(self, key_pair):
         raise NotImplementedError(
             'delete_key_pair not implemented for this driver')
 
-    def wait_until_running(self, nodes, wait_period=3, timeout=600,
-                           ssh_interface='public_ips', force_ipv4=True):
+    def wait_until_running(self, nodes, ex_cloud_opts=None, wait_period=3,
```
Kami: Can you please move this new argument to the end so it's the last argument?

This way the change is backward compatible and doesn't break existing code if using non keyword arguments.

Kami: Please also add a docstring which explains what this argument does.

Edit: Never mind, I see the docstring is already there.

libcloud/compute/base.py
```diff
@@ -1316,7 +1320,7 @@ def filter_addresses(addresses):
         uuids = set([node.uuid for node in nodes])
 
         while time.time() < end:
-            all_nodes = self.list_nodes()
+            all_nodes = self.list_nodes(ex_cloud_opts)
```
Kami: It would be better to make this argument a dict - this way it's more generic and user can pass multiple arguments to the list method. And we should also call it `ex_list_nodes_kwargs` to make it more explicit.

Then we would do something like that:

``` python
...
all_nodes = self.list_nodes(**ex_list_nodes_kwargs)
```

davidwilson2038: This is good solution. I have implemented it and edited the docstring to describe the changes.

libcloud/compute/base.py
```diff
@@ -1260,8 +1260,9 @@ def delete_key_pair(self, key_pair):
         raise NotImplementedError(
             'delete_key_pair not implemented for this driver')
 
-    def wait_until_running(self, nodes, wait_period=3, timeout=600,
-                           ssh_interface='public_ips', force_ipv4=True):
+    def wait_until_running(self, nodes, wait_period=3,
+                           timeout=600, ssh_interface='public_ips',
+                           force_ipv4=True, **ex_list_nodes_kwargs):
```
Kami, line 1265: There should be no `**` here since user should pass in a dict to the function. I will remove it when merging the patch.

Kami:
Made this method signature fix (4cc3adfef808f175647832f5a4175fcd7e105eba) and merged changes into trunk.

Thanks.

Kami:
@davidwilson2038 On a related note - Now that this is merged into trunk, you can also override `wait_until_running` method in the Azure driver - you can simply call parent class method with the correct `ex_list_nodes_kwargs` argument.

PR #549, opened 2015-07-19
[WIP] Support for pickling Driver instances and other Libcloud objects

Kami:
This pull request adds support for pickling Driver instances (e.g. `NodeDriver`) and support for pickling other Libcloud objects which hold reference to the driver instance (e.g. `Node`, `DNSRecord`, etc.).

Pickling doesn't work out of the box for NodeDriver and other objects since a driver contains a reference to the `Connection` class instance which contains many unpicklable types.

To work around this limitation, we don't pickle the `Connection` class instance, but we just pickle arguments which are needed to re-build / re-create the Connection instance and we re-create the instance when unpickling the object.

### DIFF
--- a/libcloud/common/base.py
+++ b/libcloud/common/base.py
@@ -1102,6 +1102,11 @@ class BaseDriver(object):
                             'backoff': kwargs.pop('backoff', None)})
         self.connection = self.connectionCls(*args, **conn_kwargs)
 
+        # Note: We store this attributes on the class so we can re-create
+        # connection when unpickling driver instance
+        self._connection_cls_args = args
+        self._connection_cls_kwargs = conn_kwargs
+
         self.connection.driver = self
         self.connection.connect()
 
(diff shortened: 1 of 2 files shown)


libcloud/common/base.py
```diff
@@ -1111,3 +1116,21 @@ def _ex_connection_class_kwargs(self):
         Connection class constructor.
         """
         return {}
+
+    def __getstate__(self):
+        # Note: "connection" attribute contains many unpicklable objects which
+        # is why we only store attributes which can be used to re-create the
+        # instance later when unpickling.
+        state = self.__dict__.copy()
+        del state['connection']
```
jimbobhickville, line 1125: Should we let subclasses add to this list? `def _ex_unpicklable_attrs` or something?

tonybaloney:
@Kami what are we doing with this?

Kami:
I say ditch it (I will close the PR).

I haven't found a way to make it work across all the supported Python versions and we should also encourage people to use language agnostic serialization / de-serialization protocols / formats (e.g. JSON).

If needed, we can also add some helper methods for that (e.g. to_dict and from_dict).

So perhaps we should add some documentation on that (how to handle serialization / de-serialization - ideally use something like JSON, or even better, just serialize object IDs and always retrieve objects when you need to manipulate them - this way you won't have out of date and consistency issues).

Kami:
Closing.

See the comment above.

PR #551, opened 2015-07-21
[LIBCLOUD-724] Added storage service create and delete methods to Azu…

davidwilson2038:
Added methods to create and destroy storage services in Azure and the appropriate tests. I believe this is necessary because when you use the create_node method to create a node, it will automatically create a storage service, and when you call destroy_node, it does not delete that storage service. Consequently, you must use another tool or the GUI to clean up after node creation (or to set up for it), so providing this functionality through libcloud enhances the role libcloud can play in Azure usage. Also, I think these methods belong in the compute driver rather than in the storage driver for two reasons:
1) These are higher level functions than is accessible through the current Azure storage driver (you must specify an already-created storage service to instantiate the driver)
2) The Azure API lumps these functions in with other compute functions, specifically separating them from object storage functions.

### DIFF
--- a/libcloud/compute/drivers/azure.py
+++ b/libcloud/compute/drivers/azure.py
@@ -903,7 +903,6 @@ class AzureNodeDriver(NodeDriver):
 
         :rtype: ``bool``
         """
-        # TODO: add check to ensure all nodes have been deleted
         response = self._perform_cloud_service_delete(
             self._get_hosted_service_path(name)
         )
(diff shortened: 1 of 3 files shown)


libcloud/compute/drivers/azure.py
```diff
@@ -985,6 +984,70 @@ def ex_set_instance_endpoints(self, node, endpoints,
...
+        :type       extended_properties: ``dict``
+
+        :rtype: ``bool``
+        """
+
+        response = self._perform_storage_service_create(
+            self._get_storage_service_path(),
+            AzureXmlSerializer.create_storage_service_to_xml(
+                service_name=name,
+                label=self._encode_base64(name),
```
Kami, line 1016: Hm, does the label really need to be base64 encoded?

Kami, line 1016: Ignore my comment. I've confirmed it does indeed need to be base64 encoded - https://msdn.microsoft.com/en-us/library/azure/hh264518.aspx

Kami:
Removed docstring for `tries` argument which was not actually declared in the method signature and merged patch into trunk.

Thanks.

PR #553, opened 2015-07-22
Added volume type to create_volume in openstack

ricotijsen:
Fx. to allow to choose between SATA and SSD. When None given, the default is used.

### DIFF
--- a/libcloud/compute/drivers/openstack.py
+++ b/libcloud/compute/drivers/openstack.py
@@ -164,7 +164,8 @@ class OpenStackNodeDriver(NodeDriver, OpenStackDriverMixin):
         return self._to_nodes(
             self.connection.request('/servers/detail', params=params).object)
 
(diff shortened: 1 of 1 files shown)


libcloud/compute/drivers/openstack.py
```diff
@@ -164,7 +164,8 @@ def list_nodes(self, ex_all_tenants=False):
         return self._to_nodes(
             self.connection.request('/servers/detail', params=params).object)
 
-    def create_volume(self, size, name, location=None, snapshot=None):
+    def create_volume(self, size, name, location=None, snapshot=None,
+                      volume_type=None):
```
Kami, line 168: Sorry for the delay.

This is an extension argument (not part of the standard API) so it should be prefixed with `ex_` - `ex_volume_type`.

Kami:
Made the argument name change, squashed the commits and merged changes into trunk. Thanks.

PR #554, opened 2015-07-28
[LIBCLOUD-726] Added support for custom data in create_node function …

davidwilson2038:
Azure has support for Custom Data (i.e. "User Data") in base64 format to be included on node creation (https://msdn.microsoft.com/en-us/library/azure/jj157194.aspx). I have implemented this in the create_node function for Linux VMs. This is an important feature for any workflows that rely on cloud-init.

### DIFF
--- a/libcloud/compute/drivers/azure.py
+++ b/libcloud/compute/drivers/azure.py
@@ -579,6 +579,10 @@ class AzureNodeDriver(NodeDriver):
         auth = self._get_and_check_auth(auth)
         password = auth.password
 
+        # Check for custom data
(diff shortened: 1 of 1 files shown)


Kami:
Thanks.

I've pushed some change and merged your PR intro trunk.
1. Explicitly declare the method argument, add missing docstring - d79cffd595a00c899bc57ae8eea75d4853521f52
2. Don't require user to base64 encode the value and do the encoding inside the method - cf32c2766e70ec038cf10a678e877c2c1318ca02

PR #558, opened 2015-08-02
Fix a bug in consuming streams in Paramiko ssh client 

lakshmi-kannan:
I wrote a wrapper to that wraps the SSH client to test. See https://github.com/lakshmi-kannan/st2/pull/2 and specifically https://github.com/lakshmi-kannan/st2/commit/1f0e025f8ff378eff7dd00a929cffce035e63a4b.

```
(virtualenv)/m/s/s/st2 git:paramiko_parallel_ssh ❯❯❯ st2common/bin/paramiko_ssh_evenlets_tester.py --hosts=localhost,127.0.0.1 --user=lakshmi --private-key="/home/vagrant/.ssh/id_rsa" --cmd="pwd"
"cmd results: \n{'127.0.0.1': [u'/home/lakshmi\\n', '', 0, True], 'localhost': [u'/home/lakshmi\\n', '', 0, True]}"
(virtualenv)/m/s/s/st2 git:paramiko_parallel_ssh ❯❯❯
```

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -37,6 +37,13 @@ General
   ``LIBCLOUD_RETRY_FAILED_HTTP_REQUESTS`` environment variable.
   (GITHUB-515, LIBCLOUD-360, LIBCLOUD-709)
 
+- Fix a bug in consuming stdout and stderr strams in Paramiko SSH client.
+
+  In some cases (like connecting to localhost via SSH), exit_status_ready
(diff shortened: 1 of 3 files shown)


Kami:
Good catch. Merged into trunk. Thanks.

We probably never encountered this issue because we always used it with remote servers where a higher latency would mask this bug :)

PR #559, opened 2015-08-04
[LIBCLOUD-731] Added User Data support for Digital Ocean v2 driver

davidwilson2038:
DigitalOcean v2 API supports User Data. Added ex_user_data argument to create_node function as well as expanded docstring for the method. Unlike Azure, it accepts a non-base64 encoded string.


Kami:
Merged into trunk. Thanks.

PR #560, opened 2015-08-05
Adapt onapp tests to the libcloud test framework

TooAngel:
Instead of mocking the response manually the tests use MockHttpTestCase.


Kami:
Thanks - merged into trunk :)

PR #561, opened 2015-08-09
Added Runabove volume management & fixed many little things

ZuluPro:
RunAbove has added volume management recently to their API, I added this to libcloud with tests.

I also:
- Added docstrings
- Added example of volume management in docs
- Fixed some typo and bad doc

### DIFF
--- a/libcloud/compute/drivers/runabove.py
+++ b/libcloud/compute/drivers/runabove.py
@@ -12,11 +12,13 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+"""
+RunAbove driver
+"""
 from libcloud.common.runabove import API_ROOT, RunAboveConnection
 from libcloud.compute.base import NodeDriver, NodeSize, Node, NodeLocation
-from libcloud.compute.base import NodeImage
-from libcloud.compute.types import Provider
+from libcloud.compute.base import NodeImage, StorageVolume
+from libcloud.compute.types import Provider, StorageVolumeState
 from libcloud.compute.drivers.openstack import OpenStackNodeDriver
 from libcloud.compute.drivers.openstack import OpenStackKeyPair
 
@@ -25,24 +27,9 @@ class RunAboveNodeDriver(NodeDriver):
     """
     Libcloud driver for the RunAbove API
 
-    Rough mapping of which is which:
+    For more information on the RunAbove API, read the official reference:
 
-        list_nodes              linode.list
-        reboot_node             linode.reboot
-        destroy_node            linode.delete
-        create_node             linode.create, linode.update,
-                                linode.disk.createfromdistribution,
-                                linode.disk.create, linode.config.create,
-                                linode.ip.addprivate, linode.boot
-        list_sizes              avail.linodeplans
-        list_images             avail.distributions
-        list_locations          avail.datacenters
-        list_volumes            linode.disk.list
-        destroy_volume          linode.disk.delete
-
-    For more information on the Linode API, be sure to read the reference:
-
-        http://www.linode.com/api/
+        https://api.runabove.com/console/
     """
     type = Provider.RUNABOVE
     name = "RunAbove"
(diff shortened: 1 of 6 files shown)


Kami:
Great, this will go into 0.18.0 is the voting thread for 0.18.0 passes :)

libcloud/compute/drivers/runabove.py
```diff
@@ -68,14 +62,12 @@ def __init__(self, key, secret, ex_consumer_key=None):
...
 
-        This call will return all Linodes that the API key in use has access
-        to.
-        If a node is in this list, rebooting will work; however, creation and
-        destruction are a separate grant.
+        :keyword location: Location (region) used as filter
+        :type    location: :class:`NodeLocation`
 
-        :return: List of node objects that the API key can access
+        :returns: List of node objects
```
Kami: I think it's just `return` (at least that's what most other code uses). It might also be that both works, but `return` is preferred for consistency reasons.

libcloud/compute/drivers/runabove.py
```diff
@@ -86,11 +78,41 @@ def list_nodes(self, location=None):
...
+        :type    node_id: ``str``
+
+        :returns: Created node
+        :rtype  : :class:`Node`
+        """
         action = API_ROOT + '/instance/' + node_id
         response = self.connection.request(action, method='GET')
         return self._to_node(response.object)
 
     def create_node(self, **kwargs):
```
Kami: It would also be good to explicitly declare the arguments here.

libcloud/compute/drivers/runabove.py
```diff
@@ -104,6 +126,15 @@ def create_node(self, **kwargs):
         return self._to_node(response.object)
 
     def destroy_node(self, node):
+        """
```
Kami: Docstrings for methods which are part of the standard API are not required (unless you changed the method signature) since they are inherited from the parent class.

libcloud/compute/drivers/runabove.py
```diff
@@ -52,13 +39,20 @@ class RunAboveNodeDriver(NodeDriver):
     api_name = 'runabove'
 
     NODE_STATE_MAP = OpenStackNodeDriver.NODE_STATE_MAP
+    VOLUME_STATE_MAP = OpenStackNodeDriver.VOLUME_STATE_MAP
```
Kami, line 42: I've missed this when originally reviewing the PR - it looks like their offering is based on OpenStack.

Any particular reason why you didn't inherit from the base OpenStack driver (that's what a lot of other drivers for providers which are based on the OpenStack do).

The auth is not using Keystone, but I quickly glanced over other methods and at least some of them look similar to the ones implemented in the OpenStack driver :)

libcloud/compute/drivers/runabove.py
ZuluPro: What do you want to say by "to explicitly declare the arguments here." ?

libcloud/compute/drivers/runabove.py
ZuluPro: Corrected!

libcloud/compute/drivers/runabove.py
ZuluPro, line 42: RunAbove is based on OpenStack but they propose their own API, which is more simple than the OpenStack one. It has an abstraction layer which notably make easy to deal between datacenters.

libcloud/compute/drivers/runabove.py
ZuluPro: Fixed!

libcloud/compute/drivers/runabove.py
Kami: Sorry, I just mean you should declare all the arguments in the method signature, e.g. `def create_node(self, name, image, size, location, ex_keyname=None)`.

libcloud/compute/drivers/runabove.py
Kami, line 42: Yeah, I assumed something like that was the case, but I just wanted to confirm it.

"simple" vs "standard" could be debated, but that's unrelated to this pr :)

Kami:
I've made the "create_node" changed and merged patch into trunk. Thanks!

ZuluPro:
Oh! I made a `git commit --amend` for less commit!

Kami:
Heh, thanks and no problem (changes were already merged).

PR #562, opened 2015-08-17
Support S3 sa-east-1

iurisilvio:


Kami:
Merged into trunk. Thanks!

PR #564, opened 2015-08-25
Libcloud 736 update dimension data driver api 2 0

tonybaloney:
Updated the dimension data driver to support new capabilities and a new API (2.0). All server basic functions (start, stop etc) have been updated. List nodes has been updated, also extended support for listing network domain (VRF type function), VLAN and kept compatibility with the existing driver base.
All existing unit tests were updated and extended the coverage of the test class to include the new functions along with fixtures for the expected server responses.


Kami:
Merged into trunk. Thanks!

PR #566, opened 2015-08-26
[Libcloud-732] Add implementation for World Wide DNS provider

aleGpereira:
Jira ticket is https://issues.apache.org/jira/browse/LIBCLOUD-732

### DIFF
--- a/libcloud/common/worldwidedns.py
+++ b/libcloud/common/worldwidedns.py
@@ -1,0 +1,195 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import re
+
+from libcloud.common.base import ConnectionUserAndKey
+from libcloud.common.base import Response
+from libcloud.common.types import ProviderError
+
+
+OK_CODES = ['200', '211', '212', '213']
+ERROR_CODES = ['401', '403', '405', '406', '407', '408', '409', '410', '411',
+               '412', '413', '414', '450', '451']
+
+
+class WorldWideDNSException(ProviderError):
+    def __init__(self, value, http_code, code, driver=None):
+        self.code = code
+        super(WorldWideDNSException, self).__init__(value, http_code, driver)
+
+
+class SuspendedAccount(WorldWideDNSException):
+    def __init__(self, http_code, driver=None):
+        value = "Login ID you supplied is SUSPENDED, you need to renew" + \
+                " your account"
+        super(SuspendedAccount, self).__init__(value, http_code, 401,
+                                               driver)
+
+
+class LoginOrPasswordNotMatch(WorldWideDNSException):
+    def __init__(self, http_code, driver=None):
+        value = "Login ID and/or Password you supplied is not on file or" + \
+                " does not match"
+        super(LoginOrPasswordNotMatch, self).__init__(value, http_code, 403,
+                                                      driver)
+
+
+class NonExistentDomain(WorldWideDNSException):
+    def __init__(self, http_code, driver=None):
+        value = "Domain name supplied is not in your account"
+        super(NonExistentDomain, self).__init__(value, http_code, 405,
+                                                driver)
+
+
+class CouldntRemoveDomain(WorldWideDNSException):
+    def __init__(self, http_code, driver=None):
+        value = "Error occured removing domain from name server, try again"
+        super(CouldntRemoveDomain, self).__init__(value, http_code, 406,
+                                                  driver)
+
+
+class LimitExceeded(WorldWideDNSException):
+    def __init__(self, http_code, driver=None):
+        value = "Your limit was exceeded, you need to upgrade your account"
+        super(LimitExceeded, self).__init__(value, http_code, 407,
+                                            driver)
+
+
+class ExistentDomain(WorldWideDNSException):
+    def __init__(self, http_code, driver=None):
+        value = "Domain already exists on our servers"
+        super(ExistentDomain, self).__init__(value, http_code, 408,
+                                             driver)
+
+
+class DomainBanned(WorldWideDNSException):
+    def __init__(self, http_code, driver=None):
+        value = "Domain is listed in DNSBL and is banned from our servers"
+        super(DomainBanned, self).__init__(value, http_code, 409,
+                                           driver)
+
+
+class InvalidDomainName(WorldWideDNSException):
+    def __init__(self, http_code, driver=None):
+        value = "Invalid domain name"
+        super(InvalidDomainName, self).__init__(value, http_code, 410,
+                                                driver)
+
+
+class ErrorOnReloadInNameServer(WorldWideDNSException):
+    def __init__(self, server, http_code, driver=None):
+        if server == 1:
+            value = "Name server #1 kicked an error on reload, contact support"
+            code = 411
+        elif server == 2:
+            value = "Name server #2 kicked an error on reload, contact support"
+            code = 412
+        elif server == 3:
+            value = "Name server #3 kicked an error on reload, contact support"
+            code = 413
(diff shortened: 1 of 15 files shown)


aleGpereira:
I'm not sure if i followed the contributor guideline in the correct way. I'll expect for comments.

Kami:
Great, thanks. I will look as soon as I can.

libcloud/common/worldwidedns.py
```diff
@@ -0,0 +1,200 @@
...
+
+    def parse_body(self):
+        """
+        Parse response body.
+
+        :return: Parsed body.
+        :rtype: ``str``
+        """
+        if self._code_response(self.body):
+            codes = self.body.split('\r\n')
```
Kami: Using a regular expression and `re.split` would be safer (e.g. `\r?\n`) unless the body always contains carriage return (\r).

libcloud/dns/drivers/worldwidedns.py
```diff
@@ -0,0 +1,506 @@
...
+        action = '/api_dns_list.asp'
+        if self.reseller_id is not None:
+            action = '/api_dns_list_reseller.asp'
+        zones = self.connection.request(action)
+        if len(zones.body) == 0:
+            return []
+        else:
+            return self._to_zones(zones.body)
+
+    def list_records(self, zone):
```
Kami: If you implement `iterate_records` you don't need to implement `list_records` since it's already implemented in the base driver as `list(self._iterate_records(zone))`.

libcloud/dns/drivers/worldwidedns.py
```diff
@@ -0,0 +1,506 @@
...
+        :param zone_id: ID of the required zone
+        :type  zone_id: ``str``
+
+        :param record_id: ID of the required record
+        :type  record_id: ``str``
+
+        :rtype: :class:`Record`
+        """
+        zone = self.get_zone(zone_id)
+        records = self.list_records(zone)
```
Kami: I assume there isn't any more efficient way to retrieve a single record (e.g. filter by id or similar)?

libcloud/dns/drivers/worldwidedns.py
```diff
@@ -0,0 +1,506 @@
...
+        """
+        Update en existing zone.
+
+        :param zone: Zone to update.
+        :type  zone: :class:`Zone`
+
+        :param domain: Zone domain name (e.g. example.com)
+        :type  domain: ``str``
+
+        :param raw: Mode we use to do the update using zone file or not.
```
Kami: This is an extension (driver specific) argument so please prefix it with `ex_` (`ex_raw=False`) and put it at the end after `extra`.

libcloud/dns/drivers/worldwidedns.py
```diff
@@ -0,0 +1,506 @@
...
+            action = '/api_dns_modify.asp'
+            if self.reseller_id is not None:
+                action = '/api_dns_modify_reseller.asp'
+            method = 'GET'
+        response = self.connection.request(action, params=params,  # noqa
+                                           method=method)
+        zone = self.get_zone(zone.id)
+        return zone
+
+    def update_record(self, record, name, type, data, entry):
```
Kami: Same here, please prefix `entry` with `ex_` and put it after `extra`.

``` python
def update_record(self, record, name, type, data, extra=None, ex_position=1):
...
```

libcloud/dns/drivers/worldwidedns.py
```diff
@@ -0,0 +1,506 @@
...
+        :type extra: ``dict``
+
+        :rtype: :class:`Zone`
+
+        For more info, please see
+        https://www.worldwidedns.net/dns_api_protocol_new_domain.asp
+        or
+        https://www.worldwidedns.net/dns_api_protocol_new_domain_reseller.asp
+        """
+        if type == "master":
```
Kami: Minor style thing - please use single quotes around strings (minus docstrings).

libcloud/dns/drivers/worldwidedns.py
```diff
@@ -0,0 +1,506 @@
...
+from libcloud.common.types import LibcloudError
+from libcloud.common.worldwidedns import WorldWideDNSConnection
+from libcloud.dns.types import Provider, RecordType
+from libcloud.dns.types import ZoneDoesNotExistError
+from libcloud.dns.types import RecordError
+from libcloud.dns.types import RecordDoesNotExistError
+from libcloud.dns.base import DNSDriver, Zone, Record
+
+
+MAX_RECORD_ENTRIES = 40
```
Kami: It would be good to add a comment which explains what this constant represents.

libcloud/dns/drivers/worldwidedns.py
```diff
@@ -0,0 +1,506 @@
...
+            if record.name == zone.extra['S%s' % (index + 1)]:
+                entry = index + 1
+                break
+        extra = {'S%s' % entry: '',
+                 'T%s' % entry: 'NONE',
+                 'D%s' % entry: ''}
+        self.update_zone(zone, zone.domain, extra=extra)
+        return True
+
+    def view_zone(self, domain, name_server):
```
Kami: This is an extension method so please prefix the method name with `ex_` - `def ex_view_zone(self, domain, name_server):`

libcloud/dns/drivers/worldwidedns.py
```diff
@@ -0,0 +1,506 @@
...
+        """
+        params = {'DOMAIN': domain,
+                  'NS': name_server}
+        action = '/api_dns_viewzone.asp'
+        if self.reseller_id is not None:
+            action = '/api_dns_viewzone_reseller.asp'
+        response = self.connection.request(action, params=params)
+        return response.object
+
+    def transfer_domain(self, domain, user_id):
```
Kami: Same here, method name needs `ex_` prefix.

libcloud/common/worldwidedns.py
```diff
@@ -0,0 +1,200 @@
...
+            for code in codes:
+                if code in OK_CODES:
+                    continue
+                elif code in ERROR_CODES:
```
Kami, line 159: An alternative which allows this method to be shorted would be to create a module level dict constant which maps code to the exception class (e.g. `ERROR_CODE_TO_EXCEPTION_CLS`) and using it here.

libcloud/dns/drivers/worldwidedns.py
```diff
@@ -0,0 +1,506 @@
...
+        if self.reseller_id is not None:
+            params['DYN'] = dyn
+            action = '/api_dns_new_domain_reseller.asp'
+        self.connection.request(action, params=params)
+        zone = self.get_zone(domain)
+        if ttl is not None:
+            zone = self.update_zone(zone, zone.domain, ttl=ttl)
+        return zone
+
+    def create_record(self, name, zone, type, data, entry):
```
Kami: Same here - `entry` needs to be prefixed with `ex_`.

Kami:
@aleGpereira Thanks. I've added some comments.

Overall it looks good :+1:, there are just some small things which need to be addressed.

libcloud/common/worldwidedns.py
aleGpereira: Always contains the carriage. At least the documentation says that.

aleGpereira:
Ok, changes are pushed :).

Kami:
Thanks. I just wanted to merge it, but I encountered an issue with a bunch of test failing:

```
ERROR: test_delete_zone_success (__main__.WorldWideDNSTests)
  File "libcloud/test/dns/test_worldwidedns.py", line 220, in test_delete_zone_success
    zone = self.driver.list_zones()[0]
  File "/w/lc/libcloud/libcloud/dns/drivers/worldwidedns.py", line 106, in list_zones
    return self._to_zones(zones.body)
  File "/w/lc/libcloud/libcloud/dns/drivers/worldwidedns.py", line 450, in _to_zones
    zone = self._to_zone(line)
  File "/w/lc/libcloud/libcloud/dns/drivers/worldwidedns.py", line 466, in _to_zone
    extra = {'HOSTMASTER': soa_block[0], 'REFRESH': soa_block[1],
IndexError: list index out of range
```

I had a look and it looks like something weird is going on with the splitting. As mentioned above, to be one the safe side and make it more robust I think we should also use regexed when splitting so we handle both, windows and linux new lines (with and without the carriage return).

On top of that, lint is also failing but it's a minor issue - https://travis-ci.org/apache/libcloud/builds/78253070

@aleGpereira Can you please have a look and fix it? Thanks.

aleGpereira:
Ok. I'll check for that. Thanks.

aleGpereira:
Ok. I've fix the problem. Sorry for that. I didn't ran with tox before. The regex is taking place now.

Kami:
@aleGpereira One thing I forgot - since those contributions are pretty big, can you please also sign an ICLA - https://www.apache.org/licenses/#clas

Kami:
Merged, thanks!

aleGpereira:
@Kami ICLA is signed and filed :).

PR #567, opened 2015-08-28
Libcloud 737 create dimension data load balancing driver

tonybaloney:
Created a dimension data driver for load balancing

placed proprietary logic into ex_methods
supported existing interfaces for LoadBalancer
moved common classes and the connection class used in compute into a common library
added unit tests and fixtures for the load balancer driver
updated the docs
tested against expected results and API docs


libcloud/loadbalancer/drivers/dimensiondata.py
```diff
@@ -0,0 +1,741 @@
...
+        """
+            Add the region to the kwargs before the connection is instantiated
+        """
+
+        kwargs = super(DimensionDataLBDriver,
+                       self)._ex_connection_class_kwargs()
+        kwargs['region'] = self.selected_region
+        return kwargs
+
+    def create_balancer(self, name, port, protocol, algorithm, members):
```
Kami, line 95: Is `members` a required argument or you can create a load-balancer with no initial members and attach them later?

tonybaloney, line 95: I didn't test that case of having members as None but technically you can create a virtual listener pool with no members. I'll add a check before it adds members to check members is not None and add a unit test for that scenario. 

On Sat, Aug 29, 2015 at 9:19 AM -0700, "Tomaz Muraus" notifications@github.com wrote:

In libcloud/loadbalancer/drivers/dimensiondata.py:

> -                                                    port=port,
> -                                                    api_version=api_version,
> -                                                    region=region,
> -                                                    **kwargs)
>   +
> -    def _ex_connection_class_kwargs(self):
> -        """
> -            Add the region to the kwargs before the connection is instantiated
> -        """
>   +
> -        kwargs = super(DimensionDataLBDriver,
> -                       self)._ex_connection_class_kwargs()
> -        kwargs['region'] = self.selected_region
> -        return kwargs
>   +
> -    def create_balancer(self, name, port, protocol, algorithm, members):

Is members a required argument or you can create a load-balancer with no initial members and attach them later?

Kami, line 95: Alright, thanks.

Besides that, it looks good to me.

tonybaloney:
Checked in updates and tests to support the use case you mentioned. Also added a few more tests and some docstrings for when the docs get generated

Kami:
Thanks.

I thought that the base driver and standard API defaults `members` to `None` and makes it optional, but looks like that is not the case.

Imo, it makes sense to allow users to create load balancer without any initial members (e.g. you want to pre-allocate a balancer so you don't need to wait for the creation later on when you need it), so I will also go ahead and make the change in the base API as well.

Kami:
I'm having issues (a bunch of merge conflicts) while trying to merge those changes into trunk.

Can you please synchronize your branch with the latest upstream trunk?

tonybaloney:
I must be missing something here, some scaffolding logic to add the new unit tests class to the general test suite. I was always running the tests directly.

Kami:
You mean you are not sure why tests fail on Travis CI or something else?

I can look while merging it. Could be an issue with an import or similar.

tonybaloney:
I've done the sync upstream, merged into my fork, then the branch, so auto
merge should be good now, but travis is failing:

loadTestsFromName

```
parent, obj = obj, getattr(obj, part)
```

AttributeError: 'module' object has no attribute 'test_dimensiondata'

But, it exists..

https://github.com/DimensionDataCBUSydney/libcloud/blob/LIBCLOUD-737_Create_Dimension_Data_Load_Balancing_driver/libcloud/test/loadbalancer/test_dimensiondata.py

Kami:
@tonybaloney compute driver was the one causing failure. I've fixed the issue locally and will merge it shortly:

``` python
python libcloud/test/compute/test_dimensiondata.py
  File "libcloud/test/compute/test_dimensiondata.py", line 21, in <module>
    from libcloud.compute.drivers.dimensiondata import DimensionDataAPIException
ImportError: cannot import name DimensionDataAPIException
```

I will also look into the test loading code and see if it's possible to make it report more user-friendly errors.

tonybaloney:
We just reached the same conclusion. Threw me off having 2 test classes of the same name! Pushed and passing. It should be clear to merge now.

Kami:
Merged, thanks!

PR #568, opened 2015-08-28
[LIBCLOUD-725] Fix handling of binary data in LocalStorageDriver on Py3.

torfsen:
This PR fixes issue [LIBCLOUD-725](https://issues.apache.org/jira/browse/LIBCLOUD-725): `libcloud.storage.driver.local.LocalStorageDriver` has problems with binary data on Python 3.

The fix is to make sure that files are opened in binary mode for reading and writing.

Tests and documentation have been updated.


Kami:
Changes look good to made so I went ahead and merged them into trunk. Thanks!

/cc @mahendra

PR #569, opened 2015-08-28
Adding ex_virtual_network_name and ex_network_config to create_node

jeverling:
Make it possible to use virtual network with Azure VM.


Kami:
Looks good. Merged into trunk. Thanks!

PR #571, opened 2015-09-02
GCE: Update generic auth error message

mtekel:
Google authentication will fail if your time drift (difference between your local time and time on google servers) is bigger than 5 minutes. This often happens on the VM instances running in various cloud providers. Google itself doesn't provide any helpful indication of the issue, the response is just `{u'error': u'invalid_grant'}`.

There is also problem in the `_get_error(self, body)`, line 205: the error response returned by google above will mean that err is a string, not a dictionary. `err.get` will then throw attribute error, which displays generic error message. I think it is better to display the generic error message than not so helpful 'invalid_grant' response from google. You might decide to fix or change that later, but for now I think it's important to hint about time drift to the user, as this happens quite often and it's hard to find, if you only check whether your credentials are correct (which they can be, but you'll still fail auth with big time drift).

### DIFF
--- a/docs/compute/drivers/gce.rst
+++ b/docs/compute/drivers/gce.rst
@@ -41,7 +41,9 @@ Which one should I use?
   authorization is your Project ID.
 
 Once you have set up the authentication as described below, you pass the
-authentication information to the driver as described in `Examples`_
+authentication information to the driver as described in `Examples`_. Also
+bear in mind that large clock drift (difference in time) between authenticating
+host and google will cause authentication to fail.
 
 
 Service Account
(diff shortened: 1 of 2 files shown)


Kami:
It looks like it would also be good to document this in the docs.

Besides that, LGTM. Thanks.

/cc @erjohnso

mtekel:
Hi,

I am not sure what is the best place to put this information in the docs. Should that be a separate PR?

Kami:
Ideally, it would be the same PR. 

You could put it in the section where it's document how you retrieve the credentials.

mtekel:
Added doc update. Hopefully it's in the right place.

mtekel:
Hi, 

have you had time to look at this? It's not a big change, but it can save some people a lot of time debugging issues...

Kami:
Merged, thanks!

PR #572, opened 2015-09-02
Libcloud 742 verizon compute

replicant0wnz:
Verizon Enterprise Cloud compute driver.

https://issues.apache.org/jira/browse/LIBCLOUD-742


replicant0wnz:
Still need to add tests

tonybaloney:
Also, don't forget to update docs/compute/_supported_providers.rst and create a page for your driver (otherwise the docstrings wont live anywhere).

tonybaloney:
^^ fair enough. that makes sense and aligns with most of the other drivers (e.g. https://libcloud.readthedocs.org/en/latest/compute/drivers/cloudstack.html#libcloud.compute.drivers.cloudstack.CloudStackNodeDriver.create_node)

Kami:
@tonybaloney That's actually an old leftover - a bad example of abusing kwargs before we learned better :)

A lot of times using kwargs might seem like a neat thing to do, but it turns out that there aren't actually that many scenarios where using them is a good idea (e.g. decorators). Using them where there is no need to makes the code harder to read and understand (to understand which arguments the method takes you now also need to inspect the body). On top of that it also makes "static" analysis hard / impossible - https://libcloud.readthedocs.org/en/latest/development.html#don-t-abuse-kwargs.

libcloud/compute/drivers/verizon.py
```diff
@@ -0,0 +1,890 @@
...
+
+        url = '%s/virtualmachines/computepools/%s/action/createvirtualmachine' % (API_ROOT, pool)
+
+        headers = {'Content-Length': len(json.dumps(vm))}
+
+        result = self.connection.request(url, method='POST', headers=headers, data=json.dumps(vm)).object
+        result['ip'] = ipaddr
+        return self._to_node(result)
+
+    def deploy_node(self, deploy=None, **kwargs):
```
Kami, line 493: How is this different from base `deploy_node` method?

I'm not a fan of duplicating that much functionality, especially since this method already caused a lot of issues in the past...

libcloud/compute/drivers/verizon.py
```diff
@@ -0,0 +1,890 @@
...
+        :type    node: :class:`Node` (mandatory)
+
+        :return: State of the node
+        :rtype: :class: `Node`
+        """
+        if node is None:
+            raise VerizonException('Parameter node is required')
+        href = None
+        for i in node.extra['Actions']:
+            if i is not None:
```
Kami, line 814: Would make it slightly easier to read if you combined into a single lint - `if is is not None and i['name'] == 'power:powerOff'`

libcloud/compute/drivers/verizon.py
```diff
@@ -0,0 +1,890 @@
...
+        """
+        Check to see if a libcloud row and column exist. If not, create
+        """
+        env = self.ex_get_default_env()
+        url = API_ROOT + '/layout/environments/' + env
+        rows = self.connection.request(url).object['Rows']
+        row = None
+        grp = None
+        for i in rows:
+            if i['name'] == 'libcloud':
```
Kami, line 871: Is it possible to refactor this in multiple methods to avoid that many levels of nesting which makes code hard to read?

Kami:
I've added some more comments - mostly style.

Also, how are tests coming along? Need any help?

libcloud/compute/drivers/verizon.py
replicant0wnz, line 493: By default the Verizon cloud product doesn't power on the instance after it's deployed. I had to use the ex_wait_create() and ex_poweron_node() functions in order to get it to work. There may be a cleaner way to do it if you can recommend something.

libcloud/compute/drivers/verizon.py
replicant0wnz, line 871: Actually yes, I could do a _libcloud_check_row() and _libcloud_check_group() and then reference them.

replicant0wnz:
@luckenbach @marbindrakon did y'all ever get around to writing tests? It appears @Kami is offering to help ;-)

PR #573, opened 2015-09-07
GCE: Fix check_node indentation to match check_disk

fpoms:
The _multi_check_node later if statements were somehow indented incorrectly and a try except was inserted to work around the issue. This fixes the indentation problem so that the self.ex_get_node call is only made when the original create_node call has finished creating the node.

The code now also matches the _multi_disk_node function which has the proper indentation.

### DIFF
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -4899,15 +4899,11 @@ class GCENodeDriver(NodeDriver):
             return
         if response['status'] == 'DONE':
             status['node_response'] = None
-        if error:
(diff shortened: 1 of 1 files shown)


Kami:
Sorry for the delay and thanks.

I'm not too familiar with this code so I will wait for @erjohnso before merging.

erjohnso:
Forgot to indicate last commit 80a9aa20148fdd41100d0afb2fad3243d543a25e closed this.

Thanks for the fix @abpoms!

sebgoa:
@erjohnso @Kami so we will need a blank commit with "This closes #573" to get it closed.
Or @abpoms you can close it manually.

PR #574, opened 2015-09-08
GCE: Fix calling non-existent function

fpoms:
ex_get_disk does not exist as a function on the GCEDriver object. This replaces the improper ex_get_disk call with the correct ex_get_volume call.


Kami:
LGTM. Thanks.

I also need to resume with the project of enabling pylint so those things are detected automatically.

/cc @erjohnso

Kami:
Merged, thanks.

PR #577, opened 2015-09-11
Fix traceback when simplejson < 2.1.0 is installed

terminalmage:
See https://issues.apache.org/jira/browse/LIBCLOUD-714.


Kami:
Merged into trunk. Thanks.

PR #578, opened 2015-09-14
Proxy_url & connection object fixes

MrBasset:
I have a use case that requires explicitly setting the proxy_url for accessing a vCloud API (as opposed to the default environment proxy). I noticed that the proxy_url was not being passed to the base Connection class by the base NodeDriver. I fixed this by adding the proxy_url into the constructor call where the connection class is created.

Also, within the vCouldNodeDriver a new connection class we being created with
no custom variables rather than re-using the existing connection class. Any custom values, such as the proxy_url, were not being passed to this new instance. By swapping to the self.connection class these variables are set.

I could not find any unit tests that would demonstrate this issue (I would assume that they would be integration tests anyway). So apologies that this fix does not have supporting tests.

### DIFF
--- a/libcloud/common/abiquo.py
+++ b/libcloud/common/abiquo.py
@@ -187,7 +187,8 @@ class AbiquoConnection(ConnectionUserAndKey, PollingConnection):
                                                host=host, port=port,
                                                url=url, timeout=timeout,
                                                retry_delay=retry_delay,
-                                               backoff=backoff, proxy_url=proxy_url)
+                                               backoff=backoff,
+                                               proxy_url=proxy_url)
 
         # This attribute stores data cached across multiple request
         self.cache = {}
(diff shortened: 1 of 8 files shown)


Kami:
Thanks. The changes look good to me.

Looks like the build is failing - https://travis-ci.org/apache/libcloud/builds/80160975. Can you please fix this?

MrBasset:
Sorry, will look at what I've done wrong.

MrBasset:
Kami,

I've managed to get a green light on the build - sorry it took so long. It seems that by adding that parameter into the base connection class, has highlighted that a number of the vendor specific drivers also didn't pass down the proxy_url parameter. I'm not sure if explicitly defining the proxy_url is the best approach, or whether it would be better to use **kwargs etc? I'd assumed the former, hopefully this isn't a breaking change.

If this change is too wide reaching I'm happy to take advice on how to better implement it.

Kami:
Merged, thanks.

PR #579, opened 2015-09-14
[LIBCLOUD-745] Add Vultr DNS provider implementation

aleGpereira:
I have another provider to add. Thanks for your awesome work guys.


libcloud/dns/drivers/vultr.py
```diff
@@ -0,0 +1,391 @@
...
+            raise RecordDoesNotExistError(value='', driver=self,
+                                          record_id=record_id)
+
+        for record in records:
+            if record_id == record.id:
+                ret_record = record
+
+        return ret_record
+
+    def create_zone(self, zone_id, type='master', ttl=None, extra={}):
```
Kami, line 161: Not a good idea to default to mutable types (dicts, lists), I will change it to default to `None` and do `extra = extra or {}` inside the method body when merging the PR.

Kami:
Thanks for syncing the branch with latest trunk :)

Kami:
Fixed minor Python 3 compatibility issue and merged patch into trunk. Thanks!

PR #580, opened 2015-09-15
[LIBCLOUD-749] CloudStack: fixed method ex_authorize_security_group_ingress

schaubl:
Corresponding ticket in the JIRA issue tracker: [LIBCLOUD-749](https://issues.apache.org/jira/browse/LIBCLOUD-749)

This PR fixes the following bugs:
- The docstring doesn't match parameters of the method.
- Only `ICMP` and `TCP` is allowed even if the doc of the parameter `protocol` mention `TCP` and `UDP`.
- Moreover `ICMP` cannot be used because there is no way to specify the two required parameters `icmptype` and `icmpcode`.
- The parameter `endport` is not used if it's defined.
- The return type is not a `list` but a `dict`.
- The unit test doesn't use parameters in the correct order.

This PR also does the following changes:
- Improve the docstring to be a bit more detailed (the previous one comes from the official Apache CloudStack documentation).
- Add a `kwargs` to allow to provide extra parameters defined in the API and in the docstring.
- Improve the unit test.
- Add support for other protocols like `AH`, `GRE` and `ESP` which are available on some Cloud like Exoscale.


libcloud/compute/drivers/cloudstack.py
```diff
@@ -3489,70 +3489,108 @@ def ex_delete_security_group(self, name):
                                   params={'name': name},
                                   method='GET')['success']
 
-    def ex_authorize_security_group_ingress(self, securitygroupname,
-                                            protocol, cidrlist, startport,
-                                            endport=None):
+    def ex_authorize_security_group_ingress(self, securitygroupname, protocol,
+                                            cidrlist, startport=None,
+                                            endport=None, icmptype=None,
+                                            icmpcode=None, **kwargs):
```
Kami, line 3495: Can you please explicitly declare all the supported arguments instead of using **kwargs?

Kami:
Added one comment, besides that, LGTM. Thanks.

/cc @runseb

sebgoa:
LGTM, I will merge in the coming days

schaubl:
@Kami : Thanks for the comment which allowed me to discover a small issue in the docstring (`securitygroupid` and `usersecuritygrouplist` were noted as `:param` instead of `:keyword`).

> Can you please explicitly declare all the supported arguments instead of using **kwargs?

Yes, of course. I used `kwargs` because it's the way some "_unusual_" parameters (like `domainid` and `projectid`) are passed with this driver. For example the method [create_key_pair](https://github.com/schaubl/libcloud/blob/LIBCLOUD-749_cloudstack-fixed-ex_authorize_security_group_ingress/libcloud/compute/drivers/cloudstack.py#L2292)  takes the name of the keypair and then parameters like `domainid` and `projectid` are passed via `kwargs`.

To be consistent I would suggest to keep it like this.
Please tell me if you still want than these parameters to be explicitly declared.

Kami:
Sorry for the delay - I went ahead and merged the changes.

For now, lets keep `kwargs`, but in the future it would be great if we can get rid of it and explicitly declare all the supported arguments - (ab)using kwargs makes many things such as "static" analysis hard / impossible.

PR #583, opened 2015-09-18
[google dns] update API from v1beta1 to v1

mbrukman:
The API itself is unchanged from v1beta1 to v1, but v1beta1 will be removed
after 25 Sep 2015 per https://cloud.google.com/dns/api/v1beta1/ so we need to
update the version to avoid breakage.

@erjohnso, please review and commit.


Kami:
Thanks. LGTM.

/cc @erjohnso

erjohnso:
Thanks @mbrukman!

PR #584, opened 2015-09-22
Clarify DreamHost provider is for VPS product

smaffulli:
DreamHost has a VPS product and a cloud product, based on OpenStack.
This patch clarifies that the libcloud provider for DreamHost is for
the VPS product, not for DreamCloud.

Signed-off-by: Stefano Maffulli stefano.maffulli@dreamhost.com


Kami:
Merged, thanks.

PR #586, opened 2015-09-23
Added Runabove storage

ZuluPro:


libcloud/common/runabove.py
```diff
@@ -55,12 +55,19 @@ class RunAboveResponse(JsonResponse):
     def parse_error(self):
         response = super(RunAboveResponse, self).parse_body()
 
+        status_code = int(self.status)
+        if status_code >= 300 and status_code < 600:
+            raise ProviderError('Unknown error. Status code: %d' % self.status,
+                                http_code=status_code)
         if response.get('errorCode', None) == 'INVALID_SIGNATURE':
```
Kami: Which status code do they return on invalid_signature?

Just want to make sure it's `< 300` so we don't need to move this if before the one you just added.

libcloud/storage/drivers/runabove.py
```diff
@@ -0,0 +1,289 @@
...
+        :rtype: ``None``
+        """
+        self.datacenter = None
+        self.consumer_key = ex_consumer_key
+        self.location = ex_location
+        StorageDriver.__init__(self, key, secret,
+                               ex_consumer_key=ex_consumer_key)
+        response = self.connection.request(action='%s/token' % API_ROOT)\
+            .object
+        self.auth_token = response["X-Auth-Token"]
```
Kami: Not a big fan of constructors with side-affects.

Is it possible to move this to the connection class and perform it lazily on the first request?

libcloud/storage/drivers/runabove.py
```diff
@@ -0,0 +1,289 @@
...
+    def iterate_container_objects(self, container):
+        action = '%s/storage/%s' % (API_ROOT, container.name)
+        location = container.extra['region']
+        data = {
+            'region': location,
+            'limit': 1000,
+        }
+        try:
+            response = self.connection.request(action, data=data)
+        except ProviderError as err:
```
Kami: Please use `e = sys.exc_info()[1]` instead of `as err` since we still support Python 2.5.

libcloud/storage/drivers/runabove.py
```diff
@@ -0,0 +1,289 @@
...
+        action = '%s/storage/%s' % (API_ROOT, container_name)
+        data = {
+            'region': location,
+            'limit': 1000,
+        }
+        try:
+            response = self.connection.request(action, data=data)
+            json_response = response.object
+            container = self._to_container(json_response, location)
+            obj = [o for o in json_response['objects']
```
Kami: This late filtering could be very inefficient and slow if there are many objects...

Is it possible to do it differently (e.g. only retrieve details for a single object or perform some kind of name filtering)?

libcloud/storage/drivers/runabove.py
```diff
@@ -0,0 +1,289 @@
...
+                      verify_hash=True, headers=None):
+        action = '/%s/%s' % (container.name, object_name)
+        data = open(file_path, 'rb').read()
+        self._put_data(container.extra['region'], action, data)
+        return container.get_object(object_name)
+
+    def upload_object_via_stream(self, iterator, container, object_name,
+                                 extra=None, headers=None):
+        action = '/%s/%s' % (container.name, object_name)
+        data = iterator.read()
```
Kami, line 210: Hm, so this works around the whole "memory efficient" streaming approach implemented in the base driver...

This is less than ideal since, among other things, it results in reading and buffering the whole object in memory. Can you look at other drivers for an example (e.g. CloudFiles or S3, notably `upload_object`, `upload_object_via_stream`, `_upload_object_part` and `_put_object` methods) and see if you can use the same approach / pattern here?

libcloud/storage/drivers/runabove.py
```diff
@@ -0,0 +1,289 @@
...
+            if err.http_code == httplib.NOT_FOUND:
+                raise ContainerDoesNotExistError(value=None,
+                                                 driver=self,
+                                                 container_name=None)
+            raise err
+
+    def upload_object(self, file_path, container, object_name, extra=None,
+                      verify_hash=True, headers=None):
+        action = '/%s/%s' % (container.name, object_name)
+        data = open(file_path, 'rb').read()
```
Kami: This file is never closed (you should use `with`), but the correct solution is to re-use `_upload_object` from the base driver as mentioned in the comment below.

Kami:
Thanks!

I've added some comments which need to be addressed, but besides that, it looks good :+1:

ZuluPro:
OK, I will correct!

libcloud/common/runabove.py
ZuluPro: It returns a 400.
For fix: I've put INVALID_SIGNATURE check before status code check.

Kami: Great, thanks.

ZuluPro:
Updated !

ZuluPro:
@Kami: Any news ?

Kami:
I didn't know you updated the PR - next time please mention my nick in the comment since I only get notifications if you directly reply to my comment :)

I will have a look asap, but it might take a while since I also plan to sign-up for Aurora objects and test it out.

libcloud/test/storage/test_runabove.py
```diff
@@ -0,0 +1,243 @@
...
+        container = RunAboveContainer(name='foo', extra={'region': 'BHS-1'}, driver=self.driver)
+        self.driver.upload_object(test_file_path, container, 'foo_file')
+
+    def test_upload_object_via_stream(self):
+        test_file = StringIO('bar')
+        container = RunAboveContainer(name='foo', extra={'region': 'BHS-1'}, driver=self.driver)
+        obj = self.driver.upload_object_via_stream(test_file, container, 'foo_file')
+        self.assertEqual('foo_file', obj.name)
+
+    # def test_download_object(self):
```
tonybaloney, line 211: does/would this test work?

ZuluPro, line 211: @tonybaloney 
The test should work but raise me `RuntimeError: maximum recursion depth exceeded while calling a Python object`:

```
ERROR: test_download_object (__main__.RunAboveStorageTests)
  File "libcloud/test/storage/test_runabove.py", line 217, in test_download_object
    data = self.driver.download_object(obj, '/tmp/foo2')
  File "/usr/local/lib/python2.7/dist-packages/apache_libcloud-0.18.0-py2.7.egg/libcloud/storage/drivers/runabove.py", line 266, in download_object
    callback_kwargs={'response': response.response, 'obj': obj,
  File "/usr/local/lib/python2.7/dist-packages/apache_libcloud-0.18.0-py2.7.egg/libcloud/common/base.py", line 266, in response
    if not self.success():
  File "/usr/local/lib/python2.7/dist-packages/apache_libcloud-0.18.0-py2.7.egg/libcloud/common/base.py", line 179, in success
    return self.status in [httplib.OK, httplib.CREATED]
  File "/usr/local/lib/python2.7/dist-packages/apache_libcloud-0.18.0-py2.7.egg/libcloud/common/base.py", line 273, in status
    self._status = self.response.status
  File "/usr/local/lib/python2.7/dist-packages/apache_libcloud-0.18.0-py2.7.egg/libcloud/common/base.py", line 266, in response
    if not self.success():
  File "/usr/local/lib/python2.7/dist-packages/apache_libcloud-0.18.0-py2.7.egg/libcloud/common/base.py", line 179, in success
    return self.status in [httplib.OK, httplib.CREATED]
  File "/usr/local/lib/python2.7/dist-packages/apache_libcloud-0.18.0-py2.7.egg/libcloud/common/base.py", line 273, in status
    self._status = self.response.status
  File "/usr/local/lib/python2.7/dist-packages/apache_libcloud-0.18.0-py2.7.egg/libcloud/common/base.py", line 266, in response
    if not self.success():
...
```

I tested manually, it works but I'm not able to make unit test with `_get_object` mechanism (some help would be appreciated)

ZuluPro:
I didn't finish, just don't have the time

tonybaloney:
hey @ZuluPro I'm just closing off any PRs that are >1yr old. If you want to reopen it that's fine. Some of the APIs have changed since so your unit tests probably won't pass first time

PR #587, opened 2015-09-23
LIBCLOUD-750 MCP2 deploy support

tonybaloney:
- Updated the create_node methods to support the new API, and there every datacenter.
- Updated all of the _to_x methods so that they don't lookup the locations for each iteration, significantly quickening the time to make queries for large accounts
- Updated the unit tests to reflect more production data
- The deploy returns the node id now
- Added support for filtering networks, image, network domain and vlan by location


Kami:
Looks good to me. Merged into trunk, thanks.

PR #588, opened 2015-09-24
[LIBCLOUD-748] Add Durable DNS provider implementation

aleGpereira:
Ok. Last one. Thanks again :).


aleGpereira:
I couldn't get why happen the error that travis is showing. I got that error sometimes in my local enviorenment but if i run it with the docker container, everything goes well. Any clue?

Kami:
Thanks!

Will have a look.

Kami:
It looks like issue is related to missing BeatifulSoup module:

``` python
(libcloud-trunk)kami ~/w/lc/libcloud (git:LIBCLOUD-748_durabledns_provider)$ PYTHONPATH=. python libcloud/test/dns/test_durabledns.py 
  File "libcloud/test/dns/test_durabledns.py", line 28, in <module>
    from libcloud.dns.drivers.durabledns import DurableDNSDriver
  File "/home/kami/w/lc/libcloud/libcloud/dns/drivers/durabledns.py", line 26, in <module>
    from libcloud.common.durabledns import DurableConnection, DurableResponse
  File "/home/kami/w/lc/libcloud/libcloud/common/durabledns.py", line 16, in <module>
    from bs4 import BeautifulSoup
ImportError: No module named bs4
```

Can you please use `ElementTree` module which is available in standard library same as the other drivers which handle xml (e.g. see EC2 driver for an example).

On a related note - I also need to have a look and improve `setup.py test` target, it's really annoying since it doesn't print the actual error message.

Kami:
I've pushed a change so it will be easier to identify issues like this in the future - 011d16c38b22a9c6c64137aef408b3cff9e8d2ae

libcloud/common/durabledns.py
```diff
@@ -0,0 +1,329 @@
...
+from libcloud.utils.misc import lowercase_keys
+from libcloud.common.base import ConnectionUserAndKey
+from libcloud.common.base import XmlResponse
+
+
+# API HOST to connect
+API_HOST = 'durabledns.com'
+
+# REQUEST SCHEMAS
+LIST_ZONES = """
```
Kami: Please use `ElementTree` module to construct XML documents (see EC2, S3 and other XML based drivers for an example).

Creating XML by manually concatenating strings is not safe nor robust.

aleGpereira: Alright. Noted, :+1:

libcloud/common/durabledns.py
```diff
@@ -0,0 +1,294 @@
...
+    soap = ET.Element(
+        'soap:Body',
+        {'xmlns:m': "https://durabledns.com/services/dns/%s" % method}
+    )
+    urn = ET.SubElement(soap, 'urn:%s:%s' % (urn_nid, method))
+    # Attributes specification
+    for attribute in attributes:
+        ET.SubElement(urn, 'urn:%s:%s' % (urn_nid, attribute))
+    return soap
+
```
aleGpereira, line 51: I made this builder for the xml schemas.

libcloud/common/durabledns.py
```diff
@@ -0,0 +1,294 @@
...
+            record_dict = {}
+
+        return (objects, errors)
+
+    def parse_body(self):
+        # A problem arise in the api response because there are undeclared
+        # xml namespaces. In order to fix that at the moment, we use the
+        # _fix_response method to clean up since we won't always have lxml
+        # library.
+        self._fix_response()
```
aleGpereira: I know you mentioned to try to use most of i can parent methods. But this api returns an xml with xml namespaces empty and ElementTree don't like that. I had to make this "_fix_response" method to clean those tags.

aleGpereira:
I did i bunch of changes. I believe is ready for rock and roll! Thanks a lot!

Kami:
Merged, thanks!

PR #590, opened 2015-09-30
ec2 driver support for assigning a public ip

kyle-long:
These changes add ex_assign_public_ip parameter to ec2 compute drivers create_node method.

The biggest thing to note (which is also in the comments) is that since the public ip takes a few seconds to get assigned, the node object returned will not have a public ip set on it.  A user could always use wait_until_running in the base compute driver.


Kami:
Thanks.  LGTM.

As discussed on the IRC - eventually we will want the same behavior in other drivers such as OpenStack and CloudStack. OpenStack + Neutron without nova networking results in a similar default behavior (no public IP by default).

Kami:
Merged, thanks.

PR #591, opened 2015-09-30
[LIBCLOUD-751] Add tests in Point DNS provider for Redirects and Mail…

aleGpereira:
…Redirects classes

### DIFF
--- a/libcloud/dns/drivers/pointdns.py
+++ b/libcloud/dns/drivers/pointdns.py
@@ -17,20 +17,41 @@ Point DNS Driver
 """
 
 __all__ = [
+    'PointDNSException',
+    'Redirect',
+    'MailRedirect',
     'PointDNSDriver'
 ]
+
+import sys
 
 try:
     import simplejson as json
 except ImportError:
     import json
 
+from libcloud.utils.py3 import httplib
 from libcloud.common.types import MalformedResponseError
 from libcloud.common.pointdns import PointDNSConnection
+from libcloud.common.exceptions import BaseHTTPError
 from libcloud.dns.types import Provider, RecordType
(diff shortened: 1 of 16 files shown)


Kami:
I made a change to the `PointDNSException` class so it inherits from `ProviderError` which is intended to be used for that purpose - 08556656a7d69ad3d5c696f14d8d41b95373c5ee. I also updated the constructor signature so it matches the parent one.

For the future reference - this class doesn't really provide additional value over the existing `ProviderError` one (all the information is already available via ProviderError class) so there is little value in adding / having it.

Besides that the changes look good so I've merged it into trunk.

Thanks.

libcloud/dns/drivers/pointdns.py
```diff
@@ -332,9 +371,17 @@ def update_record(self, record, name, type, data, extra=None):
...
-                                           (zone.id, record.id),
-                                           method='PUT', data=r_data)
+        try:
+            response = self.connection.request('/zones/%s/records/%s' %
+                                               (zone.id, record.id),
+                                               method='PUT', data=r_data)
+        except (BaseHTTPError, MalformedResponseError):
+            e = sys.exc_info()[1]
+            if isinstance(e, MalformedResponseError) and e.body == 'Not found':
+                raise RecordDoesNotExistError(value="Record doesn't exists",
```
Kami, line 381: Another "for the future reference" - Ideally those exceptions would be handled on the Response class - see how other classes use `set_context` method.

aleGpereira:
Yeah, you're right. My intention only was to specify the provider. If for some reason someone needs to catch an exception only for pointdns, can use it. But, yeah. Is just a little thing.

aleGpereira:
Thanks for the merges!

PR #594, opened 2015-10-09
Update README.md

freddiev4:
Add links for LICENSE and NOTICE

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -60,4 +60,4 @@ License
 =======
 
 Apache Libcloud is licensed under the Apache 2.0 license. For more information,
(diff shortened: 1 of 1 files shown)


Kami:
Thanks, but we are using restructered text and not Markdown which uses a different link syntax.

You can find an example of the link syntax - http://docutils.sourceforge.net/docs/user/rst/quickref.html#external-hyperlink-targets

freddiev4:
Oops. Sorry. I didn't realize :scream: Should I adjust it so that it fits the syntax, and resubmit a PR?

PR #595, opened 2015-10-09
Adding ex_terminate_on_shutdown to ec2 compute driver

kyle-long:
To support termination of the instance if the OS shutdown command is called


Kami:
Merged, thanks!

PR #598, opened 2015-10-10
Modify CloudStack Connection class so it looks more like other connec…

timwoo112:
…tion

classes and user can more easily specify which attributes to send as part of the
query string and as part of the request body.


Kami:
Thanks.

Can you please have a look - it looks like some tests are failing - https://travis-ci.org/apache/libcloud/builds/14251104.

/cc @runseb

PR #600, opened 2015-10-12
[LIBCLOUD-604] Add 'ex_list_subnets' to OpenStack provider

hannanabdul55:
Added ability to list subnets for libcloud OpenStack driver by fetching them from the OpenStack API and parsing them to a list of `OpenStackSubnet` class.  
This has been tested with a local devstack environment setup on my local machine.
Reference to the OpenStack API [here.](http://developer.openstack.org/api-ref-networking-v2.html#listSubnets)

### DIFF
--- a/libcloud/compute/drivers/openstack.py
+++ b/libcloud/compute/drivers/openstack.py
@@ -335,6 +335,34 @@ class OpenStackNodeDriver(NodeDriver, OpenStackDriverMixin):
         """
         return self._reboot_node(node, reboot_type='HARD')
 
+    def ex_list_subnet(self, network_id=None):
+        """
+        List all subnet resources
+        :param        network_id:   filter for subnets for this network ID
+        :type         network_id:   ``str``
+        :rtype:       ``list`` of :class:``OpenStackSubnet``
+        """
+        uri = '/subnets'
+        resp = self.connection.request(uri, method='GET')
+        if resp.status == httplib.NOT_FOUND:
+            return None
+        return self._to_subnets(resp.object)
+
+
+class OpenStackSubnet(object):
+    """
+    Represents information about an OpenStack subnet
+    NOTE: This class is openStack specific
+    """
+    def __init__(self, id, name, cidr, alloc_pools):
+        self.id = id
+        self.name = name
+        self.cidr = cidr
+        self.alloc_pools = alloc_pools
+
+    def __repr__(self):
+        return (('<OpenStackSubnet: id=%s, name=%s') % (self.id, self.name))
+
 
 class OpenStackNodeSize(NodeSize):
     """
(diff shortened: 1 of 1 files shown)


Kami:
Thanks.

Can you please also add a corresponding test case?

hannanabdul55:
Yeah sure!

hannanabdul55:
So does the Openstack test cases use rackspace driver in the backend? Because in my test case for the above code, I call the `driver.ex_list_subnets` function which throws an error saying this function is undefined in `RackspaceFirstGenNodeDriver` .  
I agree that Rackspace runs on OpenStack, but shouldn't these tests be performed on a running devstack instance?

Kami:
@hannanabdul55 All the tests right right now are unit tests where we use mock provider responses (we don't test against a live environment).

Right now OpenStack tests inherit from Rackspace oned and not vice versa and that's something which we eventually need to clean up.

I'll have a look at what could be causing the issue and let you know (we do some "weird" test inheritance which might be causing the issue).

hannanabdul55:
Yeah sure. I have the unit tests for the same ready anyways. But it breaks the testing, hence, should I withhold the commit for the test cases as of now?

Kami:
@hannanabdul55 Can you gist the tests / diff somewhere? And I will have a look.

hannanabdul55:
I have lost that local repository. I will gist it in a while after recovering it.

libcloud/compute/drivers/openstack.py
```diff
@@ -335,6 +335,34 @@ def ex_hard_reboot_node(self, node):
...
+    def ex_list_subnet(self, network_id=None):
+        """
+        List all subnet resources
+        :param        network_id:   filter for subnets for this network ID
+        :type         network_id:   ``str``
+        :rtype:       ``list`` of :class:``OpenStackSubnet``
+        """
+        uri = '/subnets'
+        resp = self.connection.request(uri, method='GET')
+        if resp.status == httplib.NOT_FOUND:
```
Kami, line 347: I assume this means that this functionality (neutron?) is not available?

In any case, it's probably better to return an empty list.

libcloud/compute/drivers/openstack.py
```diff
@@ -335,6 +335,34 @@ def ex_hard_reboot_node(self, node):
...
+            return None
+        return self._to_subnets(resp.object)
+
+
+class OpenStackSubnet(object):
```
Kami, line 352: To make it easier to eventually switch to #345, can you please make sure this class looks more similar to the one here https://github.com/apache/libcloud/pull/345/files#diff-4c75f73aaa54353918c9fc5a4fc083e5R102 https://github.com/apache/libcloud/pull/345/files#diff-5cc9e28889b2f8491dcdbc2114a213b0R65?

Kami:
Sorry for the delay.

Added some comments.

hannanabdul55:
Yeah sure. I will do this and commit!

PR #601, opened 2015-10-12
Fix method's interface in zonomi, worlwidedns, dnsimple, pointdns and…

aleGpereira:
I thought i did a review for parent compatibility. I guess not :). Also i checked-fixed some other methods in the providers that were breaking parent interface too.


Kami:
Merged, thanks :)

PR #602, opened 2015-10-15
[LIBCLOUD-758] Standardize volumesnapshot state

allardhoeve:


allardhoeve:
anyone? @kami? otherwise I'll merge :smile:

vdloo:
would like this as well

Kami:
Sorry, didn't have time to review it yet.

libcloud/compute/types.py
```diff
@@ -247,6 +247,18 @@ class StorageVolumeState(object):
     UNKNOWN = 8
 
 
+class VolumeSnapshotState(object):
```
Kami, line 250: In newer parts of the code we decided to replace numeric enum values with string values (e.g. RecordType enum).

Strings makes it easier to see what's going on since you don't need to check the enum type definition to see to which enum name a particular value maps to.

I would appreciate if you can update the code to use string values (e.g. AVAILABLE = 'available', ERROR = ' error', etc.).

In the future I also plan to clean and improve the legacy code and update `NodeState` to also use string values.

Kami:
Sorry for the delay.

I've added one comment - besides that the code looks good to me. I would appreciate if you can address that and when you do, feel free to merge the PR.

Thanks!

allardhoeve:
Changed to strings. Will merge.

libcloud/compute/types.py
```diff
@@ -236,15 +236,27 @@ class StorageVolumeState(object):
     """
     Standard states of a StorageVolume
     """
+    AVAILABLE = "available"
```
Kami, line 239: Sorry, for a late comment, I missed this in the previous pass.

For consistency, it would be great if you can change "INUSE" to "IN_USE" and "BACKUP" to "SNAPSHOTING" (this way it's a verb and it's consistent with other states).

Thanks!

allardhoeve, line 239: However, in use is also not a verb ;-)

Kami, line 239: True, but I personally find "backup" state a bit confusing (backuping is also better, imo, but snapshotting is more provider agnostic) :)

PR #603, opened 2015-10-15
LIBCLOUD-759 Updated pricing.json file to include data from SoftLayer

davidwilson2038:
Updated pricing.json file to include data from SoftLayer. I got this pricing data by launching each type of node and looking at its "hourlyRecurringFee" attribute in its extras dictionary.


Kami:
Thanks, but for this actually to work you also need to update the driver to use this information.

See:
1. https://github.com/apache/libcloud/blob/trunk/libcloud/compute/drivers/softlayer.py#L441
2. https://github.com/apache/libcloud/blob/trunk/libcloud/compute/drivers/cloudsigma.py#L251 - an example of how to retrieve the size pricing from the pricing file

davidwilson2038:
@Kami My apologies for missing that. Thanks for the hint on what to fix. My latest commit should have it working successfully (tested with list_sizes).

Kami:
Added a small test case (0ca765136c767b631272e620e27333b8ca100827) and merged it into trunk.

Thanks!

PR #604, opened 2015-10-15
DNSIMPLE: Priority typo fix

phumpal:
`'prio'` is specific to the dnsimple api however libcloud expects `'priority'`


Kami:
Merged, thanks!

aleGpereira:
@Kami @phumpal isn't the extra dictionary suppose to be driver specific?

Kami:
@aleGpereira Correct, although (sadly), right now some "standard" values in the DNS drivers are passed via the extra dictionary - notably "priority" for MX records and ttl.

I hope to eventually refactor / fix that by moving it from the extra dictionary to the method argument.

phumpal:
:+1: thanks @Kami ! I look forward to the refactor / fix :smile:

aleGpereira:
@Kami @phumpal  Alright. What matters is that it works!

PR #607, opened 2015-10-16
GCE: update list of scopes

ottob:
See https://github.com/ansible/ansible-modules-core/issues/1888


erjohnso:
Thanks for this @ottob!

PR #608, opened 2015-10-19
Make clearer travis display & clean tox.ini

ZuluPro:
I've made real usage of travis' language parameter.
It ables to have the language column used in build summary (see https://travis-ci.org/ZuluPro/libcloud/builds/86103898)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,25 +1,31 @@
 language: python
-python: 2.7
+python:
+    - 2.6
+    - 2.7
+    - 3.2
+    - 3.3
+    - 3.4
+    - "pypy"
 sudo: false
 os:
   - linux
   - osx
 env:
-  - TOX_ENV=lint
-  - TOX_ENV=py26
-  - TOX_ENV=py27
-  - TOX_ENV=pypy
-  - TOX_ENV=py32
-  - TOX_ENV=py33
-  - TOX_ENV=py34
-  - TOX_ENV=docs
 
 matrix:
   fast_finish: true
   allow_failures:
     - os: osx
+  include:
+    - python: 2.7
+      env: ENV=lint
+      before_script: TOX_ENV=lint
+    - python: 2.7
+      env: ENV=docs
+      before_script: TOX_ENV=docs
 
 install:
+  - TOX_ENV=py$TRAVIS_PYTHON_VERSION
   - pip install tox
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: python
+# Add or remove version for match with Travis support
 python:
     - 2.6
     - 2.7
--- a/tox.ini
+++ b/tox.ini
@@ -1,52 +1,28 @@
 [tox]
-envlist = py2.5,py2.6,py2.7,pypypy,py3.2,py3.3,py3.4,lint
-setenv =
-    PIP_USE_MIRRORS=1
+envlist = py{2.5,2.6,2.7,pypy,3.2,3.3,3.4,3.5},lint
 
 [testenv]
-deps = -r{toxinidir}/requirements-tests.txt
-       backports.ssl_match_hostname
-       unittest2
-       lockfile
-       paramiko
+deps =
+    -r{toxinidir}/requirements-tests.txt
+    lockfile
+    py{2.5,2.6,2.7,pypy}: paramiko
+    py{2.5,2.6}: unittest2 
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -1,25 +1,31 @@
...
   fast_finish: true
   allow_failures:
     - os: osx
+  include:
+    - python: 2.7
+      env: ENV=lint
+      before_script: TOX_ENV=lint
+    - python: 2.7
+      env: ENV=docs
+      before_script: TOX_ENV=docs
```
ZuluPro, line 26: `env` is only used for display
`before_script` allows me to override value defined in `install` section.

tox.ini
```diff
@@ -22,21 +22,21 @@ deps = -r{toxinidir}/requirements-tests.txt
...
 deps = -r{toxinidir}/requirements-tests.txt
        lockfile
 
-[testenv:py33]
+[testenv:py3.3]
 deps = -r{toxinidir}/requirements-tests.txt
        lockfile
 
-[testenv:py34]
+[testenv:py3.4]
```
ZuluPro: I added `.` for match with `$TRAVIS_PYTHON_VERSION`.

Kami:
Thanks.

I'm fine with that change, but it will break if Travis CI drops support for some Python version we still support, right?

In short - I'm fine with merging it, but we will need to revert this change in case Travis drops support for a Python version so we just need to document that - I guess we can just add a comment to the travis yaml file.

ZuluPro:
> it will break if Travis CI drops support for some Python version we still support, right?

No more things than before. We supports Python 2.5, have it in tox.ini, but can't test with Travis.

> I guess we can just add a comment to the travis yaml file.

I'm doing...

Kami:
@ZuluPro please let me know when you make those changes and I'll go ahead and merge this PR.

Kami:
@ZuluPro Bump^

Would be great to get this addressed and merged.

ZuluPro:
Sorry  I'll do it tomorow !

tox.ini
```diff
@@ -1,52 +1,28 @@
 [tox]
-envlist = py25,py26,py27,pypy,py32,py33,py34,lint
-setenv =
-    PIP_USE_MIRRORS=1
```
ZuluPro, line 4: Not supported anymore: https://github.com/chnm/serendipomatic/issues/158

tox.ini
```diff
@@ -1,52 +1,28 @@
...
+envlist = py{2.5,2.6,2.7,pypy,3.2,3.3,3.4,3.5},lint
 
 [testenv]
-deps = -r{toxinidir}/requirements-tests.txt
-       backports.ssl_match_hostname
-       unittest2
-       lockfile
-       paramiko
```
ZuluPro, line 11: All this deps are defined in `setup.py`

ZuluPro:
Done !
I also factorized tox.ini and test it for ensure great Python version is used.

Kami:
@ZuluPro Thanks. Next time please mention my nickname, since if you are not replying to my comment, I don't get a notification.

As far as tox changes go, I get this error:

``` bash
kami ~/w/lc/libcloud (git:trunk)$ tox
ERROR: unknown environment 'py3.5'
```

Is this functionality only available in newer versions of tox or is there something else going on?

Kami:
Actually, ignore my last comment. I just noticed python3.5 is not defined in `basepython =`map. I will fix this.

Kami:
There is also an issue with pypy

``` bash
ERROR:   pypypy: could not install deps [-r/home/kami/w/lc/libcloud/requirements-tests.txt, lockfile, paramiko]; v = InvocationError('/home/kami/w/lc/libcloud/.tox/pypypy/bin/pip install --pre -r/home/kami/w/lc/libcloud/requirements-tests.txt lockfile paramiko (see /home/kami/w/lc/libcloud/.tox/pypypy/log/pypypy-1.log)', 1)
```

Looks like the issue is related to paramiko (pycrypto) dependency unable to install locally (probably missing some dependency). Will have a look.

Kami:
Updated .travis.yml to also run tests with Python 3.5 and merged those changes into trunk, thanks!

P.S. I still have problems running tests under pypy locally (looks like an issue related to pycrypto on Fedora under pypy), but I went ahead and merged it since it passes on Travis.

Thanks for those changes again - I didn't know tox supports this notation, it looks much better and there is less duplication now.

PR #611, opened 2015-10-20
key_file should be azure_cert.pem

arindamchoudhury:
According to https://azure.microsoft.com/en-us/documentation/articles/cloud-services-python-how-to-use-service-management/ key_file should be '/path/to/azure_cert.pem'
azure_cert.cer raises ssl.SSLError: [SSL] PEM lib (_ssl.c:2580) error.


Kami:
You are correct, good catch - thanks.

Kami:
Merged, thanks.

PR #612, opened 2015-10-23
LIBCLOUD-762 Added pricing data for GCE

davidwilson2038:
Added pricing data for Google Compute Engine. Pricing added as listed for "Full price per hour without sustained use" (https://cloud.google.com/compute/pricing).


Kami:
Thanks.

/cc @erjohnso

erjohnso:
Hi @davidwilson2038, thanks for this!

Although I don't know of any actual differences off the top of my head, GCE's pricing _can_ vary by region.  We also have a pricing calculator that can be used to get the most current price list.

Regarding format of the pricing file, this varies a bit from what is proposed in #464?  I likely need to just get this updated and merged. What do you think about going with that PR vs this one?

davidwilson2038:
@erjohnso I like #464 , as its more complete. It might not work with the test from my commit, as I had to add in the deprecated instance names to work get it work with the test fixtures.

erjohnso:
@davidwilson2038 - ok, let's go with #464 then.  We're stepping through the release process for 0.19.0 right now and I'll get #464 merged in right after that.

Feel free to close this one out and thanks again!

PR #613, opened 2015-10-25
[LIBCLOUD-763] Simplify str2dicts util function

annp89:
Reference: https://issues.apache.org/jira/browse/LIBCLOUD-763

Note about tests:
`test_str2dicts` in (libcloud.test.compute.test_cloudsigma_v1_0.CloudSigmaAPI10IndiretTestCase) **passes**.
But there are other functions in the same test suite that are **failing**. Not sure what's going on but I'll investigate.


tonybaloney:
@annp89 do you still want to keep this open? I'd suggest running a pdb session over it to diagnose the failing test, it suggests it would fail in prod too looking at the error.

PR #614, opened 2015-10-26
[google] Set GCENodeDriver api_name to `google`.

mbrukman:
/cc: @davidwilson2038


Kami:
Thanks.

I believe this will also affect the pricing PR which will need to be updated if this is merged.

/cc @erjohnso

mbrukman:
@Kami: you're absolutely correct; I am in the process of updating the pricing PR to match this rename but wanted to separate them as they are orthogonal changes that should be treated independently.

erjohnso:
lgtm - thanks @mbrukman!

PR #616, opened 2015-10-27
doc: Update documentation for AuroraCompute

wido:
Add an example on how to instantiate a driver and where to find
the API en Secret key.


wido:
The build seems to fail, but I don't think it's related to this PR?

sebgoa:
@wido why do you say its failing, Travis seems to have passed.

wido:
@runseb The build has stabilized, no more issues :)

Kami:
Merged, thanks.

PR #617, opened 2015-10-28
cloudstack: If no volumes are found return an empty list

wido:
Otherwise we got this:

<pre>
  File "auroracompute.py", line 16, in <module>
    print(conn.list_volumes())
  File "/home/wido/repos/libcloud/libcloud/compute/drivers/cloudstack.py", line 2172, in list_volumes
    for vol in volumes['volume']:
KeyError: 'volume'
</pre>

Simply return a empty list if no volume is found

### DIFF
--- a/libcloud/compute/drivers/cloudstack.py
+++ b/libcloud/compute/drivers/cloudstack.py
@@ -2168,8 +2168,9 @@ class CloudStackNodeDriver(CloudStackDriverMixIn, NodeDriver):
                                          method='GET')
 
         list_volumes = []
+
         extra_map = RESOURCE_EXTRA_ATTRIBUTES_MAP['volume']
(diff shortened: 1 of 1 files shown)


wido:
@runseb Can you take a look?

libcloud/compute/drivers/cloudstack.py
```diff
@@ -2168,6 +2168,10 @@ def list_volumes(self, node=None):
                                          method='GET')
 
         list_volumes = []
+
+        if 'volume' not in volumes:
```
Kami: I would prefer to do `for vol in volumes.get('volume', [])`.

wido:
@Kami Yes, that's a nice syntax :) Fixed!

Kami:
Merged, thanks :)

PR #620, opened 2015-10-30
LIBCLOUD-761 network and monitoring support

tonybaloney:
Extended Dimension Data driver for compute to enable: 
- Creating, deleting and renaming "networks" MCP1 only
- Enabling, updating and disabling server monitoring (MCP 1 or 2)

All functions include tests. this is intended for the 0.20 release


libcloud/compute/drivers/dimensiondata.py
```diff
@@ -419,7 +463,84 @@ def ex_list_networks(self, location=None):
...
+        """
+        Delete a network from an MCP 1 data center
+
+        :param  network: The network to delete
+        :type   network: :class:`DimensionDataNetwork`
+
+        :rtype: ``bool``
+        """
+        response = self.connection.request_with_orgId_api_1(
+            'network/%s?delete' % network.id,
```
Kami, line 511: For query parameters, please use `params=` argument which takes care of correctly encoding query params for you.

libcloud/compute/drivers/dimensiondata.py
```diff
@@ -419,7 +463,84 @@ def ex_list_networks(self, location=None):
...
+
+        :param  new_name: The new name of the network
+        :type   new_name: ``str``
+
+        :rtype: ``bool``
+        """
+        response = self.connection.request_with_orgId_api_1(
+            'network/%s' % network.id,
+            method='POST',
+            data='name=%s' % new_name).object
```
Kami, line 531: Hm so payload in this request is neither xml or json, but just form encoded data?

Kami:
Thanks.

Besides the params thing, looks good to me so far.

As far as the networking goes, we should also look into reviving this - https://github.com/apache/libcloud/pull/345

libcloud/compute/drivers/dimensiondata.py
tonybaloney, line 531: Yes. Inconsistent with the other APIs I know but they had to maintain back compat

libcloud/compute/drivers/dimensiondata.py
tonybaloney, line 511: You can still do that when the param doesn't have a matching value? I'll try that out

Kami, line 511: Ah, that's a good point - I'm not actually sure you can do that.

tonybaloney:
+1 on #345

tonybaloney:
all good to merge+close?

Kami:
LGTM, +1 - feel free to merge it.

tonybaloney:
merged.

PR #621, opened 2015-10-30
Update worldwidedns provider with tests.

aleGpereira:
This update make worldwidedns driver to choose automatically a slot to store a record when is not specified.


Kami:
Merged, thanks.

PR #622, opened 2015-10-31
Modified Google Storage driver to use GoogleBaseConnection. This allo…

crunk1:
…ws for service accounts and installed apps to access Google Cloud Storage. Also did some cleanup: removed a few **init** methods that only pass parameters to their superclass, refactored credential_file logic with Google connections.

GoogleStorageDriver still uses the S3 API and driver, but the underlying connection and authentication has been swapped out with GoogleBaseConnection.

### DIFF
--- a/libcloud/common/base.py
+++ b/libcloud/common/base.py
@@ -520,9 +520,7 @@ class Connection(object):
             (self.host, self.port, self.secure,
              self.request_path) = self._tuple_from_url(url)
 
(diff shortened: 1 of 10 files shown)


crunk1:
Do not merge. This authentication works for Google Service Accounts and Installed Apps, but threw out the S3 XML API authentication. I need to revise this to make it work with both.

PR #623, opened 2015-11-02
auroracompute: Add support for multiple regions

wido:


sebgoa:
Might want to add some tests for this one. Trying to instantiate a driver in the different regions.
Also I am struggling to understand how the endpoint in the different regions gets selected. Your mapping only shows the paths.

wido:
@runseb Indeed, tests are a fair point, I'll do that.

We are not using the CloudStack regions as each region runs it's owner manager with accounts. This is due to legal obligations where we don't want to store information in different "legal zones".

api.auroracompute.eu is a HTTP proxy which proxies requests towards the actual CloudStack Management server.

sebgoa:
Ok so the paths proxies to the correct endpoint.
When you add a few tests I will merge it.

wido:
@runseb I just pushed with tests to verify the endpoints. My code seems to work :)

sebgoa:
looks good, merging

libcloud/compute/drivers/auroracompute.py
```diff
@@ -17,13 +17,41 @@
 from libcloud.compute.drivers.cloudstack import CloudStackNodeDriver
 
 __all__ = [
+    'AuroraComputeRegion',
     'AuroraComputeNodeDriver'
 ]
 
 
+class AuroraComputeRegion(object):
+    AMS = 'Amsterdam'
```
Kami, line 26: Minor thing - for consistency with other drivers it might be better to just use short region names (ams, rtd, mia, etc.)

That's what other drivers which support multiple regions do right now :)

PR #624, opened 2015-11-03
Turn all States into string values instead of ints

allardhoeve:
@Kami

Luckily, most tests already used the constants. I found a couple sloppy drivers, especially the ProfitBricks driver, which uses NodeState for everything, including Volumes and Snapshots. Maybe we fix that, maybe we leave it. It doesn't really matter.


Kami:
Thanks.

We should just wait with merging this until we decide when to include it in a release since it's a breaking change if people used integers directly.

I'm fine with including it in 0.20.0 (next "major" release), but we also need to update documentation (upgrade notes) to reflect this change.

libcloud/compute/types.py
```diff
@@ -207,25 +207,19 @@ class NodeState(object):
...
-    REBOOTING = 1
-    TERMINATED = 2
-    PENDING = 3
-    UNKNOWN = 4
-    STOPPED = 5
-    SUSPENDED = 6
-    ERROR = 7
-    PAUSED = 8
+    RUNNING = "running"
+    REBOOTING = "rebooting"
```
Kami: Minor style thing - for consistency, please use single quotes around simple strings :)

libcloud/compute/types.py
```diff
@@ -207,25 +207,19 @@ class NodeState(object):
...
-        values = dict([(key, string) for key, string in values.items() if
-                       not key.startswith('__')])
-
-        for item_key, item_value in values.items():
-            if value == item_value:
-                return item_key
+        return value
 
     @classmethod
     def fromstring(cls, value):
```
Kami: Is this still needed?

Kami: Would also be good to add some small test case for `tostring` and `fromstring`.

allardhoeve:
Please do not close yet. Need a little more time to get to it.

libcloud/compute/types.py
```diff
@@ -207,25 +207,19 @@ class NodeState(object):
...
-    RUNNING = 0
-    REBOOTING = 1
-    TERMINATED = 2
-    PENDING = 3
-    UNKNOWN = 4
-    STOPPED = 5
-    SUSPENDED = 6
-    ERROR = 7
-    PAUSED = 8
+    RUNNING = "running"
```
jimbobhickville: These were `ALL_CAPS` before, I don't think we should change the case on them.

libcloud/compute/types.py
jimbobhickville: Yes, please leave this in.  We have a lot of code that uses these method to convert to/from the libcloud expected values, and removing this will break us.

allardhoeve: I will create tests for this

allardhoeve: Mmmm, there are no current tests for this :cry:

allardhoeve:
So I added tests for all the objects that have strings as values now. All tests pass.

@kami: I'd like for this to go into 0.20. Do you think that is still possible?

allardhoeve:
Also, I think the changes involved are minimal. I cannot imagine a use case where you would actually use the integers in normal code. I think everyone involved either used `fromstring` for `NodeState` and `Provider`, or raw strings, because the other objects did not have consistent states anyway.

libcloud/compute/types.py
```diff
@@ -204,56 +214,42 @@ class NodeState(object):
...
-            if value == item_value:
-                return item_key
-
-    @classmethod
-    def fromstring(cls, value):
-        return getattr(cls, value.upper(), None)
-
-
-class StorageVolumeState(object):
+    RUNNING = 'running'
```
jimbobhickville, line 227: Are we ok with this behavior change?  The `NodeState.tostring(node.state)` values previously were all-caps, now they're all lower-case.  I know this means my team's codebase will require changes to keep existing behavior.  Could we  make `Type.tostring` return `value.upper()`?

allardhoeve, line 227: Ah! That is true. Let me add that test. We should've had tests for this stuff all along :cry: That would've made this much easier. Ah well, moving forward :smile:

allardhoeve, line 227: Done

jimbobhickville:
Thanks @allardhoeve - lgtm

allardhoeve:
I added docs for 0.20

docs/upgrade_notes.rst
```diff
@@ -16,6 +16,17 @@ Libcloud 0.20.0
   to the constructor (you are encouraged to use keyword arguments to avoid such
   issues in the future).
 
+* All NodeState, StorageVolumeState, VolumeSnapshotState and Provider attributes
+  are now strings instead of integers.
+
+  If you are using the ``tostring`` and ``fromstring`` methods of NodeState,
```
Kami, line 22: It would also be good to note that you are affected if you directly store the enum value in the database (integer) - I know some users who do that.

For that scenario this is a backward incompatible change. We don't have any automatic migration path for it - users need to handle that in their code - so for users who do that it's quite an invasive change.

libcloud/compute/types.py
```diff
@@ -35,7 +35,27 @@
...
-class Provider(object):
+class Type(object):
+    @classmethod
+    def tostring(cls, value):
+        """Return the string representation of the state object attribute
+        :param str value: the state object to turn into string
+        :return: the uppercase string that represents the state object
+        :rtype: str
+        """
+        return value.upper()
```
Kami, line 46: I think I'm fine with this change, but since it's a class method it might also also be a good idea to throw if users passes in an invalid enum value.

In any case, current approach works for now.

PR #625, opened 2015-11-04
Added .editorconfig file for easier editing.

mbrukman:
Editor Config is a cross-editor configuration which allows automatic per-project
configuration to conform with local style guidelines without having to add
special cases to one's own editor: it's picked up from the project's source tree
automatically.

See http://editorconfig.org for more info and how to configure various editors
to interpret this config.


sebgoa:
I don't use it but seems quite nice, so if it can help you I am merging it.

PR #627, opened 2015-11-07
[LIBCLOUD-767] Fixed CloudStackAddress parameters

atsaki:
This fixes the parameters of CloudStackAddress of node.extra['ip_addresses'].

https://issues.apache.org/jira/browse/LIBCLOUD-767

### DIFF
--- a/libcloud/compute/drivers/cloudstack.py
+++ b/libcloud/compute/drivers/cloudstack.py
@@ -1454,8 +1454,8 @@ class CloudStackNodeDriver(CloudStackDriverMixIn, NodeDriver):
 
         node = self._to_node(data=vm, public_ips=list(public_ips.keys()))
 
-        addresses = public_ips.items()
-        addresses = [CloudStackAddress(node, v, k) for k, v in addresses]
+        addresses = [CloudStackAddress(address_id, address, node.driver)
+                     for address, address_id in public_ips.items()]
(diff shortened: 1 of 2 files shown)


Kami:
Thanks.

LGTM

Kami:
Made a small change to use kwargs to make it more explicit (59ce9031b854bc29926596f627bd0a6e271a1dd5) and merged patch into trunk. Thanks!

On a related note - looks like a similar bug exists here - https://github.com/apache/libcloud/blob/71d4e5718ce5abd8f725084e3858a7075617a008/libcloud/compute/drivers/cloudstack.py#L1384, but I wasn't able to confirm it since we don't have any test fixtures which exercise that code path (ips is always an empty list) and I don't have easy access to CloudStack installation right now.

Can you please have a look?

atsaki:
Thanks @Kami .

> On a related note - looks like a similar bug exists here - https://github.com/apache/libcloud/blob/71d4e5718ce5abd8f725084e3858a7075617a008/libcloud/compute/drivers/cloudstack.py#L1384, but I wasn't able to confirm it since we don't have any test fixtures which exercise that code path (ips is always an empty list) and I don't have easy access to CloudStack installation right now.

This will be addressed in https://github.com/apache/libcloud/pull/332.

PR #629, opened 2015-11-08
LIBCLOUD-769 backup driver

tonybaloney:
https://issues.apache.org/jira/browse/LIBCLOUD-769


tonybaloney:
@Kami @erjohnso 

This is the proposed driver type, please provide initial feedback whilst I prototype implementations for 3 providers.

tonybaloney:
Proposed drivers are:
- Azure VM Backup
- Amazon EBS Snapshots
- Dimension Data Cloud Backup
- Rackspace Cloud File Backup

Azure VM Backup concept and architecture - https://azure.microsoft.com/en-us/documentation/articles/backup-azure-vms-automation/
Amazon EBS snaps -  http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-creating-snapshot.html

tonybaloney:
Also compliant to OpenStack "Freezer" proposed API https://github.com/openstack/freezer-api

tonybaloney:
Setup branched docs for review here:
http://libcloud-fork.readthedocs.org/en/libcloud-769_backup_driver/index.html

erjohnso:
I don't know other providers very well, but it sure looks like a good fit for Azure and Freezer and likely a few others.

I'm not sure I see such a fit though for AWS and Google since they don't have a "backup-as-a-service" API.  How do you see this working for those two?

tonybaloney:
Amazons solution has always been either using EBS snaps or Glacier and some scripts. This model would fit EBS snaps well, where the Target is the Volume. I stashed a quick prototype using the existing ec2 compute driver functions for snaps and it works well. 
Google I need to do research on.

tonybaloney:
@erjohnso committed an EBS driver for doing snaps (a bare bones version of one)

tonybaloney:
For google compute, there would be 2, for GCE- https://cloud.google.com/compute/docs/disks/persistent-disks#snapshots and for Google SQL there is a backup and restore function matching the same pattern.

tonybaloney:
Reviewed Rackspace API, matches to pattern suitable for a driver (need to investigate whether this service is still active) https://developer.rackspace.com/docs/cloud-backup/v1/developer-guide/#document-developer-guide

Kami:
Great, thanks.

I will hopefully have some time to look into it in more details over the weekend :)

As far as Rackspace Cloud Backup goes - I'm pretty sure that service is still active since when you launch a server they still offer an option to turn on backups and install the backups agent.

tonybaloney:
@Kami I'll put this to vote once you've had the chance to look at it. Also I'll remove the demo drivers before a merge

Kami:
@tonybaloney Thanks and I'll try to do it this week during travel.

As far as demo drivers go - initial implementation should still include at least two (ideally three) drivers when merging it into trunk. Or do you plan to do a separate PR with initial drivers?

tonybaloney:
Separate PRs, I'll stash them and push them into new PRs. I picked those 2
because they were the trickiest implementations since they didnt really fit
the mould.

I'd like to get other committers to contribute drivers for the other
providers before release. using a solid example

docs/examples/backup/create_backup_from_node.py
```diff
@@ -0,0 +1,37 @@
+import time
+from pprint import pprint
+
+from libcloud.backup.types import Provider, BackupTargetJobStatusType
+from libcloud.backup.providers import get_driver as get_backup_driver
+
+from libcloud.compute.providers import get_driver as get_compute_driver
+
+backup_driver = get_backup_driver(Provider.AZURE)('username', 'api key')
```
Kami: To be on the safe side it's better to use `libcloud.compute.types.Provider` for compute driver since constants could have a different value.

So something like this:

``` python
from libcloud.compute.types import Provider as ComputeProvider
from libcloud.backup.types import Provider as BackupProvider
```

docs/examples/backup/create_backup_from_node.py
```diff
@@ -0,0 +1,37 @@
...
+while True:
+    if job.status != BackupTargetJobStatusType.RUNNING:
+        break
+    else:
+        job = backup_driver.get_target_job(job.id)
+
+    print('Job is now at %s percent complete' % job.progress)
+    time.sleep(20)
+
+print('Job is completed with status- %s' % job.status)
```
Kami: Minor style thing - to avoid potential issues when adding more format string arguments, please always wrap them with parenthesis:

``` python
print('Job is completed with status- %s' % (job.status))
```

libcloud/backup/base.py
```diff
@@ -0,0 +1,488 @@
...
+
+    def __init__(self, id, name, address, type, driver, extra=None):
+        """
+        :param id: Target id
+        :type id: ``str``
+
+        :param name: Name of the target
+        :type name: ``str``
+
+        :param address: Hostname, FQDN, IP, file path etc.
```
Kami, line 40: Hm, trying to think of a better name for this argument. Maybe `value` or `target` (I see how `target` could be confusing though).

libcloud/backup/base.py
```diff
@@ -0,0 +1,488 @@
...
+    def update(self, name=None, address=None, extra=None):
+        return self.driver.update_target(target=self,
+                                         name=name,
+                                         address=address,
+                                         extra=extra)
+
+    def delete(self):
+        return self.driver.delete_target(target=self)
+
+    def _get_numeric_id(self):
```
Kami, line 68: Just curious - where is that used?

libcloud/backup/base.py
```diff
@@ -0,0 +1,488 @@
...
+        :type id: ``str``
+
+        :param status: Status of the job
+        :type status: :class:`BackupTargetJobStatusType`
+
+        :param progress: Progress of the job, as a percentage
+        :type progress: ``int``
+
+        :param target: BackupTarget instance.
+        :type target: :class:`BackupTarget`
```
Kami: Minor thing - since you are referring to a class in the same module you can use relative notation, so:

``` python
:class:`.BackupTarget`
```

Same goes for other classes in this module.

This way sphinx will know it refers to the class in the same module and will automatically link to it in the API docs.

libcloud/backup/base.py
```diff
@@ -0,0 +1,488 @@
...
+        :param type: Backup target type (Physical, Virtual, ...).
+        :type type: :class:`BackupTargetType`
+
+        :param extra: (optional) Extra attributes (driver specific).
+        :type extra: ``dict``
+
+        :rtype: Instance of :class:`BackupTarget`
+        """
+        return self.create_target(name=node.name,
+                                  address=node.public_ips[0],
```
Kami, line 289: Would be good to document that it uses first public IP by default.

It might be good to allow user to specify which public or private IP to use, but then again, user can already do this using a more generic, non-node specific method.

libcloud/backup/base.py
```diff
@@ -0,0 +1,488 @@
...
+                                  type=type,
+                                  extra=None)
+
+    def create_target_from_container(self, container,
```
Kami: I think it would be a good idea to merge this PR without this method first - this way we can merge it first, before we merge new container API.

libcloud/backup/drivers/gce.py
```diff
@@ -0,0 +1,478 @@
...
+
+    def pre_connect_hook(self, params, headers):
+        """
+        Update URL parameters with values from self.gce_params.
+
+        @inherits: :class:`GoogleBaseConnection.pre_connect_hook`
+        """
+        params, headers = super(GCEConnection, self).pre_connect_hook(params,
+                                                                      headers)
+        if self.gce_params:
```
Kami, line 81: I'm OK with this for now, but when possible it's better to avoid mutating class level variables.

Mutating class level variables could cause issues when using something like eventlet on a single driver instance, but then again, our documentation already says that if you want to use something like eventlet / gevent you should not share a driver instance and use one driver instance per green thread.

libcloud/backup/drivers/gce.py
```diff
@@ -0,0 +1,478 @@
...
+        """
+        if not project:
+            raise ValueError('Project name must be specified using '
+                             '"project" keyword.')
+
+        self.auth_type = auth_type
+        self.project = project
+        self.scopes = scopes
+        self.credential_file = credential_file or \
+            '~/.gce_libcloud_auth' + '.' + self.project
```
Kami, line 150: I think compute driver does something similar to this - if that's indeed the case, it would be a good idea to split this in a base `GoogleDriver` class or similar.

Kami:
I haven't tested the changes yet, but so far LGTM :+1:

libcloud/backup/base.py
tonybaloney, line 40: target_address makes more sense

libcloud/backup/base.py
tonybaloney: That's for storage containers @Kami but since I wrote it that has now become not obvious!

libcloud/backup/types.py
```diff
@@ -0,0 +1,63 @@
...
+    GCE = 'gce'
+    DIMENSIONDATA = 'dimensiondata'
+
+
+class BackupTargetType(object):
+    """
+    Backup Target type.
+    """
+
+    VIRTUAL = 'Virtual'
```
Kami, line 35: Minor thing I just noticed - for consistency with other enums, please make sure the enum values are all lowercase (`VIRTUAL = 'virtual`, etc.).

libcloud/backup/types.py
```diff
@@ -0,0 +1,63 @@
...
+    VOLUME = 'Volume'
+    """ Denotes a block storage volume """
+
+
+class BackupTargetJobStatusType(object):
+    """
+    The status of a backup target job
+    """
+
+    RUNNING = 'Running'
```
Kami, line 59: Same here - please use all lower case for enum values.

PR #630, opened 2015-11-09
[google] fix pricing script and test; fix build

mbrukman:
Revert conditional import of simplejson to unconditional: the code as
implemented only works with simplejson and not the standard json library because
it relies on the `item_sort_key` functionality which only simplejson provides
and json does not.

Also fixes indentation which breaks the build since the linter is part of the
tests.

Fixes the issues seen here:
- https://travis-ci.org/apache/libcloud/builds/90117628
- https://travis-ci.org/apache/libcloud/jobs/90117643

/cc: @erjohnso to review and merge.

### DIFF
--- a/contrib/update_google_prices.py
+++ b/contrib/update_google_prices.py
@@ -19,10 +19,8 @@ Loads Google Cloud Platform prices and updates the `pricing.json` data file.
 """
 
 import os
-try:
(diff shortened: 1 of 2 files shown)


erjohnso:
Thanks @mbrukman, sorry about the goof.

PR #631, opened 2015-11-12
LIBCLOUD 770 wait for state

tonybaloney:
Support for waiting for a request to reach a target state (i.e. provisioned)


libcloud/common/dimensiondata.py
```diff
@@ -217,6 +218,28 @@ def get_resource_path_api_2(self):
...
+        :type   func: ``function``
+
+        :param  kwargs: The arguments for func
+        :type   kwargs: Keyword arguments
+        """
+        while(True):
+            response = func(*args, **kwargs)
+            if response.status is state or response.status in state:
+                break
+            sleep(2)
```
Kami: It might be a good idea to make the sleep interval configurable with a function argument.

libcloud/common/dimensiondata.py
```diff
@@ -217,6 +218,28 @@ def get_resource_path_api_2(self):
...
+        :param  func: The function to call, e.g. ex_get_vlan
+        :type   func: ``function``
+
+        :param  kwargs: The arguments for func
+        :type   kwargs: Keyword arguments
+        """
+        while(True):
+            response = func(*args, **kwargs)
+            if response.status is state or response.status in state:
+                break
```
Kami: Probably a good idea to return original function result (node / instance) when state transition happens?

libcloud/common/dimensiondata.py
```diff
@@ -217,6 +218,28 @@ def get_resource_path_api_2(self):
...
+        :param state: Either the desired state (`str`) or a `list` of states
+        :type  state: ``str`` or ``list``
+
+        :param  func: The function to call, e.g. ex_get_vlan
+        :type   func: ``function``
+
+        :param  kwargs: The arguments for func
+        :type   kwargs: Keyword arguments
+        """
+        while(True):
```
Kami: Minor style thing - no parenthesis needed in Python :)

libcloud/compute/drivers/dimensiondata.py
```diff
@@ -815,6 +816,24 @@ def ex_get_location_by_id(self, id):
...
+        :param state: Either the desired state (`str`) or a `list` of states
+        :type  state: ``str`` or ``list``
+
+        :param  func: The function to call, e.g. ex_get_vlan
+        :type   func: ``function``
+
+        :param  kwargs: The arguments for func
+        :type   kwargs: Keyword arguments
+        """
+        self.connection.wait_for_state(state, func, *args, **kwargs)
```
Kami: Same here - if you update the original function to return the result you can also update this one.

libcloud/common/dimensiondata.py
Kami: Might also be a good idea to add some kind of timeout functionality?

Since now it could potentially block forever...

libcloud/common/dimensiondata.py
Kami: if we make this a bit more generic and safe (timeout, etc.) we could also consider it adding to the base NodeDriver class.

tonybaloney:
Updated code to reflect comments. Agree on adding it to nodedriver, I've seen similar helper methods in libraries consuming libcloud, would need to research whether all of the drivers use the same field name (status).

tonybaloney:
all done

Kami:
I have trouble applying the patch directly onto trunk.

Can you please sync this branch with latest trunk and squash the commits?

tonybaloney:
https://github.com/DimensionDataCBUSydney/libcloud/commit/609fa8541188fb3daae7653e253c77c0dd422c2f
squashed. I resynced but there weren't any differences with trunk. see if this is mergable

Kami:
I'm still having issues, I will manually apply it tomorrow by directly merging your branch locally :)

```
kami ~/w/lc/libcloud (git:trunk)$ git am --signoff 609fa8541188fb3daae7653e253c77c0dd422c2f.patch
Applying: Provide a way for users to wait for an asset to complete (+4 squashed commit)
error: patch failed: libcloud/common/dimensiondata.py:218
error: libcloud/common/dimensiondata.py: patch does not apply
error: patch failed: libcloud/compute/drivers/dimensiondata.py:815
error: libcloud/compute/drivers/dimensiondata.py: patch does not apply
error: patch failed: libcloud/loadbalancer/drivers/dimensiondata.py:823
error: libcloud/loadbalancer/drivers/dimensiondata.py: patch does not apply
Patch failed at 0001 Provide a way for users to wait for an asset to complete (+4 squashed commit)
The copy of the patch that failed is found in:
   /home/kami/w/lc/libcloud/.git/rebase-apply/patch
When you have resolved this problem, run "git am --continue".
If you prefer to skip this patch, run "git am --skip" instead.
To restore the original branch and stop patching, run "git am --abort".
```

Kami:
Alright, I merged the changes. Thanks.

Here is what I did:

``` bash
git fetch dd
git checkout LIBCLOUD-770_Wait_for_state
git rebase trunk  -> that's the important part since it replayed your changes on top of trunk so the changes were now in the mergable order
git format-patch trunk --stdout > 1.patch
git checkout trunk
git am --signoff 1.patch
```

I also made a small change - I supplied default value for poll_interval and timeout argument and changed exception message to also include response body - 4a742436775f83d73c6f0b2a762c6fe2f777c6f1. Let me know if the response.body change looks OK.

tonybaloney:
would response.body have any value?

here's an example of me using it>

so ex_get_vlan returns a DimensionDataVlan object instance, which doensn't
have a body attribute.

driver.ex_wait_for_state('NORMAL', driver.ex_get_vlan, result['id'])

Kami:
@tonybaloney Ah, you are correct.

I was confused for a second since the method is on a connection class, but it actually returns an object not the raw response from the connection class (it returns result from a calling method).

I will revert that change and change variable name from `response` to `result` to make it a bit less confusing (response in connection class usually refers to an instance of Libcloud HTTP response class).

Kami:
Alright, I've pushed a fix in - 232f9491f6ad62f8c6db8239f1291575c60b0f81.

Another minor feature improvement would be to change the method signature to `(state, poll_interval, timeout, func, *args, **kargs)`. This way you can partially apply the function, e.g.:

``` python
wait_for_vlan = functools.partial(self.wait_for_state, state=foo, func=self.ex_get_vlan)
vlan = wait_for_vlan(...)
```

Or you could even modify the method to return an already partially applied function which user can use directly.

tonybaloney:
my python knowledge doesn't extend that far. I was only just up to
functools.wrap
maybe soon I'll know what you're talking about!

pquentin:
If possible, it would be nice to disable coveralls who keeps posting random comments in old pull requests.

PR #635, opened 2015-11-18
Allow location of the data center to be supplied

joelreymont:


tonybaloney:
ok, great. I'll merge + close

joelreymont:
Do I need to open issues in the future, e.g.
https://issues.apache.org/jira/browse/LIBCLOUD-771 ?

tonybaloney:
Yes, keep opening the tickets, they should get closed automatically, it's good to keep a record.
https://libcloud.readthedocs.org/en/latest/development.html#contribution-workflow 
Also, you'll see the notes in there about maintaining branches in your own fork for each ticket, which is also good practice and comes in handy if a PR isn't closed and you want to open another one.

PR #637, opened 2015-11-21
CloudFlare DNS driver

Kami:
New driver for CloudFlare DNS (WIP).

TODO:
- [x] Support for record management (create, edit, delete)
- [x] Tests
- [ ] Pagination support


Kami:
Alright, added the tests and fixed some things.

I will go ahead and merge it and implement pagination in a future PR (driver is already functional without pagination as long as you have < page size number of zones and records).

libcloud/test/dns/fixtures/cloudflare/nul.json
```diff
@@ -0,0 +1,10 @@
+{
```
tonybaloney, line 1: @Kami nul is a reserved file name in Windows, so anyone trying to test or use this on Windows is going to have a lot a bad time.. Please can you rename it

Kami, line 1: Even if it has an extension (aka `nul.json`)?

Kami, line 1: It looks like it, that's unfortunate :P

Good catch - will rename it.

Kami, line 1: Would make it easier if we could also automatically run tests on Windows, but Travis sadly doesn't support that (yet).

dol:
@Kami Thank you for the driver. Any reason the driver uses a old version of the Cloudflare API -> https://www.cloudflare.com/docs/client-api.html .
CloudFlare is removing support for this API on November 9th, 2016.

Kami:
@dol I think that when I added it, new version of the API wasn't there yet (or I missed it).

dol:
Ok. Sorry, the question was unfortunate.
Any plans switching over to the new API version?

Kami:
I don't have any plans, but contributions are welcome (you can also ping / ask CloudFlare people).

dol:
I'm not that frequent in Python. Might be better to call your friends.

PR #634, opened 2015-11-16
Adding m4 Instance types

quasiben:

PR #638, opened 2015-11-22
C4 instance type support

amitlalani:


tonybaloney:
looks good, will merge

tonybaloney:
@amitofs any further commits?

tonybaloney:
Pricing also needs to be updated as per https://github.com/apache/libcloud/pull/634

tonybaloney:
nw, @kami already did it on the last sync, the prices are already there.

Kami:
@tonybaloney Just FYI, looks like this (29e33873d9e072f75f407cbf32f0f09c5cf3b875) slipped through since the patch was merged before @amitofs updated the PR.

In any case, it's fixed now :)

PR #640, opened 2015-11-24
LIBCLOUD-772 driver class for GoDaddy DNS

tonybaloney:
- [x] Implement driver methods
- [x] Document usage and examples
- [x] Add tests and responses based on developer.godaddy.com


Kami:
As far as test fixtures go - I would recommend to use actual responses you get back when talking to the API.

Maybe times before we encountered an issue with out of date or incorrect docs (I just encountered that issue with CloudFlare DNS driver a couple of days ago - some of the responses in the docs were incorrect and out of date). Sadly it looks like most of the provider documentation is still written manually and not automatically generated from code.

tonybaloney:
GoDaddy records don't have IDs, and when you update or query them you combine the name and type, so I turned that into the ID. Just for get_record, for lack of an alternative.

tonybaloney:
@Kami your thoughts please. this is complete now

Kami:
> GoDaddy records don't have IDs, and when you update or query them you combine the name and type, so I turned that into the ID. Just for get_record, for lack of an alternative.

Yeah, I believe GCE driver does something like that as well.

tonybaloney:
@Kami as per your suggestion i used real API responses for the tests as well. except the "purchase domain" one, that would get expensive

libcloud/dns/drivers/godaddy.py
```diff
@@ -0,0 +1,521 @@
...
+
+class GoDaddyDNSResponse(JsonResponse):
+    valid_response_codes = [httplib.OK, httplib.ACCEPTED, httplib.CREATED,
+                            httplib.NO_CONTENT]
+
+    def parse_body(self):
+        if not self.body:
+            return None
+        # json.loads doesn't like the regex expressions used in godaddy schema
+        self.body = self.body.replace('\\.', '\\\\.')
```
Kami, line 55: Hm, ok :P

tonybaloney, line 55: https://www.google.com.au/webhp?sourceid=chrome-instant&ion=1&espv=2&ie=UTF-8#q=json%20loads%20invalid%20escape yeah I'm not the only one. the other option is to introduce cjson, but adding another prereq to the project for just one regex statement seems excessive

libcloud/dns/drivers/godaddy.py
```diff
@@ -0,0 +1,521 @@
...
+
+        :param  secret: Your access key secret
+        :type   secret: `str`
+        """
+        super(GoDaddyDNSDriver, self).__init__(key=key, secret=secret,
+                                               secure=secure,
+                                               host=host, port=port,
+                                               shopper_id=str(shopper_id))
+
+    def _to_zones(self, items):
```
Kami: Minor code style thing - please put "'private" methods to the bottom - https://libcloud.readthedocs.org/en/latest/development.html#function-and-method-ordering

libcloud/dns/drivers/godaddy.py
```diff
@@ -0,0 +1,521 @@
...
+
+        for item in items:
+            records.append(self._to_record(item=item, zone=zone))
+        return records
+
+    def _to_record(self, item, zone=None):
+        ttl = item['ttl']
+        type = self._string_to_record_type(item['type'])
+        name = item['name']
+        id = '{0}:{1}'.format(name, type)
```
Kami: Please use '%s' for string interpolation. We still support Python 2.5 and 2.6 (we will probably drop support for 2.5 in the near future) where `.format()` is not available.

libcloud/dns/drivers/godaddy.py
```diff
@@ -0,0 +1,521 @@
...
+        record = Record(id=id, name=name,
+                        type=type, data=item['data'],
+                        zone=zone, driver=self,
+                        extra={'ttl': ttl})
```
Kami: I recently merged PR which adds `ttl` attribute to the `Record` class so you can use this now - https://github.com/apache/libcloud/pull/639

libcloud/dns/drivers/godaddy.py
```diff
@@ -0,0 +1,521 @@
...
+        """
+        Return a list of records for the provided zone.
+
+        :param zone: Zone to list records for.
+        :type zone: :class:`Zone`
+
+        :return: ``list`` of :class:`Record`
+        """
+        result = self.connection.request(
+            '/v1/domains/%s/records' % zone.domain).object
```
Kami: Minor code style / safety thing - always when using string interpolation wrap arguments in parenthesis - it prevents accidental errors which could arise if user adds another argument, but forgets to add parenthesis (we had those issues in production before, ideally tests should discover that, but still).

libcloud/dns/drivers/godaddy.py
```diff
@@ -0,0 +1,521 @@
...
+                'name': name,
+                'data': data,
+                'priority': 1,
+                'ttl': extra['ttl'] if hasattr(extra, 'ttl') else 5
+            }
+        self.connection.request(
+            '/v1/domains/{0}/records'.format(
+                zone.domain), method='PATCH',
+            data=[new_record])
+        id = '{0}:{1}'.format(name, type)
```
Kami: Same here - please use `%`.

Kami: Maybe utility method would be handy - `_get_id_for_record` or similar.

libcloud/dns/drivers/godaddy.py
```diff
@@ -0,0 +1,521 @@
...
+
+        :param extra: (optional) Extra attributes (driver specific).
+        :type  extra: ``dict``
+
+        :rtype: :class:`Record`
+        """
+        if extra is None:
+            extra = {}
+        new_record = {}
+        if type is RecordType.SRV:
```
Kami: It seems there is a lot of duplication code between this method and create record - would be good to refactor common stuff in a utility method.

libcloud/dns/drivers/godaddy.py
```diff
@@ -0,0 +1,521 @@
...
+            new_record = {
+                'type': type,
+                'name': name,
+                'data': data,
+                'priority': 1,
+                'ttl': extra['ttl'] if hasattr(extra, 'ttl') else 5,
```
Kami: I just noticed you use `hasattr`. This won't work since `hasattr` only works on class instances. For dictionaries you should either use `key in dict` or `dict.get(key, default_value)`.

libcloud/dns/drivers/godaddy.py
```diff
@@ -0,0 +1,521 @@
...
+                'data': data,
+                'priority': 1,
+                'ttl': extra['ttl'] if hasattr(extra, 'ttl') else 5
+            }
+        self.connection.request(
+            '/v1/domains/{0}/records'.format(
+                zone.domain), method='PATCH',
+            data=[new_record])
+        id = '{0}:{1}'.format(name, type)
+        return Record(
```
Kami, line 193: Same here, you can now also pass `ttl` to the `Record` class.

libcloud/dns/drivers/godaddy.py
```diff
@@ -0,0 +1,521 @@
...
+
+        :rtype: :class:`Record`
+        """
+        parts = record_id.split(':')
+        result = self.connection.request(
+            '/v1/domains/{0}/records/{1}/{2}'.format(
+                zone_id,
+                parts[1],
+                parts[0])).object
+        if len(result) is 0:
```
Kami: This would probably work most of the time, but it's kinda Python implementation specific so you should only use `is` to compare for True, False and None since `is` checks the reference and not the actual value.

See https://stackoverflow.com/questions/1504717/why-does-comparing-strings-in-python-using-either-or-is-sometimes-produce

libcloud/dns/drivers/godaddy.py
```diff
@@ -0,0 +1,521 @@
...
+            raise GoDaddyDNSException("Could not locate record")
```
Kami: Should throw `RecordDoesNotExistError`

Kami:
Thanks.

Added some comments.

Kami:
Ah, you are fast! :)

tonybaloney:
- [x] fix string interpolation for python 2,5
- [x] merge ttl update and use new record type
- [x] wrap string formats in parenthesis
- [x] create method to format a new record request for create and update record
- [x] dont use hasattr
- [x] use proper equality comparer
- [x] create utility method for format id of record
- [x] throw record does not exist error

tonybaloney:
ok @Kami done. very thorough code review!

libcloud/dns/drivers/godaddy.py
```diff
@@ -0,0 +1,498 @@
...
+            item_count=result['itemCount'],
+            total=result['total'],
+            currency=result['currency']
+        )
+
+    def _format_record(self, name, type, data, extra):
+        if extra is None:
+            extra = {}
+        new_record = {}
+        if type is RecordType.SRV:
```
Kami: Same here, should also use `==`.

libcloud/dns/drivers/godaddy.py
```diff
@@ -0,0 +1,498 @@
...
+        :rtype: :class:`Record`
+        """
+        parts = record_id.split(':')
+        result = self.connection.request(
+            '/v1/domains/%s/records/%s/%s' % (
+                zone_id,
+                parts[1],
+                parts[0])).object
+        if len(result) == 0:
+            raise RecordDoesNotExistError()
```
Kami: I don't think this will work - you need to pass the following arguments to the constructor - value, driver, record_id

Kami:
A couple of more things.

Besides that, LGTM, +1

Kami:
It appears that this driver never worked - https://stackoverflow.com/questions/35553386/apache-libcloud-dns-and-godaddy-name-or-service-not-known

I pushed a fix in a3ba6a4751623224f16175df9175ec06b29cdc1a

PR #642, opened 2015-11-25
Make azure driver create_node use kwargs

lkoenig:
Make AzureNodeDriver create_node use only **kwargs. That should
allow the use of the debloy_node from the base class to work.
## Motivation

Prior that change, one cannot use a deploy_node as deploy_node (defined on the libcloud.compute.base.NodeDriver) call the create_node of the driver with only keyword arguments.

### DIFF
--- a/libcloud/compute/drivers/azure.py
+++ b/libcloud/compute/drivers/azure.py
@@ -513,12 +513,7 @@ class AzureNodeDriver(NodeDriver):
         volumes = [self._to_volume(volume=v, node=node) for v in data]
         return volumes
 
-    def create_node(self, name, size, image, ex_cloud_service_name,
-                    ex_storage_service_name=None, ex_new_deployment=False,
-                    ex_deployment_slot="Production", ex_deployment_name=None,
-                    ex_admin_user_id="azureuser", ex_custom_data=None,
-                    ex_virtual_network_name=None, ex_network_config=None,
-                    auth=None, **kwargs):
+    def create_node(self, **kwargs):
         """
         Create Azure Virtual Machine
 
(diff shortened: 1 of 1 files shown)


tonybaloney:
Hi @lkoenig this goes against one of the projects code standards : https://libcloud.readthedocs.org/en/latest/development.html#don-t-abuse-kwargs

the drivers explicitly use named arguments to allow people to see from the docs what the arguments are, which are required, which are optional. etc.

create_node is intended to be used directly for all drivers. 

please can you explain your use case with a bit more context so we might be able to help?

lkoenig:
Thanks @tonybaloney for that feedback. I had read that documentation prior submitting the pull request. My issue is I didn't see that working with the implentation of libcloud.compute.base.deploy_node as that methode, line 902 of the file base.py call

``` python
        node = self.create_node(**kwargs)
```

Which, in the case of the azure driver, will throw an exception (as create_node in the azure driver expects some positional argument)

Have you a way you can recommand to fix that ?
I saw 2 ways:
- Make the create_node of the azure driver use only kwargs (like the one in openstack and other) even if that breaks the policy yo mentionned
- Change how the deploy_node takes and passes argument. To me that require a bigger change (maybe compltely worthed, but as I am new to the project, I prefer to ask).

Thanks again for your help,
Lionel
PS: I'll check also the UT failure ;)

tonybaloney:
@lkoenig everything in base.NodeDriver is meant to be overridden, if you don't override it, normally there is a NotImplementedError to catch you. 
In the case of the Azure driver it looks like it doesn't match up the arguments correctly, so you want to override deploy_node with your own implementation.

Here's an example creating a node with Azure https://github.com/mistio/mist.io/blob/master/src/mist/io/methods.py#L2315 

Is it deploy_node you want to use or create_node?

lkoenig:
I would like to leverage the deploy_node method of the drive.
According to the documentation of deploy_node, one should not overwrite that method
As stated line 829 of libcloud/compute/base.py:

```
       Deploy node is typically not overridden in subclasses.  The
        existing implementation should be able to handle most such.
```

Thus, to make then the create_node of the azure driver be able to work with the deploy_node, I did the change proposed.

Another alternatives would have been to change the deploy_node to be something like:

``` python
def deploy_node(self, *args, **kwargs):
    # (...)
    self.create_node(*args, **kwargs)
    # (...)
```

Let me know if you prefer another way to solve that issue.

tonybaloney:
@Kami  I'll refer this one to you.

Kami:
Sorry for the delay and thanks for the PR!

@tonybaloney is right - we should always explicitly declare arguments when possible. Sadly some of the old / legacy code doesn't do that, but we should avoid adding more debt in the future when possible.

There are two things we could do:
1. Leave `create_node` signature as it is and also add `**kwargs` to the end
2. Update / fix `deploy_node` so it doesn't pass `deploy_node` specific arguments to `create_node`. We could probably do something simple like this:

``` python
create_node_kwargs = copy.deepcopy(kwargs)
# remove all the deploy_node specific arguments from create_node_kwargs
node = create_node(**create_node_kwargs)
```

I think second option is the way to go. First one makes it work, but it doesn't really solve the "kwargs abuse" problem - method will just silently take in all the additional arguments and ignore them (that makes for a very bad, confusing and unfriendly API).

lkoenig:
Thanks for the inputs. I totally agree that explicit argument is way better that kwargs. I am totally for not building up technical debt.
I'll try to implement the second solution as it seems elegant and doing the work. I'll get back as soon as I have something working.

lkoenig:
Thinking, might be better to close that PR and start a new one (as the title could be misleading).

Kami:
Great - thanks!

I'm fine with opening a new PR.

PR #644, opened 2015-11-30
LIBCLOUD-775 Update node, update VMware tools, add storage, change st…

tonybaloney:
…orage size or speed, remove storage. With tests

### DIFF
--- a/libcloud/compute/drivers/dimensiondata.py
+++ b/libcloud/compute/drivers/dimensiondata.py
@@ -36,6 +36,7 @@ from libcloud.common.dimensiondata import NetworkDomainServicePlan
 from libcloud.common.dimensiondata import API_ENDPOINTS, DEFAULT_REGION
 from libcloud.common.dimensiondata import TYPES_URN
 from libcloud.common.dimensiondata import SERVER_NS, NETWORK_NS, GENERAL_NS
+from libcloud.utils.py3 import urlencode
 from libcloud.utils.xml import fixxpath, findtext, findall
 from libcloud.compute.types import NodeState, Provider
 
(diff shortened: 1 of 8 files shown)


libcloud/compute/drivers/dimensiondata.py
```diff
@@ -1134,6 +1241,118 @@ def ex_disable_monitoring(self, node):
...
+        """
+        result = self.connection.request_with_orgId_api_1(
+            'server/%s/disk/%s?delete' %
+            (node.id, disk_id)).object
+        response_code = findtext(result, 'result', GENERAL_NS)
+        return response_code in ['IN_PROGRESS', 'SUCCESS']
+
+    def ex_change_storage_speed(self, node, disk_id, speed):
+        """
+        Remove storage from a node
```
Kami: I believe this comment needs to be updated :)

libcloud/compute/drivers/dimensiondata.py
```diff
@@ -1134,6 +1241,118 @@ def ex_disable_monitoring(self, node):
...
+            'server/%s/disk/%s/changeSpeed' %
+            (node.id, disk_id),
+            method='POST',
+            data=ET.tostring(create_node)).object
+        response_code = findtext(result, 'result', GENERAL_NS)
+        return response_code in ['IN_PROGRESS', 'SUCCESS']
+
+    def ex_change_storage_size(self, node, disk_id, size):
+        """
+        Remove storage from a node
```
Kami: Same here.

libcloud/compute/drivers/dimensiondata.py
```diff
@@ -1134,6 +1241,118 @@ def ex_disable_monitoring(self, node):
...
+
+        :rtype: ``bool``
+        """
+        if image_description is None:
+            image_description = ''
+        result = self.connection.request_with_orgId_api_1(
+            'server/%s?clone=%s&desc=%s' %
+            (node.id, image_name, image_description)).object
+        response_code = findtext(result, 'result', GENERAL_NS)
+        return response_code in ['IN_PROGRESS', 'SUCCESS']
```
Kami, line 1354: Noticed all those API are async - eventually it might be handy to provide a "sync" interface in libcloud (e.g. this way you could potentially return a `NodeImage` object from this method).

libcloud/compute/drivers/dimensiondata.py
```diff
@@ -1134,6 +1241,118 @@ def ex_disable_monitoring(self, node):
...
+
+        :param  description: The description of the image
+        :type   description: ``str``
+
+        :rtype: ``bool``
+        """
+        if image_description is None:
+            image_description = ''
+        result = self.connection.request_with_orgId_api_1(
+            'server/%s?clone=%s&desc=%s' %
```
Kami, line 1351: Eventually it would also be good to refactor `request_with_orgId_api_1` to take body params as a dict and urlencode it in the `request_with_orgId_api_1` method.

This way it would be safer and more robust since right now special characters (e.g. `&` inside the description) could break the manual encoding...

Kami:
Added some comments for future improvements (not a blocker for this PR, you can address them in a future PR).

Besides that, LGTM, +1.

Please also make sure you update changelog.

tonybaloney:
Will update CHANGES.rst separately since I had a couple of other closed PRs anyway.

PR #645, opened 2015-11-30
Outscale inc & sas driver update

LordShion:
Hi every one,
This driver update permits the use of OUTSCALE's eu-west-2 & us-east-2 regions and endpoints.
needed pricing.json to be modified for driver to work with outdated prices, this should be modified later with actual prices.
I'm an Outscale user (on these new regions), and also worked there for some time.
this branch has passed trough Travis CI


tonybaloney:
Great stuff, thanks @LordShion I will merge.

PR #646, opened 2015-12-01
EC2 updates

pmhahn:
The EC2 Instance Type information contains some errors (especially the disk size for the m3 instances), conversion errors between GiB an M[i]B, disk count, ..
Also g2.8xlarge and t2.large are missing.


tonybaloney:
@pmhahn thanks for the submission,  the tests are failing because pricing.json is missing the new instance types, e.g.
KeyError: 'g2.8xlarge'
Please can you add the missing prices and resubmit

PR #647, opened 2015-12-01
Include 'service_name' support in _parse_service_catalog_auth_v3 for Openstack Drivers

steve-gregory:
The 'service_name' attribute was being left out when parsing the v3 ServiceCatalog returned by OpenStack. As a result, this error would occur: `LibcloudError: <LibcloudError in None 'Could not find specified endpoint'>`

The solution is to parse and add the service_name to the `OpenStackServiceCatalogEntry` similar to how it was done in `_parse_service_catalog_auth_v2`.

### DIFF
--- a/libcloud/common/openstack_identity.py
+++ b/libcloud/common/openstack_identity.py
@@ -419,6 +419,7 @@ class OpenStackServiceCatalog(object):
 
         for item in service_catalog:
             service_type = item['type']
+            service_name = item.get('name', None)
 
             entry_endpoints = []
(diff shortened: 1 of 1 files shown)


jimbobhickville:
+1 makes sense to me

tonybaloney:
Thanks @steve-gregory will merge

Kami:
Thanks.

It's been a while since I wrote this, but IIRC, `name` was not available in the service catalog when I was originally testing this. It's possible that things have changed since then, but some tests would also be good :)

PR #648, opened 2015-12-02
Fixing Python 2.6 unit testing and fixing Google tests.

crunk1:
@erjohnso 
@jimbobhickville

### DIFF
--- a/libcloud/common/google.py
+++ b/libcloud/common/google.py
@@ -101,35 +101,30 @@ except ImportError:
     RSA = None
     PKCS1_v1_5 = None
 
-TIMESTAMP_FORMAT = '%Y-%m-%dT%H:%M:%SZ'
+UTC_TIMESTAMP_FORMAT = '%Y-%m-%dT%H:%M:%SZ'
 
 
-def _now():
+def _utcnow():
+    """
+    Mocked in libcloud.test.common.google.GoogleTestCase.
+    """
     return datetime.datetime.utcnow()
 
 
-def _is_gce():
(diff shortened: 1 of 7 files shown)


jimbobhickville:
+1 - looks like you covered everything I found and then some.

jimbobhickville:
Btw, this should fix https://issues.apache.org/jira/browse/LIBCLOUD-776 - I forget the exact syntax you put in the commit message to have it automatically close the ticket, but I believe there's a way to do that.

tonybaloney:
:+1: looks much cleaner

crunk1:
Sorry, still working on it. I have been able to repro it on python2.6. Still fixing atm.

crunk1:
It appears that unittest in python 2.6 does not have setUpClass...

crunk1:
Interesting... libcloud/testing/__init__.py:27 shows that they try to mitigate that problem with python 2.6's unittest module by importing unittest2 (which is 2.7's module). But that seems to not be working.

@Kami comment?

crunk1:
Could be setup.py:21. It looks like the TestRunner is getting imported from Python 2.6's unittest module, which would explain why setUpClass isn't being called.

crunk1:
@Kami please take a look at my change to setup.py.

Kami:
@crunk1 setup.py change looks good to me - good catch!

erjohnso:
:+1: LGTM too, thanks @crunk1 and @jimbobhickville 

I ran through tests on both Mac and my GCE debian instance and it all checks out.  Merging it in now.

PR #650, opened 2015-12-02
Ensure rackspace lbs and servers know their uri and service_name

jimbobhickville:
This is a precursor to adding Rackspace RDNS support.  It lets you
set PTR records for loadbalancers and servers, but it needs these
two pieces of information to do so.

### DIFF
--- a/libcloud/compute/drivers/openstack.py
+++ b/libcloud/compute/drivers/openstack.py
@@ -862,6 +862,7 @@ class OpenStack_1_0_NodeDriver(OpenStackNodeDriver):
                      'uri': "https://%s%s/servers/%s" % (
                          self.connection.host,
                          self.connection.request_path, el.get('id')),
+                     'service_name': self.connection.get_service_name(),
                      'metadata': metadata})
         return n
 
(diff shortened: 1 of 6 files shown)


jimbobhickville:
Let me know if I need to open a JIRA for this bit.  I was only planning on making a JIRA for the next PR that will have full rdns support in the Rackspace DNS driver.

libcloud/compute/drivers/openstack.py
```diff
@@ -2129,7 +2131,8 @@ def _to_node(self, api_node):
                 vm_state=api_node.get("OS-EXT-STS:vm_state", None),
                 power_state=api_node.get("OS-EXT-STS:power_state", None),
                 progress=api_node.get("progress", None),
-                fault=api_node.get('fault')
+                fault=api_node.get('fault'),
```
tonybaloney: I dont think that comma is supposed to be there?

jimbobhickville: I can remove it, but trailing commas is generally considered good practice on multiline structures so it's easier to reorder and add elements without forgetting the comma.  Python just ignores it.  The only reason JSON doesn't support it is Internet Explorer at the time didn't support it, despite the Javascript spec specifically allowing it.

Anyway, I didn't mean to add it here, I just moved some stuff around and forgot to put it back to how it was.

jimbobhickville:
retest please

jimbobhickville:
Not sure what the magic incantation is, but the tests failed in the ec2 driver under pypy and nothing I changed was in any way related to either of those things, so I think they just need to be re-run.

Kami:
Thanks.

LGTM, +1

PR #652, opened 2015-12-03
[LIBCLOUD-780] Add Rackspace RDNS support

jimbobhickville:
New DNS driver methods:
- ex_iterate_ptr_records
- ex_get_ptr_record
- ex_create_ptr_record
- ex_update_ptr_record
- ex_delete_ptr_record

This should cover all of the functionality offered by the Rackspace
DNS API in regards to RDNS.

### DIFF
--- a/libcloud/dns/drivers/rackspace.py
+++ b/libcloud/dns/drivers/rackspace.py
@@ -23,6 +23,7 @@ from libcloud.utils.py3 import httplib
 import copy
 
 from libcloud.common.base import PollingConnection
+from libcloud.common.exceptions import BaseHTTPError
 from libcloud.common.types import LibcloudError
 from libcloud.utils.misc import merge_valid_keys, get_new_obj
 from libcloud.common.rackspace import AUTH_URL
@@ -34,7 +35,8 @@ from libcloud.dns.types import ZoneDoesNotExistError, RecordDoesNotExistError
 from libcloud.dns.base import DNSDriver, Zone, Record
 
 VALID_ZONE_EXTRA_PARAMS = ['email', 'comment', 'ns1']
-VALID_RECORD_EXTRA_PARAMS = ['ttl', 'comment', 'priority']
+VALID_RECORD_EXTRA_PARAMS = ['ttl', 'comment', 'priority', 'created',
+                             'updated']
 
 
 class RackspaceDNSResponse(OpenStack_1_1_Response):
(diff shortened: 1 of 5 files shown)


jimbobhickville:
This depends on #650 going in first.

jimbobhickville:
Hrm... testing this against the real API there's some issues still.  Closing it for now, will reopen when I sort 'em out.

jimbobhickville:
The documented response for listing PTR records is apparently not what is actually returned by the real API.  Adjusted the code/tests to match reality.

libcloud/dns/drivers/rackspace.py
```diff
@@ -343,6 +367,183 @@ def delete_record(self, record):
...
+
+        :param service_name: the service catalog name of the linked device(s)
+                             i.e. cloudLoadBalancers or cloudServersOpenStack
+        :param record_id: the id (i.e. PTR-12345) of the PTR record
+        :rtype: instance of :class:`RackspacePTRRecord`
+        """
+        self.connection.set_context({'resource': 'record', 'id': record_id})
+        response = self.connection.request(
+            action='/rdns/%s/%s' % (service_name, record_id)).object
+        item = next(iter(response['recordsList']['records']))
```
thesquelched, line 434: IIRC, this is only available in python 2.6+

libcloud/dns/drivers/rackspace.py
```diff
@@ -131,6 +133,28 @@ def get_endpoint(self):
...
+
+    def update(self, domain, extra=None):
+        return self.driver.ex_update_ptr_record(record=self, domain=domain,
+                                                extra=extra)
+
+    def delete(self):
+        return self.driver.ex_delete_ptr_record(record=self)
+
+    def __repr__(self):
+        return ('<%s: ip=%s, domain=%s, provider=%s ...>' %
```
jimbobhickville, line 153: I only used the `%` interpolation because other objects in libcloud were doing so already.  I'm fine switching to `.format()` as that seems to be the preferred method these days.  Let me know.

libcloud/dns/drivers/rackspace.py
thesquelched, line 434: `next`, that is

EDIT: [Confirmed](https://docs.python.org/2/library/functions.html#next)

libcloud/dns/drivers/rackspace.py
tonybaloney, line 153: .format is not compatible with some versions of python that we are compatible with, so stick with %

tonybaloney:
#650 is closed.

tonybaloney:
LGTM :+1: thanks Greg

Kami:
Great, thanks! :)

PR #653, opened 2015-12-04
Extended load balancer functionality to support persistence profiles,…

tonybaloney:
… irules and default monitors.

### DIFF
--- a/libcloud/common/dimensiondata.py
+++ b/libcloud/common/dimensiondata.py
@@ -435,13 +435,13 @@ class DimensionDataVlan(object):
         Initialize an instance of ``DimensionDataVlan``
 
         :param id: The ID of the VLAN
-        :type  id: ``string``
+        :type  id: ``str``
 
         :param name: The name of the VLAN
-        :type  name: ``string``
+        :type  name: ``str``
 
         :param description: Plan text description of the VLAN
-        :type  description: ``string``
+        :type  description: ``str``
 
         :param location: The location (data center) of the VLAN
         :type  location: ``NodeLocation``
@@ -454,11 +454,11 @@ class DimensionDataVlan(object):
 
         :param private_ipv4_range_address: The host address of the VLAN
                                             IP space
-        :type  private_ipv4_range_address: ``string``
+        :type  private_ipv4_range_address: ``str``
 
         :param private_ipv4_range_address: The size (e.g. '24') of the VLAN
                                             as a CIDR range size
-        :type  private_ipv4_range_address: ``string``
+        :type  private_ipv4_range_address: ``str``
         """
         self.id = str(id)
         self.name = name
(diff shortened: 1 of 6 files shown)


libcloud/common/dimensiondata.py
```diff
@@ -553,3 +553,64 @@ def __repr__(self):
...
+        self.name = name
+        self.compatible_listeners = compatible_listeners
+        self.fallback_compatible = fallback_compatible
+
+    def __repr__(self):
+        return (('<DimensionDataPersistenceProfile: id=%s, name=%s>')
+                % (self.id, self.name))
+
+
+class DimensionDataDefaultiRule(object):
```
Kami, line 706: Is the `i` in `Defaulti` intentional?

libcloud/common/dimensiondata.py
```diff
@@ -553,3 +553,64 @@ def __repr__(self):
...
+    """
+    A default iRule for a network domain, can be applied to a listener
```
Kami, line 708: Ah, never mind, I assume i refers to `ingress`?

Kami, line 708: If so, I would prefer to explicitly call it out - so `DimensionDataDefaultIngressRule`.

libcloud/common/dimensiondata.py
```diff
@@ -553,3 +553,64 @@ def __repr__(self):
...
+    """
+    def __init__(self, id, name, compatible_listeners):
```
Kami, line 710: Some docstrings for those arguments would be nice - I have no idea what `compatible_listeners` represents (I assume it's an array of something).

libcloud/common/dimensiondata.py
```diff
@@ -553,3 +553,64 @@ def __repr__(self):
...
+    def __repr__(self):
+        return (('<DimensionDataDefaultiRule: id=%s, name=%s>')
+                % (self.id, self.name))
+
+
+class DimensionDataVirtualListenerCompatibility(object):
+    """
+    A compatibility preference for a persistence profile
+    """
+    def __init__(self, type, protocol):
```
Kami, line 739: Same here.

libcloud/common/dimensiondata.py
tonybaloney, line 708: iRule is the feature name. Can't remember what it stands for! It's a feature of the F5 

Sent from Outlook

On Sat, Dec 5, 2015 at 8:06 PM -0800, "Tomaz Muraus" notifications@github.com wrote:

In libcloud/common/dimensiondata.py:

> -    Fallback Persistence Profile.
> -    """
> -    def **init**(self, id, name, compatible_listeners, fallback_compatible):
> -        self.id = id
> -        self.name = name
> -        self.compatible_listeners = compatible_listeners
> -        self.fallback_compatible = fallback_compatible
>   +
> -    def **repr**(self):
> -        return (('<DimensionDataPersistenceProfile: id=%s, name=%s>')
> -                % (self.id, self.name))
>   +
>   +
>   +class DimensionDataDefaultiRule(object):
> -    """
> -    A default iRule for a network domain, can be applied to a listener

Ah, never mind, I assume i refers to ingress?

PR #654, opened 2015-12-06
[WIP] LIBCLOUD-781 docker node driver

tonybaloney:
Port the node driver for Docker from @mgogoulos initial driver in Mist.IO

TODO: 
- [x] Port initial driver code
- [ ] Refactor with docstrings, method naming and project standards
- [ ] Setup tests and fixtures
- [ ] Update docs

### DIFF
--- a/libcloud/compute/drivers/docker.py
+++ b/libcloud/compute/drivers/docker.py
@@ -1,0 +1,506 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Docker (http://docker.io) driver.
+Created by Markos Gogoulos (mgogoulos@mist.io)
+"""
+
+import base64
+import datetime
+import shlex
+import re
+
+try:
+    import simplejson as json
+except:
+    import json
+
+from libcloud.utils.py3 import httplib
+from libcloud.utils.py3 import b
+
+from libcloud.compute.providers import Provider
+from libcloud.common.base import JsonResponse, ConnectionUserAndKey
+from libcloud.compute.types import (NodeState, InvalidCredsError)
+from libcloud.compute.base import (Node, NodeDriver, NodeImage,
+                                   NodeSize)
+
+VALID_RESPONSE_CODES = [httplib.OK, httplib.ACCEPTED, httplib.CREATED,
+                        httplib.NO_CONTENT]
+
+
+class DockerResponse(JsonResponse):
+
+    valid_response_codes = [httplib.OK, httplib.ACCEPTED, httplib.CREATED,
+                            httplib.NO_CONTENT]
+
+    def parse_body(self):
+        if len(self.body) == 0 and not self.parse_zero_length_body:
+            return self.body
+
+        try:
+            # error responses are tricky in Docker. Eg response could be
+            # an error, but response status could still be 200
+            body = json.loads(self.body)
+        except ValueError:
+            m = re.search('Error: (.+?)"', self.body)
+            if m:
+                error_msg = m.group(1)
+                raise Exception(error_msg)
+            else:
+                raise Exception('ConnectionError: Failed to parse JSON response')
+        return body
+
+    def parse_error(self):
+        if self.status == 401:
+            raise InvalidCredsError('Invalid credentials')
+        return self.body
+
+    def success(self):
+        return self.status in self.valid_response_codes
+
+
+class DockerConnection(ConnectionUserAndKey):
+
+    responseCls = DockerResponse
+    timeout = 60
+
+    def add_default_headers(self, headers):
+        """
+        Add parameters that are necessary for every request
+        If user and password are specified, include a base http auth
+        header
+        """
+        headers['Content-Type'] = 'application/json'
+        if self.user_id and self.key:
+            user_b64 = base64.b64encode(b('%s:%s' % (self.user_id, self.key)))
+            headers['Authorization'] = 'Basic %s' % (user_b64.decode('utf-8'))
+        return headers
+
+
+class DockerNodeDriver(NodeDriver):
+    """
+    Docker node driver class.
+
+    >>> from libcloud.compute.providers import get_driver
+    >>> driver = get_driver('docker')
+    >>> conn = driver(host='198.61.239.128', port=4243)
+    >>> conn.list_nodes()
+    or connecting to http basic auth protected https host:
+    >>> conn = driver('user', 'pass', host='https://198.61.239.128', port=443)
+
+    connect with tls authentication, by providing a hostname, port, a private
+    key file (.pem) and certificate (.pem) file
+    >>> conn = driver(host='https://198.61.239.128', port=4243, key_file='key.pem', cert_file='cert.pem')
+    """
+
+    type = Provider.DOCKER
+    name = 'Docker'
+    website = 'http://docker.io'
+    connectionCls = DockerConnection
+    features = {'create_node': ['password']}
+
+    def __init__(self, key=None, secret=None, host='localhost',
+                 port=4243, secure=False, key_file=None, cert_file=None):
+
+        super(DockerNodeDriver, self).__init__(key=key, secret=secret,
+                                               host=host, port=port, secure=secure, key_file=key_file, cert_file=cert_file)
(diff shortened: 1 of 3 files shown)


libcloud/compute/drivers/docker.py
```diff
@@ -0,0 +1,573 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Docker (http://docker.io) driver.
+Created by Markos Gogoulos (mgogoulos@mist.io)
```
sebgoa: We don't keep authorship information that way at ASF, so we can remove those lines.

libcloud/compute/drivers/docker.py
```diff
@@ -0,0 +1,573 @@
...
+    key file (.pem) and certificate (.pem) file
+    >>> conn = driver(host='https://198.61.239.128',
+        port=4243, key_file='key.pem', cert_file='cert.pem')
+    """
+
+    type = Provider.DOCKER
+    name = 'Docker'
+    website = 'http://docker.io'
+    connectionCls = DockerConnection
+    features = {'create_node': ['password']}
```
sebgoa, line 110: We don't really need this feature spec. I don't believe we should try to do a 'deploy' for a container.

libcloud/compute/drivers/docker.py
```diff
@@ -0,0 +1,573 @@
...
+
+    def list_nodes(self):
+        """
+        List running and stopped containers
+        show_all=False will show only running containers
+
+        :return:  list of node objects
+        :rtype: ``list`` of :class:`.Node`
+        """
+        return self.ex_list_nodes(True)
```
sebgoa: Why pointing to ex_list_nodes ?

libcloud/compute/drivers/docker.py
tonybaloney, line 110: that's really the whole point of the driver, that you can treat the docker host like a KVM or a vsphere hypervisor and 'create node' that pulls a container and starts it for you. at least that's the way mist.io are using it now

poliveira89:
@tonybaloney do you use it in a dev or prod environment?
If this PR is accepted, could lead to create providers for Swarm, Kubernetes and other Container implementations as well?

libcloud/compute/drivers/docker.py
poliveira89, line 110: I agree with @tonybaloney

sebgoa, line 110: a deploy in libcloud is used to provision the instance created.
I understand that you want to start container, I just don't know how you are created the underlying docker host ? Is that being created by libcloud and this driver as well ? or do you expect it to be there and accessible via a password ?

libcloud/compute/drivers/docker.py
```diff
@@ -0,0 +1,558 @@
...
+            'CpuShares': cpu_shares,
+            'WorkingDir': working_dir,
+            'MemorySwap': memswap_limit,
+            'PublishAllPorts': True,
+            'PortBindings': port_bindings,
+        }
+
+        data = json.dumps(payload)
+        try:
+            result = self.connection.request('/containers/create', data=data,
```
poliveira89, line 197: Just for curiosity, why do you have used the HTTP API instead their official [Python package](https://github.com/docker/docker-py)?

libcloud/compute/drivers/docker.py
poliveira89, line 110: I think that the need to support docker or other ecosystem in container universe, likewise any other cloud provider, you want to "provision" without the specifics or the underlying provider. 
As you do for: AWS, Openstack, Digital Ocean, etc...

On the Docker support, it's the same, you just need to provision. Does not matter if is Docker, Swarm, Kubernetes, or any other Container implementation.
At least, it's the way I think... and for that reason, that I looked into this with 100% interest. :smile: 

PS: I want to create a client for Instance Provision and not worry about the provider. Because, for my use-case docker (containers) belongs to the providers available...

mgogoulos:
hi @poliveira89, we use this driver on https://github.com/mistio/mist.io to run actions on containers and also deploy new ones (passing a name, image, command, ports to expose, port bindings and also env variables)

sebgoa:
@mgogoulos cool. Can you explain how you use the 'password' defined in the features list ? I am just not getting the workflow

poliveira89:
@mgogoulos which branch contains the stable version of libcloud with docker support?

I have only found this one: https://github.com/DimensionDataCBUSydney/libcloud/tree/LIBCLOUD-781_Docker

mgogoulos:
Hi @runseb, you are right, this line can be omitted

```
features = {'create_node': ['password']}
```

The create_node() function can be used without the need of this line of course.

libcloud/compute/drivers/docker.py
tonybaloney, line 197: Generally LibCloud does the minimal integration instead of relying on 3rd party Libs. The number of Libs would be huge and lack of control and test ability would be an issue

libcloud/compute/drivers/docker.py
tonybaloney, line 110: You would deploy the actual host wth another driver like aws or GCE then once docker daemon is installed use this driver to deploy containers. Just the same as the VMware driver assumes VMware is already there. This is the next level up

tonybaloney, line 110: And yes. I think we need to expand on this model and leverage the great foundation in the library to provide container abstraction. Most of the big cloud providers will start offering container as a service in 2016 so this will be a chance to get ahead. Kubernetes etx would be welcome

poliveira89, line 110: I know this lib a "long time" but it's the first time, I'm "obligated" to use it. 
My final question, I hope so :stuck_out_tongue:, if this PR is approved how much we have to wait to be release on a official version? 

I'm just asking for a estimation, I'm not demanding for anything. 
Thank you for your help. :smile:

libcloud/compute/drivers/docker.py
poliveira89, line 197: Makes sense! :+1: 

As I said, I asked for curiosity, and to understand your development choice! :)

libcloud/compute/drivers/docker.py
tonybaloney, line 110: About 2 months. You can of course try it now! Pip supports pulling from a git branch.

poliveira89, line 110: In this conversation, on bottom, I have asked @mgogoulos about which repo or branch, can you help me on that please?

thank you a lot, for the time you are getting to answer me! :+1:

Kami:
@poliveira89

> @tonybaloney do you use it in a dev or prod environment?
> If this PR is accepted, could lead to create providers for Swarm, Kubernetes and other Container implementations as well?

Similar to libvirt, a lot of docker functionality doesn't really nicely map to the libcloud.compute.\* API which was primarily designed to handle cloud and virtual servers.

If we decide to also add support for other container run-times (docker, rkt, FreeBsD jails, Solaris zones, etc.) it would be worth evaluating if we should create a new `libcloud.container` API which is primarily targeted and designed for containers.

In any case, if we go with a new API it's also worth noting that that's a much bigger task and commitment.

Also as far as Swarm and Kubernetes goes - those are more of container orchestrator systems so I don't think they fit well in either compute or potential new container API.

tonybaloney:
@poliveira89 @Kami @runseb 

I can see 3 options for us here (if we target docker and then rkt)
1) proceed with docker as a compute driver, the implementation is a bit wonky but it works
2) create a new driver type, ContainerNodeDriver, which INHERITS from NodeDriver, list_nodes would be overloaded to return ContainerNode's instead of Nodes. a ContainerNode (which inherits from Node) would share the same properties, but also have an additional property stating which Node it belongs to. There would be some other methods like pull_image, etc. That way users can use list_nodes etc. but also take advantage of the other functions.
3) create a new driver type, ContainerDriver and start from scratch, just design and implement which methods based on usage

For the record. I favor option 2.

poliveira89:
@Kami  let me start this comment, by saying that I fully understand that this "upgrade" or embracing Containers will be a much bigger task and require a lot effort from everyone involved (because as you stated some changes - design and implementation must be discussed). I just asked to inquire which was the roadmap regarding this topic - for Container-as-a-Service providers.

But, if you allow me to say this, before anyone start this undertaking, why don't start small? First discuss the acceptance of this PR... as @tonybaloney and @mgogoulos states, this works perfectly for their use-case scenario (which, I assume, is deploy instances/nodes regardless of the provider).
After, that the community and the Libcloud team could start the discussion of a more serious a well-developed solution regarding that alongside this there are some consortiums discussing standards for containers.

After reading, @tonybaloney last post, I still does not have a well defined opinion, because I'm questioning (because I'm not seeing at first) what is the major differences that requires a differente object Driver for Containers?

To finalize, IMHO, I think that considering Docker as a provider of Libcloud is not that farfetch, because, I may be wrong, but you support Digital Ocean and other providers which works with Docker Daemon/Client to deploy containers. 
Source: https://docs.docker.com/machine/drivers/

tetron:
One use case for a Docker compute node driver is a test environment where Docker containers are an adequate stand-in for booting full fledged cloud VMs.

tonybaloney:
@poliveira89 @tetron @runseb this PR is now #666

PR #655, opened 2015-12-07
Libcloud 782 remove dreamhost vps driver

smaffulli:
DreamHost removed support of public API for its VPS product so the driver doesn't work anymore. Documentation and driver should be removed.

DreamHost products are still supported by libcloud using the generic OpenStack driver though, so there is no need to remove DreamHost's logo from the website or other parts of the documentation.


tonybaloney:
thanks @smaffulli should we leave documentation somewhere pointing users to the openstack driver?

smaffulli:
That's a great idea, thanks for suggesting it. Any suggestion of where would be a good place to mention that?

tonybaloney:
Since the driver no longer works. I'll merge this since we don't want to ship a non-functioning driver, the removal will be included in the release notes and _supported_providers will include a mention

smaffulli:
Thank you. Do I need to do something with the Jira ticket? https://issues.apache.org/jira/browse/LIBCLOUD-782

PR #656, opened 2015-12-08
[LIBCLOUD-783] API 2.1 support for dimension data driver

tonybaloney:
- Move image functions to the 2.1 API namespace
- Update existing calls to 2.1
- Update test fixtures with response changes where appropriate
- Image listing and customer image listing
- Allow user to configure CPU and RAM specifications up front


tonybaloney:
Completed.

tonybaloney:
famous last words. I've CR'ed this but I'm sure it could do with a second pair of eyes. @Kami @runseb or @erjohnso would be appreciated.

tonybaloney:
finished final tests against prod instances and AU and NA APIs which were both updated yesterday. merging

PR #657, opened 2015-12-10
Fixed error when proxy_url is None.

juanfont:
A recently added change (https://github.com/apache/libcloud/commit/3fa2e1b4f70ed472ba39c74322915412f2d44f36) contains an error when proxy_url is None. This small change fixes it.

### DIFF
--- a/libcloud/compute/drivers/vcloud.py
+++ b/libcloud/compute/drivers/vcloud.py
@@ -848,7 +848,8 @@ class VCloud_1_5_Connection(VCloudConnection):
                      'application/vnd.vmware.vcloud.orgList+xml')).get('href')
             )
 
-            self.connection.set_http_proxy(self.proxy_url)
+            if self.proxy_url:
+                self.connection.set_http_proxy(self.proxy_url)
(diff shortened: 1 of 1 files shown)


libcloud/compute/drivers/vcloud.py
```diff
@@ -848,7 +848,8 @@ def _get_auth_token(self):
                      'application/vnd.vmware.vcloud.orgList+xml')).get('href')
             )
 
-            self.connection.set_http_proxy(self.proxy_url)
+            if self.proxy_url:
```
jimbobhickville: I think the coding standards call for strict comparisons to None, like `if self.proxy_url is not None`.  Other than that, this looks like a no-brainer fix to me.

juanfont:
You were completely right :)

jimbobhickville:
lgtm, but I'm not a committer so someone else will need to review it.

Kami:
Thanks.

LGTM, +1

Kami:
On a related note - it might also be a good idea (improvement) to throw in `set_http_proxy` method if the provided url is None / not valid.

PR #659, opened 2015-12-15
Use hasattr rather than getattr to check for existence

jimbobhickville:
I forgot getattr throws an exception.  This should have
been a hasattr.  Additionally added a check to make sure
it's a dict before treating it as such.

### DIFF
--- a/libcloud/dns/drivers/rackspace.py
+++ b/libcloud/dns/drivers/rackspace.py
@@ -667,9 +667,10 @@ def _rackspace_result_has_more(response, result_length, limit):
 
 
 def _check_ptr_extra_fields(device_or_record):
-    if not (getattr(device_or_record, 'extra') and
+    if not (hasattr(device_or_record, 'extra') and
+            isinstance(device_or_record.extra, dict) and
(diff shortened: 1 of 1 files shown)


tonybaloney:
LGTM +1

libcloud/dns/drivers/rackspace.py
```diff
@@ -667,9 +667,10 @@ def _rackspace_result_has_more(response, result_length, limit):
 
 
 def _check_ptr_extra_fields(device_or_record):
-    if not (getattr(device_or_record, 'extra') and
```
Kami, line 670: Minor thing - you can use getattr with a default value and it won't throw if the attribute is not present, it will return a default value instead (`getattr(device_or_record, 'extra', None)`).

Kami, line 670: On a related note, I find this if a bit hard to read and digest.

I would find something like this a bit more easy to understand:

``` python
extra = getattr(device_or_record, 'extra', {})
uri = extra.get('uri', None)
service_name = extra.get('service_name', None)

if not uri and not service_name:
    # throw..
```

PR #663, opened 2015-12-21
t2.nano instance for ec2

tonybaloney:


Kami:
I just noticed you didn't update region details dict - https://github.com/DimensionDataCBUSydney/libcloud/blob/1273eacf5a89547354ececf447a7684b0db38f2a/libcloud/compute/drivers/ec2.py#L647

If you don't update that, user won't be able to use this new instance type (size). You need to update region details entry for each region where this instance type is now available (you will also need to update affected tests).

Kami:
And when you make that change, please also update changelog.

tonybaloney:
done

PR #665, opened 2015-12-22
Implemented LazyObject class, which provides a 'lazy' class method to create instances with lazy initialization. Th…

crunk1:
…e lazy class method returns a Proxy object that subclasses the target object class. Upon accessing the proxy object in any way, the object is initialized.

Modified Google Compute Engine License objects, GCELicense, to be such a lazy object. This addresses https://issues.apache.org/jira/browse/LIBCLOUD-786.

Tests/Verification:
  tox -e lint
  python setup.py test
  Added test/common/test_base.py which has LazyObjectTest

### DIFF
--- a/libcloud/common/base.py
+++ b/libcloud/common/base.py
@@ -76,6 +76,41 @@ __all__ = [
 RETRY_FAILED_HTTP_REQUESTS = False
 
 
+class LazyObject(object):
+    """An object that doesn't get initialized until accessed."""
+
+    @classmethod
+    def _proxy(cls, *lazy_init_args, **lazy_init_kwargs):
+        class Proxy(cls, object):
+            _lazy_obj = None
+
+            def __init__(self):
+                # Must override the lazy_cls __init__
+                pass
+
+            def __getattribute__(self, attr):
+                lazy_obj = object.__getattribute__(self, '_get_lazy_obj')()
+                return getattr(lazy_obj, attr)
+
+            def __setattr__(self, attr, value):
+                lazy_obj = object.__getattribute__(self, '_get_lazy_obj')()
+                setattr(lazy_obj, attr, value)
+
+            def _get_lazy_obj(self):
+                lazy_obj = object.__getattribute__(self, '_lazy_obj')
+                if lazy_obj is None:
+                    lazy_obj = cls(*lazy_init_args, **lazy_init_kwargs)
+                    object.__setattr__(self, '_lazy_obj', lazy_obj)
+                return lazy_obj
+
+        return Proxy()
+
+    @classmethod
+    def lazy(cls, *lazy_init_args, **lazy_init_kwargs):
+        """Create a lazily instantiated instance of the subclass, cls."""
+        return cls._proxy(*lazy_init_args, **lazy_init_kwargs)
+
+
 class HTTPResponse(httplib.HTTPResponse):
     # On python 2.6 some calls can hang because HEAD isn't quite properly
     # supported.
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -21,6 +21,7 @@ import datetime
 import time
 import sys
 
+from libcloud.common.base import LazyObject
 from libcloud.common.google import GoogleResponse
 from libcloud.common.google import GoogleBaseConnection
 from libcloud.common.google import GoogleBaseError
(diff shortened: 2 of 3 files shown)


crunk1:
@erjohnso @Kami

libcloud/compute/drivers/gce.py
```diff
@@ -231,15 +232,38 @@ def page(self, max_results=500):
...
-        self.charges_use_fee = charges_use_fee
-        self.extra = extra or {}
-        UuidMixin.__init__(self)
+        self.charges_use_fee = None  # init in _request
+        self.extra = None  # init in _request
+
+        self._request()
+
+    def _request(self):
+        # TODO(crunkleton@google.com): create new connection? or make
```
tonybaloney, line 249: @crunk1 so what is the intention with the todo? to raise a separate PR later

crunk1, line 249: Yes, I need to look into this later.

tonybaloney:
LGTM +1

libcloud/common/base.py
```diff
@@ -76,6 +76,39 @@
...
+    @staticmethod
+    def _proxy(lazy_cls, *lazy_init_args, **lazy_init_kwargs):
+        class Proxy(lazy_cls, object):
+            lazy_obj = None
+
+            def __init__(self):
+                pass
+
+            def __getattribute__(self, attr):
+                lazy_obj = object.__getattribute__(self, 'lazy_obj')
```
jimbobhickville: Should you prefix the `lazy_obj` attr with an underscore (`_lazy_obj`) to indicate it's intended to be a private attr?

jimbobhickville: Also, I believe if you override `__getattr__` instead of `__getattribute__` then you can just check the attribute value without having to invoke `__getattribute__` directly like this:

`if self.lazy_obj is None:`

But maybe there's something subtle here I'm missing?

libcloud/common/base.py
```diff
@@ -76,6 +76,39 @@
...
+                if lazy_obj is None:
+                    lazy_obj = lazy_cls(*lazy_init_args, **lazy_init_kwargs)
+                    object.__setattr__(self, 'lazy_obj', lazy_obj)
+                return lazy_obj.__getattribute__(attr)
```
jimbobhickville: I believe it's generally preferable to use the `getattr` function than invoking `__getattribute__` directly on a foreign object.  `return getattr(lazy_obj, attr)`

libcloud/common/base.py
```diff
@@ -76,6 +76,39 @@
...
+
+            def __setattr__(self, attr, value):
+                lazy_obj = object.__getattribute__(self, 'lazy_obj')
+                if lazy_obj is None:
+                    lazy_obj = lazy_cls(*lazy_init_args, **lazy_init_kwargs)
+                    object.__setattr__(self, 'lazy_obj', lazy_obj)
+                lazy_obj.__setattr__(attr, value)
```
jimbobhickville: Same advice about `setattr(lazy_obj, attr, value)` as above.

libcloud/common/base.py
crunk1: Changed `LazyObject.Proxy.lazy_obj` to `LazyObject.Proxy._lazy_obj`.

Regarding the `__getattr__` vs `__getattribute__` discussion, see http://stackoverflow.com/a/3278104/1436495.
My considerations:
- `__getattr__` is only called if attribute resolution fails by normal means
- `__getattribute__` is ALWAYS called (even on bound methods); easier to handle given my next point
- the `Proxy` metaclass inherits from `lazy_cls` (I want `Proxy` to be as transparent as possible as a wrapper; I want the returned `Proxy` instance to have the class info for `lazy_cls`), meaning that `Proxy` inherits all methods and class attributes of `lazy_cls`. Therefore, weird things happen when you call methods or use class-level attributes intended for the lazy object instead of the `Proxy`.

libcloud/common/base.py
crunk1: True, thanks for the heads up! I got a little `object.__getattribute__` happy.

libcloud/common/base.py
crunk1: Done

libcloud/common/base.py
```diff
@@ -76,6 +76,41 @@
...
+    def _proxy(cls, *lazy_init_args, **lazy_init_kwargs):
+        class Proxy(cls, object):
+            _lazy_obj = None
+
+            def __init__(self):
+                # Must override the lazy_cls __init__
+                pass
+
+            def __getattribute__(self, attr):
+                lazy_obj = object.__getattribute__(self, '_get_lazy_obj')()
```
jimbobhickville, line 92: Wrapping it in a method was nice, but I don't think you need to use `__getattribute__` to call that method since it's defined in the same class here, so any invocation of this method is guaranteed to have `self._get_lazy_obj()` available to call.

jimbobhickville:
One last nitpick, but otherwise lgtm.

libcloud/common/base.py
crunk1, line 92: Since the Proxy class has `__getattribute__` defined, `self._get_lazy_obj()` will first call `Proxy.__getattribute__(self, '_get_lazy_obj')` to get the bound method. That leads to an infinite recursion loop. See an example here: https://repl.it/Bavg/1.

jimbobhickville, line 92: Oh, right, derp.  I should stop doing code reviews before I get any caffeine in me.

crunk1, line 92: All good, thanks for all your help!

erjohnso:
Ok, this lgtm too and I've done some additional testing. The `demos/gce_demo.py` script is also seeing about a 22% reduction in API calls too. So, :+1:, great work @crunk1!

Merging now.

crunk1:
This has now been merged: https://github.com/apache/libcloud/commit/81a8dbc324f870444e980aee523fea48fc37cbcc

PR #668, opened 2015-12-28
Fix typos

AmbientLighter:


libcloud/dns/types.py
```diff
@@ -69,11 +69,9 @@ class RecordType(object):
     TXT = 'TXT'
     PTR = 'PTR'
     SOA = 'SOA'
-    HINFO = 'HINFO'
     SPF = 'SPF'
     SRV = 'SRV'
     SSHFP = 'SSHFP'
-    PTR = 'PTR'
```
jimbobhickville, line 76: Why was this removed?  I'm pretty sure the Rackspace RDNS stuff I added recently references it.

jimbobhickville:
All the typo and formatting fixes look good, just the question about the `RecordType` removals.

libcloud/dns/types.py
AmbientLighter, line 76: This line is duplicated twice: `PTR = 'PTR'`
https://github.com/apache/libcloud/blob/trunk/libcloud/dns/types.py#L70
https://github.com/apache/libcloud/blob/trunk/libcloud/dns/types.py#L76

This line is duplicated twice: `HINFO = 'HINFO'`
https://github.com/apache/libcloud/blob/trunk/libcloud/dns/types.py#L68
https://github.com/apache/libcloud/blob/trunk/libcloud/dns/types.py#L72

jimbobhickville, line 76: Ah, derp, I missed that.  Good catch.

jimbobhickville:
lgtm

Kami:
Good catch - thanks.

docs/compute/drivers/auroracompute.rst
```diff
@@ -1,5 +1,5 @@
 AuroraCompute Computer Driver Documentation
-======================================
+===========================================
```
Kami, line 2: We need to update our build process to turn sphinx warnings into errors again and fail the build.

It used to be turned on, but then for some reason (don't recall why anymore), we disabled it.

Kami:
Merged, thanks.

PR #669, opened 2015-12-30
Treat sphinx warnings as errors

Kami:
This pull request re-enabled features which treats sphinx warnings as errors which cause build to fail.

We had this feature enabled in the past, but at some point we disabled it (probably when we enabled sphinx auto doc) so we accumulated a bunch of documentation and docstring related debts since warnings didn't cause build to fail and people simply ignored those issues.

Future TODO (up for grabs):
- [ ] Fix all the docstring related issues
- [ ] Fix all the cross reference related issue (`more than one target found for cross-reference` messages)

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -80,7 +80,7 @@ release = '0.14.0-dev'
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
 exclude_patterns = [
-    'docs/apidocs/modules.rst',  # generated during build (orphan)
+    'apidocs/modules.rst',  # generated during build (orphan)
     '_build',
     '*/_*.rst'
 ]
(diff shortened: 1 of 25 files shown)


docs/conf.py
```diff
@@ -263,3 +269,29 @@
 intersphinx_mapping = {'http://docs.python.org/': None}
 
 autoclass_content = 'both'
+
+
+# Note: For now we ignore sphinx-autodoc warnings since there are too many
```
Kami, line 275: Yes, this is pretty darn awful, but we have too much docstring related debt right now (sadly, I don't have one week to spend fixing all those docstring related issues).

I want to start with at least fix non-docstring related issues and treating those issues as fatal going forward.

Kami:
Alright, merged.

Going forward, all the sphinx warnings are not treated as error and will fail the build - this means they need to be fixed before merging changes into trunk.

PR #670, opened 2016-01-01
Fixes to __repr__ element names and allowing paging of firewall rules

tonybaloney:
Original contribution from @mingsheng36


tonybaloney:
Already reviewed. will merge

PR #672, opened 2016-01-05
auroradns: Add support for Health Checks

wido:
AuroraDNS supports Health Checks and based on the state of these
checks records will be served or not.

This way a Round-Robin DNS balancing can be achieved pointing to
only healthy servers/services.

### DIFF
--- a/docs/examples/dns/auroradns/enable_disable_record.py
+++ b/docs/examples/dns/auroradns/enable_disable_record.py
@@ -1,6 +1,6 @@
 from libcloud.dns.types import Provider
-from libcloud.dns.providers import get_driver
 from libcloud.dns.types import RecordType
+from libcloud.dns.providers import get_driver
 
 cls = get_driver(Provider.AURORADNS)
 
--- a/docs/examples/dns/auroradns/health_checks.py
+++ b/docs/examples/dns/auroradns/health_checks.py
@@ -1,0 +1,21 @@
+from libcloud.dns.types import Provider, RecordType
+from libcloud.dns.providers import get_driver
+from libcloud.dns.drivers.auroradns import AuroraDNSHealthCheckType
+
+cls = get_driver(Provider.AURORADNS)
+
+driver = cls('myapikey', 'mysecret')
+
+zone = driver.get_zone('auroradns.eu')
+
+health_check = driver.ex_create_healthcheck(zone=zone,
+                                            type=AuroraDNSHealthCheckType.HTTP,
+                                            hostname='web01.auroradns.eu',
+                                            path='/',
+                                            port=80,
+                                            interval=10,
+                                            threshold=5)
+
+record = zone.create_record(name='www', type=RecordType.AAAA,
+                            data='2a00:f10:452::1',
+                            extra={'health_check_id': health_check.id})
--- a/libcloud/dns/drivers/auroradns.py
+++ b/libcloud/dns/drivers/auroradns.py
@@ -12,6 +12,9 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+"""
+AuroraDNS DNS Driver
+"""
 
 import base64
 import json
@@ -26,6 +29,7 @@ from libcloud.utils.py3 import b
 from libcloud.common.base import ConnectionUserAndKey, JsonResponse
 
 from libcloud.common.types import InvalidCredsError, ProviderError
+from libcloud.common.types import LibcloudError
 
 from libcloud.dns.base import DNSDriver, Zone, Record
 from libcloud.dns.types import RecordType, ZoneDoesNotExistError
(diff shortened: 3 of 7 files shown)


libcloud/dns/drivers/auroradns.py
```diff
@@ -140,6 +272,14 @@ def list_zones(self):
         return zones
 
     def list_records(self, zone):
+        """
```
Kami: There is no need to add docstring for methods which are the same as the ones already defined in the base class - those docstrings are automatically inherited from the parent class methods.

docs/examples/dns/auroradns/health_checks.py
```diff
@@ -0,0 +1,18 @@
...
+from libcloud.dns.providers import get_driver
+from libcloud.dns.drivers.auroradns import AuroraDNSHealthCheckType
+
+cls = get_driver(Provider.AURORADNS)
+
+driver = cls('myapikey', 'mysecret')
+
+zone = driver.get_zone('auroradns.eu')
+
+health_check = driver.create_healthcheck(zone=zone,
```
Kami: I assume this should be `ex_create_health_check`? :)

docs/examples/dns/auroradns/enable_disable_record.py
```diff
@@ -1,6 +1,5 @@
-from libcloud.dns.types import Provider
+from libcloud.dns.types import Provider, RecordType
```
Kami: We tend to use one import per line - this makes searching and grepping a bit easier.

wido:
@Kami Thank you for the quick reviews! I just force pushed a new version

libcloud/dns/drivers/auroradns.py
```diff
@@ -222,6 +358,205 @@ def update_record(self, record, name, type, data, extra):
...
+        :type  threshold: ``int``
+
+        :param ipaddress: (optional) The IP Address of the target to monitor.
+                          You can pass a empty string if this is not required.
+        :type  ipaddress: ``str``
+
+        :param enabled: (optional) If this healthcheck is enabled to run
+        :type  enabled: ``bool``
+
+        :param extra: (optional) Extra attributes (driver specific).
```
jimbobhickville, line 437: `extra` doesn't seem to be used.  Is it needed?  Are there other 'health check' drivers that use it and this is just allowing it to be passed in for consistency?

wido, line 437: That is correct. It is indeed just there for consistency with the other APIs

A more generic approach for health checks might be nice, I think that Route 53 and Google DNS also support this.

wido:
I updated the commit with tests for AuroraDNS. They were merged in #679 

Tests are passing on my system, waiting for Travis CI :)

tonybaloney:
Thanks @wido

libcloud/dns/drivers/auroradns.py
```diff
@@ -239,18 +572,34 @@ def __res_to_record(self, zone, record):
...
+                      type=record['type'],
+                      data=record['content'], zone=zone,
+                      driver=self, ttl=record['ttl'],
+                      extra=extra)
 
     def __res_to_zone(self, zone):
-        return Zone(id=zone['id'], domain=zone['name'], type=DEFAULT_ZONE_TYPE,
+        return Zone(id=zone['id'], domain=zone['name'],
+                    type=DEFAULT_ZONE_TYPE,
                     ttl=DEFAULT_ZONE_TTL, driver=self,
```
tonybaloney: here too

libcloud/dns/drivers/auroradns.py
```diff
@@ -239,18 +572,34 @@ def __res_to_record(self, zone, record):
...
+                                    type=healthcheck['type'],
+                                    hostname=healthcheck['hostname'],
+                                    ipaddress=healthcheck['ipaddress'],
+                                    health=healthcheck['health'],
+                                    threshold=healthcheck['threshold'],
+                                    path=healthcheck['path'],
+                                    interval=healthcheck['interval'],
+                                    port=healthcheck['port'],
+                                    enabled=healthcheck['enabled'],
+                                    zone=zone, driver=self)
```
tonybaloney: I don't know how this method is used. but you should use self.connection.driver. If there are multiple drivers instantiated weird things would happen with this reference!

libcloud/dns/drivers/auroradns.py
```diff
@@ -239,18 +572,34 @@ def __res_to_record(self, zone, record):
         extra['ttl'] = record['ttl']
         extra['prio'] = record['prio']
 
-        return Record(id=record['id'], name=name, type=record['type'],
-                      data=record['content'], zone=zone, driver=self,
-                      ttl=record['ttl'], extra=extra)
+        return Record(id=record['id'], name=name,
+                      type=record['type'],
+                      data=record['content'], zone=zone,
+                      driver=self, ttl=record['ttl'],
```
tonybaloney: and here

wido:
@tonybaloney Thanks for the feedback! I wasn't aware of that. Fixed that in a new commit.

tonybaloney:
thanks. merging

PR #673, opened 2016-01-06
Removed redundant checks

ryanm101:
As far as i can tell the "and" statements are pointless as the variable on the left is the same variable as the one on the right.

### DIFF
--- a/libcloud/common/openstack_identity.py
+++ b/libcloud/common/openstack_identity.py
@@ -233,10 +233,10 @@ class OpenStackServiceCatalog(object):
         endpoints = []
 
         for entry in self._entries:
-            if service_type and entry.service_type != service_type:
+            if entry.service_type != service_type:
                 continue
 
-            if name and entry.service_name != name:
+            if entry.service_name != name:
                 continue
 
             for endpoint in entry.endpoints:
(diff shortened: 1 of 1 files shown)


Kami:
I wonder if there was a reason to make the checks the way they are (I remember some debate about it long time ago, but I can't find it now).

Kami:
Looks like the tests are failing - https://travis-ci.org/apache/libcloud/builds/100561596

So there indeed was a reason to make the check look the way they are. Can you please revert your change and add a comment to the code which clarifies why this behavior is needed - this way other people won't stumble on the same thing as you did.

Thanks!

Kami:
I just checked the code again and now I remember why the checks are the way they are - we support "partial look ups: - e.g. you can only specify `service_type`, `name`, etc.

So if this argument is not provided (it has `None` value assigned to it by default) we don't want to go into the if statement body.

ryanm101:
@Kami Sorry, I don't understand, likely i am missing something basic but...

Previous code was
     if name and entry.service_name != name:
which has to be the same as
    if entry.service_name != name:
because
    if name != name:
would never equate to true

How can name not equal name when they are the same variable?

I haven't looked at the tests yet to understand why they failed but will this evening hopefully.

jimbobhickville:
The problem is `name` and `service_type` are optional parameters that default to `None`, if you look at the method signature.  This lets you pass in only one of them, neither, or both, and it does the right thing in all cases.

Kami:
@ryanm101 Yeah what @jimbobhickville said :)

ryanm101:
Cancelling pull request Added new request with comments to explain this behaviour for others

PR #674, opened 2016-01-06
Use assertRaises for asserting if subject threw an exception

relaxdiego:
This change ensures that, should a regression happen, the test
will print a more helpful failure message than "True is not False"

### DIFF
--- a/libcloud/test/compute/test_dimensiondata.py
+++ b/libcloud/test/compute/test_dimensiondata.py
@@ -43,19 +43,14 @@ class DimensionDataTests(unittest.TestCase, TestCaseMixin):
         self.driver = DimensionData(*DIMENSIONDATA_PARAMS)
 
     def test_invalid_region(self):
-        try:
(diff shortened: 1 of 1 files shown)


tonybaloney:
Thanks Mark for the contrib. LGTM. Feedback welcome for the driver too now it's getting wider usage.

relaxdiego:
Thanks, @tonybaloney. Will do that soon. Just easing my way into the code and fixing/improving stuff as I go. :-)

Kami:
Some of the unittest library features (e.g. assertRaisesRegexp) are not available in the unittest library which ships with older versions of Python so it's better and safer to use `libcloud.utils.test.unittest` which uses `unittest2` library with the older Python version.

Would also be great if someone can document that in `docs/development.rst` :)

relaxdiego:
Glad to know there's that, @Kami! Using that also allows me to use assertRaises as a context manager which is cleaner in some cases. I noticed though that it's in `libcloud.test.unittest`, not `libcloud.utils.test.unittest`. Was there a plan to move it?

PR #675, opened 2016-01-07
auroradns: Properly parse TTL as extra data

wido:
Small typo in the list caused this not to be parsed properly.


jimbobhickville:
lgtm but I don't have a commit bit, so someone else will need to check it.

wido:
Probably @Kami or @runseb ?

Kami:
Good catch :)

Test would also be nice.

wido:
I agree @Kami, a test would indeed be a good thing :)

wido:
I just added a Unit Test to verify this case. I'll add more later in a different PR

libcloud/test/dns/test_auroradns.py
```diff
@@ -0,0 +1,58 @@
...
+
+class AuroraDNSDriverTests(LibcloudTestCase):
+
+    def setUp(self):
+        AuroraDNSDriver.connectionCls.conn_classes = \
+            (None, AuroraDNSDriverMockHttp)
+        AuroraDNSDriverMockHttp.type = None
+        self.driver = AuroraDNSDriver(*DNS_PARAMS_AURORADNS)
+
+    def test_merge_extra_data(self):
```
Kami, line 33: I was more thinking about adding a test case which uses response fixture, but that works for now.

Kami:
Tests are failing (https://travis-ci.org/apache/libcloud/jobs/101036033), need to add aurora params to `libcloud/test/secrets.py-test` :)

wido:
Fixed! Tests are no longer failing.

Kami:
Merged, thanks!

PR #677, opened 2016-01-08
Use libcloud unittest

relaxdiego:
This changeset includes two patches:
1. Modify Dimension Data tests again to use `libcloud.test.unittest` which contains logic for deciding the appropriate package version that works across Python versions. This allows us to use assertRaises as a context manager which makes the test cleaner.
2. Add an entry in `docs/development.rst` to recommend using `libcloud.test.unittest`. This is based on a suggestion from @Kami in PR #674.

### DIFF
--- a/libcloud/test/backup/test_dimensiondata.py
+++ b/libcloud/test/backup/test_dimensiondata.py
@@ -14,13 +14,12 @@
 # limitations under the License.
 
 import sys
-import unittest
 from libcloud.utils.py3 import httplib
 
 from libcloud.common.types import InvalidCredsError
 from libcloud.backup.drivers.dimensiondata import DimensionDataBackupDriver as DimensionData
 
-from libcloud.test import MockHttp
+from libcloud.test import MockHttp, unittest
 from libcloud.test.backup import TestCaseMixin
 from libcloud.test.file_fixtures import BackupFileFixtures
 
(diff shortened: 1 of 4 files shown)


tonybaloney:
Good work Mark. I'd like your eyes on #666 if you get a minute. I'm sure the tests need fixing up in the same way so I will get to that. 
This change set looks good. Getting these tests and fixtures together was a big effort as you can probably tell!

relaxdiego:
Thanks, @tonybaloney! I'll look at that PR soon. And woah on the number! ;-)

Kami:
Thanks :+1:

libcloud/test/backup/test_dimensiondata.py
```diff
@@ -35,19 +34,13 @@ def setUp(self):
...
-        try:
+        with self.assertRaises(ValueError):
             self.driver = DimensionData(*DIMENSIONDATA_PARAMS, region='blah')
-        except ValueError:
-            pass
 
     def test_invalid_creds(self):
         DimensionDataMockHttp.type = 'UNAUTHORIZED'
-        try:
+        with self.assertRaises(InvalidCredsError):
```
Kami, line 42: I'm fine with using a decorator, but another way of using this method is by passing callable + arguments to it and this method will call your function.

For example:

``` python
self.assertRaises(ValueError, DimensionData, *DIMENSIONDATA_PARAMS, region='blah')
self.assertRaises(ValueError, self.driver.list_targets)
```

Kami, line 42: On a related note, where more general exceptions such as `ValueError` are thrown, it's also a good idea to assert on the exception message (if possible).

For example:

``` python
expected_msg = 'Invalid region: blah'
self.assertRaisesRegexp(ValueError, expected_msg, DimensionData, *DIMENSIONDATA_PARAMS, region='blah')
```

relaxdiego, line 42: Thanks for the ValueError tip, @Kami. That's actually the original reason why I used assertRaises as a context manager: so that I can inspect the error message while maintaining test code that looks exactly like how one would use the API in production:

``` python
with self.assertRaises(ValueError) as err:
    DimensionData(*DIMENSIONDATA_PARAMS, region='blah')

self.assertEqual('Invalid region: blah', str(err.exception))
```

Unfortunately, I forgot to add the error message assertion. Heh. I'll add that in my next PR.

PR #678, opened 2016-01-09
Softlayer - Run script in process create node and floating_ip Open Stack

ghost:
Added post_uri parameter to the SoftLayer driver. Run script in process create node

### DIFF
--- a/libcloud/compute/drivers/softlayer.py
+++ b/libcloud/compute/drivers/softlayer.py
@@ -342,7 +342,7 @@ class SoftLayerNodeDriver(NodeDriver):
             'hourlyBillingFlag': hourly,
             'operatingSystemReferenceCode': os,
             'localDiskFlag': local_disk,
-            'postInstallScriptUri':post_uri,
+            'postInstallScriptUri': post_uri,
             'blockDevices': [
                 {
                     'device': '0',
--- a/libcloud/compute/drivers/softlayer.py
+++ b/libcloud/compute/drivers/softlayer.py
@@ -308,7 +308,7 @@ class SoftLayerNodeDriver(NodeDriver):
         if kwargs.get('ex_local_disk') is False:
             local_disk = 'false'
 
-        post_uri=''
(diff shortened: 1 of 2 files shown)


libcloud/compute/drivers/softlayer.py
```diff
@@ -306,6 +308,10 @@ def create_node(self, **kwargs):
         if kwargs.get('ex_local_disk') is False:
             local_disk = 'false'
 
+        post_uri = ''
```
tonybaloney, line 311: can post_uri be submitted with an empty value? Are there any tests to support this?

tonybaloney:
thanks for the contribution, post_install_script_uri might be a more obvious kwarg name, but thats no biggie. I've added a note with regards to testing absent values.

libcloud/compute/drivers/softlayer.py
```diff
@@ -278,6 +278,8 @@ def create_node(self, **kwargs):
         :type       ex_os: ``str``
         :keyword    ex_keyname: The name of the key pair
         :type       ex_keyname: ``str``
+        :keyword    post_uri: Post-install script to run after reload
```
Kami, line 281: It's an extension (provider specific / non-standard) argument, so please prefix it with `ex_`.

Kami, line 281: Some more documentation for this argument (and ideally example in the docs) would also be nice.

Right now it's not really clear what the value should contain - it seems like it should be a full URL to a file or something else?

Kami:
+1 on @tonybaloney's comment and thanks for the contribution :)

tonybaloney:
@onecloudportal @marcydoty @jvitorsf please see previous comment, if you can please make the changes I'd be happy to merge.

sercanyaldiz:
Hi when is this request merged?

sercanyaldiz:
@tonybaloney, is it possible to assign the public/private vlan during the vs creation?

PR #679, opened 2016-01-11
auroradns: Further improve unit tests

wido:


wido:
This contains the fix from my previous pr @runseb 

It is also covered by a UnitTest

wido:
The tests all pass, can we merge this one? It only adds unit tests

libcloud/test/dns/test_auroradns.py
```diff
@@ -47,17 +54,180 @@ def test_merge_extra_data(self):
...
+        self.assertEqual(len(zones), 2)
+
+    def test_create_zone(self):
+        zone = self.driver.create_zone('example.com')
+        self.assertEquals(zone.domain, 'example.com')
+
+    def test_get_zone(self):
+        zone = self.driver.get_zone('example.com')
+        self.assertEquals(zone.domain, 'example.com')
+        self.assertEquals(zone.id, 'ffb62570-8414-4578-a346-526b44e320b7')
```
sebgoa, line 74: I am going to merge it, but assertEquals is deprecated so its gives several warnings to use assertEqual instead.
I will let you fix that in the subsequent PR

PR #682, opened 2016-01-14
Throw a more friendly error message if establishing SSL / TLS connection fails

Kami:
With this change we now throw a more user-friendly error message in case establishing SSL / TLS connection fails.

New error message:

``` python
  File "/home/kami/w/lc/libcloud/libcloud/httplib_ssl.py", line 312, in connect
    raise new_e
socket.error: Failed to establish SSL / TLS connection ([Errno 104] Connection reset by peer). It is possible that the server doesn't support requested SSL / TLS version (TLS v1.0).
For information on how to work around this issue, please see https://libcloud.readthedocs.org/en/latest/other/ssl-certificate-validation.html#changing-used-ssl-tls-version
```

Related to https://issues.apache.org/jira/browse/LIBCLOUD-791

### DIFF
--- a/libcloud/httplib_ssl.py
+++ b/libcloud/httplib_ssl.py
@@ -299,8 +299,8 @@ class LibcloudHTTPSConnection(httplib.HTTPSConnection, LibcloudBaseConnection):
                 ca_certs=self.ca_cert,
                 ssl_version=libcloud.security.SSL_VERSION)
         except socket.error:
-            e = sys.exc_info()[1]
-            exc_msg = str(e)
+            exc = sys.exc_info()[1]
+            exc_msg = str(exc)
 
             # Re-throw an exception with a more friendly error message
             if 'connection reset by peer' in exc_msg.lower():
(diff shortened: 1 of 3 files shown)


libcloud/httplib_ssl.py
```diff
@@ -272,12 +290,31 @@ def connect(self):
...
-                                    ssl_version=libcloud.security.SSL_VERSION)
+        try:
+            self.sock = ssl.wrap_socket(
+                sock,
+                self.key_file,
+                self.cert_file,
+                cert_reqs=ssl.CERT_REQUIRED,
+                ca_certs=self.ca_cert,
+                ssl_version=libcloud.security.SSL_VERSION)
+        except Exception:
```
jimbobhickville: Shouldn't we limit this to the specific exception type(s) that might exhibit this behavior?  Looked like it was `socket.error` in the JIRA.

Kami: Yeah, that's a good idea - I will make that change.

Kami: Done.

Also committed some tests (previously I forgot to add the test file).

PR #683, opened 2016-01-16
[WIP] [DON'T MERGE] Use ssl.PROTOCOL_SSLv23 constant by default under Python >= 2.7.9 < 3 and >= 3.4

Kami:
Note: This pull request is W.I.P. and not yet ready to be reviewed and merged.

This pull request updates the code to use `ssl.PROTOCOL_SSLv23` constant for `ssl_version` attribute when establishing SSL / TLS connection when using Python >= 2.7.9 < 3 and Python >= 3.4.

This constant offers the best security and compatibility since it will pick between TLS v1.0, TLS v1.1 and TLS v1.2 depending on the version supported / requested by the server. This is only done under aforementioned Python versions since un-secure SSL v3.0 is disabled by default in those versions.

In other versions where SSL v3.0 is not disabled we still use TLS v1.0 (`ssl.PROTOCOL_TLSv1`) which means nothing has changed - that's the same as the old / existing behavior.

Another thing to keep in mind is that user can also explicitly specify which version they want to use using `libcloud.security.SSL_VERSION` variable and this value has precedence over dynamically obtained values.

Ideally, if users know that the server supports TLS v1.2 they should explicitly specify and use that aka the highest supported versions (that's also mentioned in the documentation).

### DIFF
--- a/libcloud/httplib_ssl.py
+++ b/libcloud/httplib_ssl.py
@@ -12,10 +12,12 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
 """
 Subclass for httplib.HTTPSConnection with optional certificate name
 verification, depending on libcloud.security settings.
 """
+
 import os
 import sys
 import socket
@@ -24,6 +26,7 @@ import base64
 import warnings
 
 import libcloud.security
+from libcloud.security import get_ssl_version
 from libcloud.utils.py3 import b
 from libcloud.utils.py3 import httplib
 from libcloud.utils.py3 import urlparse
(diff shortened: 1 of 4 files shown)


Kami:
I'm still contemplating if it's worth to finish this or if we should just switch to `requests`

Switching to `requests` is of course quite a bigger and potentially breaking in some places change, but I don't want us to maintain security critical code.

jimbobhickville:
I'd vote for moving to requests.  Let libcloud focus on the multi-cloud aspect and let requests handle the http heavy lifting.  We'd have to drop Python 2.5 support, but I think that's ok.

tonybaloney:
we don't support py 2.5 anymore @jimbobhickville so thats ok.

Kami:
I'm closing this.

Hopefully we can move to requests in the near future and we don't need this code anymore.

PR #684, opened 2016-01-26
EC2 support for AWS Seoul region

encyphered:
- Fixed signature version of AWS Seoul region to 4 due to that does not support
- Added provider and driver for region.

### DIFF
--- a/libcloud/compute/drivers/ec2.py
+++ b/libcloud/compute/drivers/ec2.py
@@ -1042,7 +1042,7 @@ REGION_DETAILS = {
         'endpoint': 'ec2.ap-northeast-2.amazonaws.com',
         'api_name': 'ec2_ap_northeast',
         'country': 'South Korea',
-        'signature_version': '2',
+        'signature_version': '4',
         'instance_types': [
             'c4.large',
             'c4.xlarge',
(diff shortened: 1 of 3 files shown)


tonybaloney:
Thanks @encyphered do you have instances deployed in Seoul? Amazons press release said a subset of instance sizes for EC2 was supported, is this correct?

encyphered:
@tonybaloney Yes, Amazon announced only T2, M4, C4, I2, D2, and R3 was supported. HS1 is not supported

tonybaloney:
thanks. LGTM +1

Kami:
Thanks for the PR.

Does the new region only support new signature version 4? I'm asking because right now, IIRC, we only support signature version 4 for HTTP GET methods in Libcloud which means we should use 2 if it's still supported (otherwise non GET operations won't work).

tonybaloney:
I added POST request support for v4 about a month ago 

Sent from Outlook Mobile

On Wed, Jan 27, 2016 at 1:55 AM -0800, "Tomaz Muraus" notifications@github.com wrote:

Thanks for the PR.

Does the new region only support new signature version 4? I'm asking because right now, IIRC, we only support signature version 4 for HTTP GET methods in Libcloud which means we should use 2 if it's still supported (otherwise non GET operations won't work).

Kami:
@tonybaloney Ah, nice! :)

I totally missed that.

PR #687, opened 2016-01-28
S3 support for AWS Seoul region

encyphered:
Support S3 Seoul region. In addition, this PR includes modification about Amazon signature V4 due to Seoul region does not support signature V2.

Modification that S3 requires:
- Removed HTTP method restriction from AWSRequestSignerAlgorithmV4
- Added hash of the request payload as X-AMZ-Content-SHA256 header
- Prevent Host / Accept-Encoding header duplication

### DIFF
--- a/libcloud/storage/drivers/s3.py
+++ b/libcloud/storage/drivers/s3.py
@@ -35,8 +35,8 @@ from libcloud.utils.xml import fixxpath, findtext
 from libcloud.utils.files import read_in_chunks
 from libcloud.common.types import InvalidCredsError, LibcloudError
 from libcloud.common.base import ConnectionUserAndKey, RawResponse
-from libcloud.common.aws import AWSBaseResponse, AWSDriver, AWSTokenConnection, \
-    SignedAWSConnection
+from libcloud.common.aws import AWSBaseResponse, AWSDriver, \
+    AWSTokenConnection, SignedAWSConnection
 
 from libcloud.storage.base import Object, Container, StorageDriver
 from libcloud.storage.types import ContainerError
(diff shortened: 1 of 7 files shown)


libcloud/storage/drivers/s3.py
```diff
@@ -35,7 +35,8 @@
 from libcloud.utils.files import read_in_chunks
 from libcloud.common.types import InvalidCredsError, LibcloudError
 from libcloud.common.base import ConnectionUserAndKey, RawResponse
-from libcloud.common.aws import AWSBaseResponse, AWSDriver, AWSTokenConnection
+from libcloud.common.aws import AWSBaseResponse, AWSDriver, AWSTokenConnection, \
```
tonybaloney: You need to shorten this line further.

libcloud/test/common/test_aws.py
```diff
@@ -219,6 +225,16 @@ def test_get_payload_hash_returns_digest_of_empty_string_for_GET_requests(self):
         self.assertEqual(self.signer._get_payload_hash(method='GET'),
                          'e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855')
 
+    def test_get_payload_hash_with_data_for_PUT_requests(self):
+        SignedAWSConnection.method = 'PUT'
+        self.assertEqual(self.signer._get_payload_hash(method='POST', data='DUMMY'),
```
tonybaloney: IS the method still supposed to be POST here? If so, a comment to explain why would be welcome

tonybaloney:
Hi @encyphered this PR is missing a small change before it can be merged. please see the build status light.

tonybaloney:
Thanks @encyphered LGTM :+1:

PR #691, opened 2016-02-06
DimensionData: changing image to be NodeImage or str, upping code coverage

jadunham1:


tonybaloney:
Great contribution thanks @jadunham1

PR #693, opened 2016-02-06
DimensionData: Adding new exception code/message, add backup tests

jadunham1:
Different type of BadResponse XML request happens when accessing some of the backups API
Added in ability to grab message/code from it as well as make it easier to add new elements if they come to fruition.
Added negative tests that failed before the code change and succeed after.


tonybaloney:
thanks @jadunham1 the backup driver definitely needs more tests. let me know how you get on. 
merging

PR #695, opened 2016-02-08
DimensionData: Adding in new methods/tests for backups

jadunham1:
Added in ability to add clients to a target
Added in ability to list backup info for a target
Added in positive tests


libcloud/backup/drivers/dimensiondata.py
```diff
@@ -384,6 +421,73 @@ def cancel_target_job(self, target, job):
...
+        :param target: Backup target with the backup data
+        :type  target: Instance of :class:`BackupTarget` or ``str``
+
+        :param client: Client to add to the target
+        :type  client: ``str``
+
+        :param storage_policy: The storage policy for the client
+        :type  storage_policy: ``str``
+
+        :param schedule_policy: The storage policy for the client
```
tonybaloney: should be "The schedule policy for the client"

libcloud/backup/drivers/dimensiondata.py
```diff
@@ -384,6 +421,73 @@ def cancel_target_job(self, target, job):
...
+
+        if isinstance(target, BackupTarget):
+            server_id = target.address
+        else:
+            server_id = target
+
+        backup_elm = ET.Element('NewBackupClient',
+                                {'xmlns': BACKUP_NS})
+        ET.SubElement(backup_elm, "type").text = client
+        ET.SubElement(backup_elm, "storagePolicyName").text = storage_policy
```
tonybaloney: Is the client code to fetch the storage policies and schedule policies implemented? It would be preferable to send the storagepolicy response from that client call instead of the name as a string.

jadunham1: I like the idea of passing both the obj and the string as done elsewhere in the DD client.  I'll make this change.

tonybaloney: :+1: The list of potential storage policies alters based on the client, location and the application type :8ball:

jadunham1:
I'm unhappy with how the classes for backups objects are directly defined here instead of in the common library.  I'm going to make this change as well.

jadunham1:
Added negative tests around the info call if backups don't exist.
Moved the classes to common where the classes for the node exist.
Added a bunch of docs on the classes themselves.

tonybaloney:
the additional changes look great. thanks @jadunham1 merging.

PR #696, opened 2016-02-09
Backblaze Storage Driver : Include get_container and get_object methods

jshridha:
Include get_container and  get_object methods in the BackblazeB2StorageDriver

This is required for compatibility with libraries that reference these methods.

@Kami Since you wrote the initial backblaze implementation (I know it's still under dev), it would be great if you could review this. Thanks!

### DIFF
--- a/libcloud/storage/drivers/backblaze_b2.py
+++ b/libcloud/storage/drivers/backblaze_b2.py
@@ -362,6 +362,8 @@ class BackblazeB2StorageDriver(StorageDriver):
         upload_host = parsed_url.netloc
         request_path = parsed_url.path
 
+        if isinstance(data, str):
+            data = bytearray(data)
         response = self.connection.upload_request(action=request_path,
                                                   headers=headers,
                                                   upload_host=upload_host,
--- a/libcloud/storage/drivers/backblaze_b2.py
+++ b/libcloud/storage/drivers/backblaze_b2.py
@@ -30,6 +30,7 @@ from libcloud.utils.py3 import httplib
 from libcloud.utils.py3 import urlparse
 from libcloud.utils.files import read_in_chunks
 from libcloud.utils.files import exhaust_iterator
+from libcloud.utils.escape import sanitize_object_name
 
 from libcloud.common.base import ConnectionUserAndKey
 from libcloud.common.base import JsonResponse
@@ -316,7 +317,7 @@ class BackblazeB2StorageDriver(StorageDriver):
                                 success_status_code=httplib.OK)
 
     def upload_object(self, file_path, container, object_name, extra=None,
-                      verify_hash=True, headers=None, iterator=None):
+                      verify_hash=True, headers=None):
         """
         Upload an object.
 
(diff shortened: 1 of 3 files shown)


tonybaloney:
Great contribution. is it possible to remove the dependency on the os stdlib?

tonybaloney:
Oh, and please can you develop and add some tests for this as well. We don't have CI tests for Windows with Travis CI so unittests as a minimum.

jshridha:
Can do! I just added in some unit tests, got rid of the os and sys dependencies, reformatted according style guide a bit better, and rebased the pull request. Let me know if there's any other comments, I'm happy to make changes.

libcloud/storage/drivers/backblaze_b2.py
```diff
@@ -292,7 +316,7 @@ def download_object_as_stream(self, obj, chunk_size=None):
                                 success_status_code=httplib.OK)
 
     def upload_object(self, file_path, container, object_name, extra=None,
-                      verify_hash=True, headers=None):
+                      verify_hash=True, headers=None, iterator=None):
```
tonybaloney: @Kami is that ok?

Kami:
Sorry for the delay and thanks for the contribution.

libcloud/storage/drivers/backblaze_b2.py
Kami: There is already a method for that - `upload_object_via_stream`. That method takes an iterator, while `upload_object` takes a file path.

The reason I didn't implement `upload_object_via_stream` is because Backblaze API doesn't support streaming / multi part uploads. This means we need to buffer the whole file content in memory which is not ideal (especially for large files) and implementing `upload_object_via_stream` doesn't provide much value besides convenience if you already have an iterator object.

Having said that, I'm fine with implementing `upload_object_via_stream` for convenience, we just need to document and make it clear that this method does not perform streaming upload and that the whole file is read and buffered in memory.

libcloud/storage/drivers/backblaze_b2.py
```diff
@@ -301,8 +325,13 @@ def upload_object(self, file_path, container, object_name, extra=None,
         # Note: We don't use any of the base driver functions since Backblaze
         # API requires you to provide SHA1 has upfront and the base methods
         # don't support that
-        with open(file_path, 'rb') as fp:
-            iterator = iter(fp)
+        object_name = object_name.replace('\\', '/')
```
Kami: Might be a good idea to put this into a utility method - `escape/sanitize_object_name` or similar.

libcloud/storage/drivers/backblaze_b2.py
Kami: Edit: I see you also implemented `upload_object_via_stream` below. Adding `iterator` argument to this method breaks the standard API so one thing we could do is move common functionality into a new "private" method and then utilize this private method in `upload_object` and `upload_object_via_stream`.

jshridha: @Kami I just noticed backblaze has a sneak peak of their new api method - b2_upload_part

Would it be better to implement [this](https://www.backblaze.com/b2/docs/b2_upload_part.html) as the upload_object_via_stream method?

jshridha: I just tried to implement the large file upload and it is not yet supported by backblaze. I'm going to go forward with this temporary `upload_object_via_stream` method making the changes you suggested.

Kami:
@jshridha Thanks.

Next time please ping us when you push updates since we don't get notified when you push new changes :)

Kami:
I've pushed some minor fixes / improvements and merged the patch into trunk (I also verified the patch is working).

Thanks!

jshridha:
@Kami Will do, I didn't realize that new pushes didn't result in notifications. Also, thanks for the great review!

PR #697, opened 2016-02-09
ex_creation_time method for ec2 and digitalocean

vdloo:
ec2 nodes keep the created time string in 'launch_time' and digital ocean nodes keep it in 'created_at'. ex_get_creation_time is a method that takes a node as the argument returns the created time string


tonybaloney:
Thanks for your contribution @vdloo this looks great. :+1:

PR #698, opened 2016-02-10
Add created datetime to node

allardhoeve:
Instead of the implementation in #697, I'd like to propose this instead. 

The problem with #697 is that when you want to find a list of old nodes when you have a list of `Node` objects, you'd need to write:

``` python
old_nodes = [n for n in nodes if n.driver.ex_creation_time(n) > too_old]
```

Yuk, why is that driver in there? Better would be to write:

``` python
old_nodes = [n for n in nodes if n.created > too_old]
```

I have implemented this code for OpenStack, Digital Ocean and EC2. Other drivers, I have no access to. In those cases, the behaviour is to have `created` be `NoneType`.

### DIFF
--- a/libcloud/compute/base.py
+++ b/libcloud/compute/base.py
@@ -167,7 +167,7 @@ class Node(UuidMixin):
     """
 
     def __init__(self, id, name, state, public_ips, private_ips,
-                 driver, size=None, image=None, extra=None):
+                 driver, size=None, image=None, extra=None, created_at=None):
         """
         :param id: Node ID.
         :type id: ``str``
@@ -193,6 +193,9 @@ class Node(UuidMixin):
         :param image: Image of this node. (optional)
         :type size: :class:`.NodeImage`
 
+        :param created_at: The datetime this node was created (optional)
+        :type created_at: :class: `datetime.datetime`
+
         :param extra: Optional provider specific attributes associated with
                       this node.
         :type extra: ``dict``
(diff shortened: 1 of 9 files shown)


allardhoeve:
@tonybaloney, @vdloo

tonybaloney:
this is a good addition to the base class. I agree with you @allardhoeve it is a good design. :+1:

libcloud/compute/base.py
```diff
@@ -167,7 +167,7 @@ class Node(UuidMixin):
     """
 
     def __init__(self, id, name, state, public_ips, private_ips,
-                 driver, size=None, image=None, extra=None):
+                 driver, size=None, image=None, created=None, extra=None):
```
Kami: I would prefer the attribute to be called `created_at`.

Kami:
It's also worth noting that this is a backward incompatible change for any place which doesn't use keyword arguments so we need to make sure it doesn't break any of the existing drivers and document it in the upgrade notes.

allardhoeve:
I could put the new keyword after the `extra=None`, that way it breaks nothing. But yes, it should be in the Changelog

vdloo:
I agree, this is better than using the driver and passing the node object in  :+1:

allardhoeve:
Okey, I changed the field name to `created_at` and the order of the named arguments. If everyone is on board, I'd like to commit this and revert #697.

tonybaloney:
:+1: from me

PR #699, opened 2016-02-11
DimensionData: Location to string conversion/tests/small refactor around ex_list_networks

jadunham1:
Updating some documentation
Added ex_id to filter locations to list_locations
cleaned up a function ex_network_locations that the location filter was not working and was essentially duplicate code.
Added some unit tests for more coverage while I was mucking around

### DIFF
--- a/libcloud/common/dimensiondata.py
+++ b/libcloud/common/dimensiondata.py
@@ -18,8 +18,8 @@ Dimension Data Common Components
 from base64 import b64encode
 from time import sleep
 from libcloud.utils.py3 import httplib
-from libcloud.utils.py3 import b
-
+from libcloud.utils.py3 import b, basestring
+from libcloud.compute.base import NodeLocation
 from libcloud.common.base import ConnectionUserAndKey, XmlResponse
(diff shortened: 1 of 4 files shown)


libcloud/common/dimensiondata.py
```diff
@@ -95,6 +95,29 @@
 )
 
 
+def location_to_location_id(location):
```
tonybaloney, line 98: should this be _location_to_location_id? Do you want it to be public?

libcloud/compute/drivers/dimensiondata.py
```diff
@@ -37,6 +37,7 @@
 from libcloud.common.dimensiondata import API_ENDPOINTS, DEFAULT_REGION
 from libcloud.common.dimensiondata import TYPES_URN
 from libcloud.common.dimensiondata import SERVER_NS, NETWORK_NS, GENERAL_NS
+from libcloud.common.dimensiondata import location_to_location_id
```
tonybaloney, line 40: I just answered my own question. ignore me

tonybaloney:
:+1:

PR #700, opened 2016-02-11
Expand Dimension Data regions, add new drivers

tonybaloney:
Added full list of Dimension Data regions and partners. Added drivers for BSNL, Med-1, Indosat, Cisco CCS, NTT America, Internet Solutions and documentation to boot


tonybaloney:
Passed. Merging

PR #701, opened 2016-02-12
Handle deprecated providers with a error message and a URL to follow up

tonybaloney:
See - http://libcloud.apache.org/blog/2016/02/16/new-drivers-deprecated-drivers.html


tonybaloney:
@Kami @erjohnso @jcsalterego @runseb @allardhoeve - this is an important one. Are you happy with the way this works?

allardhoeve:
Looks simple enough to me.

erjohnso:
LGTM @tonybaloney. Nice clean up!

Kami:
Yay, removing code is always great (less stuff to maintainer, less surface area for bugs, etc.) :P

LGTM

libcloud/compute/providers.py
```diff
@@ -190,6 +181,12 @@
 
 
 def get_driver(provider):
+    if provider in DEPRECATED_DRIVERS:
```
Kami, line 184: A test case for this would be nice :)

libcloud/compute/providers.py
```diff
@@ -190,6 +181,12 @@
...
+        url = DEPRECATED_DRIVERS[provider]['url']
+        reason = DEPRECATED_DRIVERS[provider]['reason']
+        msg = ('Provider no longer supported: %s, please visit: %s' %
+               (url, reason))
```
Kami, line 188: Also, looks like it should be `% (reason, url)` - reason first.

PR #702, opened 2016-02-12
DimensionData: Adding ability to remove a backup clients + small updates

jadunham1:
Added a new method ex_remove_client_from_target
After comment from @tonybaloney noticed that the translation functions should be privatized instead, can't find any use outside of the classes currently, so I privatized them.  Helps the code look cleaner.
Added both positive/negative test coverage around new backup call.
Added test coverage around privatized translation functions

### DIFF
--- a/libcloud/backup/drivers/dimensiondata.py
+++ b/libcloud/backup/drivers/dimensiondata.py
@@ -215,12 +215,13 @@ class DimensionDataBackupDriver(BackupDriver):
         Delete a backup target
 
         :param target: Backup target to delete
-        :type  target: Instance of :class:`BackupTarget`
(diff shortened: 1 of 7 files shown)


tonybaloney:
:+1: LGTM

tonybaloney:
need to investigate the use of basestring. the libcloud one isn't a superclass of unicode and str like it is in the std library. which is annoying. @Kami is that on purpose?

jadunham1:
I'm also not a huge fan as I have to make sure it isn't unicode when going in.  It was what we were using as conventions before so I kept with it as the code was this way.

tonybaloney:
The tests are looking great now @jadunham1

PR #704, opened 2016-02-12
Add support for adding a family to an image.

illfelder:
The feature is currently in beta, and will allow users to reference the latest, non-deprecated, image for a given family via the API.

### DIFF
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -97,8 +97,7 @@ class GCEConnection(GoogleBaseConnection):
                                             auth_type=auth_type,
                                             credential_file=credential_file,
                                             **kwargs)
-        self.request_path = '/compute/%s/projects/%s' % (API_VERSION,
-                                                         project)
+        self.request_path = '/compute/%s/projects/%s' % (API_VERSION, project)
         self.gce_params = None
 
     def pre_connect_hook(self, params, headers):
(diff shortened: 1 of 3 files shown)


tonybaloney:
Hi. Changes look good. So the feature in GCE is beta? Any idea when it will be completed? @erjohnso

illfelder:
The feature is already in beta, but our plan for releasing it into the v1 API includes making third party client updates. You can view the documentation for a new "getFromFamily" API endpoint here: https://cloud.google.com/compute/docs/reference/beta/images/getFromFamily

Image publishers need to create new images with the family property set before the feature will be generally useful. In order to begin the transition, we are making API client updates that enable using the feature during image creation.

tonybaloney:
Understood. Wouldn't want to introduce a chicken & egg then. Changes LGTM :+1:

libcloud/compute/drivers/gce.py
```diff
@@ -2135,11 +2141,13 @@ def ex_create_image(self, name, volume, description=None,
         image_data = {}
         image_data['name'] = name
         image_data['description'] = description
+        image_data['family'] = family
         if isinstance(volume, StorageVolume):
             image_data['sourceDisk'] = volume.extra['selfLink']
             image_data['zone'] = volume.extra['zone'].name
-        elif isinstance(volume, str) and \
-                volume.startswith('https://') and volume.endswith('tar.gz'):
+        elif (isinstance(volume, str) and
```
tonybaloney, line 2148: should make a todo here about unicode support wrt, Python 2/3

PR #708, opened 2016-02-19
DimensionData adding disk configurations to node object

jadunham1:
Adding new class DimensionDataServerDisk
Adding assert to cover that a disk is present

### DIFF
--- a/libcloud/common/dimensiondata.py
+++ b/libcloud/common/dimensiondata.py
@@ -602,6 +602,41 @@ class DimensionDataServerCpuSpecification(object):
                 % (self.cpu_count, self.cores_per_socket, self.performance))
 
 
+class DimensionDataServerDisk(object):
+    """
+    A class that represents the disk on a server
+    """
+    def __init__(self, id, scsiId, sizeGb, speed, state):
+        """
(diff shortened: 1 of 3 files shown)


tonybaloney:
@jadunham1 this def needs tests against both MCP 1 and MCP 2 locations, I don't know if the response formats are the same.

jadunham1:
Disk looks the same from both MCP1/2, added the test.

libcloud/common/dimensiondata.py
```diff
@@ -602,6 +602,41 @@ def __repr__(self):
...
+        :type  sizeGb: ``int``
+
+        :param speed: Speed of the disk (i.e. STANDARD)
+        :type  speed: ``str``
+
+        :param state: State of the disk (i.e. PENDING)
+        :type  state: ``str``
+        """
+        self.id = id
+        self.scsiId = scsiId
```
tonybaloney: these should really be snake case for consistency.

jadunham1: You're right, not sure what I was thinking, will change

tonybaloney:
Fixed conflicts and merged.

PR #709, opened 2016-02-20
Feature/dd node wait for state

jadunham1:
Adding in more node states for dimension data with fall back to the old method if we can't match
Adding tests for node states
Making it so Node objects wait_for_state looks at state instead of status
Add test for Node object in wait for state

### DIFF
--- a/libcloud/compute/drivers/dimensiondata.py
+++ b/libcloud/compute/drivers/dimensiondata.py
@@ -46,15 +46,26 @@ from libcloud.compute.types import NodeState, Provider
 # These tuples represent:
 # (<state_of_node_from_didata>, <is node started?>, <action happening>)
 NODE_STATE_MAP = {
-    ('NORMAL', 'true', None): NodeState.RUNNING,
-    ('NORMAL', 'false', None): NodeState.STOPPED,
-    ('PENDING_CHANGE', 'true', 'START_SERVER'): NodeState.STARTING,
-    ('PENDING_ADD', 'true', 'DEPLOY_SERVER'): NodeState.STARTING,
-    ('PENDING_CHANGE', 'true', 'SHUTDOWN_SERVER'): NodeState.STOPPING,
-    ('PENDING_CHANGE', 'true', 'REBOOT_SERVER'): NodeState.REBOOTING,
-    ('PENDING_CHANGE', 'true', 'RESET_SERVER'): NodeState.REBOOTING,
-    ('PENDING_CHANGE', 'true', 'RECONFIGURE_SERVER'): NodeState.RECONFIGURING,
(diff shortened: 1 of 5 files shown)


libcloud/compute/types.py
```diff
@@ -222,25 +222,31 @@ class NodeState(Type):
     Standard states for a node
 
     :cvar RUNNING: Node is running.
+    :cvar STARTING: Node is starting up.
     :cvar REBOOTING: Node is rebooting.
     :cvar TERMINATED: Node is terminated. This node can't be started later on.
+    :cvar STOPPING: Node is currently trying to stop.
```
Kami, line 228: I'm fine with adding new starting and stopping state, but not too sure about reconfiguring.

I know some load balancer as a service offerings expose a similar state, but I'm not aware of too many compute providers. In any case, not a big deal.

Kami:
Thanks.

The changes look good to me, but I will wait for @tonybaloney.

libcloud/compute/types.py
jadunham1, line 228: I was up in the air on it too, thought about maybe using PENDING for a similar effect. It won't affect my waiting for state which is what this change is trying to bring. I was just trying to be as specific as possible on node states.
I'll wait for consensus from @tonybaloney

libcloud/compute/drivers/dimensiondata.py
```diff
@@ -42,6 +42,32 @@
...
+        NodeState.STOPPING,
+    ('PENDING_CHANGE', 'true', 'POWER_OFF_SERVER'):
+        NodeState.STOPPING,
+    ('PENDING_CHANGE', 'true', 'REBOOT_SERVER'):
+        NodeState.REBOOTING,
+    ('PENDING_CHANGE', 'true', 'RESET_SERVER'):
+        NodeState.REBOOTING,
+    ('PENDING_CHANGE', 'true', 'RECONFIGURE_SERVER'):
+        NodeState.RECONFIGURING,
+}
```
tonybaloney, line 69: Full state list is "NORMAL",  "PENDING_ADD", "PENDING_CHANGE", "PENDING_DELETE",  "FAILED_ADD", "FAILED_CHANGE", "FAILED_DELETE" and  "REQUIRES_SUPPORT". 
REQUIRES_SUPPORT is my favorite. Can you map those too please

tonybaloney:
Add some extra states to the list (optional). But otherwise @jadunham1 
thumbs up

jadunham1:
As I can't test all of these in what they look like (do they have workflows associated ect?) I'd rather push as is and people can add to define states.  Things like failed adds I'm not sure I can simulate correctly.
We're definitely in a better position to react to new states ect in this model.

tonybaloney:
Agreed. :+1:

PR #711, opened 2016-02-23
Throw a more user-friendly exception on "No address associated with hostname" error

Kami:
Usually this issue represents a bug in the driver (`host` connection class attribute being set to an incorrect value), but the original exception is kinda cryptic so it's hard for the users to figure out what is going on.

Before:

``` python
  File "example.py", line 5, in <module>
    print driver.list_zones()
  File "/home/kami/w/lc/libcloud/libcloud/dns/drivers/godaddy.py", line 146, in list_zones
    '/v1/domains/').object
  File "/home/kami/w/lc/libcloud/libcloud/common/base.py", line 827, in request
    headers=headers)
  File "/usr/lib64/python2.7/httplib.py", line 1053, in request
    self._send_request(method, url, body, headers)
  File "/usr/lib64/python2.7/httplib.py", line 1093, in _send_request
    self.endheaders(body)
  File "/usr/lib64/python2.7/httplib.py", line 1049, in endheaders
    self._send_output(message_body)
  File "/usr/lib64/python2.7/httplib.py", line 893, in _send_output
    self.send(msg)
  File "/usr/lib64/python2.7/httplib.py", line 855, in send
    self.connect()
  File "/home/kami/w/lc/libcloud/libcloud/httplib_ssl.py", line 284, in connect
    self.timeout)
  File "/usr/lib64/python2.7/socket.py", line 557, in create_connection
    for res in getaddrinfo(host, port, 0, SOCK_STREAM):
socket.gaierror: [Errno -5] No address associated with hostname
```

After:

``` python
  File "example.py", line 5, in <module>
    print driver.list_zones()
  File "/home/kami/w/lc/libcloud/libcloud/dns/drivers/godaddy.py", line 146, in list_zones
    '/v1/domains/').object
  File "/home/kami/w/lc/libcloud/libcloud/common/base.py", line 842, in request
    raise socket.gaierror(msg)
socket.gaierror: [Errno -5] No address associated with hostname. Perhaphs "host" Connection class attribute (GoDaddyDNSConnection.connection) is set to an invalid, non-hostname value (http://www.google.com)?

```

### DIFF
--- a/libcloud/common/base.py
+++ b/libcloud/common/base.py
@@ -16,6 +16,7 @@
 import os
 import sys
 import ssl
+import socket
 import copy
 import binascii
 import time
@@ -818,6 +819,22 @@ class Connection(object):
                 else:
                     self.connection.request(method=method, url=url, body=data,
                                             headers=headers)
+        except socket.gaierror:
+            e = sys.exc_info()[1]
+            message = str(e)
+            errno = getattr(e, 'errno', None)
+
+            if errno == -5:
+                # Throw a more-friendly exception on "no address associated
+                # with hostname" error. This error could simpli indicate that
+                # "host" Connection class attribute is set to an incorrect
+                # value
+                msg = ('%s. Perhaphs "host" Connection class attribute (%s) '
+                       'is set to an invalid, non-hostname value?' %
+                       (message, self.host))
+                raise socket.gaierror(msg)
+            self.reset_context()
+            raise e
         except ssl.SSLError:
             e = sys.exc_info()[1]
             self.reset_context()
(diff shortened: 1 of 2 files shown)


libcloud/common/base.py
```diff
@@ -818,6 +819,24 @@ def request(self, action, params=None, data=None, headers=None,
...
+            message = str(e)
+            errno = getattr(e, 'errno', None)
+
+            if errno == -5:
+                # Throw a more-friendly exception on "no address associated
+                # with hostname" error. This error could simpli indicate that
+                # "host" Connection class attribute is set to an incorrect
+                # value
+                class_name = self.__class__.__name__
+                msg = ('%s. Perhaphs "host" Connection class attribute '
```
tonybaloney, line 833: It's too late now, but you meant 'perhaps'?

Kami, line 833: Perhaps :P

Joking aside, good catch, will push a fix.

PR #712, opened 2016-02-23
[LIBCLOUD-802] Add drivers for Aliyun cloud

samsong8610:
https://issues.apache.org/jira/browse/LIBCLOUD-802


Kami:
First thanks for the contribution! We will look into it as soon as possible.

In the mean time, since this is a larger contribution, please make sure you file an ICLA with Apache - https://libcloud.readthedocs.org/en/latest/development.html#contributing-bigger-changes and please let us know when you have done so.

samsong8610:
@Kami I have signed icla and email the pdf to secretary@apache.org.
BTW, I fixed the failed test, but FORCE pushed it to my fork branch. How to trigger the ci to test it again?

Kami:
@samsong8610 Great, thanks!

It seems like that build was already triggered on force push (https://travis-ci.org/apache/libcloud - last build was started an hour ago).

It looks like there are still same unicode related issues in Python 3.2, I will look into it later.

libcloud/test/storage/test_oss.py
```diff
@@ -0,0 +1,880 @@
...
+                                                   self.conn.key,
+                                                   action,
+                                                   'x-oss-')
+        self.assertEqual(expected, actual)
+
+
+class ObjectTestCase(unittest.TestCase):
+    def test_object_with_chinese_name(self):
+        driver = OSSStorageDriver(*STORAGE_OSS_PARAMS)
+        obj = Object(name=u'中文', size=0, hash=None, extra=None,
```
Kami: Changing this to use `libcloud.utils.py3.u` wrapper function should fix the test failure under Python 3.2.

So:

``` python

from libcloud.utils.py3 import u
...

obj = Object(name=u('中文'), size=0, hash=None, extra=None,
...
```

samsong8610: Thanks for your help. But py3.u wrapper function doesn't work when I compare an unicode literal to a variable which stores an unicode string.
I fixed the failed test cases according to [Stackoverflow](http://stackoverflow.com/questions/6625782/unicode-literals-that-work-in-python-3-and-2). Is it acceptable in libcloud?

Kami: Ah, that's weird. I quickly tested it under Python 3.2 and it seemed to work OK (I didn't test other versions though).

Problem with `unicode_literals` is that we still support Python 2.5 where this functionality is not available (sadly we don't run 2.5 tests on Travis since Travis CI doesn't support 2.5 anymore).

In any case, consider this problem solved for now. I will try to figure out something when reviewing and merging this PR. In the worst case we can just use `unicode_literals` since we will probably drop support for Python 2.5 in Libcloud v1.0.0 anyway.

libcloud/compute/drivers/ecs.py
```diff
@@ -0,0 +1,1390 @@
...
+        'Detaching': StorageVolumeState.INUSE,
+        'Creating': StorageVolumeState.CREATING,
+        'ReIniting': StorageVolumeState.CREATING}
+
+    SNAPSHOT_STATE_MAPPING = {
+        'progressing': VolumeSnapshotState.CREATING,
+        'accomplished': VolumeSnapshotState.AVAILABLE,
+        'failed': VolumeSnapshotState.ERROR}
+
+    def list_nodes(self, ex_node_ids=[], ex_filters=None):
```
Kami: Minor thing - please don't default to a mutable type (list).

Rather to `ex_node_ids=None` and inside method body: `ex_node_ids = ex_node_ids or []`

libcloud/compute/drivers/ecs.py
```diff
@@ -0,0 +1,1390 @@
...
+                  'InstanceName': name}
+
+        if 'ex_security_group_id' not in kwargs:
+            raise AttributeError('ex_security_group_id is mandatory')
+        params['SecurityGroupId'] = kwargs['ex_security_group_id']
+
+        if 'ex_description' in kwargs:
+            params['Description'] = kwargs['ex_description']
+
+        if 'ex_internet_charge_type' in kwargs:
```
Kami: Maybe this could be refactored in a utility method.

In any case, not a big deal.

libcloud/compute/drivers/ecs.py
```diff
@@ -0,0 +1,1390 @@
...
+    def list_locations(self):
+        params = {'Action': 'DescribeRegions'}
+
+        resp_body = self.connection.request(self.path, params).object
+        location_elements = findall(resp_body, 'Regions/Region',
+                                    namespace=self.namespace)
+        locations = [self._to_location(each) for each in location_elements]
+        return locations
+
+    def create_node(self, **kwargs):
```
Kami: If possible, please also explicitly declare all the arguments here instead of using kwargs (`def create_node(self, name, size, image, ex_security_group_id, ...**kwargs`)).

I know some drivers still use `**kwargs`, but we want to avoid that going forward.

libcloud/compute/drivers/ecs.py
```diff
@@ -0,0 +1,1390 @@
...
+        status_str = findtext(element, 'Status', namespace=self.namespace)
+        status = self.VOLUME_STATE_MAPPING.get(status_str,
+                                               StorageVolumeState.UNKNOWN)
+
+        extra = self._get_extra_dict(element,
+                                     RESOURCE_EXTRA_ATTRIBUTES_MAP['volume'])
+        extra['operation_locks'] = self._get_operation_locks(element)
+        return StorageVolume(_id, name, size, self, state=status, extra=extra)
+
+    def _list_to_json_array(self, value):
```
Kami: I noticed you only use this method in one place, but if I'm not mistaken, it looks like you could also use it in some other places.

Or maybe not, since it looks like that in those other places you throw a more explicit and user-friendly exception if the argument is not of a type list.

libcloud/compute/drivers/ecs.py
```diff
@@ -0,0 +1,1390 @@
...
+        return value
+    if u(value).lower() == 'true':
+        return True
+    return False
+
+
+"""
+Define the extra dictionary for specific resources
+"""
+RESOURCE_EXTRA_ATTRIBUTES_MAP = {
```
Kami, line 60: Can you please move this to the top after imports?

We tend to put global, module level constants at the top, immediately after imports.

libcloud/compute/providers.py
```diff
@@ -173,6 +173,8 @@
     ('libcloud.compute.drivers.ciscoccs', 'CiscoCCSNodeDriver'),
     Provider.NTTA:
     ('libcloud.compute.drivers.ntta', 'NTTAmericaNodeDriver'),
+    Provider.ECS:
```
Kami: Is it possible to use some other name for the provider constant? Maybe `ALIYUN_ECS`?

If we call it just `ECS` it's slightly confusing and many users will probably mistaken it for Amazon Elastic Container Service (ECS).

libcloud/loadbalancer/providers.py
```diff
@@ -42,6 +42,9 @@
     ('libcloud.loadbalancer.drivers.softlayer', 'SoftlayerLBDriver'),
     Provider.DIMENSIONDATA:
     ('libcloud.loadbalancer.drivers.dimensiondata', 'DimensionDataLBDriver'),
+    Provider.SLB:
```
Kami: Same here, probably a good idea to be explicit and add `ALIYUN_` prefix to every provider constant.

Kami:
I've added some comments. Overall, very nice work :+1: 

After you look into / address those comments, could you please also add some documentation for those drivers?

Kami:
@samsong8610 Did you select to notify Libcloud project on the ICLA form?

I'm wondering, because I still didn't receive an email notification about the signed ICLA, maybe ASF secretary is just busy...

Edit: I do see it under "Unlisted CLAs" here - https://people.apache.org/committer-index.html (unless that's some other Sam Song :)).

samsong8610:
@Kami It's my fault. I didn't fill the 'notify project' field. I will re-sign the ICLA.

libcloud/compute/drivers/ecs.py
samsong8610: I should use it all the time except `list_images` method, in which the API need a comma separated string instead of a json array string.

libcloud/compute/drivers/ecs.py
samsong8610: Done.

libcloud/compute/drivers/ecs.py
samsong8610: Done

libcloud/compute/drivers/ecs.py
samsong8610: Done

libcloud/compute/drivers/ecs.py
samsong8610, line 60: But this dict is only used in this module, I mean that it is like a private utils function, so I think it should be put with other private functions.

Anyway, I will follow our conventions.:smile:

libcloud/compute/providers.py
samsong8610: Done

libcloud/loadbalancer/providers.py
samsong8610: Done

Kami:
@samsong8610 Sorry for the delay, I was traveling.

I will look into reviewing it again and merging it ASAP (tomorrow).

libcloud/storage/types.py
```diff
@@ -64,6 +65,7 @@ class Provider(object):
     KTUCLOUD = 'ktucloud'
     AURORAOBJECTS = 'auroraobjects'
     BACKBLAZE_B2 = 'backblaze_b2'
+    ALIYUN_OSS = 'oss'
```
Kami, line 68: To avoid potential conflict in the future, I will prefix type values for all the Aliyun providers with "aliyun_", so "aliyun_oss" in this case.

Kami, line 68: 3da15748f114cea6b42bc0aa26c034598a8d8948

Kami:
I went ahead and merged changes into trunk.

Thanks again for this very complete and high-quality contribution!

samsong8610:
Glad to help the libcloud. Thanks for your review.

PR #714, opened 2016-02-26
Fixed httplib_ssl socket exception handling.

crunk1:
Addressing https://issues.apache.org/jira/browse/LIBCLOUD-803


crunk1:
@erjohnso @Kami could you PTAL at this. We need to pull a fix in for this internally at Google.

Kami:
Good catch, thanks.

I did at some test cases for this, but looks like I missed this edge case :( I will merge this into trunk and also add a test case for this.

Thanks again.

Kami:
Added a test case (f61f3cfded51077ed3b4802b1675ac7937c7d5c7) and merged patch into trunk.

Thanks.

PR #715, opened 2016-02-27
List virtual disks attached to a node

bernard357:
This is necessary to manage disks afterwards, since this is providing ids, sizes and speeds


bernard357:
All tests pass Travis-CI except one. Error message is related to the non-configuration of lint for Python 2.7. Is there something I can do on this?

Kami:
Yeah, you need to de-indent the `})` line.

``` python
        for disk in findall(element, 'disk', TYPES_URN):
            extra['disks'].append({
                'scsiId': int(disk.get('scsiId')),
                'speed': disk.get('speed'),
                'id': disk.get('id'),
                'size': int(disk.get('sizeGb'))
            })
```

It also wouldn't hurt if you split this functionality into a separate `_to_disk(self, data)` method, similar as we do in other places.

Kami:
And of course, thanks for the contribution :)

bernard357:
thanks Kami, let's do that

bernard357:
ok, now seems to be ok - thanks for the consideration of this PR

bernard357:
As a side note, this has been left in the _to_node() function for the time being. This is because disks are natural attributes of the get_node(). Like this was done for the CPU, maybe one day we will find a good reason to enrich the service and to better justify another function, a separate class, etc.

tonybaloney:
hey @bernard357 great to see some contrib back. Can you provide a test to check the values of the disk extra parameter please? You can copy the method pattern here- https://github.com/apache/libcloud/blob/trunk/libcloud/test/compute/test_dimensiondata.py#L94 the existing fixtures have the disks in them

bernard357:
Thanks Anthony for your encouragements. Libcloud tests are new to me, so please allow me to dive into these first. Keep posted...

bernard357:
Anthony, it is great that you already worked on _to_disk(), DimensionDataServerDisk, etc. For some reason this has not reached my workstation on git pull some days ago. So, at the end of the day, this PR is completely useless and should be dropped. Instead, I will change the code in plumbery to align with your contribution. Also, I may produce additional tests as suggested, if this can turn me to a better padawan. Let the Open Force be with us :-)

tonybaloney:
This test improvement is enough. I think somehow the PR code got merged already, so the disk support is already in the driver.,

tonybaloney:
@jadunham1 did the work for you @bernard357 his PR has already been merged. I suggest you close this one https://github.com/apache/libcloud/pull/708

PR #716, opened 2016-02-29
Fix to create a firewall rule with the specific source IP

kumarappanc:
added a if condition to check the source ip prefix size is not none before passing the value to the API which will enable us to create a firewall rule with the specific source ip

`if rule.source.ip_prefix_size is not None:                
                source_ip.set('prefixSize', str(rule.source.ip_prefix_size))`


kumarappanc:
Please ignore the pull request..

PR #718, opened 2016-02-29
Register priority for MX records

zepheiryan:
As is, this fails to split out the priority value from the raw record value, resulting in a key error for missing priority when trying to export in BIND format.  This is the quick hack I did to get the function to return a more useful Record.

### DIFF
--- a/libcloud/dns/drivers/gandi.py
+++ b/libcloud/dns/drivers/gandi.py
@@ -147,16 +147,20 @@ class GandiDNSDriver(BaseGandiDriver, DNSDriver):
         return res.object
 
     def _to_record(self, record, zone):
+        extras = {'ttl': record['ttl']}
+        value = record['value']
+        if record['type'] == 'MX':
+            extras['priority'] = record['value'].split(' ')[0]
+            value = record['value'].split(' ')[1]
         return Record(
             id='%s:%s' % (record['type'], record['name']),
             name=record['name'],
             type=self._string_to_record_type(record['type']),
-            data=record['value'],
+            data=value,
             zone=zone,
             driver=self,
             ttl=record['ttl'],
-            extra={'ttl': record['ttl']}
-        )
+            extra=extras
 
     def _to_records(self, records, zone):
         retval = []


Kami:
Thanks for the contribution.

libcloud/dns/drivers/gandi.py
```diff
@@ -147,16 +147,20 @@ def delete_zone(self, zone):
         return res.object
 
     def _to_record(self, record, zone):
+        extras = {'ttl': record['ttl']}
+        value = record['value']
+        if record['type'] == 'MX':
+            extras['priority'] = record['value'].split(' ')[0]
+            value = record['value'].split(' ')[1]
```
Kami, line 154: Just curious why you split the value here and take the second (and not the first item) - it would help if I could see how the value looks like :)

Kami:
Test case for `list_records` method which returns a MX record would also be nice :)

libcloud/dns/drivers/gandi.py
zepheiryan, line 154: `record['value']` looks like `15 aspmx.l.google.com` - I didn't try it, but I expect leaving the value as that raw value would spit out `...15 15 aspmx...` when converted to BIND speak.

Given that I don't have an exhaustive knowledge of DNS record types off the top of my head (or how Gandi returns them all), I expect this is an incomplete patch that happened to work for the (or my) Gandi MX case.  I'm not sure how your team prefers to handle that, but I did want it to at least start on the way in so folks (or just me) don't have to monkey patch locally.

Kami, line 154: That's for clarifying it.

This does indeed seem like a bug / inconsistency in the Gandi driver. `value` attribute should indeed only contain the value and the priority should be stored in `extra`.

Kami:
Pushed some fixes and improvements (1d0b2eeeb710ecbf3f65a34df3c38170a78a5216, 29296923046b4cecea200ee5aa3fb9e51f20ce77), added a test case (bd3c1f044cb59333e3c6bd20cc6bff4db1428316) and merged changes into trunk.

Thanks!

PR #720, opened 2016-03-02
Dimension Data:  Additions to networking

jadunham1:
Adding Additional Nics to create_node
Adding ex_primary_ipv4 to create_node so you can set static ips
Adding more filters to ex_list_vlans
Adding more filters to ex_list_network_domains
Adding tests to cover all changes

### DIFF
--- a/libcloud/test/compute/test_dimensiondata.py
+++ b/libcloud/test/compute/test_dimensiondata.py
@@ -24,13 +24,14 @@ from libcloud.utils.py3 import httplib
 from libcloud.common.types import InvalidCredsError
 from libcloud.common.dimensiondata import DimensionDataAPIException, NetworkDomainServicePlan
 from libcloud.common.dimensiondata import DimensionDataServerCpuSpecification, DimensionDataServerDisk
+from libcloud.common.dimensiondata import TYPES_URN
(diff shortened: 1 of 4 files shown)


libcloud/test/compute/test_dimensiondata.py
```diff
@@ -237,6 +238,110 @@ def test_create_node_no_network(self):
...
 
+    def test_create_node_mcp1_ipv4(self):
+        rootPw = NodeAuthPassword('pass123')
+        image = self.driver.list_images()[0]
+        node = self.driver.create_node(name='test2',
+                                       image=image,
+                                       auth=rootPw,
+                                       ex_description='test2 node',
+                                       ex_primary_ipv4='10.0.0.1',
+                                       ex_isStarted=False)
```
tonybaloney: I can't see how this passed considering ex_is_started is the arg name?

jadunham1: Because it's just passing kwargs and ex_is_started is defaulted.  I will fix, but until we have a really good mocking system and xml per validation for all values in our case, this will never be caught.

tonybaloney:
LGTM :+1:

PR #722, opened 2016-03-09
Feature/dd quick fixes

jadunham1:
Making the status field of the DimensionDataBackupDetails object state like others, that way we can wait for state like we would for other things.
Adding an new ex_get_target_by_id for getting a single target.
Fixing up update_target as it only worked really if you passed a target in.  I didn't really test out this functionality before so it was a miss on me.

### DIFF
--- a/libcloud/backup/drivers/dimensiondata.py
+++ b/libcloud/backup/drivers/dimensiondata.py
@@ -209,14 +209,11 @@ class DimensionDataBackupDriver(BackupDriver):
             'server/%s/backup/modify' % (server_id),
             method='POST',
             data=ET.tostring(request)).object
-        return BackupTarget(
-            id=server_id,
-            name=name,
-            address=address,
-            type=type,
-            extra=extra,
-            driver=self
-        )
+        if isinstance(target, BackupTarget):
+            target.extra = extra
+        else:
+            target = self.ex_get_target_by_id(server_id)
+        return target
 
     def delete_target(self, target):
         """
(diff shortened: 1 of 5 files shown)


libcloud/backup/drivers/dimensiondata.py
```diff
@@ -558,7 +575,7 @@ def _to_backup_details(self, object):
         return DimensionDataBackupDetails(
             asset_id=object.get('asset_id'),
             service_plan=object.get('servicePlan'),
```
tonybaloney, line 585: that should really be a snake case to be consistent with the line above.

libcloud/backup/drivers/dimensiondata.py
```diff
@@ -180,7 +180,8 @@ def create_target_from_container(self, container,
         return NotImplementedError(
             'create_target_from_container not supported for this driver')
 
-    def update_target(self, target, name=None, address=None, extra=None):
+    def update_target(self, target, name=None, address=None, extra=None,
+                      type=BackupTargetType.VIRTUAL):
```
Kami: `type` argument is not part of the standard API so it should be prefixed with `ex_`, right?

Kami:
Thanks.

Besides the `type` thing, LGTM.

/cc @tonybaloney

tonybaloney:
@jadunham1 yeah the parameter should be removed completely, it isn't required

jadunham1:
@tonybaloney Agreed, removed the parameters, found a few more oddities around extras and handled them better for default values as well as put in some tests so we can test defaults better.

tonybaloney:
Great :+1:

PR #723, opened 2016-03-15
Feature/customer image pass

jadunham1:
Added a few things here:
ex_get_base_image_by_id
ex_get_customer_image_by_id
ex_get_image_by_id (combines the two above, make it easier on the developer if you don't know if the img is base/customer)

The main reason for this code change is that LINUX customer images cannot be provisioned via libcloud currently for Dimension Data.  This is because the password field is not required and infact the API returns an error if you do add one. 
This change will not add the password field if you don't need a password and compatible with how we were randomly generating passwords if they are needed and aren't supplied.


tonybaloney:
Great contribution. The docstring for create_node should also be updated.

jadunham1:
Updated docstring, None is a valid type as we randomly generate a password if it is not provided.

tonybaloney:
LGTM :+1:

PR #724, opened 2016-03-15
Remove provider constants with region in the name and related classes

Kami:
This pull request removes provider constants with region in the name (e.g. `EC2_US_WEST`, etc.) and related driver classes.

We have deprecated those long time ago and it's time to fully remove the extra code which is not necessary anymore before we do the 1.0.0. release.

To provider a better user-experience, I decided to leave code which throws a more user-friendly exception if user references a deprecated / removed provider in place.
## TODO
- [x] Code for enumerating available regions for each driver (e.g. `get_available_regions` class method)
- [x] Update code which generates supported providers tables to take regions into account
- [x] Documentation / upgrade notes changes
- [ ] Check how this affects SaltCloud, etc.


tonybaloney:
@Kami what's the progress on this?

wrt to saltcloud, the aws native driver uses the boto library instead of libcloud. I think the biggest people this impacts is @mgogoulos and the mist team

mgogoulos:
Hi, we'll make sure the branch mist.io uses - https://github.com/mistio/libcloud/tree/mistio - gets synced with this. We perform syncs every few months but of course check the project's progress on a more frequent basis! 

Thanks

Kami:
As mentioned in the PR description, I have introduced a new `get_regions` driver class method in 04dbe1891f7e3c90c5f5890f67d4f41c54f3a23b (other related changes - 2b2a79ccedefaf584beb463ac371f3d07e3e1f97, 3cb03ebb0d65854924b4866debed858f2b493e40).

This approach is not ideal since we already have `list_locations` methid, but the problem is `list_locations` tends to be user account specific (connection needs to be established to retrieve locations which are available for that particular user account) while `list_regions` returns global regions which tend to be the same for all the users (and this region argument is passed to the driver constructor before the connection to the provider is established).

In short - this means we now have two concepts - location and region. Ideally we could some how unify this into a single concept (if that's even possible).

I'm open to feedback and suggestions.

tonybaloney:
IMO @Kami it doesn't make sense to consolidate the regions and locations. Most of the providers follow a hub and spoke type system, Amazon does with AZ's for example, which tie to the APIs. The 'location' is the data center and the 'region' is the geographic region, including isolated environments, like the fed-gov cloud.
I think what we have here is good.

tonybaloney:
The only change I would make it to turn the region into a class or a tuple. the region code, the location and a description string. e.g. return ['syd', 'Sydney, Australia', 'Australia public instance')... instead of ['syd'].

Kami:
@tonybaloney Yeah, I was thinking about it, but I intentionally planned to keep it simple.

People complained about Provider constants in the past, but now that the provider constants values are simple strings it's less of an issue.

I'm fine with introducing a new provider-specific Region enum class which value is a string. This allows users to still use simple strings if they wish, but they get some "static analysis" related benefits if they reference enums (e.g. region is not defined, etc.).

Kami:
I quickly glanced over salt cloud code and it seems that drivers which use libcloud already use `region` argument so it should be fine.

tonybaloney:
Reviewed final changes. :+1: I will update other drivers with list_regions as a separate PR

Kami:
I will go ahead, merge this and follow up with an additional PR at some point which introduces per provider Region enums.

PR #726, opened 2016-03-24
DimesionData: Generic pagination, anti-affinity rules, create firewall expansion

jadunham1:
Added:
ex_list_anti_affinity_rules
ex_create_anti_affinity_rule
ex_delete_anti_affinity_rule
generic pagination function
modified ex_create_firewall_rule to be able to place rules in a specific position relative to another rule

Tests for all new added features
Added some more coverage tests to ex_create_firewall rule
97% coverage in our compute driver now.

### DIFF
--- a/libcloud/common/dimensiondata.py
+++ b/libcloud/common/dimensiondata.py
@@ -436,8 +436,35 @@ class DimensionDataConnection(ConnectionUserAndKey):
 
     def paginated_request_with_orgId_api_2(self, action, params=None, data='',
                                            headers=None, method='GET',
-                                           return_generator=False,
-                                           page_count=250):
+                                           page_size=250):
+        """
+        A paginated request to the MCP2.0 API
+        This essentially calls out to request_with_orgId_api_2 for each page
+        and yields the response to make a generator
+        This generator can be looped through to grab all the pages.
+
+        :param action: The resource to access (i.e. 'network/vlan')
+        :type  action: ``str``
+
+        :param params: Parameters to give to the action
+        :type  params: ``dict`` or ``None``
+
+        :param data: The data payload to be added to the request
+        :type  data: ``str``
+
+        :param headers: Additional header to be added to the request
+        :type  headers: ``str`` or ``dict`` or ``None``
+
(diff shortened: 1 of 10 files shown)


libcloud/compute/drivers/dimensiondata.py
```diff
@@ -686,6 +687,68 @@ def ex_update_node(self, node, name=None, description=None,
...
+
+        paged_result = self.connection.paginated_request_with_orgId_api_2(
+            'server/antiAffinityRule',
+            method='GET',
+            params=params
+        )
+
+        rules = []
+        for result in paged_result:
+            rules.extend(self._to_anti_affinity_rules(result))
```
tonybaloney, line 793: would probably be quicker to use list comprehension here

tonybaloney:
changes look good. once tests are supporting I will test and merge.

libcloud/compute/drivers/dimensiondata.py
jadunham1, line 793: I Think since our _to_ methods are returning lists here themselves extend makes more sense than list comprehension.  If you have a better way let me know because all of our functions will have the same type of style in them so we should get it right the first time.

libcloud/common/dimensiondata.py
```diff
@@ -731,6 +748,19 @@ def __repr__(self):
                 % (self.id, self.status))
 
 
+class DimensionDataAntiAffinityRule(object):
+    """
+    Anti-Affinity Rule
+    """
+    def __init__(self, id, node_list):
+        self.id = id
+        self.node_list = node_list
```
tonybaloney, line 796: Can you docstring what this does/type etc please.

tonybaloney:
Excellent changes, great to see some improvement in the tests as well. We'll need this in a few weeks when the 2.2 API comes out. Only 1 small change about docstring then we're good to merge

libcloud/common/dimensiondata.py
jadunham1, line 796: Whoops missed that one

tonybaloney:
thanks LGTM

PR #728, opened 2016-03-26
Replaces httplib.request with requests library

tonybaloney:
Todo:
- [x] map out dependencies for change
- [x] replace LibcloudHTTPConnection and LibcloudHTTPSconnection with requests powered connection classes
- [x] refactor out unneeded code from old libs, e.g. SSL validation
- [x] test


Kami:
I see you made no changes to the `libcloud.utils.security` module. What's our plan for that module?

Do we plan to remove and deprecate it or do we plan to keep it for backward compatibility reasons (this mostly means making sure `VERIFY_SSL_CERT`, `SSL_VERSION` and `CA_CERT_PATH` module level constants are still respected)?

PR #729, opened 2016-03-28
add ex_encrypted and ex_kms_key_id kwargs in EC2NodeDriver.create_volume

ViktorOgnev:


libcloud/compute/drivers/ec2.py
```diff
@@ -3129,6 +3139,12 @@ def create_volume(self, size, name, location=None, snapshot=None,
         if ex_volume_type == 'io1' and ex_iops:
             params['Iops'] = ex_iops
 
+        if ex_encrypted:
```
tonybaloney: can you make value checks more explicit please e.g. `if ex_encrypted is not None:`

tonybaloney:
hi @ViktorOgnev thanks for your contribution, there are only a couple of changes before we can merge, first I made a comment about explicit "truthy" checking, can you fix that on both lines please. Second the lint is failing because 3 of the lines are too long, probably the docstring comments from a glance. you can run `flake8` locally to check the next commit.

ViktorOgnev:
hi @tonybaloney , looks like now it's ok) Should I rebase, or is it fine as is?

tonybaloney:
perfect, no need to squash, will merge.

tonybaloney:
all done, thanks for your contribution @ViktorOgnev

PR #730, opened 2016-03-28
Outscale storage providers

jmunhoz:
Add Outscale storage providers (Ceph RGW S3)

### DIFF
--- a/libcloud/storage/drivers/s3.py
+++ b/libcloud/storage/drivers/s3.py
@@ -60,6 +60,15 @@ S3_AP_NORTHEAST2_HOST = 's3-ap-northeast-2.amazonaws.com'
 S3_AP_NORTHEAST_HOST = S3_AP_NORTHEAST1_HOST
 S3_SA_EAST_HOST = 's3-sa-east-1.amazonaws.com'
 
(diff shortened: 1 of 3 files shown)


tonybaloney:
Hi @jmunhoz do they all need to be separate drivers? We are trying to make sure that there are single drivers with a `region=x` constructor.

jmunhoz:
@tonybaloney thanks for reviewing. As suggested I pushed a single driver with region support in the constructor.

libcloud/storage/drivers/s3.py
```diff
@@ -1004,3 +1013,21 @@ class S3SAEastStorageDriver(S3StorageDriver):
...
     ex_location_name = 'sa-east-1'
+
+
+class S3RGWOutscaleConnection(S3Connection):
+    pass
+
+
+class S3RGWOutscaleStorageDriver(S3StorageDriver):
+
+    def __init__(self, *args, **kwargs):
```
Kami: Please explicitly declare all the arguments where possible instead of using `*args` and `**kwargs`.

jmunhoz:
Hi @Kami, the new patch declares the arguments explicitly.

tonybaloney:
Thanks @jmunhoz. the changes look great, merging

PR #731, opened 2016-03-29
adds dns support for luadns

piratimir:


piratimir:
@tonybaloney

libcloud/dns/drivers/luadns.py
```diff
@@ -0,0 +1,293 @@
...
+from libcloud.dns.types import ZoneDoesNotExistError, ZoneAlreadyExistsError
+from libcloud.dns.types import RecordDoesNotExistError
+
+
+__all__ = [
+    'LuadnsDNSDriver'
+]
+
+
+class LuadnsDNSResponse(LuadnsResponse):
```
tonybaloney, line 21: You can just import from the common module and remove this instead

tonybaloney:
Looks great otherwise. Let's see what the build says. You also want to add a doc page to explain the driver? You can copy existing ones and just update the logo and the website

piratimir:
@Yes, but let's see the builds first :dancer:

PR #735, opened 2016-04-01
correct rST filename for NFSN DNS driver docs

ktdreyer:
I imported the incorrect filename. Correct this here.


Kami:
Merged into trunk, thanks!

PR #736, opened 2016-04-04
Add AWS4 support in the S3_RGW_OUTSCALE provider

jmunhoz:
This patch adds AWS4 auth protocol support in the S3_RGW_OUTSCALE
provider. It is needed to use AWS4 with the upcoming version of Ceph RGW
(Jewel)

Ceph Jewel will ship with AWS2 and AWS4 enabled by default.

In the case of regions and signature binding, Ceph does not enforce any
signature version per region. Every region supports AWS2 and AWS4. Ceph
detects the signature version per request in order to authenticate
properly.

More information on the Ceph's AWS4 implementation:

http://docs.ceph.com/docs/master/release-notes/#v10-1-0-jewel-release-candidate
http://blogs.igalia.com/jmunhoz/blog/2016/03/01/aws-signature-version-4-goes-upstream-in-ceph.html

### DIFF
--- a/libcloud/storage/drivers/s3.py
+++ b/libcloud/storage/drivers/s3.py
@@ -36,7 +36,7 @@ from libcloud.utils.files import read_in_chunks
 from libcloud.common.types import InvalidCredsError, LibcloudError
 from libcloud.common.base import ConnectionUserAndKey, RawResponse
 from libcloud.common.aws import AWSBaseResponse, AWSDriver, \
-    AWSTokenConnection, SignedAWSConnection
+    AWSTokenConnection, SignedAWSConnection, DEFAULT_SIGNATURE_VERSION
 
 from libcloud.storage.base import Object, Container, StorageDriver
 from libcloud.storage.types import ContainerError
@@ -834,7 +834,7 @@ class BaseS3StorageDriver(StorageDriver):
         bytes_transferred = result_dict['bytes_transferred']
         headers = response.headers
         response = response.response
-        server_hash = headers['etag'].replace('"', '')
+        server_hash = headers.get('etag', '').replace('"', '')
 
         if (verify_hash and result_dict['data_hash'] != server_hash):
             raise ObjectHashMismatchError(
@@ -1015,8 +1015,37 @@ class S3SAEastStorageDriver(S3StorageDriver):
     ex_location_name = 'sa-east-1'
 
 
-class S3RGWOutscaleConnection(S3Connection):
-    pass
+class S3RGWOutscaleConnectionAWS4(SignedAWSConnection, BaseS3Connection):
+    service_name = 's3'
+    version = API_VERSION
+
+    def __init__(self, user_id, key, secure=True, host=None, port=None,
+                 url=None, timeout=None, proxy_url=None, token=None,
+                 retry_delay=None, backoff=None, **kwargs):
+
+        super(S3RGWOutscaleConnectionAWS4, self).__init__(user_id, key,
+                                                          secure, host,
+                                                          port, url,
+                                                          timeout,
+                                                          proxy_url, token,
+                                                          retry_delay,
+                                                          backoff,
+                                                          4)  # force aws4
+
+
+class S3RGWOutscaleConnectionAWS2(S3Connection):
+
+    def __init__(self, user_id, key, secure=True, host=None, port=None,
+                 url=None, timeout=None, proxy_url=None, token=None,
+                 retry_delay=None, backoff=None, **kwargs):
+
+        super(S3RGWOutscaleConnectionAWS2, self).__init__(user_id, key,
+                                                          secure, host,
+                                                          port, url,
+                                                          timeout,
+                                                          proxy_url, token,
+                                                          retry_delay,
+                                                          backoff)
 
 
 class S3RGWOutscaleStorageDriver(S3StorageDriver):
@@ -1029,9 +1058,19 @@ class S3RGWOutscaleStorageDriver(S3StorageDriver):
         self.name = 'OUTSCALE Ceph RGW S3 (%s)' % (region)
         self.ex_location_name = region
         self.region_name = region
-        self.connectionCls = S3RGWOutscaleConnection
-        self.connectionCls.host = S3_RGW_OUTSCALE_HOSTS_BY_REGION[region]
+        self.signature_version =\
+            kwargs.pop('signature_version', DEFAULT_SIGNATURE_VERSION)
+        self.connectionCls = S3RGWOutscaleConnectionAWS2
+        if self.signature_version == '4':
+            self.connectionCls = S3RGWOutscaleConnectionAWS4
+        host = S3_RGW_OUTSCALE_HOSTS_BY_REGION[region]
+        self.connectionCls.host = host
         super(S3RGWOutscaleStorageDriver, self).__init__(key, secret,
                                                          secure, host, port,
                                                          api_version, region,
                                                          **kwargs)
+
+    def _ex_connection_class_kwargs(self):
+        kwargs = {}
+        kwargs['signature_version'] = self.signature_version
+        return kwargs


tonybaloney:
Hey @jmunhoz I remember the AWS4 auth algorithm works with GET and POST, but I'm not so sure about PUT. Can you please check that you can upload files using these changes

jmunhoz:
@tonybaloney I would expect S3RGWOutscaleConnectionAWS4 (https://github.com/jmunhoz/libcloud/blob/s3-rgw-outscale-aws4/libcloud/storage/drivers/s3.py#L1021) having the same behaviour as S3APNE2Connection (https://github.com/jmunhoz/libcloud/blob/s3-rgw-outscale-aws4/libcloud/storage/drivers/s3.py#L987). I ran a smoke test listing objects (GET) and the testing suite. They were ok.

As suggested I will check the PUT case explicitly ASAP. Thanks!

jmunhoz:
@tonybaloney I reviewed the PUT case. The Libcloud code required one minor fix to handle the different behaviour related to ETags in the Amazon S3 and Ceph RGW implementations. Ceph RGW is not shipping ETag headers in this case so the code needs to handle the case where ETags don't exist.

BTW, Ceph RGW was not handling the case for unsigned payloads in the PUT method so I pushed this support in Ceph and it is upstream now with Jewel 10.2.0

The PUT method is working now. Tests are passing. The header dumps for Amazon S3 and Ceph RGW using the new patch follow...

<pre>
Amazon S3
---------
PUT /my-container-12345678901234/my-name-12345678901234 HTTP/1.1
X-AMZ-Content-SHA256: UNSIGNED-PAYLOAD
Content-Length: 12
Accept-Encoding: gzip,deflate
X-AMZ-Date: 20160501T164719Z
x-amz-storage-class: STANDARD
Content-Type: text/plain
Host: s3.eu-central-1.amazonaws.com
Authorization: AWS4-HMAC-SHA256 Credential=12345678901234567890/20160502/eu-central-1/s3/aws4_request, SignedHeaders=accept-encoding;content-length;content-type;host;user-agent;x-amz-content-sha256;x-amz-date;x-amz-storage-class, Signature=3200c1e45af4ef5acde510877d1b78541337b7c90f420012d694a04a13cebcd5
User-Agent: libcloud/1.0.0-rc2 (OUTSCALE Ceph RGW S3 (eu-central-1)) 
Hello test!
HTTP/1.1 200 OK
x-amz-id-2: FiEe9rH1Rmgb7+palW18yvo6cVNt1TE90UbLwC3syxRiu0DNuDWMwzGSZMQT0ao7srkGyqOmQi4=
x-amz-request-id: CC7007FFB4A71C44
Date: Sun, 01 May 2016 16:47:20 GMT
ETag: "8ac91dd17dd64e95271601b773859aaa"
Content-Length: 0
Server: AmazonS3
Ceph RGW
---------
PUT /my-container-12345678901234/my-name-12345678901234 HTTP/1.1
X-AMZ-Content-SHA256: UNSIGNED-PAYLOAD
Content-Length: 12
Accept-Encoding: gzip,deflate
X-AMZ-Date: 20160501T185440Z
x-amz-storage-class: STANDARD
Content-Type: text/plain
Host: osu.eu-west-1.outscale.com
Authorization: AWS4-HMAC-SHA256 Credential=12345678901234567890/20160502/eu-west-1/s3/aws4_request, SignedHeaders=accept-encoding;content-length;content-type;host;user-agent;x-amz-content-sha256;x-amz-date;x-amz-storage-class, Signature=fb7503d5014ce581723d9695a1395bb668a3a5fcee713663de3d7a316d9b4b27
User-Agent: libcloud/1.0.0-rc2 (OUTSCALE Ceph RGW S3 (eu-west-1)) 
Hello test!
HTTP/1.1 200 OK
ETag: "8ac91dd17dd64e95271601b773859aaa"
Content-Length: 0
Accept-Ranges: bytes
x-amz-request-id: tx000000000000000000012-00572650f0-1017-default
Date: Sun, 01 May 2016 18:54:40 GMT
</pre>

tonybaloney:
Great work. LGTM :+1:

libcloud/storage/drivers/s3.py
```diff
@@ -1015,8 +1015,35 @@ class S3SAEastStorageDriver(S3StorageDriver):
     ex_location_name = 'sa-east-1'
 
 
-class S3RGWOutscaleConnection(S3Connection):
-    pass
+def _get_s3_rgw_outscale_conn_by_sign_version(signature_version):
+
+    # aws4 connection support
+    class S3RGWOutscaleConnectionAWS4(SignedAWSConnection, BaseS3Connection):
```
Kami: I would prefer to make those classes top-level and define them outside of this function, but that's mostly a style thing.

libcloud/storage/drivers/s3.py
```diff
@@ -1029,7 +1056,9 @@ def __init__(self, key, secret=None, secure=True, host=None, port=None,
         self.name = 'OUTSCALE Ceph RGW S3 (%s)' % (region)
         self.ex_location_name = region
         self.region_name = region
-        self.connectionCls = S3RGWOutscaleConnection
+        signature_version = kwargs.get('signature_version', 'aws2')
+        self.connectionCls =\
```
Kami: Since `connectionCls` is a class-level attribute (and not an instance one), I believe this won't work correctly if multiple instances of this class are instantiated and each one uses a different signature version.

For example:

``` python
cls1 = S3RGWOutscaleStorageDriver(..., signature_version='aws2')
cls2 = S3RGWOutscaleStorageDriver(..., signature_version='aws4')
```

The version of the class which was instantiated last wins so connectionCls of cls1 would also be set to `S3RGWOutscaleConnectionAWS4` after the second class is instantiated.

Another option, which I believe should work correctly, is to utilize `ex_connection_class_kwargs` method and pass correct value for the `version` argument to the `connectionCls` depending on the requested signature version.

Kami: Here is an example of how we handle `region` connection class argument in the same manner - https://github.com/apache/libcloud/blob/trunk/libcloud/compute/drivers/rackspace.py#L142

libcloud/storage/drivers/s3.py
jmunhoz: ok

libcloud/storage/drivers/s3.py
jmunhoz: I see, thanks!

jmunhoz:
@Kami I added the suggested improvements. After some tests I finished adding the dispatching code between the connection classes as part of the _ex_connection_class_kwargs method. It is working with the case you mentioned.

libcloud/storage/drivers/s3.py
```diff
@@ -1015,23 +1015,50 @@ class S3SAEastStorageDriver(S3StorageDriver):
...
-        self.connectionCls = S3RGWOutscaleConnection
-        self.connectionCls.host = S3_RGW_OUTSCALE_HOSTS_BY_REGION[region]
+        self.signature_version = signature_version
+        self.host = S3_RGW_OUTSCALE_HOSTS_BY_REGION[region]
         super(S3RGWOutscaleStorageDriver, self).__init__(key, secret,
                                                          secure, host, port,
                                                          api_version, region,
                                                          **kwargs)
+
+    def _ex_connection_class_kwargs(self):
```
Kami, line 1073: With this approach, the same issue exists as before - you need to do something along the lines of:

``` python
def _ex_connection_class_kwargs(self):
    kwargs = {}
    kwargs['signature_version'] = self.signature_version
    return kwargs
```

And you also need to modify connection class to take `signature_version` argument (or similar) and use this argument to correctly set host attribute on the connection class inside the constructor.

I'm also fine with merging this approach for now, as long as we add some test cases for it (a test case where you instantiate two drivers with different signature versions and verify that the connectionCls and connectionCls.host is set correctly) so it's easy to refactor and fix in the near future.

jmunhoz:
@kami updated!

I tested the 'last connectionCls wins' issue with the following code:

<pre>
$ cat ex.py  
#!/usr/bin/env python
from libcloud.storage.types import Provider, ContainerDoesNotExistError
import libcloud
container_name = 'container-example'
object_name = 'object-example'
cls = get_driver(Provider.S3_RGW_OUTSCALE)
driver1 = cls('id', 'secret', region='eu-west-1', secure=False, signature_version='4')
driver2 = cls('id', 'secret', region='us-east-2', secure=False, signature_version='2')
print "driver1.connectionCls = %s" % (driver1.connectionCls)
print "driver1.connectionCls.host = %s" %( driver1.connectionCls.host)
print "driver2.connectionCls = %s" % (driver2.connectionCls)
print "driver2.connectionCls.host = %s" %( driver2.connectionCls.host)
$ ./ex.py 
driver1.connectionCls = &lt;class 'libcloud.storage.drivers.s3.S3RGWOutscaleConnectionAWS4'&gt;
driver1.connectionCls.host = osu.eu-west-1.outscale.com
driver2.connectionCls = &lt;class 'libcloud.storage.drivers.s3.S3RGWOutscaleConnectionAWS2'&gt;
driver2.connectionCls.host = osu.us-east-2.outscale.com
</pre>

libcloud/storage/drivers/s3.py
jmunhoz, line 1073: @kami done. Added the modifications. I tested with the shared snippet and it looks okay now.

Kami, line 1073: @jmunhoz Thanks for the update and sorry to bug you again, but this still won't work correctly.

You are still manipulating connectionCls inside the driver constructor. You should get rid of this code and as mentioned above, update connection class constructor to take `signature_version` argument and correctly set host inside the connection class constructor.

This `_ex_connection_class_kwargs` method doesn't actually do anything at this point, it doesn't throw because connection class declares `**kwargs`. So in short, connection class constructor accepts this argument, but doesn't do anything with it.

In any case, I can go ahead, merge this into master as is and add a (failing) test case so this can be fixed in the future.

jmunhoz, line 1073: @kami Thanks for the review. It is very appreciated.

Yes, please. Feel free to merge. We can continue improving the code over the upstream version. I guess my next patch would build on a similar bug so they could be fixed together.

I guess the example code I used at https://github.com/apache/libcloud/pull/736#issuecomment-218434459 is not catching the issue properly although it looks working ok? I am manipulating the connectionCls inside the driver constructor cause it looks the only place to choose between the two different connection classes without a further modification.

Kami:
Made a small change (09638edf4ddac08de38f7894f817dc83ac51fb1a) and merged it into trunk - thanks!

P.S. I forgot to close the PR in the commit message, can you please close this PR?

jmunhoz:
Thanks! closing this PR

PR #737, opened 2016-04-06
Added location filter in list_nodes of CloudStack

schaubl:


libcloud/compute/drivers/cloudstack.py
```diff
@@ -1348,7 +1348,7 @@ def list_locations(self):
 
         return locations
 
-    def list_nodes(self, project=None):
+    def list_nodes(self, project=None, location=None):
```
Kami, line 1351: It looks like we missed this while merging the original changes - project should really be prefixed with `ex_` since it's not part of the standard API.

In any case, I'm fine with this for now.

Kami:
Thanks.

As noted in the in-line comment, change looks good to me, but it would also be great if you can add a test case for it :)

schaubl:
Ok, I will add a test case.

Kami:
I merged this change into trunk.

Please follow up with tests as soon as you can.

Thanks!

schaubl:
@Kami: I added a test. Should I create a new PR since this one is already merged ?

schaubl:
@Kami I created a new PR for the test: #754

PR #738, opened 2016-04-06
Fix OpenStack IP type resolution

schaubl:
Checking the type of the IP with `explicit_ip_type` is not reliable.
A fixed IP is not always a private IP and a floating IP is not always a public one.

So I moved up the check of the public subnets.

I also added the `addresses` element of the reply in the `extra` dict.
This allows users to implement their own algorithm to find the right IP (e.g. by network name).

### DIFF
--- a/libcloud/compute/drivers/openstack.py
+++ b/libcloud/compute/drivers/openstack.py
@@ -2074,15 +2074,15 @@ class OpenStack_1_1_NodeDriver(OpenStackNodeDriver):
                 # public and 'fixed' for private
                 explicit_ip_type = value.get('OS-EXT-IPS:type', None)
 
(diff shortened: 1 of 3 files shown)


Kami:
Thanks.

The change looks good to me, but it would be great if you can also add some tests cases for it.

schaubl:
Ok, I will add a test case.

Kami:
Mention me when the test case has been added and I will go ahead and merge it.

Thanks!

schaubl:
@Kami I improved a bit the implementatino of the IP type resolution and updated tests accordingly.

Kami:
Merged into trunk. Thanks!

PR #739, opened 2016-04-06
DimensionData: Adding in ability to cancel a backup job

jadunham1:
Added ability to cancel a backup job
Added tests for full coverage
Removed target from canceling backup job, in base classes, target is in TargetBackupJob class
Refactored the DD info object a little bit to use the TargetBackupJob class instead of a DD specific class


tonybaloney:
beautiful. thanks Jeff

PR #740, opened 2016-04-06
Fix/create firewall rule

kumarappanc:
script sends 'True' as boolean value in the request xml for the enabled parameter to the CAAS API  but 'True' is not a valid value  of the atomic type 'xs:boolean'.


tonybaloney:
Quickly validated that https://github.com/DimensionDataCBUSydney/libcloud/blob/fix/create_firewall_rule/libcloud/common/dimensiondata.py#L629 is a property. Thanks @kumarappanc LGTM

PR #744, opened 2016-04-12
Fix openstack v3 authentication

schaubl:
### Description

This PR allows to define the OpenStack `domain` to another value that the default `Default`.
It also adds the ability to define the `scope` of the token.

With the code for the OpenStack Identity API v3, [two new parameters were added](https://github.com/apache/libcloud/blob/v0.20.1/libcloud/common/openstack_identity.py#L925): `domain_name` and `token_scope` but it was impossible to define them to a value other than their respective default.
### Status
- done, ready for review
- [x] Documentation
- [ ] ~~[ICLA](http://libcloud.readthedocs.org/en/latest/development.html#contributing-bigger-changes) (required for bigger changes)~~

### DIFF
--- a/libcloud/common/openstack_identity.py
+++ b/libcloud/common/openstack_identity.py
@@ -571,14 +571,13 @@ class OpenStackIdentityConnection(ConnectionUserAndKey):
     timeout = None
 
     def __init__(self, auth_url, user_id, key, tenant_name=None,
+                 domain_name=None, token_scope=None,
                  timeout=None, parent_conn=None):
         super(OpenStackIdentityConnection, self).__init__(user_id=user_id,
                                                           key=key,
                                                           url=auth_url,
                                                           timeout=timeout)
 
-        self.auth_url = auth_url
-        self.tenant_name = tenant_name
         self.parent_conn = parent_conn
 
         # enable tests to use the same mock connection classes.
@@ -590,6 +589,10 @@ class OpenStackIdentityConnection(ConnectionUserAndKey):
 
         self.auth_url = auth_url
         self.tenant_name = tenant_name
+        self.domain_name = domain_name if domain_name is not None else \
+            'Default'
+        self.token_scope = token_scope if token_scope is not None else \
+            OpenStackIdentityTokenScope.PROJECT
         self.timeout = timeout
 
         self.urls = {}
@@ -926,8 +929,7 @@ class OpenStackIdentity_3_0_Connection(OpenStackIdentityConnection):
     ]
 
     def __init__(self, auth_url, user_id, key, tenant_name=None,
-                 domain_name='Default',
-                 token_scope=OpenStackIdentityTokenScope.PROJECT,
+                 domain_name=None, token_scope=None,
                  timeout=None, parent_conn=None):
         """
         :param tenant_name: Name of the project this user belongs to. Note:
@@ -940,8 +942,8 @@ class OpenStackIdentity_3_0_Connection(OpenStackIdentityConnection):
                             domain to scope the token to.
         :type domain_name: ``str``
 
-        :param token_scope: Whether to scope a token to a "project" or a
-                         "domain"
+        :param token_scope: Whether to scope a token to a "project", a
+                            "domain" or "unscoped"
         :type token_scope: ``str``
         """
         super(OpenStackIdentity_3_0_Connection,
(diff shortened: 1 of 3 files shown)


Kami:
Great, thanks!

I will do a proper review later, but from a quick glance, it looks like it would be great to add some more tests to cover all the (edge) cases.

schaubl:
@Kami I don't know how to add tests for this fix because I need to check the body of the request but all tests seems to test the reply.

To test the body of the request I have to do it in `OpenStackIdentity_3_0_MockHttp` which is not a subclass of `unittest.TestCase` and doesn't have the `assert...` methods.

libcloud/common/openstack_identity.py
```diff
@@ -571,14 +571,13 @@ class OpenStackIdentityConnection(ConnectionUserAndKey):
...
 
     def __init__(self, auth_url, user_id, key, tenant_name=None,
+                 domain_name=None, token_scope=None,
                  timeout=None, parent_conn=None):
         super(OpenStackIdentityConnection, self).__init__(user_id=user_id,
                                                           key=key,
                                                           url=auth_url,
                                                           timeout=timeout)
 
-        self.auth_url = auth_url
```
Kami, line 580: (just thinking out loud)

I assume you removed those attributes since they are duplicated aka are declared again a couple of lines below, right?

libcloud/common/openstack_identity.py
```diff
@@ -926,8 +929,7 @@ class OpenStackIdentity_3_0_Connection(OpenStackIdentityConnection):
     ]
 
     def __init__(self, auth_url, user_id, key, tenant_name=None,
-                 domain_name='Default',
-                 token_scope=OpenStackIdentityTokenScope.PROJECT,
+                 domain_name=None, token_scope=None,
```
Kami: Any particular reason why you changed the default values? Doesn't this make it backward incompatible?

libcloud/common/openstack_identity.py
```diff
@@ -940,32 +942,29 @@ def __init__(self, auth_url, user_id, key, tenant_name=None,
...
-                             (token_scope))
+                             (self.token_scope))
 
-        if (token_scope == OpenStackIdentityTokenScope.PROJECT and
-                (not tenant_name or not domain_name)):
+        if (self.token_scope == OpenStackIdentityTokenScope.PROJECT and
+                (not self.tenant_name or not self.domain_name)):
             raise ValueError('Must provide tenant_name and domain_name '
                              'argument')
-        elif (token_scope == OpenStackIdentityTokenScope.DOMAIN and
```
Kami, line 962: Any particular reason you removed this? Shouldn't we still have this check here?

libcloud/common/openstack_identity.py
schaubl, line 580: Right

libcloud/common/openstack_identity.py
schaubl, line 962: Because `domain_name` has a default value it cannot be None nor empty.
See:
- https://github.com/schaubl/libcloud/blob/openstack-fix_auth_v3/libcloud/common/openstack_identity.py#L592
- https://github.com/schaubl/libcloud/blob/openstack-fix_auth_v3/libcloud/common/openstack.py#L420
- https://github.com/schaubl/libcloud/blob/openstack-fix_auth_v3/libcloud/common/openstack.py#L443
- https://github.com/schaubl/libcloud/blob/openstack-fix_auth_v3/libcloud/common/openstack.py#L203

libcloud/common/openstack_identity.py
schaubl: Before this PR the only [place where it is called](https://github.com/apache/libcloud/pull/744/files#diff-cb42abf0c253c7c80a89e7a9293d15a3R199) wasn't passing these two parameters so their default value was the only possible value.
With the way this parameter is passed from the driver's `__init__` to this method, it will be None if it is omitted. 
So I changed the logic from

> use the default value if it is not passed during the instanciation of OpenStackIdentity_3_0_Connection

(which is a case that cannot happens)  
to 

> use the default value if it is None

This is also why I removed the `if` below.

libcloud/common/openstack_identity.py
Kami, line 962: Ah, thanks to for the clarification.

Also re to the comment above - I would prefer to specify default values for the arguments in the method signature instead of doing it in the method body.

Doing it in the body kinda defeats the purpose of default argument values and it also breaks things such as static code analysis (now if you want to know the default value you need to be aware of the implementation details and inspect the method body).

Would it be possible to change that back to assign default values in the method signature?

schaubl, line 962: Ok, I will try to do something about that.

Kami, line 962: Great, thanks.

schaubl:
@Kami: Now default values are specified in the method signature.
I also added back a test case and the condition which check if `domain_name` is defined.

libcloud/common/openstack_identity.py
```diff
@@ -940,32 +942,32 @@ def __init__(self, auth_url, user_id, key, tenant_name=None,
...
         """
         super(OpenStackIdentity_3_0_Connection,
               self).__init__(auth_url=auth_url,
                              user_id=user_id,
                              key=key,
                              tenant_name=tenant_name,
+                             domain_name=domain_name,
+                             token_scope=token_scope,
                              timeout=timeout,
                              parent_conn=parent_conn)
```
schaubl, line 957: I would suggest to change that code to:

``` python
    # Docstring
    super(OpenStackIdentity_3_0_Connection, self).__init__(*args, **kwargs)
```

So the signature doesn't have to be specified twice.
But I know that `kwargs` are not appreciated, so tell me if you want me to do this change or not.

tonybaloney, line 957: @schaubl this is a clear case to use kwargs, its not that the project dictates not using kwargs but in the case where the keyword arguments can be listed they can.
this is fine as is.

Kami:
@schaubl Thanks.

I will review the changes again and if everything looks good go ahead and merge them intro trunk.

Kami:
Merged, thanks!

PR #745, opened 2016-04-12
sort dns types

ktdreyer:
Sort the `Provider` and `RecordType` lists alphabetically. This makes it simpler to insert new providers and types.


tonybaloney:
thanks @ktdreyer

ktdreyer:
thanks, I've rebased on the latest trunk and re-pushed the branch.

Kami:
Are the types already sorted for other APIs (compute, loadbalancer, storage)? If not, we should do it there as well to keep it consistent.

As far as existing order goes - we sorted providers based on the date they were added (from oldest to the newest).

ktdreyer:
You're right @Kami , they are not already sorted for the other providers, so we should do it.

The issue I've found with the oldest-to-newest ordering is that it makes it slightly more complicated when adding in-progress drivers to the codebase, because each new in-progress driver is guaranteed to introduce a Git merge conflict with all the other in-progress drivers. With alphabetical sorting, merge conflicts could still arise, but at least they won't happen every single time.

PR #748, opened 2016-04-13
Sort providers

ktdreyer:
By sorting the providers alphabetically, we make it less likely that each new in-progress driver will conflict with each other when merging in Git.

DNS was already sorted in #745 , so this PR sorts the `Providers` for the rest.

### DIFF
--- a/libcloud/compute/types.py
+++ b/libcloud/compute/types.py
@@ -60,111 +60,111 @@ class Provider(Type):
     Defines for each of the supported providers
 
     :cvar DUMMY: Example provider
+    :cvar ABIQUO: Abiquo driver
+    :cvar ALIYUN_ECS: Aliyun ECS driver.
+    :cvar AURORACOMPUTE: Aurora Compute driver.
+    :cvar AZURE: Azure driver.
+    :cvar BLUEBOX: Bluebox
+    :cvar CLOUDSIGMA: CloudSigma
+    :cvar CLOUDSIGMA_US: CloudSigma US Las Vegas
+    :cvar CLOUDSTACK: CloudStack
+    :cvar DIMENSIONDATA: Dimension Data Cloud
+    :cvar EC2_EU_WEST: Amazon AWS EU Ireland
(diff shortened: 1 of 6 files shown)


Kami:
Thanks.

Please also add a comment to each doc string class otherwise things will get out-of-sync soon :)

Ideally this would also be auto-sorted on commit, but comment is fine to start with.

ktdreyer:
thanks @Kami , I've added another commit to this branch which adds comments above every `Provider` class.

ktdreyer:
Shoot you said to add it to the _doc string_, and I just added it as an inline comment. I will fix that and re-push.

ktdreyer:
I've re-pushed with the doc string comment additions

ktdreyer:
Ok, this passes Travis CI now.

libcloud/compute/types.py
```diff
@@ -59,112 +59,115 @@ class Provider(Type):
...
+    :cvar ABIQUO: Abiquo driver
+    :cvar ALIYUN_ECS: Aliyun ECS driver.
+    :cvar AURORACOMPUTE: Aurora Compute driver.
+    :cvar AZURE: Azure driver.
+    :cvar BLUEBOX: Bluebox
+    :cvar CLOUDSIGMA: CloudSigma
+    :cvar CLOUDSIGMA_US: CloudSigma US Las Vegas
+    :cvar CLOUDSTACK: CloudStack
+    :cvar DIMENSIONDATA: Dimension Data Cloud
+    :cvar EC2_EU_WEST: Amazon AWS EU Ireland
```
tonybaloney, line 75: The compute driver for EC2 is now a single driver. these cvars are wrong @ktdreyer

ktdreyer:
Yeah, there's several inconsistencies in the cvars. I can fix those in a separate PR, if you merge this one?

Kami:
Merged into trunk, thanks.

I've fixed the cvar docstring locally.

ktdreyer:
Thanks @Kami !

PR #753, opened 2016-04-15
Firewall rule update

tonybaloney:
### Description

Minor change @jadunham1 
### Status
- done, ready for review
- [ ] Documentation


libcloud/test/compute/fixtures/dimensiondata/network_firewallRule.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<firewallRules xmlns="urn:didata.com:api:cloud:types" pageNumber="1" pageCount="13" totalCount="13" pageSize="50">     
+<firewallRules xmlns="urn:didata.com:api:cloud:types" pageNumber="1" pageCount="14" totalCount="14" pageSize="50">
```
jadunham1, line 2: For the file I think you'll lose some coverage by removing those rules.  Rules are all over the place, I just added some more to make sure it at least was able to parse all of the different types of rules we have.

Also page count and total count are way off now with those deleted rules (probably were before)

libcloud/compute/drivers/dimensiondata.py
```diff
@@ -2015,12 +2032,19 @@ def _to_firewall_rule(self, element, locations, network_domain):
     def _to_firewall_address(self, element):
         ip = element.find(fixxpath('ip', TYPES_URN))
         port = element.find(fixxpath('port', TYPES_URN))
```
jadunham1, line 2034: same element port set twice

libcloud/compute/drivers/dimensiondata.py
```diff
@@ -2015,12 +2032,19 @@ def _to_firewall_rule(self, element, locations, network_domain):
...
+        port = element.find(fixxpath('port', TYPES_URN))
+        port_list = element.find(fixxpath('portList', TYPES_URN))
+        address_list = element.find(fixxpath('ipAddressList', TYPES_URN))
         return DimensionDataFirewallAddress(
             any_ip=ip.get('address') == 'ANY',
             ip_address=ip.get('address'),
             ip_prefix_size=ip.get('prefixSize'),
             port_begin=port.get('begin') if port is not None else None,
```
jadunham1, line 2041: Just curious on these lines...
.get() returns None if it doesn't exist I believe, so wondering what the if is not None else None is for.

tonybaloney, line 2041: .get returns None if the dictionary key doesn't exist, but if port is not None is when the instance is None

libcloud/test/compute/fixtures/dimensiondata/network_firewallRule.xml
tonybaloney, line 2: I'll use yours later. The rules I had in my results were probably a security issue to commit to the repo

libcloud/common/dimensiondata.py
```diff
@@ -806,12 +806,15 @@ class DimensionDataFirewallAddress(object):
     The source or destination model in a firewall rule
     """
     def __init__(self, any_ip, ip_address, ip_prefix_size,
-                 port_begin, port_end):
```
Kami, line 809: Just a heads up - looks like this change is backward incompatible since it broke st2contrib action (pack there is using latest trunk) - https://travis-ci.org/StackStorm/st2contrib/jobs/125555233#L455 :)

Seems like the solution would be to default new constructor arguments to `None` if they are optional and / or update the pack action to supply those arguments.

PR #755, opened 2016-04-15
Fix regression with "_ex_connection_class_kwargs" kwargs getting overridden with None in the BaseDriver class

Kami:
This issue was originally reported by Jay Rolette on the mailing list.

fe81fef807d0f26b0fe94d55aa59b6979d133ae8 introduced a regression where arguments such as `timeout`, etc. supplied via `_ex_connection_class_kwargs` method would get overridden with `None` if the values for those arguments weren't explicitly passed to the driver constructor.

This pull request fixes that and adds some test cases to make sure we don't introduce the same regression or break it again in the future.


libcloud/common/base.py
```diff
@@ -1153,10 +1153,19 @@ def __init__(self, key, secret=None, secure=True, host=None, port=None,
...
         conn_kwargs = self._ex_connection_class_kwargs()
-        conn_kwargs.update({'timeout': kwargs.pop('timeout', None),
-                            'retry_delay': kwargs.pop('retry_delay', None),
-                            'backoff': kwargs.pop('backoff', None),
-                            'proxy_url': kwargs.pop('proxy_url', None)})
+
+        # Note: We do that to make sure those additional arguments which are
+        # provided via "_ex_connection_class_kwargs" are not overriden with
+        # None
+        additional_kwargs = ['timeout', 'retry_delay', 'backoff', 'proxy_url']
```
Kami, line 1160: I really dislike kwargs abuse here so I will try to move those arguments to the actual method signature and make sure it doesn't break anything.

Sadly, I think the original author had a reason to do that (putting it in the method signature might break some drivers) so maybe that won't work out...

PR #757, opened 2016-04-16
test: fix spelling in _get_response_if_not_available()

ktdreyer:
"availale" -> "available"


Kami:
Heh, that happens when you rely on auto-completion a lot :)

In any case, thanks.

PR #758, opened 2016-04-16
[LIBCLOUD-810] add PowerDNS driver

ktdreyer:
implements the PowerDNS driver, https://issues.apache.org/jira/browse/LIBCLOUD-810


tonybaloney:
@ktdreyer this contribution is excellent, where did you get the test fixtures from?

ktdreyer:
Thank you. I have a local PowerDNS server here, so I executed the raw REST commands using curl, and then I pasted the JSON output from curl into the .json fixtures.

PR #759, opened 2016-04-16
doc: bump number of providers

ktdreyer:
It's 74 unique providers by my count from grepping the tree. Let's be conservative and say "over 60" at this point.


tonybaloney:
ha. nice. merging

PR #760, opened 2016-04-16
doc: fix grammar in working-with-oo-apis

ktdreyer:
Use the definite article "the" when referring to methods.

Use the conjunction "and" instead of the typo "an" when referring to two method calls, and pluralize the HTTP requests that these calls make.


tonybaloney:
Thanks Ken

PR #762, opened 2016-04-18
[Ideas welcome] Should lxml be properly supported or removed?

pzrq:
### Description

This PR asks the question - is [lxml](http://lxml.de/) supported as part of libcloud? Should it be?

So far it has found little evidence lxml is in practice supported, and real-world errors if lxml is present.
Known to affect some real-world AWS S3 cases, and probably lots of other drivers based on the tests which detect 52 errors (Py2.7) and 595 errors (Py3.5). 

Unfortunately [lxml](http://lxml.de/) often sneaks its way into real-world virtual environments via dependencies such as [premailer](https://github.com/peterbe/premailer) so removing it entirely from a solution as a workaround would often be difficult to do in practice.

It would be appreciated if libcloud's authors made a decision about whether to drop lxml in places like this one [here](https://github.com/apache/libcloud/blob/v1.0.0-rc2/libcloud/backup/drivers/dimensiondata.py#L16-L19), or whether the time should be taken to properly support both Python's builtin `xml` module, and `lxml`.

See also https://github.com/apache/libcloud/commit/0619c9257556b52794f3a2f70760abc01798faf5
### Status
- work in progress, ideas and suggestions on how to fix this would be welcome
### History

Original script that triggered the discovery:

```
from libcloud.storage.types import ObjectDoesNotExistError, Provider
driver_cls = get_driver(Provider.S3_US_WEST)
# Get keys from following
# http://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSGettingStartedGuide/AWSCredentials.html
driver = driver_cls('your_access_key','your_secret_key')
driver.list_containers()
```

Results in **MalformedResponseError**

```
# Note: Added some whitespace for readability and redacted some possibly sensitive info
MalformedResponseError: <MalformedResponseException in <libcloud.storage.drivers.s3.S3USWestStorageDriver object at 0x10856e128> 'Failed to parse XML'>:
'<?xml version="1.0" encoding="UTF-8"?>
<ListAllMyBucketsResult xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
<Owner>
    <ID>redacted</ID>
    <DisplayName>redacted</DisplayName>
</Owner>
<Buckets>
    <Bucket><Name>redacted</Name><CreationDate>2011-06-23T16:49:03.000Z</CreationDate></Bucket>
    <Bucket><Name>redacted</Name><CreationDate>2016-04-14T06:49:08.000Z</CreationDate></Bucket>
    <Bucket><Name>redacted</Name><CreationDate>2011-06-24T06:57:52.000Z</CreationDate></Bucket>
</Buckets>
</ListAllMyBucketsResult>'
```

WIP
- [ ] Documentation

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,9 @@ addons:
 matrix:
   fast_finish: true
   include:
+    - env: ENV=3.5-lxml
+      python: 3.5
+      before_script: TOX_ENV=py3.5-lxml
     - env: ENV=lint
       python: 2.7
       before_script: TOX_ENV=lint
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,9 +22,30 @@ addons:
 matrix:
   fast_finish: true
   include:
+    - env: ENV=2.6-lxml
+      python: 2.6
+      before_script: TOX_ENV=py2.6-lxml
+    - env: ENV=2.7-lxml
+      python: 2.7
+      before_script: TOX_ENV=py2.7-lxml
+    - env: ENV=3.2-lxml
+      python: 3.2
+      before_script: TOX_ENV=py3.2-lxml
+    - env: ENV=3.3-lxml
+      python: 3.3
+      before_script: TOX_ENV=py3.3-lxml
+    - env: ENV=3.4-lxml
+      python: 3.4
+      before_script: TOX_ENV=py3.4-lxml
     - env: ENV=3.5-lxml
       python: 3.5
       before_script: TOX_ENV=py3.5-lxml
+    - env: ENV=pypy-lxml
+      python: pypy
+      before_script: TOX_ENV=pypypy-lxml
+    - env: ENV=pypy3-lxml
+      python: pypy3
+      before_script: TOX_ENV=pypypy3-lxml
     - env: ENV=lint
       python: 2.7
       before_script: TOX_ENV=lint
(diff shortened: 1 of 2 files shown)


Kami:
Thanks for the PR.

I agree with you and if nothing else, we should provide a way (module level setting / environment variable) to disable lxml in case user for whatever reason doesn't want to use it.

And yeah, as far as Python 3 + lxml goes, I haven't done much testing of that combination.

Would you be OK with something like that to begin with?

Kami:
To begin with, I will merge https://github.com/apache/libcloud/pull/763 into trunk and then we can decide how to proceed.

First step would probably be allowing user to disable lxml.

As far as "Unfortunately lxml often sneaks its way into real-world virtual environments via dependencies such as premailer so removing it entirely from a solution as a workaround would often be difficult to do in practice." comment goes - usually the best thing to do is to isolate different applications using virtual environments and have a virtual environment per application (this way there won't be a any dependency clashes, that's usually a best practice).

I know that's not always possible that's why I'm fine with adding a way to disable lxml.

Kami:
Sorry, didn't want to close this PR, merge commit automatically closed it.

In any case, let's continue discussion on this problem in this issue / thread :)

Kami:
One option would also be to make it opt-in instead of opt-out by default.

What do others think?

pzrq:
My 2c is having an option would be an acceptable short term fix, especially if there's a strong desire to get `v1.0.0` shipped. Some considerations:
- Security https://bitbucket.org/tiran/defusedxml#rst-header-python-xml-libraries
- Speed http://lxml.de/performance.html

pzrq:
Came up in some Googling https://github.com/ministryofjustice/salt-cloud-provider-vcloud/issues/7

Kami:
Thanks for additional clarification, I didn't know it was causing issues in Salt as well.

In this case I think it's reasonable to make it opt-in by default (aka disabling it by default).

PR #767, opened 2016-04-19
PY3: Change libcloud.common.base.Response.body from str to bytes type

pzrq:
### Description

Let's at least fix the case where if `lxml` is present, AWS S3 `driver.list_containers()` raises a `MalformedResponseError` in normal operation, and roll forward as we discover other things that actually need fixing, i.e. 

```
import this
...
Although practicality beats purity.
```

See also #762
### Status
- done, ready for review (I'm sure someone will let me know if it needs more testing in light of #762)
- [ ] Documentation

### DIFF
--- a/libcloud/common/base.py
+++ b/libcloud/common/base.py
@@ -170,9 +170,6 @@ class Response(object):
             self.body = self._decompress_response(body=response.read(),
                                                   headers=self.headers)
 
-        if PY3:
-            self.body = b(self.body).decode('utf-8')
-
         if not self.success():
             raise exception_from_message(code=self.status,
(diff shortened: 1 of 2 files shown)


Kami:
Thanks a lot.

It would be great if you can add some test case for this. I know this was causing issues under Python 3 and that's why we added that code.

pzrq:
Thanks @kami going to close this as copying the change into the relevant `MockResponse` classes reveals this won't work with JSON responses, e.g. 
`json.loads(b'hi')` => TypeError: the JSON object must be str, not 'bytes'

PR #768, opened 2016-04-19
Hook up pylint and run pylint target on Travis C

Kami:
This is probably like 5th attempt to get pylint working in some form or another on the Libcloud code base.

I tried it a couple of times myself, but I always started with `libcloud/compute/` which has tons of false positives so I always kinda lost the motivation to continue.

Best approach to avoid that is to do it step by step (baby steps!). In this PR I'm starting with `libcloud/dns/` and we can slowly continue from there towards `libcloud/compute/` which will be a massive beast :)

There are still tons of false positives in other places which will need to be addressed. For now, we can disable some of those warnings, but down the road we will need to change some code and / or write some pylint plugins which let pylint know how to handle some of the more convoluted code.


Kami:
I will merge this into trunk so at least part of the code base is covered by pylint (right now that is `libcloud/{common,dns}`).

pylint also indicated some bigger underlying issues with wrong usage of multiple inheritance, etc. but I decided to only fix small things and added pragmas to ignore those other parts (I don't want to break more code).

PR #770, opened 2016-04-20
[LIBCLOUD-811] Fix to mishandled shopper_id at GoDaddy DNS Driver

ghost:
Link to the issue tracked: https://issues.apache.org/jira/browse/LIBCLOUD-811
### Description

GoDaddy DNS driver was not handling shopper id properly. The X-Shopper-Id which is supposed to tell the shopper id was change to True. For that reason the whole GoDaddy driver was unusable. The fix aws to add _ex_connection_class_kwargs method to GoDaddyDNSDriver. Arguments of GoDaddyDNSConnection constructor were reordered. 
### Status

Done, ready for review
- [ x ] [Code linting](http://libcloud.readthedocs.org/en/latest/development.html#code-style-guide) (required, can be done after the PR checks)
- [  ] Documentation
- [ x ] [Tests](They were executed) (http://libcloud.readthedocs.org/en/latest/testing.html)
- [ NA ] [ICLA](http://libcloud.readthedocs.org/en/latest/development.html#contributing-bigger-changes) (required for bigger changes)

### DIFF
--- a/libcloud/dns/drivers/godaddy.py
+++ b/libcloud/dns/drivers/godaddy.py
@@ -71,7 +71,7 @@ class GoDaddyDNSConnection(ConnectionKey):
 
     allow_insecure = False
 
-    def __init__(self, key, secret, shopper_id, secure=True, host=None,
+    def __init__(self, key, secret, secure=True, shopper_id=None, host=None,
                  port=None, url=None, timeout=None,
                  proxy_url=None, backoff=None, retry_delay=None):
         super(GoDaddyDNSConnection, self).__init__(
@@ -130,6 +130,7 @@ class GoDaddyDNSDriver(DNSDriver):
         :param  secret: Your access key secret
         :type   secret: ``str``
         """
+        self.shopper_id = shopper_id
         super(GoDaddyDNSDriver, self).__init__(key=key, secret=secret,
                                                secure=secure,
                                                host=host, port=port,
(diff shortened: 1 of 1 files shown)


tonybaloney:
Hey @teemuvesula thanks for the contribution  I think I know what happened. There are 3 types of accounts in godaddy API. I tested on my account, which is the type that apparently doesn't require that header. Which type of account are you using? I'm going to pull this fork and check it locally with my API keys as well because I think the addition of the header is optional but if you include it I'm not sure if they ignore it. 
https://developer.godaddy.com/getstarted

ghost:
@tonybaloney - Yes, it is optional at GoDaddy API, but if the header exists, then it has to be valid. Unfortunately in at the current code the LibCloud tries to send it always - no matter what the value is. 

I was thinking that the X-Shopper-Id should be totally optional, but it would need more code changes. Also at that point the constructor of GoDaddyDNSDriver should be changed so that whole shopper_id is optional. Unfortunately that would break the whole backward compatibility. I think that it's bad idea.

tonybaloney:
ok so the change should just include a small `if self.shopper_id is not None:` around 
add_default_headers line for adding the header.

ghost:
Ok - now that change is done also.

tonybaloney:
:+1: merging

Kami:
Thanks :+1: 

I also noticed some issues related to that just a couple of days ago when I enabled pylint on `libcloud/dns/` (it turned out shopper_id was not even used, or something along those lines).

PR #774, opened 2016-04-23
[WIP] Lint script for checking driver compliance to the base / standard Libcloud API and quality checking

Kami:
This pull request works on the great initial work and idea from @Jc2k (#211).
## Background

I've talked with @tonybaloney the other day on how we could make the library better and more user-friendly. Usually the biggest problem and "unfriendliness" is when the driver doesn't fully comply to the base API. From a high level perspective those small things (e.g. argument is missing an `ex_` prefix, etc.) don't seen like a big deal, but they actually are (death by the thousand cuts).

This causes problems because it makes building truly provider agnostic applications harder and means that user needs to build another layer on top of Libcloud which handles those differences and this makes the library less attracting (hey, Libcloud is supposed to abstract those differences away!).

As of now, we do this "compliance against the base / standard API" checking manually during code reviews. Obviously manual process is error prone and wastes a tons of time which could be spend on reviewing things which actually add value.

I think the first step for improving the quality is having a script and automated check (linter or whatever you might call it) which runs on every commit and reports all the violations.
## TODO
- [ ] Finish it up (still tons of work and refactoring to be done)
- [ ] Hook it up to the CI (start with the existing warnings not being fatal)
- [ ] Start fixing and improving the existing code (not part of this PR)

### DIFF
--- a/contrib/lint.py
+++ b/contrib/lint.py
@@ -14,91 +14,191 @@
 # limitations under the License.
 
 """
-libcloud linter
+Script which checks a driver for for compliance against the base API.
 
-This script checks the libcloud codebase for registered drivers that don't
-comply to libcloud API guidelines.
+Right now it checks for the following things:
+
+1. Driver methods which are not part of the base API need to be prefixed with
+   "ex_"
+2. Additional arguments for the methods which are part of the standard API need
+   to be prefixed with "ex_"
+3. Method signature for the methods which are part of the standard API needs to
+   match the signature of of the standard API (ignoring the extension arguments).
 """
 
-import inspect
 import os
+import argparse
+import hashlib
+import inspect
+
+from collections import defaultdict
 
 import libcloud
 
-import libcloud.compute.providers
+from libcloud.compute.providers import get_driver as get_compute_driver
 from libcloud.compute.base import NodeDriver
 
 import libcloud.dns.providers
 from libcloud.dns.base import DNSDriver
 
 import libcloud.loadbalancer.providers
-from libcloud.loadbalancer.base import Driver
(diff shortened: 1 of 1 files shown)


contrib/lint.py
```diff
@@ -0,0 +1,204 @@
...
+
+from collections import defaultdict
+
+import libcloud
+
+from libcloud.compute.providers import get_driver as get_compute_driver
+from libcloud.compute.base import NodeDriver
+
+import libcloud.dns.providers
+from libcloud.dns.base import DNSDriver
```
tonybaloney, line 42: Add `# pylint: disable=unused-import` to lines 42, 45 and 48

contrib/lint.py
```diff
@@ -0,0 +1,204 @@
...
+
+        extra_args = (driver_args - core_args)
+        for extra in extra_args:
+            if not extra.startswith('ex_'):
+                message = "Core API function shouldn't take arg '%s'. Should it be prefixed with ex_?" % extra
+                warning = get_warning_object(obj=method, message=message)
+                warnings.append(warning)
+
+    # Filter out empty warning objects (dupes)
+    warnings = [warning for warning in warnings if warning is not None]
```
tonybaloney, line 170: call the temp var (`warning`) something else because it conflicts with the previous scope

tonybaloney:
@Kami are you going to merge this and then keep the fixes to the drivers seperate?

Kami:
The script / code itself still needs quite a bit more of (refactoring) love before I plan to merge it.

I mostly just wanted to get some feedback on the overall approach to begin with :)

Kami:
Old and out of date PR. Closing.

PR #775, opened 2016-04-25
[LIBCLOUD-813] create record and update record was not working

ghost:
### Description

Link to the bug report: https://issues.apache.org/jira/browse/LIBCLOUD-813

Following changes were required to get those working:
- Priority entry exists only at MX and SRV records
- Content type of submitted data had to be application/json
- Data parameter of connection request has to be string 
  Only MX and SRV records have pririoty. Cotent-type for data has to be json.  data param of connection request must be string, changed that.
### Status
- done, ready for review
- [X ] [Code linting](http://libcloud.readthedocs.org/en/latest/development.html#code-style-guide) (required, can be done after the PR checks)
- [NA ] Documentation
- [X] [Tests](http://libcloud.readthedocs.org/en/latest/testing.html) - executed
- [ NA] [ICLA](http://libcloud.readthedocs.org/en/latest/development.html#contributing-bigger-changes) (required for bigger changes)

### DIFF
--- a/libcloud/dns/drivers/godaddy.py
+++ b/libcloud/dns/drivers/godaddy.py
@@ -89,6 +89,7 @@ class GoDaddyDNSConnection(ConnectionKey):
     def add_default_headers(self, headers):
         if self.shopper_id is not None:
             headers['X-Shopper-Id'] = self.shopper_id
+        headers['Content-type'] = 'application/json'
         headers['Authorization'] = "sso-key %s:%s" % \
             (self.key, self.secret)
         return headers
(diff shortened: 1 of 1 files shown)


Kami:
Thanks, the changes look good to me.

If you get a chance it would also be good to add some tests which verify that the payload is indeed correct (you would need to assert this into the MockHttp class methods).

ghost:
Actually all tests should be re-done because they didn't notice any critical bugs. Unfortunately I don't have time for that at the moment. LIBCLOUD-811 and -813 delayed me too much with my work.

tonybaloney:
LGTM

Kami:
@teemuvesala Yeah, I agree.

I recently also enabled pylint so this should also help with catching at least part of the similar issues in the future.

PR #776, opened 2016-04-25
Fix gce_demo for loadbalancing and Python 3

wrigri:
### Description

Firewall for load balancing needs target tags set instead of source tags.
The 'print' statement was causing this code to not run in Python 3.x.
### Status
- done, ready for review
- [NA] Documentation
- [NA] [Tests](http://libcloud.readthedocs.org/en/latest/testing.html)

### DIFF
--- a/demos/gce_demo.py
+++ b/demos/gce_demo.py
@@ -553,7 +553,7 @@ def main_load_balancer():
     name = '%s-firewall' % DEMO_BASE_NAME
     allowed = [{'IPProtocol': 'tcp',
                 'ports': ['80']}]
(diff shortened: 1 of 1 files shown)


wrigri:
- @erjohnso

tonybaloney:
erm. why are you printing an empty string in the first place @wrigri ?

wrigri:
@tonybaloney The reason for this print is that right above we are calling a 'sys.stdout.write()' with only a CR (no LF) to display output from the instances in a loop (continually overwriting the line). The print makes sure that we get a final line feed in there so that the next output properly shows up on the next line (instead of the same one).

PR #785, opened 2016-05-11
Updated create_volume routine in the cloudstack.py file to support ex_volume_type

gregorybishop:
### Description

As part of the deployment of an internal cloud using CloudStack, we discovered that
one cannot use the "DiskOffering" capability of CloudStack to select volumes with
specific attributes. When reviewing the OpenStack portion of the library, this is
supported using the ex_volume_type variable in the call to create_volume.

The CloudStack DiskOffering capability is required for this deployment, so the create_volume
routing in cloudstack.py was updated to add this as an option using the same variable name
as used for OpenStack.

When a DiskOffering is specified, the input size to create_volume is over-riden by the 
size of the DiskOffering, unless the DiskOffering size is customizable. This was chosen
to minimize errors, as if someone calls create_volume with the ex_volume_type option,
then it is likely that is what they want, even if the specified size is different from the request.
### Status
- done, ready for review
- [ ] Documentation

### DIFF
--- a/libcloud/compute/drivers/cloudstack.py
+++ b/libcloud/compute/drivers/cloudstack.py
@@ -2111,7 +2111,7 @@ class CloudStackNodeDriver(CloudStackDriverMixIn, NodeDriver):
 
         return projects
 
-    def create_volume(self, size, name, location=None, snapshot=None, 
+    def create_volume(self, size, name, location=None, snapshot=None,
                       ex_volume_type=None):
         """
         Creates a data volume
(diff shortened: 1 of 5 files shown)


Kami:
Thanks and sorry for the delay.

The change looks good to me.

/cc @runseb

Kami:
The changes look good to me so I went and merged them into trunk - thanks!

In the future, please make sure that git author field is correctly configured for each commit (looks like a second commit was made from a different (root) user account which didn't have git user name and email configured so I squashed that commit into the previous one).

gregorybishop:
Thank you. 

I had to stand up a new VM to get the test harness working and pushed from the wrong place. Ugh.

PR #786, opened 2016-05-13
Add S3_RGW storage driver

jmunhoz:
Add S3_RGW storage driver

### DIFF
--- a/libcloud/storage/drivers/rgw.py
+++ b/libcloud/storage/drivers/rgw.py
@@ -1,0 +1,137 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from libcloud.common.types import LibcloudError
+
+from libcloud.storage.drivers.s3 import BaseS3Connection, S3Connection,\
+    S3StorageDriver, API_VERSION
+
+from libcloud.common.aws import SignedAWSConnection, DEFAULT_SIGNATURE_VERSION
+
+S3_RGW_DEFAULT_REGION = 'default'
+
+S3_RGW_OUTSCALE_HOSTS_BY_REGION =\
+    {'eu-west-1': 'osu.eu-west-1.outscale.com',
+     'eu-west-2': 'osu.eu-west-2.outscale.com',
+     'us-west-1': 'osu.us-west-1.outscale.com',
+     'us-east-2': 'osu.us-east-2.outscale.com',
+     'cn-southeast-1': 'osu.cn-southeast-1.outscale.hk'}
+
+S3_RGW_OUTSCALE_DEFAULT_REGION = 'eu-west-2'
+
+
+class S3RGWConnectionAWS4(SignedAWSConnection, BaseS3Connection):
+    service_name = 's3'
+    version = API_VERSION
+
+    def __init__(self, user_id, key, secure=True, host=None, port=None,
+                 url=None, timeout=None, proxy_url=None, token=None,
+                 retry_delay=None, backoff=None, **kwargs):
+
+        super(S3RGWConnectionAWS4, self).__init__(user_id, key,
+                                                  secure, host,
+                                                  port, url,
+                                                  timeout,
+                                                  proxy_url, token,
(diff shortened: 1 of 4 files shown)


libcloud/storage/drivers/s3.py
```diff
@@ -69,6 +69,8 @@
 
 S3_RGW_OUTSCALE_DEFAULT_REGION = 'eu-west-2'
 
+S3_RGW_DEFAULT_REGION = 'default'
```
Kami: I would prefer to put non-s3 related stuff and RGW driver in a separate file / module (e.g. drivers/rgw.py or similar).

This module should only contain base classes and S3 driver related code.

jmunhoz: @kami makes sense! thanks

jmunhoz: @kami done. All RGW stuff lives in a separated module (drivers/rgw.py) now.

Kami:
@jmunhoz Thanks, pylint step is failing, can you please have a look - https://travis-ci.org/apache/libcloud/builds/130791984

jmunhoz:
@kami I updated the PR to adjust the hierarchy although pylint is expected to fail with this new update. With this PR I am suggesting the following hierarchy:

S3RGWOutscaleStorageDriver -> S3RGWStorageDriver -> S3StorageDriver

The Outscale code is looking for reusing all methods in the RGW driver (_ex_connection_class_kwargs, etc) but the constructor. That's the reason to see pylint failing. In the case of S3RGWOutscaleStorageDriver the code calls the S3RGWStorageDriver's parent instead of the S3RGWOutscaleStorageDriver's parent as expected by pylint. It is also passing 2.x and 3.x series together with lint checks so perhaps the approach could be considered good enough.

jmunhoz:
@kami would it be possible merging this PR as it is and then iterating over this pylint error if needed?

Libcloud is using the PUT + UNSIGNED-PAYLOAD + Auth header combination to sign uploads via S3. This combo is not available in the Ceph RGW's s3-tests in this moment, so having this RGW driver in place would be useful to add the Libcloud coverage in Ceph. This way everybody (s3-tests, developers and users) could use the same upstream code without modifications.

The Libcloud support in Ceph (PUT + UNSIGNED-PAYLOAD + Auth header) was upstream several weeks ago. I commented on this issue. It is available at:

https://github.com/ceph/ceph/pull/8601

Thanks!

Kami:
> @kami I updated the PR to adjust the hierarchy although pylint is expected to fail with this new update. With this PR I am suggesting the following hierarchy: S3RGWOutscaleStorageDriver -> S3RGWStorageDriver -> S3StorageDriver

@jmunhoz I think the correct class-hierarchy should be  S3StorageDriver -> S3RGWStorageDriver -> S3RGWOutscaleStorageDriver.

The reason for that is that S3 is the generic S3 implementation and other S3 based providers should inherit from it (and not vice-versa).

If needed, we could also refactor base S3 functionality into BaseS3Driver and then have the following class hierarchy - BaseS3Driver -> S3Driver, BaseS3Driver -> S3RGWStorageDriver, S3RGWStorageDriver -> S3RGWOutscaleStorageDriver (that's similar to what we do with EC2 driver right now).

libcloud/storage/drivers/rgw.py
```diff
@@ -0,0 +1,137 @@
...
+                             (self.signature_version))
+
+        if self.signature_version == '2':
+            self.connectionCls = S3RGWOutscaleConnectionAWS2
+        elif self.signature_version == '4':
+            self.connectionCls = S3RGWOutscaleConnectionAWS4
+
+        host = S3_RGW_OUTSCALE_HOSTS_BY_REGION[region]
+        self.connectionCls.host = host
+        super(S3RGWStorageDriver, self).__init__(key, secret,
```
Kami, line 134: You should pass `S3RGWOutscaleStorageDriver` as the first argument to `super()`. This will fix pylint issue.

Kami:
Also, the lint issue is actually a very minor one, please see https://github.com/apache/libcloud/pull/786/files#r63909503.

Kami:
I fixed the lint issue and went ahead and merged PR into trunk - thanks!

When you get a chance, please also submit a corresponding PR with tests.

jmunhoz:
@kami Thanks for taking care of this review. I was working on your review comments the last hours but I was not fast enough to update the PR before merging and closing this PR.

I think I have a PR addressing your previous comments now. It has a better design and it is not forced by the kwargs issue.

I reviewed the class-hierarchy stuff and I think we are on the same page. I was using UML notation (maybe not the best choice to communicate my idea in plain text, sorry!) with the following meaning subclass -> superclass. As mentioned I think we are on the same page. The code follows this class hierarchy.

By the way,  I think the lint fix could be breaking the driver cause it will call on the parent constructor without the signature_version parameter (it was pop'ed in the subclass) so the parent will set the default signature (v2). The previous code was jumping over the parent's constructor. It was the reason to lint complaining.

As commented, I will test the link fix and I will open a new PR to fix it if needed together with the improvements suggested by you on the previous revision.

Kami:
@jmunhoz Sounds good & thanks :)

jmunhoz:
@kami I opened a new PR https://github.com/apache/libcloud/pull/792 It fixes the default signature issue.

PR #788, opened 2016-05-16
[LIBCLOUD-820] added check if libvirt uri is local

Katana-Steel:
### Description

without this listing nodes on a remote libvirtd server will fail to
lookup the IPs in the arp cache.
### Status
- done, ready for review
- [X] Documentation

### DIFF
--- a/libcloud/compute/drivers/libvirt_driver.py
+++ b/libcloud/compute/drivers/libvirt_driver.py
@@ -263,11 +263,20 @@ class LibvirtNodeDriver(NodeDriver):
 
         mac_addresses = self._get_mac_addresses_for_domain(domain=domain)
 
-        cmd = ['arp', '-an']
-        child = subprocess.Popen(cmd, stdout=subprocess.PIPE,
+        arp_table = {}
+        try:
+            cmd = ['arp', '-an']
+            child = subprocess.Popen(cmd, stdout=subprocess.PIPE,
                                  stderr=subprocess.PIPE)
-        stdout, _ = child.communicate()
-        arp_table = self._parse_arp_table(arp_output=stdout)
+            stdout, _ = child.communicate()
+            arp_table = self._parse_arp_table(arp_output=stdout)
+        except OSError as e:
+            if e.errno == 2:
+                cmd = ['ip', 'neigh']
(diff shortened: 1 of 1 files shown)


Katana-Steel:
a thing to note running arp on the systems I have access to (CentOS 6+7) requires root access.
maybe iproute2 should be a consideration to get IP from MAC address?

Kami:
@Katana-Steel Thanks!

The change looks good to me.

As far as root requirement goes - I would prefer to stick to a tool which is available by default on most Linux distributions. Having said that, I would be fine with trying to use iproute2 first and falling back to arp if iproute2 is not available.

libcloud/compute/drivers/libvirt_driver.py
```diff
@@ -257,13 +257,27 @@ def _get_ip_addresses_for_domain(self, domain):
...
 
-        cmd = ['arp', '-an']
-        child = subprocess.Popen(cmd, stdout=subprocess.PIPE,
-                                 stderr=subprocess.PIPE)
-        stdout, _ = child.communicate()
-        arp_table = self._parse_arp_table(arp_output=stdout)
+        arp_table = {}
+        try:
+            cmd = ['arp', '-an']
+            child = subprocess.Popen(cmd, stdout=subprocess.PIPE,
```
Kami, line 269: Minor improvement which would also make it easier to test separately would be to refactor this into two methods (e.g. `_get_arp_table_using_arp_cmd` and `_get_arp_table_using_ip_cmd` or something along those lines).

Anyways, not a blocker.

Kami:
Thanks, I went ahead and merged changes into trunk.

I will add some tests in a separate commit (I was the one who originally added the driver without any tests, yeah, bad me!) :)

libcloud/compute/drivers/libvirt_driver.py
Katana-Steel, line 269: Hi Tomaz,

I had a version calling a different function for parsing the arp table of
'ip neighbor' but it looked like code duplication to me. So I rolled it
into the existing one.

I'd be happy to refactor it out into 2 functions. And update the tests when
they get in.

thanks
Rene

PR #791, opened 2016-05-20
Fix delete and create record

beres:
- delete_record, won't work if _others_records are present, now delete it but leave _others_records 
- create_record wont' work if a record with the same id already exists
- fix ex_delete_all_records 

test performed on aws

### DIFF
--- a/libcloud/dns/drivers/powerdns.py
+++ b/libcloud/dns/drivers/powerdns.py
@@ -16,13 +16,14 @@ PowerDNS Driver
 """
 import json
 import sys
+import copy
 
 from libcloud.common.base import ConnectionKey, JsonResponse
 from libcloud.common.exceptions import BaseHTTPError
 from libcloud.common.types import InvalidCredsError, MalformedResponseError
 from libcloud.dns.base import DNSDriver, Zone, Record
 from libcloud.dns.types import ZoneDoesNotExistError, ZoneAlreadyExistsError
-from libcloud.dns.types import Provider, RecordType
+from libcloud.dns.types import Provider, RecordType, RecordDoesNotExistError
 from libcloud.utils.py3 import httplib
 
 __all__ = [
@@ -45,6 +46,7 @@ class PowerDNSResponse(JsonResponse):
         except MalformedResponseError:
             e = sys.exc_info()[1]
             body = '%s: %s' % (e.value, e.body)
+
         try:
             errors = [body['error']]
         except TypeError:
@@ -76,7 +78,7 @@ class PowerDNSDriver(DNSDriver):
     RECORD_TYPE_MAP = {
         RecordType.A: 'A',
         RecordType.AAAA: 'AAAA',
-        RecordType.AFSDB: 'AFSDB',
+        # RecordType.AFSDB: 'AFSDB',
         RecordType.CERT: 'CERT',
         RecordType.CNAME: 'CNAME',
         RecordType.DNSKEY: 'DNSKEY',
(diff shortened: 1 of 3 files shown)


Kami:
Thanks for the contribution.

Can you please provide some more context on this change (since it's a pretty big one)?

beres:
The change involves mainly the creation and deletion process of a record.

If you have a record "foo.bar.com A 127.0.0.1" and you do want to add "foo.bar.com A 10.0.0.1" the method create_record() fails, because does not consider that a record with the same id could be already present, A:foo in this case.

Insted if you already have two record with the same id (A:foo) and you do want to delete one, the method delete_record() fails because does not check _others_records, so it does not correctly format the field data in batch command.

Finally ex_delete_all_records fails because first it creates a delete command for each record, while amazon groups records by id, and secondly the command was not correcly formatted because it does not consider _others_record

The patch modifies _post_changeset in order to correctly consider _other_records

The tests work successfully, but it does not check that cases and in the real world  without the patch they fail.

jvitorsf:
Any updates if this will be merged? I'm having exactly the problem @beres described for Route53

tonybaloney:
@ktdreyer could you please test this change as well please

libcloud/dns/drivers/powerdns.py
```diff
@@ -76,7 +78,7 @@ class PowerDNSDriver(DNSDriver):
     RECORD_TYPE_MAP = {
         RecordType.A: 'A',
         RecordType.AAAA: 'AAAA',
-        RecordType.AFSDB: 'AFSDB',
+        # RecordType.AFSDB: 'AFSDB',
```
ktdreyer, line 81: This AFSDB change looks like an error. Please remove it from your PR. PowerDNS supports AFSDB record types.

libcloud/dns/drivers/powerdns.py
```diff
@@ -182,31 +189,47 @@ def create_record(self, name, zone, type, data, extra=None):
                                              zone.id)
         if extra is None or extra.get('ttl', None) is None:
             raise ValueError('PowerDNS requires a ttl value for every record')
-        record = {
+
+        record_name = self._fmt_name(name)
+        id = ':'.join((self.RECORD_TYPE_MAP[type], record_name))
```
ktdreyer, line 194: Would you please add some explanation in the documentation about this id change? I used "None" for a reason - PowerDNS itself does not expose IDs for records (via the HTTP API, anyway).

libcloud/dns/drivers/powerdns.py
```diff
@@ -428,8 +470,85 @@ def update_record(self, record, name, type, data, extra=None):
...
+        try:
+            record = self.ex_filter_records(zone.list_records(), id=record_id)[0]
+        except:
+            raise RecordDoesNotExistError(value='', driver=self,
+                                            record_id=record_id)
+        return record
+
+    def ex_filter_records(self, records, **kwargs):
+        """
+        Given a list of records, it filter by record attribute passed as kwargs
```
ktdreyer, line 489: "it filter by" does not make sense to me. Mind rephrasing this?

ktdreyer:
For PowerDNS, it would be nice to have comments in each new private method that describe what these methods do. It's hard to tell at a glance the difference between `_to_records_3x()` and `_to_records_4x()` for example, and comments would really make this part more readable.

libcloud/dns/drivers/powerdns.py
beres, line 194: You are right, PowerDNS does not expose ID via HTTP api, however even Route53 does not have an ID, but the driver generates it with the same method. I've used the same technique because I found it useful.

PR #792, opened 2016-05-20
Ceph S3 Outscale storage driver fixes

jmunhoz:
- Simplify the Outscale driver
- Remove unused Outscale connection classes
- Fix default signature v2 issue with the Outscale driver

### DIFF
--- a/libcloud/storage/drivers/rgw.py
+++ b/libcloud/storage/drivers/rgw.py
@@ -72,12 +72,16 @@ class S3RGWStorageDriver(S3StorageDriver):
 
     def __init__(self, key, secret=None, secure=True, host=None, port=None,
                  api_version=None, region=S3_RGW_DEFAULT_REGION, **kwargs):
+
         if host is None:
             raise LibcloudError('host required', driver=self)
-        self.name = 'Ceph RGW S3 (%s)' % (region)
+
+        self.name = kwargs.pop('name', None)
+        if self.name is None:
+            self.name = 'Ceph RGW S3 (%s)' % (region)
+
         self.ex_location_name = region
         self.region_name = region
-
         self.signature_version = str(kwargs.pop('signature_version',
                                                 DEFAULT_SIGNATURE_VERSION))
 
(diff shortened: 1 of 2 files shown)


Kami:
@jmunhoz Thanks!

Can you please also add some tests - at least the one which instantiate both drivers and verify that host, etc. on the connection class is set correctly (this should be easy to do and be a bare minimum).

jmunhoz:
@kami I pushed some test cases covering both drivers.

Kami:
Thanks.

I've added some additional asserts (084ce3cf7eaeeb4bd6bb0b5472b9f75ca59ec6cd) and merged changes into trunk. I believe there is still a bug with they way the host is handled though:

``` python
       ....
        self.driver_v2 = self.driver_type(*self.driver_args,
                                          signature_version='2',
                                          host='host1')
        self.driver_v4 = self.driver_type(*self.driver_args,
                                          signature_version='4',
                                          host='host2')

        print self.driver_v2.connectionCls.host
        print self.driver_v4.connectionCls.host
```

And when running it:

``` bash
osu.eu-west-2.outscale.com
osu.eu-west-2.outscale.com
```

As you can see, the host argument is ignored. That's probably because when `connectionCls` is instantiated, the `host` argument you set in the driver class constructor is overridden - as mentioned before, I recommend you to use `_ex_connection_class_kwargs` method. Overriding class variables is usually not a good idea, because it results in bugs and issues like that.

It would be great if you can look into it and also add corresponding tests (I didn't include those tests in the commit so it would cause the build to fail). When working on a fix, you can start with a failing test case :)

jmunhoz:
Thanks @kami I reviewed the test case and I think it is working as expected. The Outscale driver maps valid Outscale regions to the Outscale hosts. It should not let arbitrary hosts.

In this case, if you don't add any explicit valid region the Outscale driver works with the default region and it assigns the default host 'osu.eu-west-2.outscale.com'. The default region/host is the same for AWS2 and AWS4. It is the reason to see the same default host twice.

You can have a look in this line https://github.com/apache/libcloud/blob/trunk/libcloud/storage/drivers/rgw.py#L116

Maybe it makes sense raising an expection when host is not none? Something similar to:

<pre>
   if region not in S3_RGW_OUTSCALE_HOSTS_BY_REGION:
      raise LibcloudError('Unknown region (%s)' % (region), driver=self)
   if host is not None:
      raise LibcloudError('This driver does not accept arbitrary host', driver=self)
   host = S3_RGW_OUTSCALE_HOSTS_BY_REGION[region]
</pre>

PR #793, opened 2016-05-26
[WIP] Support POST requests in cloudstack create_node (LIBCLOUD-822)

sfriesel:
### Description

This allows sending longer parameter values (like ex_user_data) that
would not fit into the request line limit.
### Status

This is an initial proposal, tests are not adapted yet but I tested it manually. I'm not really happy with the abstraction layers yet.
- [ ] Documentation

### DIFF
--- a/libcloud/compute/drivers/cloudstack.py
+++ b/libcloud/compute/drivers/cloudstack.py
@@ -18,6 +18,7 @@ from __future__ import with_statement
 import sys
 import base64
 import warnings
+import functools
 
 from libcloud.utils.py3 import b
 from libcloud.utils.py3 import urlparse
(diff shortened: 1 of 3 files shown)


sfriesel:
open questions:
- The cloudstack documentation is quite handwave-y about POST [1]; probably because tomcat treats urlencoded POST requests transparently [2]. So I'm not sure if it makes sense to expose the HTTP verb to the user or if libcloud should just do The Right Thing automatically
- In `common.cloudstack.CloudStackConnection`, `_sync_request` and `_async_request` are basically treated like overrides of `request` and `async_request` but without actually using the same name (`_sync_request` even gets defined as `request_method`). Is this intended and just the names are different for historical reasons? Or are they meant to be "higher-level" request methods? In the former case I would propose renaming them as they are private anyway; in the latter case I would simplify their interface so the caller doesn't need to set all of `method`, `data` and `headers` to get a working POST request

[1] http://docs.cloudstack.apache.org/en/latest/dev.html#making-api-requests
[2] https://github.com/apache/tomcat/blob/e5168242ad7600e19ac14eef87aced2cd6ab7325/java/org/apache/catalina/connector/Request.java#L3179

tonybaloney:
cc @runseb

Kami:
@sfriesel How is this PR looking?

It's great if we could get it finished (get the tests passing, etc.) so we can merge it into trunk.

libcloud/compute/drivers/cloudstack.py
```diff
@@ -1570,14 +1571,24 @@ def create_node(self, **kwargs):
...
+
         :rtype:     :class:`.CloudStackNode`
         """
 
         server_params = self._create_args_to_params(None, **kwargs)
 
-        data = self._async_request(command='deployVirtualMachine',
-                                   params=server_params,
-                                   method='GET')['virtualmachine']
+        method = kwargs.get('ex_http_method', 'GET')
```
Kami, line 1585: Does it make sense to make this user-configurable?

Can we just always use POST, or will this cause issues with some CloudStack installations?

Kami:

PR #795, opened 2016-05-27
Better support for Dimension Data private cloud

tonybaloney:
Improved the support for Dimension Data private cloud and public cloud driver instantiation


Kami:
Wait, so private cloud is actually a thing? :trollface:

libcloud/backup/drivers/dimensiondata.py
```diff
@@ -1,690 +1,691 @@
-# Licensed to the Apache Software Foundation (ASF) under one or more
```
Kami, line 1: What happened to this file? Did indentation change from spaces to tabs or similar?

tonybaloney, line 1: shrug. which line ending configuration are you using? I've got autoclrf = false in git, but the file is CLRF

PR #797, opened 2016-06-02
Fix cloudstack user data on Python 3

sfriesel:
### Description

A wrongly nested string decode caused a TypeError as b64encode
expects a byte string in Python 3 and conversely the url parameter
needs to be a normal string.
### Status

done
- [ ] Documentation

### DIFF
--- a/libcloud/test/compute/test_cloudstack.py
+++ b/libcloud/test/compute/test_cloudstack.py
@@ -221,6 +221,18 @@ class CloudStackCommonTestCase(TestCaseMixin):
         self.assertEqual(node.name, 'test')
         self.assertEqual(node.extra['key_name'], 'foobar')
 
+    def test_create_node_ex_userdata(self):
+        size = self.driver.list_sizes()[0]
+        image = self.driver.list_images()[0]
+        location = self.driver.list_locations()[0]
+        CloudStackMockHttp.fixture_tag = 'deploykeyname'
+        node = self.driver.create_node(name='test',
+                                       location=location,
+                                       image=image,
+                                       size=size,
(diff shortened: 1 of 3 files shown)


Kami:
Good catch, thanks!

libcloud/test/compute/test_cloudstack.py
```diff
@@ -221,6 +221,18 @@ def test_create_node_ex_keyname(self):
         self.assertEqual(node.name, 'test')
         self.assertEqual(node.extra['key_name'], 'foobar')
 
+    def test_create_node_ex_userdata(self):
```
Kami, line 224: Is it also possible to add a test / assert which verifies that user data is encoded correctly (you will need to do this inside the corresponding mock http class method).

sfriesel, line 224: sure, I'll look into it

sfriesel:
@Kami added the assertion; it seems odd that it is so far away from the test, is that how it's supposed to be used?

sfriesel:
@Kami is there still hope this fix makes it into 1.0? :)

libcloud/test/compute/test_cloudstack.py
```diff
@@ -1269,6 +1282,11 @@ def _test_path(self, method, url, body, headers):
             body, obj = self._load_fixture(fixture)
             return (httplib.OK, body, obj, httplib.responses[httplib.OK])
 
+    def _test_path_userdata(self, method, url, body, headers):
```
tonybaloney, line 1285: :+1:

tonybaloney:
thanks @sfriesel merging

PR #798, opened 2016-06-02
GCE: Add support for creating multiple nodes on a subnetwork

jbaublitz:
### Description

Currently for GCE, the multi node creation API does not support creating multiple nodes on a subnetwork. I have made changes to the ex_create_multiple_nodes function and the functions it calls to support creation of nodes on a given subnetwork. If subnetwork is None, the changes fall back to legacy network behavior on GCE so it is backwards compatible. There was one change to _create_node_req that was needed to get disk types which is simply passing in the location provided as an argument to ex_get_disktype.
### Status
- ready for review
- [ ] Documentation

### DIFF
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -2590,12 +2590,11 @@ class GCENodeDriver(NodeDriver):
             size = self.ex_get_size(size, location)
         if not hasattr(ex_network, 'name'):
             ex_network = self.ex_get_network(ex_network)
-        if ex_subnetwork:
-            if not hasattr(ex_subnetwork, 'name'):
-                ex_subnetwork = \
-                    self.ex_get_subnetwork(ex_subnetwork,
-                                           region=self._get_region_from_zone(
-                                               location))
+        if ex_subnetwork and not hasattr(ex_subnetwork, 'name'):
+            ex_subnetwork = \
+                self.ex_get_subnetwork(ex_subnetwork,
+                                       region=self._get_region_from_zone(
+                                           location))
         if ex_image_family:
             image = self.ex_get_image_from_family(ex_image_family)
         if image and not hasattr(image, 'name'):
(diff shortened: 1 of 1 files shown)


Kami:
Thanks.

LGTM.

/cc @erjohnso

libcloud/compute/drivers/gce.py
```diff
@@ -5158,7 +5166,8 @@ def _create_node_req(self, name, size, image, location, network=None,
             if not ex_disk_type:
                 ex_disk_type = 'pd-standard'
             if not hasattr(ex_disk_type, 'name'):
-                ex_disk_type = self.ex_get_disktype(ex_disk_type)
+                ex_disk_type = self.ex_get_disktype(ex_disk_type,
```
Kami, line 5167: I assume this is a bug-fix? :)

jbaublitz, line 5167: Yes from what I can tell looking at the code, it is. :)

jbaublitz, line 5167: When I looked at the code, it seems passing in None for zone will cause the disk type not to be detected so my code didn't work without that.

jbaublitz:
I have made one final change to clean up an unnecessary if statement.

erjohnso:
Thanks for this @jbaublitz!

@supertom - do you mind looking this over and testing it out?

Kami:
@erjohnso @supertom How is this change looking, can we get it merged (change looks OK to me)?

libcloud/compute/drivers/gce.py
```diff
@@ -2802,6 +2802,12 @@ def ex_create_multiple_nodes(self, base_name, size, image, number,
             size = self.ex_get_size(size, location)
         if not hasattr(ex_network, 'name'):
             ex_network = self.ex_get_network(ex_network)
+        if ex_subnetwork:
+            if not hasattr(ex_subnetwork, 'name'):
```
supertom: Can we flatten this if statement as we did in create_node?

jbaublitz: Sure thing. I missed that and will push that when I get the chance.

jbaublitz:
@supertom Not enough coffee... I pushed with a syntax error. The next build should be fine. Anything else I should look at?

supertom:
@jbaublitz LGTM.  Thanks for the contribution!

jbaublitz:
@Kami @supertom I read over the libcloud development guide again and it seems that I need to submit a patch and tests for changes. I can definitely squash the commits and submit the patch but do these changes need any tests?

PR #799, opened 2016-06-04
refactored _parse_arp_table

Katana-Steel:
refactoring of the arp table parsing for libvirt
### Description

passing each command output to their own parsing function
and add unittest for parsing the tables
### Status
- done, ready for review
- [X] Documentation

split _parse_arp_table into 1 general and 2 specialized functions

### DIFF
--- a/libcloud/compute/drivers/libvirt_driver.py
+++ b/libcloud/compute/drivers/libvirt_driver.py
@@ -269,15 +269,14 @@ class LibvirtNodeDriver(NodeDriver):
             child = subprocess.Popen(cmd, stdout=subprocess.PIPE,
                                      stderr=subprocess.PIPE)
             stdout, _ = child.communicate()
-            arp_table = self._parse_arp_table(arp_output=stdout)
+            arp_table = self._parse_ip_table_arp(arp_output=stdout)
         except OSError as e:
             if e.errno == 2:
                 cmd = ['ip', 'neigh']
                 child = subprocess.Popen(cmd, stdout=subprocess.PIPE,
                                          stderr=subprocess.PIPE)
                 stdout, _ = child.communicate()
-                arp_table = self._parse_arp_table(arp_output=stdout,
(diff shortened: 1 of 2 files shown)


Katana-Steel:
I looked at other compute driver tests, but wasn't able to determine how to write a stub test for this driver

Kami:
Thanks for doing that!

As far as the tests go - it will be a different in this case since libvirt driver uses libvirt library and doesn't talk directly to an HTTP API.

One thing which would help is if you can gist (or paste it here) the output of both commands (arp and ip neigh) - this way we can use this as fixtures for ip table parse test case.

Katana-Steel:
I've only obfuscated the IP addresses a little, and only included the first 4 entries...
I think that should be enough to see the pattern

```
~ $ ip neighbor 
1.2.10.80 dev br0 lladdr 52:54:00:bc:f9:6c STALE
1.2.10.33 dev br0 lladdr 52:54:00:04:89:51 REACHABLE
1.2.10.97 dev br0 lladdr 52:54:00:c6:40:ec DELAY
1.2.10.40 dev br0 lladdr 52:54:00:77:1c:83 STALE

~ # arp -an
? (1.2.10.80) at 52:54:00:bc:f9:6c [ether] on br0
? (1.2.10.33) at 52:54:00:04:89:51 [ether] on br0
? (1.2.10.97) at 52:54:00:c6:40:ec [ether] on br0
? (1.2.10.40) at 52:54:00:77:1c:83 [ether] on br0
```

Katana-Steel:
@Kami  added the command output

Katana-Steel:
@Kami I've added a skeleton for a libvirt_driver test class with 2 actual tests.
Let me know what you think

Kami:
Thanks and sorry for the delay - I will look into it asap.

tonybaloney:
Changes LGTM :+1:

PR #803, opened 2016-06-06
loadbalancer/rackspace/list_balancers: add params

raittes:
### Description

Allow to set params to be submitted in the query string to API request.
By default, Rackspace API [return a maximum of 100 items at a time](https://developer.rackspace.com/docs/cloud-load-balancers/v1/developer-guide/#paginated-collections).
This change allow to set "offset" in the params, to paginate requests.
Ex. `list_balancers(ex_params={'offset': 100})`
### Status
- done, ready for review
### Checklist
- [X] [Code linting](http://libcloud.readthedocs.org/en/latest/development.html#code-style-guide) 
- [X] Documentation

### DIFF
--- a/libcloud/loadbalancer/drivers/rackspace.py
+++ b/libcloud/loadbalancer/drivers/rackspace.py
@@ -349,7 +349,8 @@ class RackspaceLBDriver(Driver, OpenStackDriverMixin):
         return self._to_protocols_with_default_ports(
             self.connection.request('/loadbalancers/protocols').object)
 
-    def list_balancers(self, ex_member_address=None):
+    def list_balancers(self, ex_member_address=None, ex_status=None,
+                       ex_changes_since=None, ex_params={}):
         """
         @inherits: :class:`Driver.list_balancers`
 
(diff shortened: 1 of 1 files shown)


tonybaloney:
thanks for the contribution @raittes - libcloud convention is that any arguments which are not present in the base class methods are prepending with `ex_`.
Please could you rename the parameter to ex_params.

Ideally, there would be named keyword arguments for typical arguments, according to the docs https://developer.rackspace.com/docs/cloud-load-balancers/v1/developer-guide/#list-load-balancers they would include **status**, **nodeaddress** and **changes-since**.
With a simple dictionary param, it's leaving it to the developer consuming to have to read the API documentation.

raittes:
Hi @tonybaloney, I've updated the PR, please check it out!
I've included the params for /loadbalancers, but also keep the ex_params, to allow setting general api params like limit and offset.
Thx!

raittes:
@tonybaloney, something is missing?

libcloud/loadbalancer/drivers/rackspace.py
```diff
@@ -349,20 +349,42 @@ def ex_list_protocols_with_default_ports(self):
...
         if ex_member_address:
             params['nodeaddress'] = ex_member_address
 
+        if ex_status:
+            params['status'] = ex_status
+
+        if ex_changes_since:
+            params['changes-since'] = ex_changes_since
+
+        for key, value in ex_params.iteritems():
```
tonybaloney: `dict.iteritems()` is not valid in Python 3, you must use items().

Which probably highlights that this code doesn't have a supporting unit test as TravisCI should have picked that up. please can you add a test.

raittes: Fixed!

Tests are ok, I forced push and TravisCI may have missed something.
Here another build: https://travis-ci.org/raittes/libcloud/jobs/140894397

```
ERROR: test_destroy_balancer (libcloud.test.loadbalancer.test_rackspace.RackspaceLBTests)
  File "/home/raittes/projects/raittes/libcloud/libcloud/test/loadbalancer/test_rackspace.py", line 187, in test_destroy_balancer
    balancer = self.driver.list_balancers()[0]
  File "/home/raittes/projects/raittes/libcloud/libcloud/loadbalancer/drivers/rackspace.py", line 385, in list_balancers
    for key, value in ex_params.iteritems():
AttributeError: 'dict' object has no attribute 'iteritems'
```

tonybaloney:
the commit came in after your comment, but the build passed now that the py3 compatibility issue was fixed up

PR #804, opened 2016-06-06
auroradns: Set 'priority' key in extra dict instead of 'prio'.

wido:
The AuroraDNS API returns the DNS record priority as the field 'prio', but libcloud expects 'priority'.

In the AuroraDNS driver set the 'priority' key in the extra dictionary when parsing return values from the AuroraDNS API.

Also add a additional test to verify this behavior.


tonybaloney:
Change look good. thanks for the contribution :+1:

PR #808, opened 2016-06-07
[LIBCLOUD-827] Add support for automatic SNI

lochiiconnectivity:
### Description

Adds support for automatic SNI using the hostname
supplied to connect to.

Python SSL has had SNI support for around six years
(https://hg.python.org/cpython/rev/846c0e1342d0/) 

We need to convert old ssl socket wrap into a context
to effect this.
### Status
- done, ready for review
- [N/A] Documentation
- [N/A] [ICLA](http://libcloud.readthedocs.org/en/latest/development.html#contributing-bigger-changes) (required for bigger changes)

### DIFF
--- a/libcloud/httplib_ssl.py
+++ b/libcloud/httplib_ssl.py
@@ -292,21 +292,50 @@ class LibcloudHTTPSConnection(httplib.HTTPSConnection, LibcloudBaseConnection):
 
         ssl_version = libcloud.security.SSL_VERSION
 
+        # Connect, using contexts and SNI if we can
+
         try:
-            self.sock = ssl.wrap_socket(
-                sock,
-                self.key_file,
-                self.cert_file,
-                cert_reqs=ssl.CERT_REQUIRED,
-                ca_certs=self.ca_cert,
-                ssl_version=ssl_version)
-        except socket.error:
-            exc = sys.exc_info()[1]
-            # Re-throw an exception with a more friendly error message
-            exc = get_socket_error_exception(ssl_version=ssl_version, exc=exc)
-            raise exc
+            self.tls_context = ssl.SSLContext(ssl_version)
+            self.tls_context.verify_mode = ssl.CERT_REQUIRED
+            if self.cert_file and self.key_file:
+                self.tls_context.load_cert_chain(
+                    self.cert_file, self.key_file, None)
+            if self.ca_cert:
+                self.tls_context.load_verify_locations(self.ca_cert)
+        except AttributeError:
+            self.tls_context = None
+
+        if self.tls_context and ssl.HAS_SNI:
+            try:
+                self.sock = self.tls_context.wrap_socket(
+                    sock,
+                    server_hostname=self.host,
+                )
+            except:
+                exc = sys.exc_info()[1]
+                exc = get_socket_error_exception(ssl_version=ssl_version,
+                                                 exc=exc)
+                raise exc
+        else:
+            try:
+                self.sock = ssl.wrap_socket(
+                    sock,
+                    self.key_file,
+                    self.cert_file,
+                    cert_reqs=ssl.CERT_REQUIRED,
+                    ca_certs=self.ca_cert,
+                    ssl_version=ssl_version
+                )
+            except:
+                exc = sys.exc_info()[1]
+                exc = get_socket_error_exception(ssl_version=ssl_version,
+                                                 exc=exc)
+                raise exc
 
         cert = self.sock.getpeercert()
+
+        # Verify Hostname
+
         try:
             match_hostname(cert, self.host)
         except CertificateError:
--- a/libcloud/httplib_ssl.py
+++ b/libcloud/httplib_ssl.py
@@ -292,9 +292,10 @@ class LibcloudHTTPSConnection(httplib.HTTPSConnection, LibcloudBaseConnection):
 
         ssl_version = libcloud.security.SSL_VERSION
 
-        # Connect, using contexts and SNI if we can
+        # If we support SNI, use SSLContext and tls_context.wrap_socket()
+        # else revert to older behaviour with ssl.wrap_socket()
 
-        try:
+        if getattr(ssl, 'HAS_SNI', False):
             self.tls_context = ssl.SSLContext(ssl_version)
             self.tls_context.verify_mode = ssl.CERT_REQUIRED
             if self.cert_file and self.key_file:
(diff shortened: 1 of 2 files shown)


lochiiconnectivity:
Need to modify test_connect_throws_friendly_error_message_on_ssl_wrap_connection_reset_by_peer

Kami:
Thanks a lot for this contribution!

I thought we already support SNI, but it looks like we doesn't.

Since this code touches security related code, it will need thorough reviews and also extensive tests (I see you updated existing tests which is great, but we also need additional tests coverage which cover those new changes - including all the common and edge cases).

Having said that, we are working on replacing our http layer with the `requests` library in the near feature (https://github.com/apache/libcloud/pull/728). We should decide if it's worth spending a lot more time on this change or we should instead focus on the requests change which is the future for us.

Again, thanks a lot.

lochiiconnectivity:
Thanks for taking a look, we need SNI support pretty quickly to support some some compute platforms for our customers , which are now adopting SNI-only API endpoints, so its pretty urgent / important for us. 

The patch is really simple, it is to start using the SSLContext wrapper introduced in python 2.7. In the [security considerations](https://docs.python.org/3/library/ssl.html#ssl-security) statement, they recommend using a default context (_create_default_context_), but doing so would significantly change the behaviour of the ssl functionality today, by specifying defaults that the user didn't have before (so really, nothing has changed).

If the system is not supporting SSLContext (i.e, python 2.6), then we fall back to the previous method.

With regards to tests, can you tell me what additional coverage you need, bearing in mind the interface , which produces a wrap_socket, is practically identical? 

Also, with regards to moving to requests, that's great news, but unfortunately, we need this functionality now, and can't really wait... I hope you understand.

lochiiconnectivity:
Hi Kami, do you have an update on this P/R?

tonybaloney:
:+1: from me @Kami I went over this with our resident security guy yesterday.

libcloud/httplib_ssl.py
```diff
@@ -292,21 +292,50 @@ def connect(self):
...
-                self.cert_file,
-                cert_reqs=ssl.CERT_REQUIRED,
-                ca_certs=self.ca_cert,
-                ssl_version=ssl_version)
-        except socket.error:
-            exc = sys.exc_info()[1]
-            # Re-throw an exception with a more friendly error message
-            exc = get_socket_error_exception(ssl_version=ssl_version, exc=exc)
-            raise exc
+            self.tls_context = ssl.SSLContext(ssl_version)
```
Kami, line 299: Does this need to me an instance variable (e.g. for ease of testing or some other reason) or can we just make it method local?

libcloud/httplib_ssl.py
```diff
@@ -292,21 +292,50 @@ def connect(self):
...
+            self.tls_context.verify_mode = ssl.CERT_REQUIRED
+            if self.cert_file and self.key_file:
+                self.tls_context.load_cert_chain(
+                    self.cert_file, self.key_file, None)
+            if self.ca_cert:
+                self.tls_context.load_verify_locations(self.ca_cert)
```
Kami, line 305: Is this functionality also supported in older versions of Python (2.6 and 2.5) by default?

libcloud/httplib_ssl.py
```diff
@@ -292,21 +292,50 @@ def connect(self):
...
+        except AttributeError:
+            self.tls_context = None
+
+        if self.tls_context and ssl.HAS_SNI:
```
Kami: Hm, looks like `HAS_SNI` is only available in Python >= 2.7.0 (https://docs.python.org/2/library/ssl.html#ssl.HAS_SNI), but we still support older versions (2.6 at the very least).

Or am I missing something?

Kami: If that's the case, we need to make code more robust and only use that functionality in newer versions of Python when available.

That way we don't break Libcloud's compatibility with older versions of Python.

Kami: Actually, taking a look at it again, I believe `self.tls_context` should already be `None` when used with older version of Python, right?

If so that's fine, but we should add some comment to the code which clarifies that. And maybe also doing `has_sni = getattr(sssl, 'HAS_SNI', False)` is also not such a bad idea :)

Kami:
@lochiiconnectivity Sorry for the delay.

I've added some comments, looks like there might be some Python version compatibility issues.

If we can address that (make sure it also works on older Python versions we support) and if we add tests for all the edge cases, I'm fine with merging this :+1:

libcloud/httplib_ssl.py
Kami: And maybe even put this check at the top and only try to instantiate `ssl.SSLContext`, etc. if `has_sni is True`.

I know this is a style / preference thing, but I found current approach a bit confusing / hard to understand. Making it very clear what is going on (we only try to instantiate `ssl.SSLContext` and use SNI when `has_sni` is True aka this functionality is available, etc.) would make it more clear, imo.

And I think this is especially important since it's security sensitive code - we should try to make it as simple and obvious / easy to read and understand as possible :)

Kami:
@lochiiconnectivity I know this is not a big change, but I would still ask you if you can please file an ICLA - https://libcloud.readthedocs.io/en/latest/development.html#contributing-bigger-changes

Thanks!

lochiiconnectivity:
ICLA should already be in place from the Abiquo contribution, no?

lochiiconnectivity:
regarding the style change, will address this today and re-submit

Kami:
@lochiiconnectivity I didn't found one for "David Freedman" on ASF site, but I could be wrong :)

Under which name and email did you submit your ICLA?

lochiiconnectivity:
that's me.

lochiiconnectivity:
pretty sure I filled this out already but ill do another one just in case.

Kami:
Thanks.

To clarify it, I looked at https://people.apache.org/unlistedclas.html and I couldn't find one in your name.

lochiiconnectivity:
Ok, I've pushed the style change, the original was meant to use SSLContext even if SNI wasn't available, but now both have to be used together, as you've suggested.
I can't test this on python 2.5 as tox doesn't support it any more , and your CI (travis) doesn't test it. I'm presuming that this means you will be dropping support for it?
I have submitted another ICLA.

Kami:
@tonybaloney Probably best to disable those coveralls comments since they can be a bit annoying - can you please look into it?

@lochiiconnectivity Thanks, I will look into it tomorrow. It looks like we are still missing more unit tests though :)

As far as Python 2.5 support goes, yeah it's hard to keep it continuously tested since most services dropped support for it.

lochiiconnectivity:
coveralls broke travis for 2.7, looks to be transient, trying again

lochiiconnectivity:
with regards to the unit tests, now we activate contexts only on the presence of ssl.HAS_SNI, so the logic in the socket wrapping in httplib_ssl.py has been updated to do this, and the logic in test_httplib_ssl.py has been updated to mock the sslcontext wrap_socket in the presence of ssl.HAS_SNI also (so the test is aligned with the real implementation). There are no additional SNI tests, because we are not connecting to a real endpoint, we can't simulate this.

Kami:
I made a minor change to use keyword argument when calling the method, etc. and merged change into trunk.

I will also look if it's possible to add some more tests.

Thanks again!

PR #810, opened 2016-06-10
 Aliyun ecs: Add support for creating security group

netgenius18:
### Description

Currently for Aliyun ECS, if you did not create a security group through web browser, you would encounter 'list index out of range' error when running example_aliyun_ecs.py to create a VM instance. 
# 

Found 0 security groups
  File "example_aliyun_ecs.py", line 54, in <module>
    sg = sgs[0]
IndexError: list index out of range
# 

So I add the ex_create_security_group function in ecs.py to solve it.


libcloud/compute/drivers/ecs.py
```diff
@@ -768,6 +768,27 @@ def ex_stop_node(self, node, ex_force_stop=False):
         return resp.success() and \
             self._wait_until_state([node], NodeState.STOPPED)
 
+    def ex_create_security_groups(self, ex_description=None,
```
Kami: For consistency with other drivers, please name it `ex_create_security_group` (remove `s` suffix) and remove `ex_` prefix from argument names since method name is already prefixed.

This way signature should similar to the AWS one -> `def ex_create_security_group(self, description=None, client_token=None)`

netgenius18:
I am sorry, @ Kami. Now,  it was renamed ex_create_security_group(self, description=None, client_token=None), please check it

PR #812, opened 2016-06-14
Add certifi CA bundle path to Libcloud CA bundle search path if certifi is available

Kami:
This change will add certifi CA bundle path to Libcloud CA bundle search path if `certifi` library (https://pypi.python.org/pypi/certifi) is installed and available in the virtual environment / PYTHONPATH where Libcloud is used.

This path is injected into the CA bundle search path which means it has precedence over other paths.

This behavior is enabled by default, but it can be disabled by setting `LIBCLOUD_SSL_USE_CERTIFI` environment variable to `false` or any other non-truthy value.

Some context can be found in this PR - https://github.com/conda-forge/staged-recipes/pull/694, but even without it, I think that's a good change until we switch to `requests`.
## TODO
- [ ] Docs

### DIFF
--- a/libcloud/security.py
+++ b/libcloud/security.py
@@ -36,6 +36,10 @@ VERIFY_SSL_CERT = True
 
 SSL_VERSION = ssl.PROTOCOL_TLSv1
 
+# True to use certifi CA bundle path when certifi library is available
+USE_CERTIFI = os.environ.get('LIBCLOUD_SSL_USE_CERTIFI', True)
+USE_CERTIFI = str(USE_CERTIFI).lower() in ['true', '1']
+
 # File containing one or more PEM-encoded CA certificates
 # concatenated together.
 CA_CERTS_PATH = [
@@ -60,6 +64,21 @@ CA_CERTS_PATH = [
     # opensuse/sles: openssl
     '/etc/ssl/certs/YaST-CA.pem',
 ]
+
+# Insert certifi CA bundle path to the front of Libcloud CA bundle search
+# path if certifi is available
+try:
+    import certifi
+except ImportError:
+    has_certifi = False
+else:
+    has_certifi = True
+
+if has_certifi and USE_CERTIFI:
+    certifi_ca_bundle_path = certifi.where()
+
+    if certifi_ca_bundle_path not in CA_CERTS_PATH:
+        CA_CERTS_PATH.insert(0, certifi_ca_bundle_path)
 
 # Allow user to explicitly specify which CA bundle to use, using an environment
 # variable
--- a/libcloud/security.py
+++ b/libcloud/security.py
@@ -76,9 +76,7 @@ else:
 
 if has_certifi and USE_CERTIFI:
     certifi_ca_bundle_path = certifi.where()
-
-    if certifi_ca_bundle_path not in CA_CERTS_PATH:
-        CA_CERTS_PATH.insert(0, certifi_ca_bundle_path)
+    CA_CERTS_PATH.insert(0, certifi_ca_bundle_path)
 
 # Allow user to explicitly specify which CA bundle to use, using an environment
 # variable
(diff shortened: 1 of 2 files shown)


libcloud/security.py
```diff
@@ -61,6 +65,19 @@
...
+try:
+    import certifi
+except ImportError:
+    has_certifi = False
+else:
+    has_certifi = True
+
+if has_certifi and USE_CERTIFI:
+    certifi_ca_bundle_path = certifi.where()
+    CA_CERTS_PATH.insert(0, certifi_ca_bundle_path)
```
jakirkham, line 79: Should one issue a warning if the user sets `USE_CERTIFI` to `true`, but we can't find it?

Kami, line 79: The thing is that `USE_CERTIFI` actually defaults to True right now. So in this case issuing a warning would be confusing.

I'm personally not totally opinionated on this one (yet) and I could be convinced to default it to False and issue a warning if set to True.

Kami, line 79: Defaulting it to False would require user interaction which is not ideal, but then again Python packaging / virtual environment world is far from ideal (ideally, every app, etc. would always run in an isolated virtual environment, but in real life that's not always the case) so making it explicit opt-in and defaulting to False might indeed be a better choice (this way it will be an explicit choice and people won't be surprised).

jakirkham, line 79: Sorry, this was more meant to be a question and not a suggestion.

> The thing is that `USE_CERTIFI` actually defaults to True right now. So in this case issuing a warning would be confusing.

Good point.

> I'm personally not totally opinionated on this one (yet) and I could be convinced to default it to False and issue a warning if set to True.

Personally, I prefer the current behavior (for somewhat selfish reasons). Basically, `conda` has activate scripts that one can use to set environment variables, but they are a bit flaky. There has been a lot of work recently to improve them and I do expect they will get better. Unfortunately, they just are not there yet.

While I can understand the need for explicit over implicit, `certifi` ends up being a simple and reliable way to provide what you are looking for here IIUC. Even `setuptools` provides it as an [option](https://github.com/pypa/setuptools/blob/a73ca4fdcd1ba386159eb6ceb66d40af674a034c/setup.py#L147). So, I think trying to use that first by default and searching elsewhere is completely reasonable.

tonybaloney, line 79: So do we have an agreement to proceed with this change @jadunham1 @Kami ?

tonybaloney, line 79: I think it would make sense to issue that warning alongside the warning that no SSL certificates could be found (in libcloud.security module)

jadunham1, line 79: I'd prefer having certifi as a default if it's present and not show the error message.
I'd be fine with just updating https://github.com/apache/libcloud/blob/trunk/docs/other/ssl-certificate-validation.rst if the ssl certs can be found as that is what it will point to anyways.

tonybaloney, line 79: Thanks @jadunham1 I meant @jakirkham but your input is always valued!

tonybaloney, line 79: 👍 for @jadunham1's selection

Jc2k:
Have you seen https://github.com/certifi/python-certifi/issues/26? Is it worth adding some docs/release notes about expected behaviour with older openssl?

jakirkham:
At conda-forge, we are using `openssl` 1.0.2 currently. Same story with `defaults`. So, if they are using a `conda` install, they should be fine.

If not, you could try using `old_where` in some places or provide some option to set that to the default.

libcloud/security.py
jakirkham, line 79: I wasn't sure if that was a typo or not, @tonybaloney. So, I stayed quiet.

Yes, 👍 on this change.

tonybaloney:
OK. Merging. This is holding up 1.0 btw.

jakirkham:
So, another way of checking for an ssl certificate that may be of value would be to query `ssl.get_default_verify_paths()`. In particular, `ssl.get_default_verify_paths().cafile` and `ssl.get_default_verify_paths().openssl_cafile`. It seems the former (but not the latter 😕) can be overridden by setting the environment variable `SSL_CERT_FILE`.

Kami:
@jakirkham Thanks, I will look into it.

IIRC, some of that functionality is only available in newer versions of Python so we also need to be careful with that (we still support 2.5 and 2.6).

PR #813, opened 2016-06-15
[LIBCLOUD-826] [GCE]: Improve performance of list nodes by caching volume information

supertom:
### Description

When listing nodes, the GCE driver currently calls the disk API for each disk attached to the node.  This PR changes that behavior by using the aggregatedLIst call for disks (once per list_node request) and using that information to provide disk details.

See sample performance info in [LIBCLOUD-826](https://issues.apache.org/jira/browse/LIBCLOUD-826)
#### Implementation:

For disk information, aggregated calls are now always used and the disk information is stored in a dictionary, called 'volume_dict'.  If the user would like the most current information, they may set the use_cache keyword to false and the call (and subsequent population of volume_dict) will be made prior to returning disk information.

Code was added/changed in two classes.  In GCENodeDriver, added two methods and an additional parameter to build, lookup and toggle the refresh of the volume cache.   In GCEConnection, added convenience and helper methods to the class, which not only support this performance improvement but also support the longer term vision of leveraging aggregatedList calls elsewhere.
##### GCEConnection
- _(new method)_ **def request_aggregated_items(self, api_name)** - make all necessary calls, handling maxresults and saving the 'items' portion of the response.
- _(new method)_ **def _merge_response_items(self, list_name, response_list)** - helper method to merge responses into a single dictionary
##### GCENodeDriver
- _(new member)_ **volume_dict** - dictionary organized by name, zone.  Name is always available to us, but is not unique across zones.  Zones, though, are optionally supplied.  By organizing by name, we remove the need to search through the entire list of disks each time and can do a single hash lookup to have access to all disks by that name.  If we have the zone, another hash lookup, if not, we take the first key alphabetically.
- _(new method)_ **_build_name_zone_dict(self, zone_dict)** - internal method to populate volume dict
- _(new method)_ **_ex_populate_volume_dict(self)** - Void method to call API and build volume dictionary
- _(new method)_ **_ex_lookup_volume(self, name, zone=None)** - implements the actual lookup.  If zone is not provided, take the disk with that name from the (alphabetical) first zone (this is only an issue if there are more than two disks with the same name).
- _(new parameter)_ **list_nodes(self, ex_zone=None, use_disk_cache=True)** - use_disk_cache parameter for list_nodes to pass through, defaults to True.  If set to true, no more than one call per 500 disks would be made to populate all disk info for nodes
- _(new parameter, revision)_ **ex_get_volume(self, name, zone=None, use_cache=False)** - revised to check if volume_dict has been populated or should be used, followed by returning the call to _ex_lookup_volume
### Status
- done, ready for review
- [ ] Documentation

/cc @erjohnso /cc @tonybaloney

### DIFF
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -72,6 +72,7 @@ class GCEConnection(GoogleBaseConnection):
     GCEConnection extends :class:`google.GoogleBaseConnection` for 2 reasons:
       1. modify request_path for GCE URI.
       2. Implement gce_params functionality described below.
+      3. Add request_aggregated_items method for making aggregated API calls.
 
     If the parameter gce_params is set to a dict prior to calling request(),
     the URL parameters will be updated to include those key/values FOR A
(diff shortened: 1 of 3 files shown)


libcloud/compute/drivers/gce.py
```diff
@@ -4769,6 +4872,63 @@ def _ex_connection_class_kwargs(self):
...
+
+        :keyword  zone: The zone to search for the volume in (set to 'all' to
+                        search all zones)
+        :type     zone: ``str`` or ``None``
+
+        :return:  A StorageVolume object for the volume.
+        :rtype:   :class:`StorageVolume` or raise ``ResourceNotFoundError``.
+        """
+        if volume_name not in self.volume_dict:
+            raise ResourceNotFoundError(
```
erjohnso: Should this instead make an API call in case the cache is missing a recently created disk?

libcloud/compute/drivers/gce.py
```diff
@@ -38,6 +38,7 @@
 from libcloud.utils.iso8601 import parse_date
 
 API_VERSION = 'v1'
+DEFAULT_AGGREGATED_MAX_RESULTS = 500
```
erjohnso: iirc, this is already the default, so perhaps this is redundant.

supertom: Yes, definitely the default is 500 for all the APIs.  I have conflicting views about being explicit, but I'll take it out for simplicity.

libcloud/compute/drivers/gce.py
```diff
@@ -4934,23 +4932,45 @@ def _ex_lookup_volume(self, volume_name, zone=None):
...
         :rtype:   :class:`StorageVolume` or raise ``ResourceNotFoundError``.
         """
-        if volume_name not in self.volume_dict:
-            raise ResourceNotFoundError(
-                'Volume name: \'%s\' not found. Zone: %s' % (volume_name,
-                                                             zone), None, None)
+        if volume_name not in self._volume_dict:
+            # Possibly added through another thread/process, so re-populate
+            # _volume_dict and try again.  If still not found, raise exception.
+            self._ex_populate_dict()
```
sayap, line 7349: Should this be `_ex_populate_volume_dict`?

libcloud/compute/drivers/gce.py
```diff
@@ -131,6 +132,80 @@ def request(self, *args, **kwargs):
...
+        api_responses = []
+
+        params = {}
+        more_results = True
+        while more_results:
+            self.gce_params = params
+            response = self.request(request_path, method='GET').object
+            if 'items' in response:
+                api_responses.append(response)
+            more_results = 'pageToken' in params
```
sayap: To get all items, we should check `nextPageToken` from response, and assign it to `params['pageToken']` if it is not empty.

libcloud/compute/drivers/gce.py
```diff
@@ -131,6 +132,80 @@ def request(self, *args, **kwargs):
 
         return response
 
+    def request_aggregated_items(self, api_name):
```
sayap, line 134: I think this always returns only up to 500 items, as the `if self.gce_params:` condition few lines above will always evaluate to false.

supertom: Thanks @sayap, fixed.

PR #814, opened 2016-06-16
Fix dimensiondata lb port

relaxdiego:
### Description

The documentation for the Dimension Data Cloud REST API v2.2-20160222 states that a load balancer port can be an integer in the range of 1-65535 and, if not supplied, will be taken to mean 'Any Port'

This change fixes the libcloud driver to conform to the docs.
### Status
- done, ready for review
- [x] Documentation

### DIFF
--- a/libcloud/loadbalancer/drivers/dimensiondata.py
+++ b/libcloud/loadbalancer/drivers/dimensiondata.py
@@ -98,16 +98,17 @@ class DimensionDataLBDriver(Driver):
         kwargs['region'] = self.selected_region
         return kwargs
 
-    def create_balancer(self, name, port, protocol, algorithm, members):
+    def create_balancer(self, name, port=None, protocol=None,
+                        algorithm=None, members=None):
         """
         Create a new load balancer instance
 
         :param name: Name of the new load balancer (required)
         :type  name: ``str``
 
-        :param port: Port the load balancer should listen on,
-                    defaults to 80 (required)
-        :type  port: ``str``
+        :param port: An integer in the range of 1-65535. If not supplied,
+                     it will be taken to mean 'Any Port'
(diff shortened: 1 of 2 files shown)


libcloud/loadbalancer/drivers/dimensiondata.py
```diff
@@ -559,8 +560,8 @@ def ex_create_virtual_listener(self,
                                    network_domain_id,
                                    name,
                                    ex_description,
-                                   port,
                                    pool,
+                                   port=None,
```
tonybaloney, line 561: I think this would break the scenario where you specify a port?

libcloud/loadbalancer/drivers/dimensiondata.py
```diff
@@ -122,7 +123,7 @@ def create_balancer(self, name, port, protocol, algorithm, members):
         """
         network_domain_id = self.network_domain_id
         if port is None:
-            port = 80
+            port = None
```
tonybaloney: That isnt' required anymore

relaxdiego: DOH! I blame late night coding.

libcloud/loadbalancer/drivers/dimensiondata.py
relaxdiego, line 561: No, it won't. We're just setting the value to None here if port is not specified. (This is the argument declaration part of the method)

jadunham1, line 561: But you're changing the order of variables.  If keyword args aren't used here it would break.

relaxdiego:
I'll submit another PR later. Closing this for now.

PR #820, opened 2016-06-21
LIBCLOUD-606 Can not connect to AWS with IAM role temporary credentials

alejandrogr:
### Description

Can not connect to AWS with IAM role temporary credentials. Doesn't seem like the supplied token is in the header

From: [LIBCLOUD-606](https://issues.apache.org/jira/browse/LIBCLOUD-606)
### Status

done, ready for review
- [ ] Documentation

### DIFF
--- a/libcloud/compute/drivers/ec2.py
+++ b/libcloud/compute/drivers/ec2.py
@@ -5398,7 +5398,12 @@ class BaseEC2NodeDriver(NodeDriver):
 
     def _ex_connection_class_kwargs(self):
         kwargs = super(BaseEC2NodeDriver, self)._ex_connection_class_kwargs()
-        kwargs['signature_version'] = self.signature_version
+        if self.token is None:
+            kwargs['signature_version'] = self.signature_version
+        else:
+            kwargs['token'] = self.token
+            # Force signature_version 4 for tokens or auth breaks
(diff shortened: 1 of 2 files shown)


tonybaloney:
hi @alejandrogr thanks for the contribution, if you look at the test runs, every class that derived from this connection class (and didn't create a property `self.token`) is failing.

could you please add a supporting test for this scenario as well, the docs say that this is supported, so is this a bugfix or a new behaviour? http://libcloud.readthedocs.io/en/latest/compute/drivers/ec2.html#using-temporary-security-credentials

Kami:
Good catch - thanks!

And yeah, what Anthony said wrt tests :)

tonybaloney:
@alejandrogr can you run `flake8 --ignore=E402 --exclude=test libcloud/` locally and just check the linting errors please? Travis is giving a failure.

alejandrogr:
Hi,

I've just checked the linting problems and solved them. I'm pretty new to python and libcloud so please let me know if something can be improved.

Should other derived drivers from BaseEC2NodeDriver supports the temporary token too?

Cheers

Kami:
Merged into trunk, thanks.

For now, I think it's the best to only support token in the EC2 driver since I believe it's EC2-specific functionality.

PR #821, opened 2016-06-24
Docker bug fixes

tonybaloney:
## Bug fixes for the Docker driver

The docker driver was initially tested on v1.21 (1.9.x), Docker API supports versioning in the URI path now. I've updated the tests with 2 sets of fixtures, the 1.21 and the 1.24 (latest) API.

section of our documentation.
### Status

- work in progress
- done, ready for review
- [ ] Documentation


Kami:
Nice :+1: 

And yeah, hopefully including version in the URL will allow us to keep the driver a bit more sane and not break when new docker (API) version is released.

Down the road, we also need decide if we should (does it make sense, maybe it's too much overhead, etc.) to support multiple API versions. Maybe we should only do it for major versions or similar.

PR #822, opened 2016-06-24
Lazy entry

tonybaloney:
### Description

I find the current mechanism for fetching drivers a little cumbersome, 2 namespaces to import. After using the `requests` library I really like the module `__init__` accessors for common attributes. The most common factory is `get_driver` and the most common enum is the Provider enum.

This code allows for this example:

``` python
import libcloud
cls = libcloud.get_driver(libcloud.DriverType.COMPUTE, libcloud.DriverType.COMPUTE.RACKSPACE)
```
### Status
- done, ready for review
- [x] Documentation

### DIFF
--- a/libcloud/__init__.py
+++ b/libcloud/__init__.py
@@ -34,7 +34,8 @@ from libcloud.dns.providers import Provider as DnsProvider
 from libcloud.dns.providers import get_driver as get_dns_driver
 
 from libcloud.loadbalancer.providers import Provider as LoadBalancerProvider
-from libcloud.loadbalancer.providers import get_driver as get_loadbalancer_driver
+from libcloud.loadbalancer.providers import get_driver as \
+    get_loadbalancer_driver
 
 from libcloud.storage.providers import Provider as StorageProvider
 from libcloud.storage.providers import get_driver as get_storage_driver
--- a/libcloud/__init__.py
+++ b/libcloud/__init__.py
@@ -21,24 +21,9 @@ libcloud provides a unified interface to the cloud computing resources.
 import os
 import codecs
 
-from libcloud.backup.providers import Provider as BackupProvider
-from libcloud.backup.providers import get_driver as get_backup_driver
-
-from libcloud.compute.providers import Provider as ComputeProvider
-from libcloud.compute.providers import get_driver as get_compute_driver
-
-from libcloud.container.providers import Provider as ContainerProvider
-from libcloud.container.providers import get_driver as get_container_driver
-
-from libcloud.dns.providers import Provider as DnsProvider
-from libcloud.dns.providers import get_driver as get_dns_driver
-
-from libcloud.loadbalancer.providers import Provider as LoadBalancerProvider
-from libcloud.loadbalancer.providers import get_driver as \
-    get_loadbalancer_driver
-
-from libcloud.storage.providers import Provider as StorageProvider
-from libcloud.storage.providers import get_driver as get_storage_driver
+from libcloud.base import DriverType  # NOQA
+from libcloud.base import DriverTypeFactoryMap  # NOQA
+from libcloud.base import get_driver  # NOQA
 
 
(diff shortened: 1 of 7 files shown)


Kami:
I'm not too opinionated about of and I'm fine with the overall change and "external API".

libcloud/__init__.py
```diff
@@ -75,3 +94,40 @@ def _init_once():
             paramiko.common.logging.basicConfig(level=paramiko.common.DEBUG)
 
 _init_once()
+
+
+class DriverType:
```
Kami: Good idea to be consistent and inherit from object even if it's just used as an ENUM class.

libcloud/__init__.py
```diff
@@ -75,3 +94,40 @@ def _init_once():
...
+    DriverType.LOADBALANCER: get_loadbalancer_driver,
+    DriverType.STORAGE: get_storage_driver
+}
+
+
+def get_driver(type, provider):
+    """
+    Get a driver
+    """
+    return DriverTypeFactoryMap[type](provider)
```
Kami: We should probably re-throw a more user-friendly exception in case API / driver is not found, right?

Since I imagine KeyError or similar might be a bit confusing and not immediately obvious to the user what is going on.

libcloud/__init__.py
```diff
@@ -75,3 +94,40 @@ def _init_once():
             paramiko.common.logging.basicConfig(level=paramiko.common.DEBUG)
 
 _init_once()
+
+
+class DriverType:
+    """ Backup-as-a-service driver """
```
Kami: I would also prefer to move this code to `base.py` or similar module since usually the convention is for init to either contain initialization code or be empty (and this is not really initialization code).

We can of course also still import DriverType and get_driver from other module here so user can still do "from libcloud import get_driver" for convenience.

tonybaloney:
@Kami updated

libcloud/base.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def get_driver(type, provider):
+    """
+    Get a driver
+    """
+    try:
+        return DriverTypeFactoryMap[type](provider)
+    except KeyError:
+        raise DriverTypeNotFoundError()
```
Kami: I think it would also be good to include `type` in the error message.

Kami: And test for it would be nice :)

Besides that, LGTM, :+1:

Kami:
Would also be good to be update upgrade notes if want this to be a preferred approach for retrieving a driver down the road :)

For that though, it might be worth soliciting feedback and reaching consensus from more people.

erjohnso:
Looks good to me. A good follow-on would be to update site docs / examples for this new approach.

JamieCressey:
Observation from an outsider...

If you're defining a provider class as `libcloud.DriverType.SERVICE.PROVIDER`. Can we not determine the service from the provider, negating the requirement to also provide a service object, therefore simplifying the instantiation from:

`cls = libcloud.get_driver(libcloud.DriverType.COMPUTE, libcloud.DriverType.COMPUTE.RACKSPACE)`

to:

`cls = libcloud.get_driver(libcloud.DriverType.COMPUTE.RACKSPACE)`

Just a thought, either way seems much cleaner than the current method.

tonybaloney:
@JamieCressey that would be better using the Enum base type in Python 3.4+ but once it has been called, its already been converted into the value (string), and there's no difference in values between `storage.Providers.RACKSPACE` and `loadbalancer.Providers.RACKSPACE`, they're both 'rackspace'.
Instead I created seperate factory methods, `get_compute_driver`, `get_storage_driver` with the type inbuilt.
There are some cases where people might want to generate a driver dynamically without knowing the type, so `get_driver` is there as a catch-all.

tonybaloney:
I'm pushing this in unless any serious objections?

tonybaloney:
merging

PR #827, opened 2016-06-29
LIBCLOUD-830 Adding S3 support for cn-north-1

JamieCressey:
### Description

This PR adds S3 support for AWS China (cn-north-1)
### Status
- done, ready for review
- [x] Documentation


Kami:
Thanks for the contribution!

That's actually the old approach which has been mostly removed in v1.0.0 (we moved away from class per region to "region" constructor argument approach), but S3 is one of the drivers which still needs to be updated...

I will go ahead and merge this for now, but hopefully we can also update S3 driver soon and move it to the new model (then we can get rid of this code).

Kami:
Merged, thanks!

PR #829, opened 2016-07-02
auroradns: Implement iterator for zones and records

wido:
By implementing iterate_zones() and iterate_records() you can
iterate over zones and records returned by the API.

The list_zones() and list_records() methods simply talk to the
iteration functions and create a list of it and return that.

The test cases should still match since they call the list functions
and those call the iterators on their turn.

### DIFF
--- a/libcloud/dns/drivers/auroradns.py
+++ b/libcloud/dns/drivers/auroradns.py
@@ -261,24 +261,23 @@ class AuroraDNSDriver(DNSDriver):
         AuroraDNSHealthCheckType.TCP: 'TCP'
     }
 
-    def list_zones(self):
-        zones = []
-
+    def iterate_zones(self):
         res = self.connection.request('/zones')
         for zone in res.parse_body():
-            zones.append(self.__res_to_zone(zone))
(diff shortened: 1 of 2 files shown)


tonybaloney:
thanks @wido it would be great to update the test case to validate that the responses are infact tuples or iterate the responses. the current case I think just checks the first response.
e.g.
`assertEqual(tuple(generator_object), (1, 2, ...))`

wido:
@tonybaloney Thanks for the feedback! I expanded one test-case slightly, but the records were already iterated.

Or I might be missing your point here. But the list_records/zones calls call the iterate functions, so that tests it from beginning to end, right?

tonybaloney:
cool, 👍 then. you need to make sure you make multiple assertions for a generator method, that was all I was suggesting.

Kami:
Thanks for this change.

I know I am a bit late, but anyway. I removed the `list_*` methods from the driver since they are a bit redundant - they are already implemented on the base driver (52b037bc21aee9a7c7b84c20c6dc4a9f7cdb6363).

PR #830, opened 2016-07-02
auroradns: The 'extra' parameter in update_record() should be optional

wido:
In base update_record() is defined with extra=None but this was not properly
done in the Aurora DNS driver.

Update the method signature so that it matches the base DNS driver of libcloud.


tonybaloney:
👍

PR #833, opened 2016-07-04
Add support for setting the VM admin password from vCloud 

juanfont:
### Description

These commits allow the user to establish the admin password of a virtual machine on its creation without having to connect to it. It uses the API provided by VMware and its VMware Tools.
### Status

Done and ready for review.
- [x] Documentation

### DIFF
--- a/libcloud/compute/drivers/vcloud.py
+++ b/libcloud/compute/drivers/vcloud.py
@@ -2011,6 +2011,21 @@ class VCloud_1_5_NodeDriver(VCloudNodeDriver):
         return res
 
     def ex_change_vm_admin_password(self, vapp_or_vm_id, ex_admin_password):
+        """
+        Changes the admin (or root) password of VM or VMs under the vApp. If
+        the vapp_or_vm_id param represents a link to an vApp all VMs that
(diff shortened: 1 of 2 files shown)


libcloud/compute/drivers/vcloud.py
```diff
@@ -1986,6 +1993,78 @@ def _change_vm_ipmode(self, vapp_or_vm_id, vm_ipmode):
...
+                tag = '{http://www.vmware.com/vcloud/v1.5}%s' % prev_section
+                if e.tag == tag:
+                    break
+            e = ET.Element(
+                '{http://www.vmware.com/vcloud/v1.5}%s' % section)
+            e.text = text
+            res.object.insert(i, e)
+        return res
+
+    def _change_vm_admin_password(self, vapp_or_vm_id, ex_admin_password):
```
tonybaloney: do you not think this might be a function that users might want to call publically? If so I would rename it ex_change_vm_admin_password

juanfont:
Yes, you are right :)

tonybaloney:
thanks 👍

PR #834, opened 2016-07-06
auroradns: Change valid record types and implement listing

wido:
This commit changes the supported record types by Aurora DNS and it also
supports listing the valid record types.

It merely returns the keys of the RECORD_TYPE_MAP dict which is already
available. In the future Aurora DNS should return this in a API call
so this is no longer a static dictionary.

This dict also contained a type 'SPF', which is not valid. SPF records
are defined as TXT records in DNS. Remove this from the list of valid
record types.

Further it adds support for the SSHFP, DS, TLSA and PTR type since
they are also supported.

Also add tests to verify if this functions returns the proper values


Kami:
Merged, thanks!

PR #835, opened 2016-07-06
auroracompute: Add Barcelona (Spain) region to driver

wido:
This PR only adds a new region to the Aurora Compute driver, it does not change functionality of it.

Barcelona is the soon to open new region of Aurora Compute and we would like to have it supported in libcloud prior to launch which is ~2 weeks away


Kami:
Merged, thanks.

PR #836, opened 2016-07-07
docs: Update AuroraCompute examples

wido:
Might be a bit redundant, but gives end-users a overview on a single
page.

Easy for people to refer to.

### DIFF
--- a/docs/examples/compute/auroracompute/create_node.py
+++ b/docs/examples/compute/auroracompute/create_node.py
@@ -1,0 +1,22 @@
+from libcloud.compute.types import Provider
+from libcloud.compute.providers import get_driver
+
+apikey = 'mykey'
+secretkey = 'mysecret'
+
+Driver = get_driver(Provider.AURORACOMPUTE)
+driver = Driver(key=apikey, secret=secretkey)
+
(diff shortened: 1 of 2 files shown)


docs/examples/compute/auroracompute/create_node.py
```diff
@@ -0,0 +1,22 @@
...
+sizes = driver.list_sizes()
+
+# Find a Agile Offering with 2GB of Memory
+size = [s for s in sizes if s.ram == 2048 and s.name.startswith('Agile')][0]
+
+# Search for the Ubuntu 16.04 image
+image = [i for i in images if i.name == 'Ubuntu 16.04'][0]
+
+# Create the new Virtual Machine
+node = driver.create_node(image=image, size=size)
```
Kami, line 20: Hm, is the name automatically generated if the `name` argument is not explicitly provided?

Kami:
Merged into trunk - thanks!

docs/examples/compute/auroracompute/create_node.py
wido, line 20: Yes, CloudStack in this case will generate a name if not provided.

PR #837, opened 2016-07-08
Fix incorrect doc string type information.

illfelder:
Fix type information in doc string.
### Description

The get from family function iterates through the project list. If you
pass in a string, it will iterate through the string instead of considering
the string a list with one element.
### Status

done, ready for review
- [ ] Documentation


Kami:
I merged the change into trunk - thanks!

You could also add isinstance check and throw `ValueError` is the argument is not of a correct type, but that's optional.

``` python
if not isinstance(ex_project_list, (list, tuple, type(None)):
    raise ValueError(....)
```

Feel free to open a new PR though if you do want to add that change :)

PR #838, opened 2016-07-09
Libvirt/libvirt node driver cleanup and tests

Katana-Steel:
### Description

Code clean up and added more tests for the libvirt_driver module.
Cleaner driver creation with less console output when using SASL
authentication.
### Status
- done, ready for review
- [X] Documentation

### DIFF
--- a/libcloud/compute/drivers/libvirt_driver.py
+++ b/libcloud/compute/drivers/libvirt_driver.py
@@ -82,7 +82,7 @@ class LibvirtNodeDriver(NodeDriver):
         self._uri = uri
         self._key = key
         self._secret = secret
-        if '+tcp' in self._uri:
+        if uri is not None and '+tcp' in self._uri:
             if key is None and secret is None:
                 raise RuntimeError('The remote Libvirt instance requires ' +
                                    'authentication, please set \'key\' and ' +
@@ -92,8 +92,8 @@ class LibvirtNodeDriver(NodeDriver):
             self.connection = libvirt.openAuth(uri, auth, 0)
         else:
             self.connection = libvirt.open(uri)
-        if self.connection is None:
-            raise RuntimeError('Unable to establish a connection to libvirtd')
+        if uri is None:
+            self._uri = self.connection.getInfo()
 
     def _cred_callback(self, cred, user_data):
         """
--- a/libcloud/compute/drivers/libvirt_driver.py
+++ b/libcloud/compute/drivers/libvirt_driver.py
@@ -82,16 +82,18 @@ class LibvirtNodeDriver(NodeDriver):
         self._uri = uri
         self._key = key
         self._secret = secret
-        try:
-            self.connection = libvirt.open(uri)
-        except libvirt.libvirtError:
-            if key is None or secret is None:
+        if '+tcp' in self._uri:
+            if key is None and secret is None:
                 raise RuntimeError('The remote Libvirt instance requires ' +
                                    'authentication, please set \'key\' and ' +
                                    '\'secret\' parameters')
             auth = [[libvirt.VIR_CRED_AUTHNAME, libvirt.VIR_CRED_PASSPHRASE],
                     self._cred_callback, None]
             self.connection = libvirt.openAuth(uri, auth, 0)
+        else:
+            self.connection = libvirt.open(uri)
+        if self.connection is None:
+            raise RuntimeError('Unable to establish a connection to libvirtd')
 
     def _cred_callback(self, cred, user_data):
         """
--- a/libcloud/test/compute/test_libvirt_driver.py
+++ b/libcloud/test/compute/test_libvirt_driver.py
@@ -16,129 +16,26 @@
 import sys
 
 from libcloud.compute.drivers.libvirt_driver import LibvirtNodeDriver
-from libcloud.compute.base import NodeState
-from libcloud.compute.types import Provider
 
 
 from libcloud.test import unittest
+from unittest import mock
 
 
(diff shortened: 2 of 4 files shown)


libcloud/test/compute/test_libvirt_driver.py
```diff
@@ -26,11 +28,19 @@ class virConnect:
...
         return 0
 
+    def __yes(self, *args, **kwargs):
+        return True
+
+    def __ary(self, *args, **kwargs):
+        return []
+
     def __init__(self):
-        stub = self.stub
```
Kami, line 33: I'm fine with this code, but I believe you could also utilize Mock for that by specifying a class you want to mock for `spec` argument - http://www.voidspace.org.uk/python/mock/mock.html#mock.Mock.

This would mean less code to maintain and keep up to date :)

libcloud/test/compute/test_libvirt_driver.py
```diff
@@ -65,103 +75,66 @@ def __init__(self):
...
+            else:
+                self.__dict__[f] = stub
 
 
 class LibvirtNodeDriverTestCase(LibvirtNodeDriver, unittest.TestCase):
+    type = Provider.LIBVIRT
+    name = 'Libvirt'
+    website = 'http://libvirt.org/'
+
+    NODE_STATE_MAP = {
```
Kami: Hm, do we need this attributes here on the test class? They are already defined in the `LibvirtNodeDriver` class.

Also, usually the idea is not to have the test class inherit from the Driver class, but the test case should instantiate the driver class and use that.

For example:

``` python
def setUp(self):
    ....
    self.driver = LibvirtNodeDriver(...)

def some_test_method(self):
    nodes = self.driver.list_nodes()
    ...
```

Kami: It would be great if you could make those changes so it's consistent with other drivers tests and it's a bit easier to maintain it - thanks!

libcloud/compute/drivers/libvirt_driver.py
```diff
@@ -82,16 +82,18 @@ def __init__(self, uri, key=None, secret=None):
...
+        if '+tcp' in self._uri:
+            if key is None and secret is None:
                 raise RuntimeError('The remote Libvirt instance requires ' +
                                    'authentication, please set \'key\' and ' +
                                    '\'secret\' parameters')
             auth = [[libvirt.VIR_CRED_AUTHNAME, libvirt.VIR_CRED_PASSPHRASE],
                     self._cred_callback, None]
             self.connection = libvirt.openAuth(uri, auth, 0)
+        else:
+            self.connection = libvirt.open(uri)
```
Kami, line 94: Hm, so `open` simply returns `None` if it's unable to establish a connection?

If so, that's a bit unfriendly and I guess if you want to debug it you need to set `LIBVIR_DEBUG` environment variable, or is there any other, more user-friendly way to do it?

Kami:
Thanks for this contribution.

I've added some comments. It would be great if you could at least update the tests to follow the same approach as other driver test (sorry for not mentioning in your first tests PR, I missed it :().

libcloud/compute/drivers/libvirt_driver.py
Katana-Steel, line 94: my bad... actually it will raise libvirt.libvirtError in case of a failed connection, fixing

Katana-Steel:
reviewed the suggestions and I did read up on Mock for testing
and updated the test code to mock essentially the entire libvirt module just to be safe ;)

Katana-Steel:
the have_libvirt check makes using the mock framework hard when libvirt isn't installed.

I'll take any suggestions... I could go back to inherit and just use MagicMock for virConnection

it seems bad to assume libvirt is installed for tests to pass. Or would that be an acceptable change?

tox.ini
```diff
@@ -37,10 +38,12 @@ deps = -r{toxinidir}/requirements-tests.txt
 [testenv:py2.6-lxml]
 deps = -r{toxinidir}/requirements-tests.txt
        unittest2
+       mock
```
Kami, line 41: mock is already specified in requirements-tests.txt so no need to also specify it here.

In any case, not a big deal, I will remove it while merging :)

Kami:
@Katana-Steel Ideally we can make it work without requiring libvirt to be installed, at least for the basic unit tests.

I'm fine with having additional integration tests which can require libvirt to be installed as long as those don't run by default if libvirt is not installed (we can use unittest.skipIf and skip if libvirt is not installed).

Kami:
Thanks for making those changes.

I will make the mock change and get the tests passing locally and then merge it into trunk.

Kami:
I made those changes and merged it into trunk - 01c6db45e532958d114033ab1b8e271351bf1c26 , d9fc1edaf6b3f52b1141ab80c1d88d97a964380a, a4bddfec0b26606258f0e1da94c289d0ab573369

Thanks!

Kami:
Also - 6626808f2e301685b88a7cf3cbe3a13d14ef78ce

Some time in the next couple of days when I get a chance I will look into running libvirt tests on Travis CI.

For one, we need to make sure all the system level libvirt dependencies are installed on Travis and after that we need to install `libvirt` library in all the virtual environments used by tox.

Kami:
It looks like that Travis doesn't have libvirt system package whitelisted so sadly it doesn't look like we can get this easily to run on new Travis container based infrastructure :/

https://travis-ci.org/apache/libcloud/jobs/144437976#L118

PR #844, opened 2016-07-25
[LIBCLOUD-835] Fix caching of Google auth tokens

paultiplady:
### Description

The `GoogleOAuth2Credential. _write_token_to_file()` method writes a copy of the latest OAuth token to disk. Prior to this fix, the token was being written to disk without truncating the file first, which is fine in the case where the new token has the same number of characters (or more) as the old one. However, in some situations Google OAuth returns a shorter token string, which was causing the library to crash when loading the corrupted token.
### Status

Fixed, needs tests.
- [ ] Documentation

_write_token_to_file was not zeroing the file before writing
a new token, causing corruption.

FIXES: LIBCLOUD-835

### DIFF
--- a/libcloud/common/google.py
+++ b/libcloud/common/google.py
@@ -688,7 +688,7 @@ class GoogleOAuth2Credential(object):
         """
         filename = os.path.realpath(os.path.expanduser(self.credential_file))
         data = json.dumps(self.token)
-        with os.fdopen(os.open(filename, os.O_CREAT | os.O_WRONLY,
+        with os.fdopen(os.open(filename, os.O_CREAT | os.O_WRONLY | os.O_TRUNC,
                                int('600', 8)), 'w') as f:
             f.write(data)
 


Kami:
Thanks.

The change looks good to me. /cc @erjohnso 

It's also worth noting that there is still a possibility of a race condition (short window) if multiple instances of script which uses the driver run on the same host. Sadly there is not much we can (and probably should) do.

One option would be to use file based locking, but that's not totally trivial doing it cross-platform if we want to avoid additional dependencies and it's something we should probably outsource to the user (there will always be some possible edge cases and race conditions and sadly we can't take care of all of it).

paultiplady:
I think this code could also be made more robust; I believe we're just writing the token as an optimization to speed up the next startup, so failure here shouldn't really tank the app.

Could probably defensively catch all exceptions in the _get_token_from_file() method, instead of just IOErrors; that way if multiple writers clobber the file, the driver will fallback to fetching a new auth token from the API.

Kami:
> I think this code could also be made more robust; I believe we're just writing the token as an optimization to speed up the next startup, so failure here shouldn't really tank the app.

I agree.

It should just degrade and re-authenticate each time instead of exploding like it does right now.

If possible - it would also be great if you can add a test case for it :)

And again thanks and a good catch.

supertom:
LGTM.  I agree, we should be more robust in this area.  No reason why we can't be defensive here.

Thanks @paultiplady for the fix!

paultiplady:
I've started looking at tests, but I'm afraid I probably don't have time to
finish today. Not sure how you're prioritizing this, but if the merge can
wait a day or two then I should be able to come up with a UT in that
timeframe.

Kami:
I've merged PR into trunk and also made changes so failure on writing and reading token to / from file is not fatal - 2f02aecebc4ba371b1239378961675a477ab9086, 25c414e33526bb647f93db4f2adb8ad0120b4957.

I also added logging statement on failure, but we don't really have any common logging standards and practices established in Libcloud yet so that's something we need to improve in the future (among other things, we need to make sure that when `LIBCLOUD_DEBUG` is set that those messages are written to sys.stderr, etc.).

Thanks again!

PR #846, opened 2016-08-01
[FIX] CloudStack: Values with wildcards will fail signature validation

The-Loeki:
### Description

Thanks to a small bug in the request signer the signature will be invalid for any request with any value containing a wildcard; it should be considered 'safe'.

This is identical to apache/cloudstack-cloudmonkey@38b68fb
### Status
- done, ready for review

### DIFF
--- a/libcloud/common/cloudstack.py
+++ b/libcloud/common/cloudstack.py
@@ -80,7 +80,7 @@ class CloudStackConnection(ConnectionUserAndKey, PollingConnection):
         pairs = []
         for pair in signature:
             key = urlquote(str(pair[0]), safe='[]')
-            value = urlquote(str(pair[1]), safe='[]')
+            value = urlquote(str(pair[1]), safe='[]*')
             item = '%s=%s' % (key, value)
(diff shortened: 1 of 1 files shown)


tonybaloney:
thanks @The-Loeki have you tested this? I don't really understand the issue but wondered if the string should be "_" instead of "[]_"?

The-Loeki:
@tonybaloney Thanks & we're running it in production already ;) 

As I mentioned, the same bug surfaced in CloudMonkey (CloudStack's own API-driven CLI tool), and the fix was the same.

All API calls are signed & verified based on, amongst others, the encoded URL. This must apparently exclude HTML-encoding asterisks or the sig is invalid (as will the entire req be).

Of course it doesn't happen very often that a \* is in the values, so that explains why this wasn't noticed for so long.
The `safe` param is just a list of exceptions pretending to be a `str`, so the order shouldnt make a difference (it's not a regex or anything)

tonybaloney:
ok, thats a strange bug. :+1:

PR #848, opened 2016-08-08
[LIBCLOUD-843] Added fetch tags support in elb driver

lrvan:
### Description

Currently ELB driver doesn't provide a way to fetch tags for load balancer instance(s). Appropriate method calling DescribeTags AWS API action is implemented to address this.

Issue:
https://issues.apache.org/jira/browse/LIBCLOUD-843
### Status

done, ready for review
- [ ] Documentation

### DIFF
--- a/libcloud/loadbalancer/base.py
+++ b/libcloud/loadbalancer/base.py
@@ -222,19 +222,6 @@ class Driver(BaseDriver):
         raise NotImplementedError(
             'get_balancer not implemented for this driver')
 
-    def get_tags(self, balancer_id):
-        """
-        Return a dict of tag/value
-
-        :param balancer_id: id of a load balancer you want to fetch tags for
-        :type  balancer_id: ``str``
-
-        :rtype: ``dict``
-        """
-
-        raise NotImplementedError(
-            'get_tags not implemented for this driver')
-
     def update_balancer(self, balancer, **kwargs):
         """
         Sets the name, algorithm, protocol, or port on a load balancer.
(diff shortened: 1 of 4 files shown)


libcloud/loadbalancer/base.py
```diff
@@ -222,6 +222,19 @@ def get_balancer(self, balancer_id):
         raise NotImplementedError(
             'get_balancer not implemented for this driver')
 
+    def get_tags(self, balancer_id):
```
tonybaloney: would be better named `list_balancer_tags`, to be more consistent with the other base methods.

libcloud/loadbalancer/drivers/elb.py
```diff
@@ -115,6 +124,14 @@ def get_balancer(self, balancer_id):
         data = self.connection.request(ROOT, params=params).object
         return self._to_balancers(data)[0]
 
+    def get_tags(self, balancer_id):
```
tonybaloney: would you ever want to call `get_tags` without calling `list_balancers`? If not, it would be better to leave this as an internal method and remove it from the base class.

tonybaloney:
thanks @lrvan for the contribution, I had a couple of questions (commented) about the method names and approach.

libcloud/loadbalancer/base.py
lrvan: renamed it.

libcloud/loadbalancer/drivers/elb.py
lrvan: Since in most cases we either operate with single balancer instance or list of them and they should already have tags data - no, we won't :) 

Made _ex_list_balancer_tags & _ex_populate_balancer_tags private extra (cloud api should normally return tags for get/list tags) and removed from the base.

tonybaloney:
Great. thanks 👍

PR #849, opened 2016-08-09
Libcloud 845 delete tags

foriegnbody:
Support deleting tags without values

ex_delete_tags won't delete a tag without the value - it passes 'None' as a string.

Modified ex_delete_tags to check for None as a passed Value of a tag and if found don't supply Tag.%d.Value to params (and on to EC2 API). See http://docs.aws.amazon.com/AWSEC2/latest/CommandLineReference/ApiReference-cmd-DeleteTags.html For API Reference - Value is not required and if supplied must match for tag to be deleted. By always passing a value, a tag cannot be deleted if you don't know it's value.

Status: done, ready for review

Checklist: 
- [x ] [Code linting](http://libcloud.readthedocs.org/en/latest/development.html#code-style-guide) (required, can be done after the PR checks)


tonybaloney:
👍 thanks @foriegnbody

PR #850, opened 2016-08-09
Fix a missing / in requests made by the DockerContainerDriver

j-mie:
### Description

Following the examples provided by the documentation would result in a "400 Bad Request" error due to the fact that the requests didn't begin with a '/' for example:

```
GET v1.24/images/json HTTP/1.1
```

would result in a 400 Bad Request error, but:

```
GET /v1.24/images/json HTTP/1.1
```

does not.
### Status

done, ready for review
- [ ] Documentation

### DIFF
--- a/libcloud/container/drivers/docker.py
+++ b/libcloud/container/drivers/docker.py
@@ -205,7 +205,7 @@ class DockerContainerDriver(ContainerDriver):
         }
         data = json.dumps(payload)
 
-        result = self.connection.request('v%s/images/create?fromImage=%s' %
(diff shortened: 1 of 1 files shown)


tonybaloney:
thanks @JamieH, the tests are broken because of the way it checks for mock methods (that's simple to fix). But, I'm pretty sure in all my tests that I left a `/` at the end of my connection string.
Can you try that for me too? Maybe it makes more sense to check for the existence of it in `__init__`, because users may or may not put a / at the end of the host.

tonybaloney:
now relates to #874 since both users were reporting the same issue (with different ways of fixing it)

PR #852, opened 2016-08-11
Fix image undeprecation in GCE

illfelder:
### Description

An empty request body in a deprecation POST request will undeprecate the image. Specifying the "ACTIVE" state preserves the current "DEPRECATE" default.
### Status
- done, ready for review
- [ ] Documentation

### DIFF
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -3879,22 +3879,25 @@ class GCENodeDriver(NodeDriver):
             raise ValueError('state must be one of %s'
                              % ','.join(possible_states))
 
-        image_data = {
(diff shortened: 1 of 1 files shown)


erjohnso:
Thanks @illfelder! Could you help improve testing a bit around images and cover this change? https://github.com/apache/libcloud/blob/trunk/libcloud/test/compute/test_gce.py#L1177

illfelder:
I've played around with testing and spoken about it with @wrigri. It appears that existing tests for libcloud only verify that the correct API is called, not the body of the request. As a result, we can't find a supported way to verify that using "ACTION" translates to an empty request body. If you know of one, please let us know.

PR #853, opened 2016-08-19
[LIBCLOUD-848] Fix NameError - 'body' is not defined

steve-gregory:
### Description

Small change to add, but would be a big help. When a response from Openstack fails, the missing 'body' variable caused an uncaught exception. I remedied this by adding the body variable, similar to what was done above the else statement.
### Status
- done, ready for review
- [x] [Link to Jira](https://issues.apache.org/jira/browse/LIBCLOUD-848)


tonybaloney:
👍 thanks @steve-gregory

PR #857, opened 2016-09-16
Create volume is more robust on newer OpenStacks

allardhoeve:
### Description

On newer OpenStacks, things will fail if you do not provide `location` or if you do not provide `ex_volume_type`. The driver will pass the default `None` values to `self.driver.request`, which serializes them to `null`, which is an invalid value for `/os-volumes/` on newer incarnations of OpenStack.

This PR makes it so the OpenStack driver does not include these fields if they are None, so they are not refused by OpenStack's validation.
### Status
- done, ready for review
- [x] Documentation

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,18 @@
 ﻿Changelog
 =========
 
+Changes in Apache Libcloud 1.2.2
+
+Compute
+~~~~~~~
+
+- When creating volumes on OpenStack with defaults for `location` or `volume_type`,
+  newer OpenStack versions would throw errors. The OpenStack driver will now only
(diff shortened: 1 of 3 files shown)


tonybaloney:
sorry assertNotIn was only added in py2.7, you'll have to use `assertFalse(x in y)` instead.

allardhoeve:
Yeah, forgot about that ancient version ;) Changed it.

allardhoeve:
Thumbs up, anyone?

vdloo:
lgtm

tonybaloney:
Changes look good. thanks @allardhoeve +1

PR #859, opened 2016-09-16
[LIBCLOUD-852] Support filtering in ec2 list_volumes

jbmchuck:
### Description

There currently is not a way to filter volumes in the EC2 driver list_volumes.  This is a useful feature for identifying volumes with certain attributes - e.g. orphaned volumes.  The filtering param is already implicitly used in list_volumes to specify an attached instance-id, if node arg is provided.  This change allows additional explicit filters to be provided.

The particular issue I have run into motivating this change is - if I create a volume using create_volume and immediately attach it to a node using attach_volume I occasionally run into errors where the volume was created but is not yet ready for attachment.  I can use list_volumes to fetch everything and filter on my end, but I'd rather filter server side as I am with list_images and other similar functions.

I also added a docstring. 
### Status

done, ready for review
- [x] Documentation

### DIFF
--- a/libcloud/compute/drivers/ec2.py
+++ b/libcloud/compute/drivers/ec2.py
@@ -2892,13 +2892,37 @@ class BaseEC2NodeDriver(NodeDriver):
                     )
         return locations
 
-    def list_volumes(self, node=None):
+    def list_volumes(self, node=None, ex_filters=None):
+        """
+        List volumes
+        @inherits: :class:`NodeDriver.list_volumes`
+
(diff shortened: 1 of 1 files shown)


tonybaloney:
LGTM. thanks for your contribution

jbmchuck:
Thanks, creating and attaching patch to https://issues.apache.org/jira/browse/LIBCLOUD-852

jbmchuck:
Any ETA on having this merged?

PR #860, opened 2016-09-21
Google Storage permissions support

crunk1:
- Added ex_(get|set|delete)_permissions to libcloud.storage.drivers.google_storage.GoogleStorageDriver
- Added a JSON connection to the GoogleStorageDriver.
- The JSON connection is only used in the new methods I added. Added tests for changes.

Tests:
python setup.py test
python3.4 setup.py test

### DIFF
--- a/libcloud/storage/drivers/google_storage.py
+++ b/libcloud/storage/drivers/google_storage.py
@@ -14,22 +14,41 @@
 # limitations under the License.
 
 import copy
+import json
 
 import email.utils
 
 from libcloud.common.base import ConnectionUserAndKey
 from libcloud.common.google import GoogleAuthType
 from libcloud.common.google import GoogleOAuth2Credential
+from libcloud.common.google import GoogleResponse
+from libcloud.common.types import ProviderError
 from libcloud.storage.drivers.s3 import BaseS3Connection
 from libcloud.storage.drivers.s3 import BaseS3StorageDriver
 from libcloud.storage.drivers.s3 import S3RawResponse
 from libcloud.storage.drivers.s3 import S3Response
+from libcloud.utils.py3 import httplib
 
 # Docs are a lie. Actual namespace returned is different that the one listed
 # in the docs.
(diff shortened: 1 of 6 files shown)


crunk1:
@adjackura @erjohnso

crunk1:
@rickard-von-essen

rickard-von-essen:
Don't think I should have been pinged here?

crunk1:
Ah, yes. Sorry, about that, Rickard. I was looking over a previous bug patch we had discussed.

crunk1:
@zmarano

tonybaloney:
@crunk1 can you explain the motivations behind the PR please?

crunk1:
Google Cloud Storage doesn't have a method exposed to query for, or set, permissions on containers or objects. This PR adds the functionality to do so. Internally at Google, I have built some libraries on top of libcloud and I need a way to get and set permissions on GCS entities.
IntEnum was a helper class that I made for my changes, but I purposefully made it more generalized and put it in common/types.py with the idea to share it.

crunk1:
@illfelder

erjohnso:
The IntEnum addition looks like it could be generically useful across more of the code base. I'm pretty sure I've used dicts and sorted key lists for things like this. But I wonder if it would be better to decouple this and introduce IntEnum more broadly (across more of libcloud) to help us all move to this model?  My concern with keeping it joined with the GCS changes, we'd all forget about it and the generic solution would not get adopted elsewhere at which point, this becomes a more complex solution than needed purely for the GCS changes.

wdyt?

/cc @supertom

crunk1:
I think it is a little beyond the scope of this PR, but I agree with your opinion. Could I put it in a subsequent PR?

erjohnso:
Yes please! I think teasing that out separately and doing something more basic for GCS now would be easier.

crunk1:
PTAL, IntEnum logic and other cleanup removed from this PR.

erjohnso:
Thanks @crunk1! 👯

PR #864, opened 2016-09-23
Fix case where node does not exist on OpenStack

allardhoeve:
### Description

```
Fix case where node does not exist on OpenStack

  - Add case where the node does not exist on OS
  - Fix code that tries to return None if the node does not exist
```
### Status
- done, ready for review
- [x] Documentation

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,6 +13,11 @@ Compute
   (GITHUB-857)
   [Allard Hoeve]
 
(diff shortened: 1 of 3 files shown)


allardhoeve:
Branch has the wrong name. Fixing OS stop node will be the next fix :smile:

tonybaloney:
:+1:

PR #874, opened 2016-09-28
[LIBCLOUD-850] Docker Driver missing '/' prior to URI

InAnimaTe:
### Description

This is a quick 2 line change that enables a proper check for prepending slash on a uri so drivers can successfully still work without explicitly defining a `/`. This change was inspired by the `docker` driver not working when I first instantiated it.

See [this](https://issues.apache.org/jira/browse/LIBCLOUD-850) issue for more.
### Status
- done, ready for review

### DIFF
--- a/libcloud/common/base.py
+++ b/libcloud/common/base.py
@@ -867,6 +867,8 @@ class Connection(object):
         return response
 
     def morph_action_hook(self, action):
+        if not action.startswith("/")
+            action = "/" + action
         return self.request_path + action
 
     def add_default_params(self, params):
--- a/libcloud/common/base.py
+++ b/libcloud/common/base.py
@@ -867,7 +867,7 @@ class Connection(object):
         return response
 
     def morph_action_hook(self, action):
-        if not action.startswith("/")
+        if not action.startswith("/"):
             action = "/" + action
         return self.request_path + action
 
--- a/libcloud/test/container/test_docker.py
+++ b/libcloud/test/container/test_docker.py
@@ -129,101 +129,101 @@ class DockerMockHttp(MockHttp):
             raise AssertionError('Unsupported method')
         return (httplib.OK, body, {}, httplib.responses[httplib.OK])
 
-    def vlinux_121_images_search(
+    def _vlinux_121_images_search(
             self, method, url, body, headers):
         return (httplib.OK, self.fixtures.load('linux_121/search.json'), {}, httplib.responses[httplib.OK])
 
-    def vmac_124_images_search(
+    def _vmac_124_images_search(
             self, method, url, body, headers):
         return (httplib.OK, self.fixtures.load('mac_124/search.json'), {}, httplib.responses[httplib.OK])
 
-    def vlinux_121_images_json(
+    def _vlinux_121_images_json(
             self, method, url, body, headers):
         return (httplib.OK, self.fixtures.load('linux_121/images.json'), {}, httplib.responses[httplib.OK])
 
-    def vmac_124_images_json(
+    def _vmac_124_images_json(
             self, method, url, body, headers):
         return (httplib.OK, self.fixtures.load('linux_121/images.json'), {}, httplib.responses[httplib.OK])
 
-    def vlinux_121_images_create(
+    def _vlinux_121_images_create(
             self, method, url, body, headers):
         return (httplib.OK, self.fixtures.load('linux_121/create_image.json'), {'Content-Type': 'application/json'},
                 httplib.responses[httplib.OK])
 
-    def vmac_124_images_create(
+    def _vmac_124_images_create(
             self, method, url, body, headers):
         return (httplib.OK, self.fixtures.load('mac_124/create_image.json'), {'Content-Type': 'application/json'},
                 httplib.responses[httplib.OK])
 
-    def vlinux_121_containers_json(
+    def _vlinux_121_containers_json(
             self, method, url, body, headers):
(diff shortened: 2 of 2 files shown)


libcloud/common/base.py
```diff
@@ -867,6 +867,8 @@ def request(self, action, params=None, data=None, headers=None,
         return response
 
     def morph_action_hook(self, action):
+        if not action.startswith("/")
```
tonybaloney: missing `:` at the end of the line 👍

tonybaloney (commented):
1 change to get this complete.

InAnimaTe:
Wow, I must be getting old.

libcloud/common/base.py
InAnimaTe: btw, this is fixed now. although checks seem to be failing ;( not sure what that's about.

Let me know if there is anything else I can do.

tonybaloney:
@InAnimaTe see https://travis-ci.org/apache/libcloud/builds/164054258 or run `tox -e py27` locally to check the output of the tests. Because there is now always a `/` at the beginning of the request, the mock http methods all need to be renamed with a `_` at the beginning. a simple find and replace should do the trick.

libcloud/common/base.py
```diff
@@ -867,6 +867,8 @@ def request(self, action, params=None, data=None, headers=None,
         return response
 
     def morph_action_hook(self, action):
+        if not action.startswith("/"):
```
tonybaloney, line 870: I still have an issue with this IF the host provided by the user surely that would be a request with a double slash?

http://example.com/ - host
containers/ - action

http://example.com//containers - resulting URL?

Can you try this and see what happens, if so this logic should check the last character in the host string as well

tonybaloney (commented):
1 last thing to check

libcloud/common/base.py
InAnimaTe: Yep just tested this. Tried a few different values
- `localhost` - works normally as expected
- `localhost/` - Doesn't work; can create connection object but any actual calls fail (see below)
- `http://localhost/` Doesn't work, same as above.

It seems the connection handling portion of libcloud does a thorough check to ensure the provided host is exactly that: a proper hostname.

```
In [9]: connection = driver(host="http://localhost/", port=4243)

In [10]: connection.list_containers()
gaierror                                  Traceback (most recent call last)
<ipython-input-10-52f6b96a7188> in <module>()
----> 1 connection.list_containers()

/home/mloria/arrovenv/libcloud/libcloud/container/drivers/docker.py in list_containers(self, image, all)
    277         try:
    278             result = self.connection.request(
--> 279                 "v%s/containers/json%s" % (self.version, ex)).object
    280         except Exception as exc:
    281             errno = getattr(exc, 'errno', None)

/home/mloria/arrovenv/libcloud/libcloud/common/base.py in request(self, action, params, data, headers, method, raw)
    845                 raise socket.gaierror(msg)
    846             self.reset_context()
--> 847             raise e
    848         except ssl.SSLError:
    849             e = sys.exc_info()[1]

/home/mloria/arrovenv/libcloud/libcloud/common/base.py in request(self, action, params, data, headers, method, raw)
    827                 else:
    828                     self.connection.request(method=method, url=url, body=data,
--> 829                                             headers=headers)
    830         except socket.gaierror:
    831             e = sys.exc_info()[1]

/usr/lib/python3.5/http/client.py in request(self, method, url, body, headers)
   1104     def request(self, method, url, body=None, headers={}):
   1105         """Send a complete request to the server."""
-> 1106         self._send_request(method, url, body, headers)
   1107
   1108     def _set_content_length(self, body, method):

/usr/lib/python3.5/http/client.py in _send_request(self, method, url, body, headers)
   1149             # default charset of iso-8859-1.
   1150             body = _encode(body, 'body')
-> 1151         self.endheaders(body)
   1152
   1153     def getresponse(self):

/usr/lib/python3.5/http/client.py in endheaders(self, message_body)
   1100         else:
   1101             raise CannotSendHeader()
-> 1102         self._send_output(message_body)
   1103
   1104     def request(self, method, url, body=None, headers={}):

/usr/lib/python3.5/http/client.py in _send_output(self, message_body)
    932         del self._buffer[:]
    933
--> 934         self.send(msg)
    935         if message_body is not None:
    936             self.send(message_body)

/usr/lib/python3.5/http/client.py in send(self, data)
    875         if self.sock is None:
    876             if self.auto_open:
--> 877                 self.connect()
    878             else:
    879                 raise NotConnected()

/usr/lib/python3.5/http/client.py in connect(self)
    847         """Connect to the host and port specified in __init__."""
    848         self.sock = self._create_connection(
--> 849             (self.host,self.port), self.timeout, self.source_address)
    850         self.sock.setsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)
    851

/usr/lib/python3.5/socket.py in create_connection(address, timeout, source_address)
    691     host, port = address
    692     err = None
--> 693     for res in getaddrinfo(host, port, 0, SOCK_STREAM):
    694         af, socktype, proto, canonname, sa = res
    695         sock = None

/usr/lib/python3.5/socket.py in getaddrinfo(host, port, family, type, proto, flags)
    730     # and socket type values to enum constants.
    731     addrlist = []
--> 732     for res in _socket.getaddrinfo(host, port, family, type, proto, flags):
    733         af, socktype, proto, canonname, sa = res
    734         addrlist.append((_intenum_converter(af, AddressFamily),

gaierror: [Errno -2] Name or service not known

```

tonybaloney:
@Kami what say you?

PR #884, opened 2016-10-03
fix path to SUSE CA bundle

mcalmer:
### Description

Fix the path to the SUSE CA bundle. The old path is a special CA file and not the CA bundle.
### Status
- done, ready for review

### DIFF
--- a/libcloud/security.py
+++ b/libcloud/security.py
@@ -63,6 +63,9 @@ CA_CERTS_PATH = [
 
     # opensuse/sles: openssl
     '/etc/ssl/ca-bundle.pem',
+
+    # SLES11 imported CA certificate
+    '/etc/ssl/certs/YaST-CA.pem',
 ]
 
 # Insert certifi CA bundle path to the front of Libcloud CA bundle search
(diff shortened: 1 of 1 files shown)


tonybaloney:
@mcalmer this path was used based on a live instance of suse and testing an issue with a consumer (salt) - see related issue https://github.com/saltstack/salt/issues/32743

Are there any differences with the files?

mcalmer:
SUSE bring a yast module called yast2 ca_mgm to create own CA certificates and sign certificates.
You can import such a certificate including the CA. In this case the CA certificate is stored with the name /etc/ssl/certs/YaST-CA.pem in SLES11. This does not apply anymore with SLES12 and current openSUSE distributions.

This file contains only 1 certificate and is not a bundle. Older SLES never used bundles. We used openssl directory structure. Every file contain only one certificate.
With current SLES and openSUSE, we introduce a pki structure to generate any kind of trust stores. This include also a ca-bundle with the path I added with this commit. In case a CA is imported the ca-bundle contain also the certificate which was called YaST-CA.pem in the past.

Of cause in SLES11 there is still a problem. But including a single CA is only a workaround which will help only in a few cases.

tonybaloney:
The path list is a collection so you can add both and we can cover 12 and 11 cases? They are done in sequence so put the newer one first

```
    _____________________________
```

mcalmer:
Agreed. This will at least not break older installations. I will adapt the PR.

tonybaloney:
👍

PR #889, opened 2016-10-05
Remove DigitalOcean v1 driver as it has reached EOL.

andrewsomething:
### Description

DigitalOcean API v1 reached EOL November 9, 2015. It's code paths can/should be removed.

https://developers.digitalocean.com/documentation/changelog/api-v1/sunsetting-api-v1/
### Status

WIP: I still need to clean up the docs, but want to make sure that this is on the right track.
- [ ] Documentation
- [ ] [ICLA](http://libcloud.readthedocs.org/en/latest/development.html#contributing-bigger-changes)  (Doubt I need this for removing code. I might have signed this before, but I'm not sure.)


tonybaloney:
thanks @andrewsomething for the contribution. I was just looking through the change. we have an error that get raised when you run get_driver on a provider which no longer exists:
https://github.com/apache/libcloud/blob/trunk/libcloud/compute/deprecated.py and https://github.com/apache/libcloud/blob/trunk/libcloud/common/providers.py

Since this is just for deprecating the old API, that doesn't make sense but it would be good to have the errors more aligned. Please can you review the existing deprecated driver code and adjust this message (we include a URL for example).

andrewsomething:
@tonybaloney How's that look?

tonybaloney:
@andrewsomething great. I'll make sure this is in the release notes for the next release, scheduled next couple of weeks.

PR #890, opened 2016-10-05
Add new FCU API support in Outscale EC2 driver

jmunhoz:
Add libcloud support for the following new APIs available in the Flexible Computing Unit (FCU)
- DescribeInstanceTypes
- DescribeProductTypes
- GetProductType
- DescribeQuota
- ModifyInstanceKeypair

The patchset includes tests and fixtures for the new APIs.

### DIFF
--- a/libcloud/compute/drivers/ec2.py
+++ b/libcloud/compute/drivers/ec2.py
@@ -6908,6 +6908,47 @@ class OutscaleNodeDriver(BaseEC2NodeDriver):
 
         return is_truncated, quota
 
+    def _to_product_types(self, elem):
+
+        product_types = []
+        for product_types_item in findall(element=elem,
+                                          xpath='productTypeSet/item',
+                                          namespace=OUTSCALE_NAMESPACE):
+            productTypeId = findtext(element=product_types_item,
+                                     xpath='productTypeId',
+                                     namespace=OUTSCALE_NAMESPACE)
+            description = findtext(element=product_types_item,
+                                   xpath='description',
+                                   namespace=OUTSCALE_NAMESPACE)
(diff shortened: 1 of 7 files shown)


libcloud/compute/drivers/ec2.py
```diff
@@ -6806,6 +6811,268 @@ def list_sizes(self, location=None):
...
+        :param      marker: Set quota marker
+        :type       marker: ``string``
+
+        :return:    (is_truncated, quota) tuple
+        :rtype:     ``(bool, dict)``
+        """
+
+        if filters:
+            raise NotImplementedError(
+                'quota filters are not implemented')
```
tonybaloney, line 6912: when will this be implemented?

tonybaloney (commented):
Great thanks @jmunhoz for the updates.

libcloud/compute/drivers/ec2.py
jmunhoz: @tonybaloney thanks for reviewing. we hope a new release of FCU at the end of November or December. Filters in this API are being reworked cause the original implementation is raising some non expected issues in the server side and we had to rollback although the API doc was published and the interface spec is stable. This way users get a clear message and they would not report this well-know bug.

BTW this first patchset would be also a first update to catch up with the new features and improvements in the FCU. Outscale would like to contribute new APIs and available features implemented along the previous months in the server side. As expected, with each new FCU release we would revisit the non implemented bits and bugfixes to support the new features as soon as they hit production in the server side.

tonybaloney:
Ok. I'm happy 👍

jmunhoz:
@tonybaloney thanks! :+1:

PR #891, opened 2016-10-05
Replace RunAbove driver by OVH

ZuluPro:
### Description

RunAbove is the OVH's test platform. Now its compute is closed, I replaced by their final solution at OVH.
### Status

Done
- [x] Documentation


libcloud/common/ovh.py
```diff
@@ -115,17 +116,23 @@ def request_consumer_key(self, user_id):
 
     def get_timestamp(self):
         if not self._timedelta:
-            url = 'https://%s/%s/auth/time' % (API_HOST, API_ROOT)
+            url = 'https://%s%s/auth/time' % (API_HOST, API_ROOT)
```
tonybaloney, line 119: can you just clarify for me that it was intentional to drop `/` in between %s and %s

libcloud/compute/drivers/ovh.py
```diff
@@ -113,29 +118,31 @@ def create_node(self, name, image, size, location, ex_keyname=None):
...
-        action = API_ROOT + '/instance/' + node.id
+        action = '%s/cloud/project/%s/instance/%s' % (
+            API_ROOT, self.project_id, node.id)
         self.connection.request(action, method='DELETE')
         return True
 
     def list_sizes(self, location=None):
-        action = API_ROOT + '/flavor'
-        data = {}
+        action = '%s/cloud/project/%s/flavor' % (API_ROOT, self.project_id)
```
tonybaloney: You make this transformation a lot. I would either store this as an instance level attribute or have a private method for `_request_for_project`.

libcloud/compute/providers.py
```diff
@@ -124,8 +124,8 @@
     ('libcloud.compute.drivers.packet', 'PacketNodeDriver'),
     Provider.ONAPP:
     ('libcloud.compute.drivers.onapp', 'OnAppNodeDriver'),
-    Provider.RUNABOVE:
-    ('libcloud.compute.drivers.runabove', 'RunAboveNodeDriver'),
+    Provider.OVH:
+    ('libcloud.compute.drivers.ovh', 'OvhNodeDriver'),
```
tonybaloney, line 128: `Provider.RUNABOVE` should be added to the deprecated dictionary with a message about using the OVH service instead of RA so that when someone runs `get_driver(Provider.RUNABOVE)` they get a deprecated driver exception

tonybaloney (commented):
Thanks very much @ZuluPro I've made some comments but overall this PR is excellent.

libcloud/common/ovh.py
ZuluPro: Because `API_ROOT` already starts with `'/'` (https://github.com/apache/libcloud/pull/891/files/a5ed7e1f907cd1fcb5dde3f873bf4855012f3fe7#diff-b98d1474f4f20fc646a3ac82fcbd0387R37) . So it will make a doublet.
Maybe it worked with RunAbove, but not with OVH

libcloud/compute/drivers/ovh.py
ZuluPro: I refactor code with `_get_project_action`.

libcloud/compute/providers.py
ZuluPro: Done! Say me if I miss something!

tonybaloney:
@ZuluPro OK. good to merge. great work. thanks

PR #894, opened 2016-10-09
LIBCLOUD-858: Fix Listing Libvirt Nodes with Python 3

femnad:
### Description

Listing nodes with the `Libvirt` provider fails on Python 3 because of a `split` operation on a `bytes` object where the delimiter is not a `bytes` object.

Use the `ensure_string` function from the `py3` module to ensure that the output of a `subprocess.Popen` call is always a string for Python 2 and 3. Also add a test to convert the mock command output results to `bytes` objects if the active Python version is 3.
### Status
- done, ready for review
- [ ] Documentation


tonybaloney:
great, thanks for also contributing the test to assert the issue exists and the change resolves it. 👍

PR #898, opened 2016-10-10
Fix bug in Dimension Data driver

tonybaloney:
The code for ex_audit_log calls the wrong URL. this updates fixes that.


tonybaloney:
fyi @benobermayer @manishkp

PR #899, opened 2016-10-10
Better Error Message on Fieldnames; Add start_container

InAnimaTe:
### Description

This PR extends the Rancher Driver to add `start_container` support. It follows the base class so providing a container object is necessary (and you get one back as expected).

Additionally, I ran into an a scenario where fields which caused errors would return something like:

```
In [12]: driver.ex_deploy_service(name='empty_test', image=image, environment_id="", service_description="testing", environment={"ohnoes":"maybe"})
BaseHTTPError                             Traceback (most recent call last)
<ipython-input-12-e9850d5a4dbf> in <module>()
----> 1 driver.ex_deploy_service(name='empty_test', image=image, environment_id="", service_description="testing", environment={"ohnoes":"maybe"})
/home/mloria/arrovenv/libcloud/libcloud/container/drivers/rancher.py in ex_deploy_service(self, name, image, environment_id, start, assign_service_ip_address, service_description, external_i
d, metadata, retain_ip, scale, scale_policy, secondary_launch_configs, selector_container, selector_link, vip, **launch_conf)
    394         data = json.dumps(dict((k, v) for (k, v) in service_payload.items()
    395                                if v is not None))
--> 396         result = self.connection.request('%s/services' % self.baseuri,
    397                                          data=data, method='POST').object
    398
/home/mloria/arrovenv/libcloud/libcloud/common/base.py in request(self, action, params, data, headers, method, raw)
    860
    861         try:
--> 862             response = responseCls(**kwargs)
    863         finally:
    864             # Always reset the context after the request has completed
/home/mloria/arrovenv/libcloud/libcloud/common/base.py in __init__(self, response, connection)
    177             raise exception_from_message(code=self.status,
    178                                          message=self.parse_error(),
--> 179                                          headers=self.headers)
    180
    181         self.object = self.parse_body()

BaseHTTPError: None - None
```

So I edited `parse_error` to check for this and print something a bit nicer:

```
BaseHTTPError: name is NotUnique: None - None
```
### Status
- done, ready for review

### DIFF
--- a/libcloud/test/container/test_rancher.py
+++ b/libcloud/test/container/test_rancher.py
@@ -157,6 +157,10 @@ class RancherContainerDriverTestCase(unittest.TestCase):
         self.assertEqual(container.extra['environment'],
                          {'STORAGE_TYPE': 'file'})
 
+    def test_start_container(self):
+        container = self.driver.get_container("1i31")
+        container.start()
+
     def test_stop_container(self):
         container = self.driver.get_container("1i31")
         container.stop()
--- a/libcloud/test/container/test_rancher.py
+++ b/libcloud/test/container/test_rancher.py
@@ -160,10 +160,14 @@ class RancherContainerDriverTestCase(unittest.TestCase):
     def test_start_container(self):
         container = self.driver.get_container("1i31")
         container.start()
+        self.assertEqual(container.id, "1i31")
+        self.assertEqual(container.name, "newcontainer")
 
     def test_stop_container(self):
         container = self.driver.get_container("1i31")
         container.stop()
+        self.assertEqual(container.id, "1i31")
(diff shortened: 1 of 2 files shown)


libcloud/test/container/test_rancher.py
```diff
@@ -157,6 +157,10 @@ def test_get_container(self):
         self.assertEqual(container.extra['environment'],
                          {'STORAGE_TYPE': 'file'})
 
+    def test_start_container(self):
+        container = self.driver.get_container("1i31")
+        container.start()
```
tonybaloney: You should really put assertions in these test methods, assert the returned instance of Container and the ID/name that it returns. same applies for the other methods. I should have spotted that last time.

tonybaloney:
👍

PR #904, opened 2016-10-13
Use sslv23 constant in newer versions of python

lin31504:


tonybaloney:
please stop spamming our project. 👎

PR #905, opened 2016-10-13
Return Object Changes and State Handling Fixup

InAnimaTe:
### Description

> This should hopefully be my last PR for a long-time regarding the Rancher driver. At least until `v2` comes out ;)

This PR changes the way the `start_container`, `stop_container`, and `destroy_container` functions work. They still return a container object as expected, but that object is now made from the actual response we receive from Rancher (which is not just a empty 204, but the full json for the container question), not another `get_container` call.

This is most certainly the better way to handle this:

```
In [1]: con = driver.get_container('1i24')

In [2]: con
Out[2]: <Container: id=1i24, name=newconbr0,state=running, provider=Rancher ...>

In [3]: stopcon = driver.stop_container(con)

In [4]: stopcon
Out[4]: <Container: id=1i24, name=newconbr0,state=terminated, provider=Rancher ...>

In [5]: stopcon.extra['state']
Out[5]: 'stopping'

In [6]: startcon = driver.start_container(stopcon)

In [7]: startcon
Out[7]: <Container: id=1i24, name=newconbr0,state=terminated, provider=Rancher ...>

In [8]: startcon.extra['state']
Out[8]: 'starting'

In [9]: destroycon = driver.destroy_container(startcon)

In [10]: destroycon
Out[10]: <Container: id=1i24, name=newconbr0,state=terminated, provider=Rancher ...>

In [11]: destroycon.extra['state']
Out[11]: 'stopping'

```
#### State Handling Fix

While the above output shows that our return objects are much better, it shows another issue. For some reason, `state=terminated` when we're in the process of a transition (starting, stopping, etc..). Looking deeper, I found my `elif` for terminated state doesn't work as expected. I've fixed this by using `any`. I've also modified the tests to properly check for this.

Now it works as expected:

```
In [4]: con = driver.get_container('1i31')                                          

In [5]: con                                                                         
Out[5]: <Container: id=1i31, name=newcontainer,state=stopped, provider=Rancher ...> 

In [6]: startcon = driver.start_container(con)                                      

In [7]: startcon                                                                    
Out[7]: <Container: id=1i31, name=newcontainer,state=pending, provider=Rancher ...> 

In [8]: startcon.extra['state']        
Out[8]: 'starting'                     

In [9]: startcon.extra['transitioning']
Out[9]: 'yes'                          
```
### Status
- done, ready for review


tonybaloney:
agreed. this is a far better approach.

PR #912, opened 2016-10-17
DimensionData - Handle exception when powering off a node that is already powered off

samuelchong:
### Description

-Handle exception when powering off a node that is already stopped
- This will be no longer throw exception if it already powered off
### Status
- done, ready for review
- [ ] Documentation


libcloud/common/dimensiondata.py
```diff
@@ -561,7 +563,7 @@ def wait_for_state(self, state, func, poll_interval=2, timeout=60, *args,
             sleep(poll_interval)
             cnt += 1
 
-        msg = 'Status check for object %s timed out' % (result)
+        msg = 'Status check for object %s timed out' % result
```
Kami: For safety reasons I would leave parenthesis there - if you add another format string and format to add parenthesis there it will break during run-time so it's usually safer to just always include parenthesis (this way it's impossible to forget to add them later if and when you add another format string argument).

libcloud/compute/drivers/dimensiondata.py
```diff
@@ -638,12 +639,18 @@ def ex_power_off(self, node):
...
+            body = self.connection.request_with_orgId_api_2(
+                'server/powerOffServer',
+                method='POST',
+                data=ET.tostring(request_elm)).object
+            response_code = findtext(body, 'responseCode', TYPES_URN)
+        except (DimensionDataAPIException, NameError, BaseHTTPError):
+            r = self.ex_get_node_by_id(node.id)
+            response_code = r.state.upper()
+
+        return response_code in ['IN_PROGRESS', 'OK', 'STOPPED', 'STOPPING']
```
Kami, line 653: This looks good to me, but maybe we should consider throwing an exception if node is already in stopping / stopped state.

That's more just me thinking out loud, I'm also fine with current approach.

Kami:
LGTM, thanks.

libcloud/compute/drivers/dimensiondata.py
samuelchong: Thx @Kami  for your comment. My intension is to make it more user friendly; As long as the aim is achieved, doesn't matter if the state a precondition or the result of the operation, as long as it is 'STOPPED',  I'm happy, no exception. 

Throwing exception will means one will have to check the state every time. More code. Open for discussion.

tonybaloney:
:+1:

PR #918, opened 2016-10-20
[LIBCLOUD-864] Fix Docker Driver install_image response parsing

ptzianos:
### Description

As reported in the corresponding bug, the docker daemon will respond in an install_image call with all the messages produced during the procedure parsed as json docs. In that case the response headers also contain the value 'transfer-encoding':'chunked'. That kind of response can now be parsed properly by the DockerResponse parse_body method. Also, another small change is that previously the id of the new image was marked in the json document as id, but now it's marked as sha256, so the regex used to discover the id has been updated. Finally the test data have been updated both for linux and for mac for install_image tests where this behavior has been observed and the version of the test fixtures has been bumped up to 124 to show the change more clearly.
### Status

Done and ready for review.


libcloud/test/container/test_docker.py
```diff
@@ -124,108 +124,108 @@ class DockerMockHttp(MockHttp):
...
+    def _vlinux_124_images_create(
             self, method, url, body, headers):
-        return (httplib.OK, self.fixtures.load('linux_121/create_image.json'), {'Content-Type': 'application/json'},
+        return (httplib.OK, self.fixtures.load('linux_124/create_image.txt'), {'Content-Type': 'application/json', 'transfer-encoding': 'chunked'},
                 httplib.responses[httplib.OK])
 
     def _vmac_124_images_create(
             self, method, url, body, headers):
-        return (httplib.OK, self.fixtures.load('mac_124/create_image.json'), {'Content-Type': 'application/json'},
+        return (httplib.OK, self.fixtures.load('mac_124/create_image.txt'), {'Content-Type': 'application/json', 'transfer-encoding': 'chunked'},
```
tonybaloney, line 155: 👍

PR #919, opened 2016-10-20
add ex_start_node for the openstack driver

vdloo:
#### Currently only ex_resume_node is implemented, but it does not start stopped instances.

http://developer.openstack.org/api-ref-compute-v2.1.html

```
BaseHTTPError: 409 Conflict Cannot 'resume' instance 39ec1234-1234-1234-ba28-123412349d9c while it is in vm_state stopped
```


allardhoeve (approved):

Kami:
LGTM, thanks.

tonybaloney:
👍

PR #921, opened 2016-10-24
google: Prevent GCE auth to hide S3 auth

pquentin:
### Description

We currently authenticate to Google Cloud Storage using Amazon S3 compatibility auth. Our code runs in Kubernetes on Google Container Engine. We tried to upgrade libcloud recently but 3849f65 from @crunk1 prevented us to authenticate. (Interestingly, it's also the commit that made us want to upgrade, since we eventually want to use service accounts.)

The issue happened for two reasons:
- `GoogleAuthType._is_gce()` always returns True when the code is run on the Google Container Engine, regardless of the authentication provided (which makes the issue impossible to reproduce in a local Docker environment)
- `GoogleAuthType._is_gcs_s3()` is always checked _after_ `_is_gce()`, so it could not be used on Google Container Engine

This pull request simply changes the order to give S3 higher priority. Note that Installed Applications auth has lower priority still, because it's the default auth when everything else fails. That's OK because I guess it's not possible on GCE. Still, I think the documentation should recommend to always specify the auth type, because explicit is better than implicit and it helps to avoid unclear errors.

### DIFF
--- a/libcloud/common/google.py
+++ b/libcloud/common/google.py
@@ -583,10 +583,10 @@ class GoogleAuthType(object):
     def guess_type(cls, user_id):
         if cls._is_sa(user_id):
             return cls.SA
-        elif cls._is_gce():
-            return cls.GCE
         elif cls._is_gcs_s3(user_id):
             return cls.GCS_S3
+        elif cls._is_gce():
+            return cls.GCE
         else:
             return cls.IA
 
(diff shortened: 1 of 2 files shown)


tonybaloney:
seeking input from @crunk1

crunk1 (approved):

crunk1:
LGTM. Sorry for the hassle!

pquentin:
@crunk1 No worries! Thanks for your work on libcloud.

pquentin:
@tonybaloney Anything else I should do? Thanks.

pquentin:
@asfgit Anything I can do to help get this merged? Thanks.

tonybaloney:
hey @pquentin back from travelling. let's get this merged.

PR #922, opened 2016-10-24
add ex_change_kernel in DigitalOcean_v2 driver

vdloo:
#### So the kernel can be managed using the API

> To change the kernel of a Droplet, send a POST request to
> /v2/droplets/$DROPLET_ID/actions. Set the "type" attribute to
> change_kernel and the "kernel" attribute to the new kernel's ID.

https://developers.digitalocean.com/documentation/v2/#change-the-kernel

The `DigitalOcean GrubLoader` can also be set with this allowing for [Internal Kernel Management](https://www.digitalocean.com/community/tutorials/how-to-update-a-digitalocean-server-s-kernel) in a Droplet.

### DIFF
--- a/libcloud/test/compute/test_digitalocean_v2.py
+++ b/libcloud/test/compute/test_digitalocean_v2.py
@@ -154,6 +154,12 @@ class DigitalOcean_v2_Tests(LibcloudTestCase):
         result = self.driver.destroy_node(node)
         self.assertTrue(result)
 
+    def test_ex_change_kernel_success(self):
+        node = self.driver.list_nodes()[0]
+        DigitalOceanMockHttp.type = 'KERNELCHANGE'
+        result = self.driver.ex_change_kernel(node, 7515)
+        self.assertTrue(result)
+
     def test_ex_rename_node_success(self):
         node = self.driver.list_nodes()[0]
         DigitalOceanMockHttp.type = 'RENAME'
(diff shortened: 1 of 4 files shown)


libcloud/test/compute/test_digitalocean_v2.py
```diff
@@ -154,6 +154,12 @@ def test_destroy_node_success(self):
         result = self.driver.destroy_node(node)
         self.assertTrue(result)
 
+    def test_ex_change_kernel_success(self):
+        node = self.driver.list_nodes()[0]
+        DigitalOceanMockHttp.type = 'KERNELCHANGE'
```
tonybaloney, line 159: "AttributeError: 'DigitalOceanMockHttp' object has no attribute '_v2_droplets_3164444_actions_KERNELCHANGE'"

libcloud/test/compute/test_digitalocean_v2.py
```diff
@@ -271,6 +277,11 @@ def _v2_droplets_3164444_DESTROY(self, method, url, body, headers):
         return (httplib.NO_CONTENT, body, {},
                 httplib.responses[httplib.NO_CONTENT])
 
+    def _v2_droplets_3164444_actions_CHANGEKERNEL(self, method, url, body, headers):
```
tonybaloney: that should be KERNELCHANGE, not CHANGEKERNEL

vdloo: fixed, thanks!

PR #923, opened 2016-10-25
Continue - Replace httplib with requests #728

tonybaloney:
See #728


tonybaloney:
`requests` explicitly does not support Python 3.2. https://github.com/kennethreitz/requests/issues/3507
This PR will mean dropping 3.2 support.

tonybaloney:
@Kami what do you think? This PR is now done and everything is passing

tonybaloney:
download this package here- http://people.apache.org/~anthonyshaw/libcloud/1.5.0.post0/

Kami:
@tonybaloney Maybe 2.0.0-rc1 and if everything is fine, call it 2.0.0?: :)

tonybaloney:
That package had a dumb error in it, I've since fixed it and against a live API (GoDaddy). I've tested the following scenarios 

- Applying a custom proxy via the environment variable
- Using libcloud.security to disable SSL verification
- Using libcloud.security to set a custom CA certificate
- Combining all of those scenarios
- Verification of custom headers applied by the driver using Charles Proxy and inspecting the HTTP messages manually
- Decoding JSON messages - although this still uses the existing methods, not the requests own json() decoder.

IMO, this is ready to merge. I would like to test the raw connections and file uploads if anyone has an account on one of those providers?

tonybaloney:
Tested `download_object` and `download_object_as_stream()` in the `GOOGLE_STORAGE` driver, which uses the S3 code. Both work.

tonybaloney:
`upload_object_as_stream` is complicated. I'm not sure the iterators are correctly handling the sessions or if the Google Storage API is just broken.

tonybaloney:
ok, figured it out. uploads don't work, because the current storage driver base code starts a `raw` connection session, sends the request and the headers but doesn't read the response from httplib, the callback (which could be one of 5 weird and crazy methods) then sends the data. 
`requests` does not support sending part of a request. So the options are:
* rewrite the storage driver base code to take advantage of `requests` native support for chunking via yield generators or passing a file stream object. http://docs.python-requests.org/en/master/user/advanced/#streaming-uploads 
* use a "raw" httplib connection and use the old code. So requests is used for the non-storage APIs and httplib for the storage APIs.

I favour the first but it seems too much to change all in 1 go so for the sake of sanity and stability, option 2 would be the more logical choice. @Kami ?

Kami:
Yeah, ideally we could do 1 (public API and behavior stays the same, just the implementation details change).

From a quick glance it looks like that by switching to requests we do lose some flexibility (like ability to specify chunk size, etc.), but I only done a quick glance so I could be wrong. It also looks like we might be able to simplify upload and download via stream methods a lot if we switch to requests so all those base class methods might not be needed anymore or at least we can simplify them.

In any case, I don't care much about the implementation detail as long as the API and all the behavior stays the same.

libcloud/httplib_ssl.py
```diff
@@ -165,7 +179,8 @@ def request(self, method, url, body=None, headers=None, raw=False):
             data=body,
             headers=headers,
             allow_redirects=1,
-            stream=raw
+            stream=raw,
+            verify=self.ca_cert if self.ca_cert is not None else self.verify
```
pquentin, line 193: verify expects a boolean, but self.ca_cert is a list, so this should be `bool(self.ca_cert) ...`.

libcloud/httplib_ssl.py
```diff
@@ -187,17 +186,19 @@ def request(self, method, url, body=None, headers=None, raw=False):
...
         prepped.body = body
-        
-        self.response = self.session.send(prepped,
+
+        self.response = self.session.send(
+            prepped,
             stream=raw,
-            verify=self.ca_cert if self.ca_cert is not None else self.verify
-        )
+            verify=self.ca_cert if self.ca_cert is not None else self.verify)
```
pquentin, line 208: ditto, verify expects a boolean, but self.ca_cert is a list, so this should be `bool(self.ca_cert) ...`.

pquentin:
Tested this pull request with our libcloud wrapper that we use for GCS. There are two verify issues (sorry, it looks like I commented on the wrong commits because GitHub shows them as "outdated"). Fixed them locally, there's only one remaining issue: when doing `next(storage.download_object_as_stream(obj))` in our own code we get this:

```
iterator = <Response [200]>, chunk_size = 8096, fill_size = False, yield_empty = False

    def read_in_chunks(iterator, chunk_size=None, fill_size=False,
                       yield_empty=False):
        ...
        while not empty or len(data) > 0:
            if not empty:
                try:
>                   chunk = b(get_data(*args))
E                   TypeError: 'Response' object is not an iterator

libcloud/libcloud/utils/files.py:74: TypeError
```

libcloud/httplib_ssl.py
tonybaloney: It can be a string or a boolean. This statement says to allow the cert or the flag http://docs.python-requests.org/en/master/user/advanced/#ssl-cert-verification the string option is hidden in the docs. I did test it out

tonybaloney:
Ok. Download object as stream doesn't respond the correct type. I'll sort that out

libcloud/httplib_ssl.py
tonybaloney: As with other comment, this is actually correct if it is a string. http://docs.python-requests.org/en/master/user/advanced/#ssl-cert-verification
I see your comment about a list now I didn't try that

libcloud/httplib_ssl.py
pquentin: In my case it was a list, not a string, but I don't know why.

pquentin: `libcloud.security.CA_CERTS_PATHS` is indeed defined as a list, not a string. Am I missing something?

tonybaloney: I've just checked in a fix to check if it's a list, if it is, then use the first one, if its bigger than 1, throw an error. People shouldn't need to use this anymore (that was the whole point) since requests comes with the Mozilla cert bundle pre-installed.

tonybaloney:
@pquentin https://github.com/apache/libcloud/pull/923/commits/109ecad94257615b57f196f3c36d13d7b3c8e8fd should fix the download as stream iterator bug.

libcloud/httplib_ssl.py
```diff
@@ -149,7 +149,12 @@ def _setup_ca_cert(self):
         if self.verify is False:
             pass
         else:
-            self.ca_cert = libcloud.security.CA_CERTS_PATH
+            if isinstance(libcloud.security.CA_CERTS_PATH, list):
+                if len(libcloud.security.CA_CERTS_PATH) > 1:
+                    raise ValueError('Only 1 certificate path is supported')
```
pquentin: I guess this was intended to be a warning? The next line will never be executed.

When removing the raised exception, I get a 'Not such file a directory' exception, unless the only line left in CA_CERTS_PATH is a path that exists on my machine. When it's a path that does not exist, the exception is raised by [ssl.SSLContext.load_verify_locations](https://docs.python.org/3/library/ssl.html#ssl.SSLContext.load_verify_locations) which is called by urllib3 which is called by requests.

pquentin:
@tonybaloney Thank you for those commits 👍 

Unfortunately they don't quite work: libcloud now tries to parse the content of the file as XML:

```_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <libcloud.storage.drivers.s3.S3Response object at 0x108616240>

    def parse_body(self):
        if len(self.body) == 0 and not self.parse_zero_length_body:
            return self.body

        try:
            try:
                body = ET.XML(self.body)
            except ValueError:
                # lxml wants a bytes and tests are basically hard-coded to str
                body = ET.XML(self.body.encode('utf-8'))
        except:
            raise MalformedResponseError('Failed to parse XML',
                                         body=self.body,
>                                        driver=self.connection.driver)
E           libcloud.common.types.MalformedResponseError: <MalformedResponseException in <libcloud.storage.drivers.google_storage.GoogleStorageDriver object at 0x1091885f8> 'Failed to parse XML'>: 'Hello World!'

../../Software/libcloud/libcloud/common/base.py:267: MalformedResponseError```

**Comment by tonybaloney** [comment]
Frustratingly, the Storage mock system relies on the mocked http methods being in the response class. For the regular mocked http client, they belong to the connection class. For what I assume was a mistake, is now part of every unit test that requires a raw connection. So the tests are still broken but the code works fine!

File: libcloud/httplib_ssl.py [hunk 4]
```diff
@@ -152,6 +152,7 @@ def _setup_ca_cert(self):
             if isinstance(libcloud.security.CA_CERTS_PATH, list):
                 if len(libcloud.security.CA_CERTS_PATH) > 1:
                     warnings.warn('Only 1 certificate path is supported')
+                    print(libcloud.security.CA_CERTS_PATH)
```
pquentin, line 155: nitpick: print() left

tonybaloney: gaaah! I did a search for `pdb.set_trace()`

pquentin:
With the latest changes, the HTTPS issues went away. Thanks! 💯 

I have an issue left with the download as stream method. Here's the relevant pytest output:

```
iterator = <bound method Response.iter_content of <Response [200]>>, chunk_size = 8096, fill_size = False, yield_empty = False

    def read_in_chunks(iterator, chunk_size=None, fill_size=False,
                       yield_empty=False):
        chunk_size = chunk_size or CHUNK_SIZE

        if isinstance(iterator, (file, httplib.HTTPResponse)):
            get_data = iterator.read
            args = (chunk_size, )
        else:
            get_data = next
            args = (iterator, )

        data = b('')
        empty = False

        while not empty or len(data) > 0:
            if not empty:
                try:
>                   chunk = b(get_data(*args))
E                   TypeError: 'method' object is not an iterator
```

Should I report such issues on JIRA now that the PR got merged?

tonybaloney:
@pquentin Can you gist the script you're using to test the download of a file?

tonybaloney:
@pquentin this is what I'm doing : https://gist.github.com/tonybaloney/e7159fbe72b81decce0395ceb87aa358

pquentin:
It's part of our integration tests. Until I can reduce this to a minimal script, all I can say is that I call `next()` on the iterator returned by `download_object_as_stream`.

PS: maybe your gist above can be made part of libcloud test suite?

libcloud/common/ovh.py
```diff
@@ -102,7 +102,7 @@ def request_consumer_key(self, user_id):
             'Content-Type': 'application/json',
             'X-Ovh-Application': user_id,
         }
-        httpcon = LibcloudHTTPSConnection(self.host)
+        httpcon = LibcloudConnection(host=self.host, port=443)
```
allardhoeve, line 105: Does this automatically enable HTTPS on port 443? What if the port is something like 8080? Or maybe even 4443 where the TLS port is on a non-standard port? How would you then enable TLS for that connection?

libcloud/httplib_ssl.py
```diff
@@ -141,213 +142,99 @@ def _parse_proxy_url(self, proxy_url):
...
+class LibcloudConnection(LibcloudBaseConnection):
+    timeout = None
+    host = None
+    response = None
 
-class LibcloudHTTPConnection(httplib.HTTPConnection, LibcloudBaseConnection):
-    def __init__(self, *args, **kwargs):
+    def __init__(self, host, port, **kwargs):
+        self.host = '{0}://{1}'.format(
+            'https' if port == 443 else 'http',
```
allardhoeve, line 168: I think this is an inflexible interface. We can have smart defaults, but there now is no way to have, for example, a TLS enabled port 4443.

libcloud/httplib_ssl.py
```diff
@@ -141,213 +142,99 @@ def _parse_proxy_url(self, proxy_url):
...
+    def request(self, method, url, body=None, headers=None, raw=False,
+                stream=False):
+        self.response = self.session.request(
+            method=method.lower(),
+            url=''.join([self.host, url]),
+            data=body,
+            headers=headers,
+            allow_redirects=1,
+            stream=stream,
+            verify=self.ca_cert if self.ca_cert is not None else self.verify
```
allardhoeve, line 193: This should probably be in a property so the logic for this is not embedded here

libcloud/httplib_ssl.py
```diff
@@ -141,213 +142,99 @@ def _parse_proxy_url(self, proxy_url):
...
-        except CertificateError:
-            e = sys.exc_info()[1]
-            raise ssl.SSLError('Failed to verify hostname: %s' % (str(e)))
+        self.session.timeout = kwargs.get('timeout', 60)
+
+    def request(self, method, url, body=None, headers=None, raw=False,
+                stream=False):
+        self.response = self.session.request(
+            method=method.lower(),
+            url=''.join([self.host, url]),
```
allardhoeve, line 188: If we use this, we'll probably run into issues for people that expect 'https://example.com/' to join nicely with '/api/', so it results in 'https://example.com/api/', like `urlparse.urljoin` does:

```python
>>> a = 'https://example.com/'
>>> b = '/api/'
>>> import urlparse
>>> urlparse.urljoin(a, b)
'https://example.com/api/'
>>> 
```

Can be achieved without urljoin by doing a `rstrip` and `lstrip`.

libcloud/httplib_ssl.py
```diff
@@ -141,213 +142,99 @@ def _parse_proxy_url(self, proxy_url):
...
+            data=body,
+            headers=headers,
+            allow_redirects=1,
```
allardhoeve, line 191: Do you really want to follow redirects by default here?

libcloud/storage/base.py
```diff
@@ -622,200 +624,62 @@ def _upload_object(self, object_name, content_type, upload_func,
...
-            **upload_func_kwargs)
-
-        if not success:
+        if stream:
+            response = self.connection.request(
+                request_path,
+                method=request_method, data=stream,
+                headers=headers, raw=True)
+            stream_hash, stream_length = self._hash_buffered_stream(
+                stream,
```
allardhoeve, line 634: So, if stream is a generator (it has `next`), does consuming it twice actually work? I would expect it to be empty the second time around, as generators generally do not generate the same content twice. This contrary to the file code below, which is clear.

libcloud/storage/base.py
```diff
@@ -622,200 +624,62 @@ def _upload_object(self, object_name, content_type, upload_func,
...
-                    response.connection.connection.send(b(chunk))
-                    response.connection.connection.send(b('\r\n'))
+        if upload_func:
+            upload_func(**upload_func_kwargs)
+
+        return {'response': response,
+                'bytes_transferred': stream_length,
+                'data_hash': stream_hash}
+
+    def _hash_buffered_stream(self, stream, hasher, blocksize=65536):
```
allardhoeve, line 658: I wonder why we have two implementations floating around in this file. Why the special casing for `__exit__` and `__next__`? This is a code smell for me. I feel that if we would refactor this function a little bit, we could do with a lot less complexity.

Can you perhaps add type hinting to this method so that it is clear what this method expects?

libcloud/storage/base.py
```diff
@@ -622,200 +624,62 @@ def _upload_object(self, object_name, content_type, upload_func,
...
+                total_len = total_len + len(s)
+            return (hasher.hexdigest(), total_len)
+        with stream:
+            buf = stream.read(blocksize)
+            while len(buf) > 0:
+                total_len = total_len + len(buf)
+                if isinstance(buf, str):
+                    hasher.update(buf)
+                else:
+                    hasher.update(buf)
```
allardhoeve, line 680: This if/else statement is equivalent to:

```python
hasher.update(buf)
```

libcloud/storage/base.py
```diff
@@ -622,200 +624,62 @@ def _upload_object(self, object_name, content_type, upload_func,
...
-                    response=response,
-                    iterator=iter(file_handle),
-                    chunked=chunked,
-                    calculate_hash=calculate_hash))
-
-        return success, data_hash, bytes_transferred
+                    hasher.update(s)
+                total_len = total_len + len(s)
+            return (hasher.hexdigest(), total_len)
+        with stream:
```
allardhoeve, line 673: This block can read:

```python
with stream:  # why though?
    for buf in iter(partial(stream.read, blocksize), ''):
          hasher.update(buf)
          total_len = total_len + len(buf)
```

This makes it easier to fold into the case where you use `for` already.

allardhoeve (commented):
So I finished reading the first parts of the connection code and streaming stuff. (up to libcloud/storage/drivers/azure_blobs.py).

I think we need to document the interface of the uploading well. There are (I feel) three cases that are handled in that function and it needs disentangling. Can you document the methods so that they have type hinting? I think that would make it clear what you expect this function to take without having to look too deeply into the tests.

I spot a generator, file_name and context manager handing there, but I don't know why.

libcloud/httplib_ssl.py
tonybaloney: this was fixed in #972  and #971 along with a test to do what you just suggested.

libcloud/common/ovh.py
tonybaloney: See #972

libcloud/httplib_ssl.py
tonybaloney: agreed. I'll update that.

libcloud/httplib_ssl.py
tonybaloney: I'll find some py2/3 code for doing that and add a test

tonybaloney:
@allardhoeve with regards to the comments on the hasher.

The existing library has tests for some really odd scenarios. I still don't understand why you would want to upload a generator to an S3 bucket?

I do understand why you would want to consume a buffered reader, like a `file` object, or a `StringIO` or `BytesIO` stream. That makes total sense. The existing drivers have checks in them assume that the storage base class will return a hash. Since there were 3 callbacks in the past, which implemented seperate logic for the hashing, one supported file_path. I understand that, one supported iterators (don't understand but it's existing functionality so I don't want to break back compat) and the other supported streams, but there wasn't really any testing for it.

This base class still needs significant refactoring, so far I have an account for GCP which I'm using to test file streams, but I need more integration scenarios. #970 should provide a good baseline of existing functionality and allow for better refactoring.

libcloud/httplib_ssl.py
allardhoeve: k

allardhoeve:
I hadn't noticed that you had already merged this. Guess the review isn't
need any more then :)

libcloud/storage/base.py
allardhoeve: I see that this has been entirely refactored. So never mind

tonybaloney:
@allardhoeve it absolutely is! we'll have to put together an 2.0 alpha release to get more feedback some users of the drivers. but all of your comments are being added to seperate PRs. This changeset just became absolutely ridiculous. But it was near impossible to make such a significant change without so many related changes. There will be a number of small PRs now to impact all the feedback and any issues.

allardhoeve:
Okey. But I just spent half an hour wrapping my head around the _upload
code and then the trunk already had an entirely refactored version. So it's
pretty hard to review now, because it is pretty much a moving target.

pquentin:
@tonybaloney Here's what I'm doing which still fails as of 3dd6d018:

```

driver = GoogleStorageDriver(key='GOOG***', secret='***')
remote_object = driver.get_object('containername', 'objectname')
print(remote_object)
stream = driver.download_object_as_stream(remote_object)
next(stream)
```

libcloud/httplib_ssl.py
tonybaloney: eeek. that one is dangerous. 
```
>>> urlparse.urljoin('http://test.com', '/test')
'http://test.com/test'
>>> urlparse.urljoin('http://test.com', '//test')
'http://test'
```

tonybaloney: ha. that found an error in morph_action_hook when the Connection object is instantiated from a URL. It would have a double slash. Pretty sure that would apply for <1.5.0

PR #925, opened 2016-10-27
[LIBCLOUD-867] Add st1 and sc1 volume types to valid types

smgoller:
### Description

EC2 supports additional storage types geared toward throughput and cold storage. We have a need for the cold storage volume type but libcloud currently doesn't support it. It seemed a good idea to just put up a pull request for this instead of asking for the change since it's so small.

We didn't feel there was a need to create a test for this change, nor additional documentation examples. If this is not the case we can add them.
### Status
- done, ready for review
- [*] [Code linting](http://libcloud.readthedocs.org/en/latest/development.html#code-style-guide) (required, can be done after the PR checks)
- [ ] Documentation


tonybaloney:
Seems simple enough, 👍 thanks for the PR.

PR #938, opened 2016-11-02
[LIBCLOUD-873] Updated ProfitBricks Compute Driver (REST api v3)

matfin:
### Description

This PR includes changes to the ProfitBricks compute driver to drop support for the old SOAP api (now end of life) and provide support for v3 of the REST api.

All public operations provided by the new REST api are supported in this driver. 

New test fixtures have been added and all tests have been amended.

A change has been added to the libcloud/common/base.py file to ensure that a request that returns with a HTTP status of 202 is not flagged as a failed request. Before this, only 200 and 201 were flagged as successful requests.

### Status

- done, ready for review

- Code linting for the compute driver implementation and the accompanying test file has been completed using the ```flake8``` tool.

- Documentation has been updated consisting of examples of using all core and extended functionality specific to ProfitBricks. A demo file with instructions on creating two servers has been put together and these will be published on the ProfitBricks community website.

- All test fixtures for api calls have been added in JSON format, replacing the older XML fixtures. All tests are passing.

### Summary of changes

Refactoring the compute  driver to remove support for the EOL SOAP api and modifying api calls to use the REST api (v3).

Removing XML test fixtures and replacing them with JSON fixtures.

Providing wrappers for all new functionality provided in v3 of the Cloud API.

Addressing an issue in libcloud/common/base.py where a http status code of 202 would not be flagged as a successful request.


tonybaloney:
thanks @matfin for the detailed PR, the changes look good to me. Has this been tested against an active endpoint using a real account as well?

matfin:
Hi @tonybaloney - It has. It was also developed against a test account I was using.

tonybaloney:
then 👍

PR #940, opened 2016-11-08
Dimension Data - Enable api version

samuelchong:
### Description
Allow specifying api version to use. such as 
driver = cls('xxx', 'xxx', region='dd-au', api_version='2.4')
if no api version is provided, it will use the default version (usually the latest).

Validation is included to check for the oldest and latest version supported. 

### Status

- done, ready for review

- [X] Documentation


tonybaloney:
thanks @samuelchong won't the API behaviours available affect the choice of API version? What's the other benefit of using this override?

samuelchong:
Thanks @tonybaloney for the review. 

We are in a situation where the data contract changes from version 2.3 and 2.4. In view of support and maintainability, I think it's better to have version control.

The ideas behind this is
- if no api version is specified, then the latest version is used. This will imply breaking changes if it occurs. But use will benefit from the latest release
- Having specify a version, consistency is assured with no changes required until the version is out dated and no longer supported. 
- We will avoid breaking changes as much as we can. 

An example of use of this version control for backward compatibility 
https://github.com/samuelchong/libcloud/commit/9cb7785e1707b7c688820a7a7a6a8a1df9c19597

Some other benefit:
-    oldest_api_version = 2.2
-    latest_api_version = 2.4
-    active_api_version = 2.3
  - Deprecate a version when it is outdated, or give warning for a period of time warning user to upgrade before it fall out of support.
  - Avoid complex logic to support X number of versions which will get messy when more versions are being released.
  - Allow testing of beta version by using latest_api_version
  - Default version (stable version) is set to active_api_version unless user specify one

Further feedback and comment welcome.

tonybaloney:
ah i see, yes if you're also planning on building logic into the driver around the models and behaviours of the API.
the `distutils` package has a `LooseVersion` class which has a `__cmp__` method for comparing values, this would be cleaner than using the `float(` converter all over the place

http://epydoc.sourceforge.net/stdlib/distutils.version.LooseVersion-class.html

I don't know if it's supported in 2.6, but work checking. Otherwise implement a separate version class method to do `__cmp__` comparison within `libcloud.common.dimensiondata`

tonybaloney:
also implement `__lt__` and `__eq__` for python 3. https://docs.python.org/3/whatsnew/3.0.html#ordering-comparisons

samuelchong:
thx @tonybaloney. I will make the changes and resubmit for review.

samuelchong:
Tested  LooseVersion('2.2') > LooseVersion('2.1') with python 2.6, 2.7 and 3.5.

samuelchong:
Tested with python version 2.6, 2.7, 3.5. 
LooseVersion('2.4') > LooseVersion('2.2')

samuelchong:
Hi @tonybaloney, I have used a customised version of LooseVersion in DD common class. A workaround to unsupported disutils.version issues in travis ci pylint.

tonybaloney:
ok :+1:

PR #952, opened 2016-11-23
[LIBCLOUD-879] Add support of node without public IP in LB

charly37:
### Description

Link to jira ticket 879 
https://issues.apache.org/jira/browse/LIBCLOUD-879

I had a condition to only try to grab the publicIP of a node if it had one since publicIP is not mandatory.

### Status

This PR is ready to be review. I run the test (tox) before and after the change and no impact.

- [X] [Code linting] I tried to followed the guidline
- [X] [Documentation] No change require
- [X] [Tests] Added 1 test case to cover the new behavior
- [X] [ICLA] Minor change so I skip it

### DIFF
--- a/libcloud/loadbalancer/drivers/gce.py
+++ b/libcloud/loadbalancer/drivers/gce.py
@@ -338,9 +338,12 @@ class GCELBDriver(Driver):
         # would be found if it was there.
         if hasattr(node, 'name'):
             member_id = node.name
-            member_ip = node.public_ips[0]
         else:
             member_id = node
+
+        if hasattr(node, 'public_ips') and len(node.public_ips) > 0:
+            member_ip = node.public_ips[0]
+        else:
             member_ip = None
 
         extra = {'node': node}
--- a/libcloud/test/loadbalancer/test_gce.py
+++ b/libcloud/test/loadbalancer/test_gce.py
@@ -189,6 +189,13 @@ class GCELoadBalancerTest(GoogleTestCase):
         self.assertEqual(member.id, node.name)
         self.assertEqual(member.port, balancer.port)
 
+    def test_node_to_member_no_pub_ip(self):
+        node = self.driver.gce.ex_get_node('libcloud-lb-nopubip-001',
+                                           'us-central1-b')
+        balancer = self.driver.get_balancer('lcforwardingrule')
+        member = self.driver._node_to_member(node, balancer)
+        self.assertIsNone(member.ip)
+
     def test_forwarding_rule_to_loadbalancer(self):
         fwr = self.driver.gce.ex_get_forwarding_rule('lcforwardingrule')
         balancer = self.driver._forwarding_rule_to_loadbalancer(fwr)
--- a/libcloud/test/compute/fixtures/gce/zones_us-central1-b_instances_libcloud-lb-nopubip-001.json
+++ b/libcloud/test/compute/fixtures/gce/zones_us-central1-b_instances_libcloud-lb-nopubip-001.json
@@ -1,0 +1,49 @@
+{
+  "canIpForward": false,
+  "creationTimestamp": "2013-12-13T10:51:25.165-08:00",
+  "disks": [
+    {
+      "boot": true,
+      "deviceName": "libcloud-lb-demo-www-001",
+      "index": 0,
+      "kind": "compute#attachedDisk",
+      "mode": "READ_WRITE",
+      "source": "https://www.googleapis.com/compute/v1/projects/project_name/zones/us-central1-b/disks/libcloud-lb-demo-www-001",
+      "type": "PERSISTENT"
+    }
+  ],
+  "id": "11523404878663997348",
+  "kind": "compute#instance",
+  "machineType": "https://www.googleapis.com/compute/v1/projects/project_name/zones/us-central1-b/machineTypes/n1-standard-1",
+  "metadata": {
+    "fingerprint": "09vSzO6KXcw=",
+    "items": [
+      {
+        "key": "startup-script",
+        "value": "apt-get -y update && apt-get -y install apache2 && hostname > /var/www/index.html"
+      }
+    ],
+    "kind": "compute#metadata"
+  },
+  "name": "libcloud-lb-nopubip-001",
+  "networkInterfaces": [
+    {
+      "name": "nic0",
+      "network": "https://www.googleapis.com/compute/v1/projects/project_name/global/networks/default",
+      "networkIP": "10.240.94.66"
+    }
+  ],
+  "scheduling": {
+    "automaticRestart": true,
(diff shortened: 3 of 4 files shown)


libcloud/loadbalancer/drivers/gce.py
```diff
@@ -338,9 +338,12 @@ def _node_to_member(self, node, balancer):
         # would be found if it was there.
         if hasattr(node, 'name'):
             member_id = node.name
-            member_ip = node.public_ips[0]
         else:
             member_id = node
+
+        if (hasattr(node, 'public_ips') and (len(node.public_ips) > 0)):
```
tonybaloney: no need for parens here and it's not very Pythonic to use them either ;-)

charly37:
Updated following @tonybaloney review.

charly37:
I was wandering if I need to provide any extra information to move forward on this PR following the changes I made last week ?

libcloud/loadbalancer/drivers/gce.py
```diff
@@ -338,9 +338,12 @@ def _node_to_member(self, node, balancer):
         # would be found if it was there.
         if hasattr(node, 'name'):
             member_id = node.name
-            member_ip = node.public_ips[0]
         else:
             member_id = node
+
```
allardhoeve, line 343: Although a correct solution, I think we should solve this by fixing the `Node` object, which should always have a `public_ips` attribute that is empty if there are no public IPs. This has the benefit of us needing to code that only once instead of all over the place.

Also, I would like to see a testcase, so this functionality never gets accidentally removed.

allardhoeve: Testcases are added by adding the Google API output for a node without public_ips to the fixture, then adding a test to the Node object to make sure the node has an empty list, then this function needs no conditionals, just:

```python
try:
    member_ip = node.public_ips[0]
except IndexError:
    member_ip = None
```

allardhoeve (changes requested):

libcloud/loadbalancer/drivers/gce.py
allardhoeve: The reason I ask for a testcase is that it makes it possible to refactor this kind of thing without introducing new bugs.

allardhoeve (changes requested):
I'd like a new test for this behaviour and also some refactoring of the logic into the `Node` object.

allardhoeve:
Hey Charles, sorry to spring more work on you after 20 days, but I'd like this to keep working next year :)

libcloud/loadbalancer/drivers/gce.py
charly37: @allardhoeve Thx a lot for the feedback. 

1/ node object : In fact the node object is good.  The node IP has indeed an empty "publicIP" which lead to the error since we try to access the first element. Maybe I did not fully understand your feedback ?
2/ test case : I can definitively try to add a test case (will be a good experience for me). 

Will update the PR with a test case ASAP and wait for your clarification about the point 1/

Thx again guys for the feedback.

allardhoeve: 1/ you check if the `Node` object has a `public_ips` attribute, why? Does it not always have the attribute? If not, we should make sure it always has the attribute. If so, you can remove the `hasattr`.

charly37: Thx @allardhoeve for your clarification. 
In fact checking the length of "public_ips" was my first idea but it do not work (break some test case). The "Node" object can in fact be a string and not a proper node object. That is probably why the actual code test for the attribute "name". This is also written in a comment that I saw after my initial test break the regression 
```
        # A balancer can have a node as a member, even if the node doesn't
        # exist.  In this case, 'node' is simply a string to where the resource
        # would be found if it was there.
```
That is why I keep the same design as the actual code where the attribute "name" is tested before accessing it for my new change. I hope it clarify ?
PS : test case in progress
PSS : Node is a string when the load balancer reference a non longer existing node

allardhoeve: Huh, I think that is not a really good design from a user's perspective. Maybe we can check with the original author what the intent was? I think having incompatible mixed types is a smell, so maybe we should set node to none and have a different field for the original node name?

charly37: I think it would be indeed possible to design it another way to avoid mixing String and Node objects nevertheless it would change the actual behavior (that could be use by some existing application) and that is why I limited my change to fixing the publicIP issue with the actual design. 

What if I open another dedicated JIRA ticket to discuss this point ?

charly37:
@allardhoeve I added a test case as suggested.

libcloud/test/loadbalancer/test_gce.py
```diff
@@ -188,6 +188,12 @@ def test_node_to_member(self):
         self.assertEqual(member.ip, node.public_ips[0])
         self.assertEqual(member.id, node.name)
         self.assertEqual(member.port, balancer.port)
+        # Test node without publicIP
```
allardhoeve: Can you make this a test of its own?

allardhoeve (changes requested):
:+1: looks good, but I think tests should be separate.

allardhoeve:
Thanks Charles! I'll merge this momentarily.

allardhoeve:
The tests are failing, but that has nothing to do with this PR

allardhoeve:
@tonybaloney, any idea why the tests might be failing? They fail on `apt-get install libvirt`, but I cannot find that provisioning code in the repo anywhere.

charly37:
@allardhoeve Tests are failing because of the Travis setup. Looking at some other PR it is the same issue but it seems someone find a workaround by using libvirt-bin package instead of libvirt (see https://github.com/apache/libcloud/commit/20ab4cbf89c855a2e8b67f49ce239de506eab865 )I was thinking to do it in my PR too but i m afraid it generate conflict.

libcloud/test/loadbalancer/test_gce.py
charly37: Sure. Will do it today

allardhoeve:
Please try to merge master into this branch or rebase against it, thanks.

charly37:
@allardhoeve Tests have been split as suggest. Rebase from trunk to fix the TRAVIS jobs.
I think it is all clean to be merge ;)

allardhoeve:
Yes! Will merge today

charly37:
Thx again for the review and help.
I also open a ticket to discuss the design of mixing string and node object as discuss previously in the PR review : https://issues.apache.org/jira/browse/LIBCLOUD-883

allardhoeve:
Merged, thanks. Should be closed in a minute.

pquentin:
Nice catch @charly37!

PR #955, opened 2016-11-28
Bug fix for ex_create_multiple_nodes Google Cloud disk auto delete

jbaublitz:
### Description

There seems to be a small bug in `ex_create_multiple_nodes` that will cause the function to create disks that are auto delete regardless of the parameter passed in. This patch should fix this behavior.

### Status

- ready for review

- [ ] Documentation


supertom:
LGTM!  @erjohnso, OK to merge.

PR #967, opened 2017-01-04
fix vsphere driver issue LIBCLOUD-806

tonybaloney:
Tested locally, but not pushed test case as it only works on Python2.7


tonybaloney:
Merging changes

PR #968, opened 2017-01-06
Ensure that request_path is restored on exception 

tetron:
### Description

When requesting a GCE license, restore the request_path in a finally block.  Uses same pattern as used in other parts of the driver.

### Status

Ready for review


tonybaloney:
it would be better to put that in a context manager, but if that's how it's dealt with in other places it'll have to do for now.
Thanks for the contribution 👍

PR #969, opened 2017-01-06
Add support for selecting alternate cloud_environments

tetron:
### Description

Add support for selecting alternate cloud_environments via "cloud_environments" keyword option when creating compute driver object. Supports default, AzureChinaCloud, AzureUSGovernment, and AzureGermanCloud.  Affects endpoints used by azure_arm compute driver.

### Status

- In testing

- [x] Documentation

### DIFF
--- a/docs/compute/drivers/azure_arm.rst
+++ b/docs/compute/drivers/azure_arm.rst
@@ -49,8 +49,8 @@ You can select an alternate cloud environment using the "cloud_environment"
 parameter to AzureNodeDriver constructor.  Available alternate cloud
 environments are 'AzureChinaCloud', 'AzureUSGovernment' and 'AzureGermanCloud'.
 You can also supply explicit endpoints by providing a dict with the keys
-'resourceManagerEndpointUrl', 'activeDirectoryEndpointUrl' and
-'activeDirectoryResourceId'.
+'resourceManagerEndpointUrl', 'activeDirectoryEndpointUrl',
+'activeDirectoryResourceId' and 'storageEndpointSuffix'.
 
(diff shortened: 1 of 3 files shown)


tonybaloney:
`urlparse` is a Python 2.7 specific standard library. We have a compatibility option in `libcloud.utils.py3`

tetron:
Thanks, I'll see what I can do.

tetron:
@tonybaloney is this ready to merge?

tonybaloney:
@tetron please can you change the `__init__` keyword to `region` to keep it consistent with the ec2 driver and others.

tetron:
Azure has two concepts: "location" and "cloud environment".  A cloud environment is a separately operated installation of Azure, it has a distinct set of endpoints and separate accounts.  Within a cloud environment there can be many locations (data centers).

When setting up a connection to Azure, it is only necessary to know the "cloud environment".  The "location" supplied in API requests that need it.

Currently `AzureNodeDriver` maps the `region` keyword to `location`, and accepts `cloud_environment` as an additional parameter which is passed through to `AzureResourceManagementConnection`.  If we renamed `cloud_environment` to `region` in `AzureResourceManagementConnection` that would give `region` an inconsistent meaning between the two classes.  

So we can either:

* Leave it as-is
* Merge cloud_environment and location into a single value for region (which gets split by the driver): user would specify `centralus` (use default) or `AzureChinaCloud/southchina` (use an alternate cloud environment). 
* Something else?

tonybaloney:
sorry that's just my lack of understand of Azure, agreed 👍  merging.

PR #970, opened 2017-01-09
[WIP] Introduce a standard set of integration tests to libcloud

tonybaloney:
### Description

Testing of libcloud drivers is primarily done using the libcloud unit test suite, which mocks out all of the HTTP request functionality to a set of static methods.

It would be very beneficial to both community developers and release managers to have an integration test to check the end-to-end functionality of libcloud

### Status

- work in progress

### Checklist

- [ ] Documentation

## Components

- [x] Create a `test` compute driver and implement basic functions
- [x] - Create a test data package to automatically ensure data is correctly encoded and decoded by the library
- [x] - Create tests for propagation of ports, hosts and IPs
- [x] - Create tests for instantiation of drivers
- [x] - Use a micro web framework to run the web service

CC @pquentin @Kami @erjohnso @supertom


pquentin:
Great idea!

If this is also extended to storage, maybe the test storage driver could be the local storage driver? It would need to support metadata, though, but that should not be too hard, either with xattr or a hidden folder containing JSON or pickle files with the metadata.

I also see that libcloud tests support plugging our own secrets, in theory this would allow to run real end-to-ends tests without mocks, right? That would be too slow for normal usage, but would allow to catch more issues. Can you please confirm that this is not what this issue is about?

tonybaloney:
yes, I want it for storage, file uploads, hashing, streaming downloads, checking things like 'raw' connections to csv files and the like, header propagation, JSON and XML validation.

Libcloud does support your own secrets, but all of the test modules are currently using the MockHttp class, which doesn't look hardly anything like the actual implementation. I've introduced `requests_mock` which allows you to introduce a http-handler as a context manager so you're only mocking out the base http call. 

This PR is about running a real end-to-end test, without having to spend $$ on a cloud provider. I hope that I can get it to work with Travis. But as a minimum it'll be part of the release process.

PR #971, opened 2017-01-09
fix issue where port is not propagated down to connection URL

tonybaloney:
Noticed that a custom port does not roll down properly to the connection class.

Also, there is still an outstanding issue that unless the port is 443, the connection will be assumed as plain HTTP.

E.g. setting port as 9999 but Secure=True will still issue a plain HTTP connection

### DIFF
--- a/libcloud/httplib_ssl.py
+++ b/libcloud/httplib_ssl.py
@@ -163,9 +163,10 @@ class LibcloudConnection(LibcloudBaseConnection):
     response = None
 
     def __init__(self, host, port, **kwargs):
-        self.host = '{0}://{1}'.format(
+        self.host = '{0}://{1}{2}'.format(
             'https' if port == 443 else 'http',
-            host
+            host,
+            ":{0}".format(port) if port not in (80, 443) else ""
         )
         # Support for HTTP proxy
         proxy_url_env = os.environ.get(HTTP_PROXY_ENV_VARIABLE_NAME, None)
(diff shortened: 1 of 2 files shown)


tonybaloney:
mergey merge

Kami:
Good catch.

libcloud/httplib_ssl.py
```diff
@@ -163,9 +163,10 @@ class LibcloudConnection(LibcloudBaseConnection):
     response = None
 
     def __init__(self, host, port, **kwargs):
-        self.host = '{0}://{1}'.format(
+        self.host = '{0}://{1}{2}'.format(
             'https' if port == 443 else 'http',
```
Kami, line 167: Do we have access to `secure` argument here?

Because if we do, it looks like we still don't correctly handle one scenario - https on a non default port (protocol is not set correctly). For example, secure=True, port=8080 should result in `https://<host>:8080`.

Or am I missing something?

Kami: Edit: Sorry, I should have read the description better - you already addressed this scenario there.

Also checked the code and it looks like "secure" argument is not passed to the constructor right now so yeah, we would need to change some code to make it work.

tonybaloney: @Kami yes, absolutely. That's a separate PR. I'm back to making one change at a time!

PR #972, opened 2017-01-10
Fix secure propagation to connection class

tonybaloney:
Since there is no longer a secure/insecure connection class, they are one and the same.

This change adds new tests to check that

- When a connection is defined as secure, the connection class will use the https scheme.
- When a connection scheme implies a port, the port is not included in the URL
- When a connection defines secure and an unusual port, the scheme is propagated

It uncovered 1 issue with secure connections not being explicit for unusual ports.

### DIFF
--- a/libcloud/httplib_ssl.py
+++ b/libcloud/httplib_ssl.py
@@ -162,9 +162,10 @@ class LibcloudConnection(LibcloudBaseConnection):
     host = None
     response = None
 
-    def __init__(self, host, port, **kwargs):
+    def __init__(self, host, port, secure=None, **kwargs):
+        scheme = 'https' if secure is not None and secure else 'http'
         self.host = '{0}://{1}{2}'.format(
-            'https' if port == 443 else 'http',
+            'https' if port == 443 else scheme,
             host,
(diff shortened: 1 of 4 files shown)


tonybaloney:
@Kami this is the second change to fix the issue you mentioned.

tonybaloney:
merging

libcloud/httplib_ssl.py
```diff
@@ -162,9 +162,10 @@ class LibcloudConnection(LibcloudBaseConnection):
     host = None
     response = None
 
-    def __init__(self, host, port, **kwargs):
+    def __init__(self, host, port, secure=None, **kwargs):
+        scheme = 'https' if secure is not None and secure else 'http'
```
allardhoeve, line 166: I think the defaults would be more clear if this read:

```
def __init__(self, host, port, secure=None, **kwargs):
    if secure is None and port == 443:
        secure = True
    scheme = 'https' if secure else 'http'

    # ... rest of the code
```

I dislike inline logic because it is hard to find for the casual reader.

Also, a docstring would be nice.

tonybaloney: :+1:

PR #973, opened 2017-01-10
Fix access to httplib.HTTPResponse class from RawResponses

tonybaloney:
I've noticed that some drivers rely on access to a `httplib.HTTPResponse` object as a property of `RawResponse`. This was updated in #728 to the requests response class, which would, in turn, break a few implementations.

To avoid backward compatibility issues, this introduces a proxy class with `HTTPResponse` methods and properties and tests to validate access.

None of the existing unit tests would have caught this behaviour since the response classes are mocked. This issue was caught in #970

### DIFF
--- a/libcloud/httplib_ssl.py
+++ b/libcloud/httplib_ssl.py
@@ -24,7 +24,7 @@ import warnings
 import requests
 
 import libcloud.security
-from libcloud.utils.py3 import urlparse
+from libcloud.utils.py3 import urlparse, PY3
 
 
 __all__ = [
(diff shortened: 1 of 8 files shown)


tonybaloney:
build fixed. PR complete https://travis-ci.org/tonybaloney/libcloud/builds/190789008

tonybaloney:
@pquentin does this fix your script?

pquentin:
@tonybaloney Sorry, no. I double-checked to ensure I had the code from this PR. I still get the same issue, both with my code and the minimal example from https://github.com/apache/libcloud/pull/923#issuecomment-271784099

libcloud/httplib_ssl.py
```diff
@@ -238,6 +238,50 @@ def close(self):  # pragma: no cover
...
+
+    def read(self, amt=None):
+        return self._response.text
+
+    def getheader(self, name, default=None):
+        """
+        Get the contents of the header name, or default
+        if there is no matching header.
+        """
+        if name in self._response.headers.keys():
```
Kami, line 257: Does this handle mixed case (aka, should we return correct entry no matter what casing is used)?

IIRC, built-in library lower cases all the header keys, but I could be wrong...

tonybaloney: It's a case insensitive dictionary yes http://docs.python-requests.org/en/master/api/#requests.Response.headers. Although I need to check if that's not just on the protocol method and requesting keys will just ignore it.

tonybaloney:
@pquentin https://github.com/apache/libcloud/pull/973/commits/1eb58838df91ff8d9156a341b1ef560f601f7277 should fix the issue, I've been testing that script you sent me against my GCP account and it works. The unit tests are broken, but only because the mocks are missing certain properties..

pquentin:
@tonybaloney Awesome, it works! The only difference is that the generator yields one byte per `next()` call, while the old one was yielding byte chunks (8096 bytes is the default value I think). Not an issue for me, but I thought I would mention it.

libcloud/storage/drivers/s3.py
```diff
@@ -419,7 +419,7 @@ def download_object_as_stream(self, obj, chunk_size=None):
         return self._get_object(
             obj=obj, callback=read_in_chunks,
             response=response,
-            callback_kwargs={'iterator': response.iter_content,
+            callback_kwargs={'iterator': response.iter_content(),
```
tonybaloney: @pquentin I suspect adding chunk size parameter here would do the trick

tonybaloney: @pquentin I'm also curious to see a performance difference between the 2. The old code buffered out the download to get the stream size and then returned the iterator. I think for a large file that would really hurt both time and memory consumption

pquentin:
@tonybaloney Perfect, this fixes it! I no longer see any differences between the old libcloud and the requests-based libcloud! 👏 Thank you for you work and fixing those issues this quickly.

tonybaloney:
merging

PR #974, opened 2017-01-10
Bug fix for deletion against asynchronous Google subnetwork API

jbaublitz:
### Description
libcloud assumes that the Google subnetwork API is synchronous which does not appear to be the case. This causes breakage in ansible where the libcloud API for `ex_destroy_subnetwork` reports that is has successfully destroyed the subnetwork. Immediate calls to `ex_destroy_network` seem to indicate that the subnetwork has not been destroyed as the network is reported to still be in use by the subnetwork. This bug fix is also verified to fix this race condition that causes ansible to fail when sequentially executing the code in PR ansible/ansible-modules-core#3773 first for subnetwork deletion and then for network deletion.

### Status

- done, ready for review

- [ ] Documentation

### DIFF
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -6900,7 +6900,7 @@ class GCENodeDriver(NodeDriver):
                 region_name = region.name
 
         request = '/regions/%s/subnetworks/%s' % (region_name, subnet_name)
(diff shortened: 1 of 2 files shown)


tonybaloney:
makes sense. thanks for the contribution 👍 

/cc @supertom @erjohnso

PR #976, opened 2017-01-13
Refactoring LibcloudConnection and Connection classes from feedback

tonybaloney:
Issues Raised:

- Use URLJoin instead of `''.join` : https://github.com/apache/libcloud/pull/923#discussion_r95320130 raised by @allardhoeve 
- Create a test to ensure creating Connection instances with a trailing / and then using request with a method with a leading / does not create malformed URL requests
- Make allow_redirects configurable  (but true by default?) https://github.com/apache/libcloud/pull/923#discussion_r95320171 @allardhoeve 
- Move verification logic into property in LibcloudConnection https://github.com/apache/libcloud/pull/923#discussion_r95319749
- Fix if/else in hash buffer code https://github.com/apache/libcloud/pull/923#discussion_r95324871
- Remove SSL/TLS error messages and custom socket.error handling code as it's no longer needed

### DIFF
--- a/libcloud/common/base.py
+++ b/libcloud/common/base.py
@@ -674,9 +674,7 @@ class Connection(object):
         return response
 
     def morph_action_hook(self, action):
-        if not action.startswith("/"):
-            action = "/" + action
-        return self.request_path + action
+        return urlparse.urljoin(self.request_path, action)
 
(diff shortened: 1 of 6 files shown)


tonybaloney:
merging

PR #977, opened 2017-01-13
Documentation in preparation for 2.0.0RC1

tonybaloney:


tonybaloney:
merging when build completes.

PR #978, opened 2017-01-18
Fix LIBCLOUD-887

tonybaloney:
If a user sets LIBCLOUD_DEBUG, they will receive an error and no requests will be made. This is happening on trunk and v2.0.0rc1-tentative

```
In ./libcloud/utils/loggingconnection.py
LoggingConnection.request does not accept the stream keyword which causes an error:
TypeError: request() got an unexpected keyword argument 'stream'
```

Reported by @supertom


tonybaloney:
Passed. Merging.

PR #979, opened 2017-01-20
[LIBCLOUD-888] TypeError occurred when calling ex_list_nics in Azure ARM Driver

zoopi01:
### Description

In Azure ARM mode, 'TypeError: not all arguments converted during string formatting' occurs when calling ex_list_nics() method.
According to https://msdn.microsoft.com/en-us/library/mt163627.aspx, The request URI form requires two parameters subscription-id and resource-group-name. ex_list_nics() has resource-group-name as a parameter, but it is not presented in the arguments list to format.

### Status

done, ready for review

- [ ] Documentation

### DIFF
--- a/libcloud/compute/drivers/azure_arm.py
+++ b/libcloud/compute/drivers/azure_arm.py
@@ -868,8 +868,8 @@ class AzureNodeDriver(NodeDriver):
         :rtype: ``list`` of :class:`.AzureNic`
         """
 
-        action = "/subscriptions/%s/providers/Microsoft.Network" \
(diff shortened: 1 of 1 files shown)


tonybaloney:
thanks @zoopi01 have you checked this fix against the live API?

cc @tetron

tetron:
I haven't tried it, but the fix looks correct.

zoopi01:
Yes, I have. I am using it now. @tonybaloney @tetron

tonybaloney:
:+1:

PR #980, opened 2017-01-26
Add Security Token Service (STS) support for Route 53

vaygr:
### Description

This adds support for [Temporary Security Credentials](http://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp.html) for Route 53, extending it beyond [existing for EC2](http://libcloud.readthedocs.io/en/latest/compute/drivers/ec2.html#using-temporary-security-credentials) and S3.

### Status

- done, ready for review


tonybaloney:
great contribution thanks @vaygr 👍

PR #993, opened 2017-02-24
Add Cloudscale Host Keys And Documentation Improvements

davidhalter:
### Description

This is a pretty small set of changes.

It only includes two things: 

- We have had some issues where we added stuff like host keys to our API. This should be reflected in Libcloud. Basically the default behavior is to store all keys in the `extra` field that are not already part of the server object.
- cloudscale.ch's name was never cloudscale. It's always written as the full URL.

### Status

Done, ready for review

- [x] Documentation


tonybaloney:
thanks @davidhalter changes LGTM 👍

PR #1003, opened 2017-03-07
Azure ARM optimizations (get all results on list_nodes and fix actions)

mgogoulos:
### Description

- show all results on list_nodes (current API call gets only first batch)
- fix destroy_node (contains while statements that never exit)
- correct ex_stop_node API call
- optimize states on _to_node function

### Status

done, ready for review

- [ ] Documentation

### DIFF
--- a/libcloud/compute/drivers/azure_arm.py
+++ b/libcloud/compute/drivers/azure_arm.py
@@ -342,8 +342,18 @@ class AzureNodeDriver(NodeDriver):
                      % (self.subscription_id)
         r = self.connection.request(action,
                                     params={"api-version": "2015-06-15"})
+
+        nodes_data = r.object["value"]
+        # get paginated results
+        # remove host from nextLink
+        while r.object.get("nextLink"):
+            next_url = r.object.get("nextLink").split(self.connection.host)[1]
+            r = self.connection.request(next_url,
+                                        params={"api-version": "2015-06-15"})
+            nodes_data.extend(r.object["value"])
+
         return [self._to_node(n, fetch_nic=ex_fetch_nic)
-                for n in r.object["value"]]
+                for n in nodes_data]
 
     def create_node(self,
                     name,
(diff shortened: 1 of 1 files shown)


libcloud/compute/drivers/azure_arm.py
```diff
@@ -649,63 +659,48 @@ def destroy_node(self, node, ex_destroy_nic=True, ex_destroy_vhd=True):
...
-                        # LibcloudError.  Wait a bit and try again.
-                        time.sleep(10)
-                    else:
-                        raise
+            try:
+                resourceGroup = node.id.split("/")[4]
+                self._ex_delete_old_vhd(
+                    resourceGroup,
+                    vhd_uri)
+                break
```
tonybaloney, line 693: you can't break here.

tonybaloney (changes requested):

tetron:
The loops in destroy_node are there so it can keep trying to delete the NIC and VHD objects, which are not actually released until some time later after accepting the DELETE request for the VM.  The current code doesn't have "while statements that never exit", they break out of the loop after a successful delete of the NIC and VHD.

I recommend restoring the retry loops but making them bounded (maybe retry 12 times with a 10 second wait between attempts).

pquentin:
Thank you for this @mgogoulos. Unfortunately we can't merge this as-is right now:

 * the pagination in list_nodes is a welcome addition but it needs to be tested (using fixtures for example)
 * as explained by @tetron, the while loops can probably be encapsulated somewhere else by making them bounded and possibly by adding exponential back-off but the should not be simply removed
 * The powerOff is a good one [according to the docs](https://docs.microsoft.com/en-us/rest/api/compute/virtualmachines/virtualmachines-stop).
 * The changes to _to_node make sense. However, I will leave the `pass` on `"ProvisioningState/succeeded"` because on success, the real state will be set later on. This a harmless bug in your PR hidden by the fact that you don't break after setting the state to NodeState.RUNNING. If this was done for another reason, then please tell me.
  * `created_at` use the `time` field, but I see `timeCreated` in the fixtures that seems to be more suited, can you please explain why you're using `time`?

In other words, I plan to:
  * wait for tests before adding pagination
  * add the powerOff change
  * wait to understand time vs timeCreated
  * add ProvisioningState/updating and PowerState/stopped support to _to_node

pquentin:
Executed the plan outlined above. Thank you! Please open a new PR for the other changes that could not be accepted yet.

PR #1006, opened 2017-03-09
LIBCLOUD-901 Correctly wrap requests library Responses with 4xx/5xx status codes

richardxia:
### Description

This fixes the constructor of `RawResponse` to explicitly check if `response is not None`. This is necessary because the requests library overloads the `__nonzero__`/`__bool__` methods on the `Response` object to return False if a 4xx or 5xx status code is returned. This [has been fixed](https://github.com/kennethreitz/requests/pull/2587) in the 3.0 branch of requests, but that version hasn't been released yet.

### Status
done, ready for review

I haven't checked any of these because all I did was make a 10 character change.

- [ ] Documentation

### DIFF
--- a/libcloud/common/base.py
+++ b/libcloud/common/base.py
@@ -282,7 +282,7 @@ class RawResponse(Response):
         self._error = None
         self._reason = None
         self.connection = connection
-        if response:
+        if response is not None:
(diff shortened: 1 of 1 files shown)


tonybaloney:
great spot. thanks @richardxia for the contribution

PR #1015, opened 2017-04-02
Fix CertificateConnection in >2.0Rc1

tonybaloney:
### Description

In 2.0, the httplib module was used as a low-level module for making requests to the APIs, some drivers used a `CertificateConnection` class in `libcloud.common.base`, which accepted an x509 certificate to be used for encoding the requests.

Those drivers are : 
- Azure (classic)
- Docker (with TLS)
- OpenStack

This PR will reimplement the CertificateConnection class inside `libcloud.httplib_ssl` by mounting a `HTTPAdapter` to the requests session, detecting TLS and signing requests for the cert file.

We also need a test for this issue.

### Status

- done, ready for review

- [x] Documentation

### DIFF
--- a/libcloud/common/base.py
+++ b/libcloud/common/base.py
@@ -468,8 +468,6 @@ class Connection(object):
         if not hasattr(kwargs, 'cert_file') and hasattr(self, 'cert_file'):
             kwargs.update({'cert_file': getattr(self, 'cert_file')})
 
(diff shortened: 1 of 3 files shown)


tonybaloney:
tested against my actual Azure account and verified that I can call `list_nodes` now as it worked in 1.5.0

PR #1021, opened 2017-04-04
Add price_monthly in extra dict to digitalocean sizes

fjros:
### Description

DigitalOcean sizes are commonly referred to by their monthly price. This PR makes libcloud expose such information within the extra dictionary of Size. This new field is named 'price_monthly', as in the DigitalOcean API.

### Status

- done, ready for review


tonybaloney:
Thanks @fjros what about `price_hourly`? I can see that in the API too

fjros:
Hi @tonybaloney! `price_hourly` is already exposed as the `price` attribute of class `NodeSize`. For the sake of uniformity wrt other drivers, I think it's ok to leave it there and don't duplicate such value in the `extra` dict.

tonybaloney:
:+1:

PR #1022, opened 2017-04-04
Fix number of ec2 t2.small cpus to 1

fjros:
### Description

Minor fix: libcloud was incorrectly stating that t2.small sizes of EC2 feature 11 vcpus. This PR fixes the number of cpus to 1.

### Status

- done, ready for review


tonybaloney:
thanks @fjros good spot.

PR #1023, opened 2017-04-05
Added Import Snapshot and Describe Import Snapshot to EC2 compute driver

nirzari:
### Description

On Importing Snapshot in aws, it returns the import snapshot task which needs to be monitored by describe import snapshot task until it is ready. Hence, added a wait block method. Finally, it returns a volume snapshot.  

### Status
- done, ready for review

- [x] Documentation

### DIFF
--- a/libcloud/compute/drivers/ec2.py
+++ b/libcloud/compute/drivers/ec2.py
@@ -22,6 +22,7 @@ import sys
 import base64
 import copy
 import warnings
+import time
 
 try:
     from lxml import etree as ET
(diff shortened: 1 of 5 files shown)


libcloud/compute/drivers/ec2.py
```diff
@@ -3906,6 +3924,138 @@ def ex_delete_placement_group(self, name):
...
 
+    def ex_import_snapshot(self, client_data=None,
+                           client_token=None, description=None,
+                           disk_container=None, dry_run=None, role_name=None):
+        """
+        Imports a disk into an EBS snapshot. More information can be found
+        at https://goo.gl/sbXkYA.
+
+        :param  client_data: Describes the client specific data (optional)
+        :type   client_data: ''dict''
```
tonybaloney: these should be backticks, instead of single quotes. I'm not sure RST would mark them as an internal type otherwise.

tonybaloney (changes requested):
thanks for the contribution. changes look good. One small request is to change from singlequote to backtick in the docstrings

tonybaloney:
perfect 👍

PR #1025, opened 2017-04-06
LIBCLOUD-910 error with LIBCLOUD_DEBUG and compressed responses

tonybaloney:
As reported in LIBCLOUD-910, the loggingconnection, used with `LIBCLOUD_DEBUG` is enabled still inspects the `content-encoding` header and tries to decode the responses.

`requests` handles this via `urllib3` which has inbuilt decompression so this entire code block and module is not needed.

- [x] provide test
- [x] Fix issue
- [x] check test


tonybaloney:
@mgogoulos this relates to the LINODE issue as well, that driver has a whole bunch of copy+paste from the base libraries.

tonybaloney:
@supertom please can you run your script against this branch

tonybaloney:
nw, I ran it myself. it works

tonybaloney:
👍 merging

PR #1026, opened 2017-04-06
Fix Linode for 2.0x 

tonybaloney:
Changes in #1025 highlighted a problem with the Linode response class.

It has a lot of copy+paste from the Response `__init__` and no call to super. This looks related to some legacy code but its a common anti-pattern.

https://github.com/apache/libcloud/commit/fe72fc13a131b446e1f18cd22005f9c8fd03ce79

This changeset removes the duplicate (but now broken) code and calls `super` correctly.


tonybaloney:
merging.

PR #1028, opened 2017-04-09
rename libcloud.httplib_ssl to libcloud.http

tonybaloney:
rename libcloud.httplib_ssl to libcloud.http since it has nothing to do with httplib anymore


tonybaloney:
:+1:

PR #1030, opened 2017-04-10
Fix impossible location condition

techhat:
### Description
When requesting a list of publishers from other parts of the code (such as when listing images), a condition can be encountered where the `location` has been passed in (and therefore is not `None`), but  `self.default_location` is also set. The `location` that is passed in should be used, but the condition doesn't allow it to be.

### Status

- done, ready for review

- [ ] Documentation


tonybaloney:
makes sense. thanks @techhat and thanks for making your first contribution to the project!

tonybaloney:
👍  merging

PR #1035, opened 2017-04-13
Fix VhdImage and ex_list_public_ips in ARM driver

tonybaloney:
I noticed that `AzureVhdImage` `__repr__` would never work because it doesn't have a location field.

Also, this string would never render. 
https://github.com/apache/libcloud/blame/trunk/libcloud/compute/drivers/azure_arm.py#L1040-L1055

What is the correct URL? @techhat @tetron do you know?

### DIFF
--- a/libcloud/compute/drivers/azure_arm.py
+++ b/libcloud/compute/drivers/azure_arm.py
@@ -67,8 +67,8 @@ class AzureVhdImage(NodeImage):
         super(AzureVhdImage, self).__init__(urn, name, driver)
 
     def __repr__(self):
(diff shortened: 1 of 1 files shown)


tetron:
If it's consistent with the other API endpoints then it probably should be:

```
        action = "/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Network" \
                 "/publicIPAddresses" % (self.subscription_id, resource_group)
```

tetron:
Reference: https://docs.microsoft.com/en-us/rest/api/network/publicipaddresses

tonybaloney:
merging

PR #1037, opened 2017-04-21
Fix issues with OpenStack auth raised in LIBCLOUD-912 and LIBCLOUD-904

tonybaloney:
Noticed so far:

The existing test fixture returns the host as 127.0.0.1 which also happens to be the default host for a connection class. 

The test I just added fails to demonstrate the issue with

requests_mock.exceptions.NoMockAddress: No mock address: GET https://127.0.0.1/v2/1337/servers/detail

### DIFF
--- a/libcloud/common/openstack.py
+++ b/libcloud/common/openstack.py
@@ -297,6 +297,7 @@ class OpenStackBaseConnection(ConnectionUserAndKey):
     def _set_up_connection_info(self, url):
         result = self._tuple_from_url(url)
         (self.host, self.port, self.secure, self.request_path) = result
+        self.connect()
 
     def _populate_hosts_and_request_paths(self):
         """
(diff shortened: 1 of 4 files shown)


tonybaloney:
So the issue was the connect() would always be called in <2.0, because it meant something in httplib. Now requests is a higher level API so connect is only called lazily if the connection instance hasn't been set. In the case of the openstack auth, it has. So the host property would be loaded at init but then when the catalog is fetched, the property doesn't trickle down to the instance of the requests session.
Adding `connect()` to each call for the internal method that unpacks the catalog would fix this issue but might add a small overhead to each call since it's in morph_action_hook. But that takes it back to the speed of <2.0 anyway.

tonybaloney:
I noticed abiquo tests are being skipped. seperate issue.

tonybaloney:
merging.

PR #1038, opened 2017-04-21
Add i3 instance types for AWS

smullins7:
Followed the r4 commit, adding i3 instances types for AWS. First time caller so let me know if I missed anything, please and thank you!


libcloud/test/compute/test_ec2.py
```diff
@@ -429,23 +429,23 @@ def test_list_sizes(self):
...
             if region_name == 'us-east-1':
-                self.assertEqual(len(sizes), 61)
+                self.assertEqual(len(sizes), 67)
                 self.assertTrue('cg1.4xlarge' in ids)
                 self.assertTrue('cc2.8xlarge' in ids)
                 self.assertTrue('cr1.8xlarge' in ids)
                 self.assertTrue('x1.32xlarge' in ids)
             elif region_name == 'us-west-1':
-                self.assertEqual(len(sizes), 52)
+                self.assertEqual(len(sizes), 56)
```
tonybaloney: it's now 58, according to the tests.

tonybaloney (changes requested):
thank you for your first contribution! all looks ok except the tests are failing because the number of sizes is now wrong. run `tox -e py2.7` for example to run the test suite locally making sure it passes first or `python -m libcloud.test.compute.test_ec2` to just run the ec2 module

libcloud/test/compute/test_ec2.py
smullins7: Cool thank you! I fixed it and a couple more and just pushed. Fyi I had to copy the secrets file to get it to import. I'm not sure if there are setup steps I'm supposed to follow but I just did the following:
`cp libcloud/test/secrets.py-dist libcloud/test/secrets.py`

smullins7:
@tonybaloney looks like the tests pass, if this gets pulled in do you know when the next RC version would be released?

tonybaloney:
👍  this will go into 2.0.1

PR #1045, opened 2017-04-26
Add test to check that can list sizes for all valid ec2 regions

fjros:
### Description

EC2 compute driver fails to list sizes in some regions. The reason is that pricing hasn't been updated in such cases.

This PR adds a test to force the build fail if not all supported regions are working properly. I'd like to start a discussion about how to avoid this failure from happening again in the future (what about scraping prices as part of the build process?).

### Status

- start discussion

- [ ] Documentation


tonybaloney:
thanks @fjros theres a build error that needs fixing first.

fjros:
Hi @tonybaloney, that was the intended behavior of this PR :) I mean, this is something that I experience on a regular basis: Cannot list sizes of new EC2 regions because their pricing isn't available on libcloud.

I haven't found the time to look into the solution, just to create this test that shows the issue. Can try to fix it in a few days, but any suggestion would be welcome ;-)

pquentin:
@fjros Since this was added in #1056, can you please close this PR? It turns out normal libcloud committers can't do this without a commit that asks the PR to be closed. Thanks!

PR #1049, opened 2017-04-28
Fixies for docker

johnnyWalnut:
### Description

Docker remote has change its call for start a container. As we can check here:
https://docs.docker.com/engine/api/version-history/#v124-api-changes
The previous implementation doesn't work and throws the following error:
```
/home/johnny/Documents/mylibcloud/libcloudfine/fork_libcloud/libcloud/libcloud/common/base.pyc in __init__(self, response, connection)
    159         if not self.success():
    160             raise exception_from_message(code=self.status,
--> 161                                          message=self.parse_error())
    162
    163         self.object = self.parse_body()

BaseHTTPError: {"message":"starting container with non-empty request body was deprecated since v1.10 and removed in v1.12"}
(where v refers to docker version)
```
So i changed the implementation to work like:
```POST /containers/(id or name)/start``` without data

Also, i add one more line which checks container status, because
this changed also, ```status=(created	restarting	running	paused	exited	dead)```
in API v 1.24 ```https://docs.docker.com/engine/api/v1.24/```

### Status

- done, ready for review

### DIFF
--- a/libcloud/container/drivers/docker.py
+++ b/libcloud/container/drivers/docker.py
@@ -386,10 +386,16 @@ class DockerContainerDriver(ContainerDriver):
 
         data = json.dumps(payload)
         if start:
-            result = self.connection.request(
-                '/v%s/containers/%s/start' %
-                (self.version, id_), data=data,
-                method='POST')
+            if float(self._get_api_version()) > 1.22:
+                result = self.connection.request(
+                    '/v%s/containers/%s/start' %
+                    (self.version, id_),
+                    method='POST')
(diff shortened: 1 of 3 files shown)


tonybaloney:
thanks @johnnyWalnut for the fix. This API is a constantly moving target. Can you please add a small test to demonstrate the functionality?

johnnyWalnut:
Yes, sure i am going to add a test in the next few days..

tonybaloney:
Thanks for your contribution 👍

PR #1051, opened 2017-05-02
Fix checking for "location is None" in several functions

SElsharkawy:
This suggested update fixes the condition that checks for parameter "location" being None. If the parameter is not None, the original code would still exit and raise ValueError exception.

### Description

This suggested update fixes the condition that checks for parameter "location" being None. If the parameter is not None, the original code would still exit and raise ValueError exception.

In the file libcloud/compute/drivers/azure_arm.py.
There are some functions as ex_create_public_ip function that it needs a location parameter, when this parameter is send, a error happen: "Location is required".
Looking at the code I found with the following:

```
       if location is None and self.default_location:
            location = self.default_location
        else:
            raise ValueError("location is required.")
```
This does not evaluate a possible parameter, only if the value is None.

I suggested the following change:

```
        if location is None:
            if self.default_location:
                location = self.default_location
            else:
                raise ValueError("location is required.")
```

The affected methods are:
```
   ex_create_network_security_group
   ex_delete_network_security_group
   ex_create_public_ip
```
The same code has been used in several other places.

### Status

OPEN

- [ ] Documentation

### DIFF
--- a/libcloud/compute/drivers/azure_arm.py
+++ b/libcloud/compute/drivers/azure_arm.py
@@ -910,10 +910,11 @@ class AzureNodeDriver(NodeDriver):
         :type location: :class:`.NodeLocation`
         """
 
-        if location is None and self.default_location:
-            location = self.default_location
-        else:
-            raise ValueError("location is required.")
+        if location is None:
+            if self.default_location:
+                location = self.default_location
+            else:
+                raise ValueError("location is required.")
 
         target = "/subscriptions/%s/resourceGroups/%s/" \
                  "providers/Microsoft.Network/networkSecurityGroups/%s" \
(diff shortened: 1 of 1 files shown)


SElsharkawy (commented):
When using SSH key authentication, the randomly generated HEX password for admin account is encoded as bytes which results in a JSON serialization error. Changed the encoding to UTF-8.

SElsharkawy:
When using SSH key authentication, the randomly generated HEX password for admin account is encoded as bytes which results in a JSON serialization error. Changed the encoding to UTF-8.

SElsharkawy:
Pull request is covered by later pull request #1098. Closing this one.

PR #1053, opened 2017-05-04
Restore RawResponse speed

pquentin:
RawResponse uses response.body from requests, which is pathologically
slow due to chardet (requests#2359). Since we're only interested in the
bytes, using response.content is faster. On a 3Mib file, this makes
a download_object() go from 40s to 4s.

### Description

RawResponse uses response.body from requests, which is pathologically
slow due to chardet (requests#2359). Since we're only interested in the
bytes, using response.content is faster. On a 3Mib file, this makes
a download_object() go from 40s to 4s.

### Status

- done, ready for review

- [ ] Documentation

### DIFF
--- a/libcloud/common/base.py
+++ b/libcloud/common/base.py
@@ -278,7 +278,7 @@ class RawResponse(Response):
         if not self._response:
             response = self.connection.connection.getresponse()
             self._response = HttpLibResponseProxy(response)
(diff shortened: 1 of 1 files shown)


tonybaloney:
Nice! :+1:

pquentin:
Thank you for the quick merge.

PR #1056, opened 2017-05-09
Fix list_sizes() for new regions in EC2 compute driver

fjros:
### Description

Calls to driver.list_sizes() were broken for new EC2 regions, namely 'eu-west-2', 'ca-central-1' and 'cn-north-1'. There were two different problems:
- The api name employed to scrape prices was incorrect in the case of 'eu-west-2' and 'ca-central-1'.
- Pricing for the Beijing region doesn't seem to be available at the moment.

This PR fixes these issues in the following way and closes #1045:
1. Fix api names in `scrape-ec2-prices.py`
2. Remove 'cn-north-1' from the VALID_EC2_REGIONS list

### Status

- done, ready for review

- [ ] Documentation


fjros:
@tonybaloney this build is failing for the pypy env due to something that doesn't seem to be related with the PR: `ERROR: InvocationError: '/home/travis/build/apache/libcloud/.tox/pypypy/bin/python setup.py test'`.

Any hint on how to solve this?

fjros:
What's the rationale of checking the coverage of the tests themselves?

pquentin:
@fjros What makes you say that the coverage of the tests is checked?

fjros:
@pquentin coveralls is failing in this PR because the coverage of `test_ec2.py` has decreased:
https://coveralls.io/builds/11760946

pquentin:
Ah, right.

A simple fix is to use `with pytest.raises()` instead of the `try ... except ... else` hack and remove the unused methods. It would be interesting to know why they are unused (a path that was covered no longer is?), so maybe tracking the coverage of tests is actually valuable.

fjros:
@pquentin thanks for your comments :)

I'm afraid I cannot see any value yet in checking the coverage of test files. I think such kind of meta-coverage is a bug, not a feature ;)

Whenever you want a failing test to output some descriptive message you're going to have an 'uncovered path' when such test passes. In the particular case of this PR, there's a test that lets the developer know that he/she forgot to include pricing info for the EC2 regions he/she added. Unless we get rid of such message, the 'coverage' will decrease.

pquentin:
Also, sorry, I don't use AWS, which is why I can't comment on the actual fixes here.

tonybaloney:
👍  don't worry about coveralls for now

PR #1061, opened 2017-05-24
Temporarily disable sphinx warnings as errors

fjros:
### Description

Doc generation is currently broken in the build pipeline, thus blocking other PRs from passing tests.

This PR temporarily removes `-W` flag from the call to `sphinx-build` so that warnings are not treated as errors. Ideally we should get rid of such warnings.

### Status

- work in progress

- [x] Documentation

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -44,7 +44,7 @@ deps = sphinx
        backports.ssl_match_hostname
 changedir = docs
 commands = python ../contrib/generate_provider_feature_matrix_table.py
-           sphinx-build -W -b html -d {envtmpdir}/doctrees . _build/html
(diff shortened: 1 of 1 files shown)


tonybaloney:
👍

Kami:
Thanks for your contribution and sorry for being late to the game.

I don't think removing this flag is a good idea - it will just mask real issues. Instead of removing this flag we should fix the actual root cause which is causing the warning.

I quickly checked the docs build locally and since we merged this change, quite some documentation "regressions" were introduced. We should fix all those issues and then re-enable the flag to make sure we won't commit and merge bad code into master again.

cc @tonybaloney

PR #1064, opened 2017-05-26
Fix bug with GCE instance_templates disk_type selection

evancarter-iex:
### Description

Using instance templates it always used pd-standard as disk type, not allowing you to choose between pd-ssd or pd-standard. It always chose pd-standard.

### Status
done, ready for review


tonybaloney:
thanks for the fix @evancarter-iex  👍

PR #1074, opened 2017-06-20
Fix #LIBCLOUD-925

micafer:
### Description
Fixes an error in _get_error function in common/google.py file that assumes that the body['error'] value is a dict, but in some cases it contains an string (unicode).
So it tries to do a err.get and it failts.

### Status
done, ready for review

- [ ] Documentation


tonybaloney:
👍  thanks for the contribution

PR #1079, opened 2017-06-22
Add loadBalancingScheme parameter for ex_create_forwarding_rule method in GCE driver.

sT331h0rs3:
loadBalancingScheme parameter exists in ForwardingRules resource: https://cloud.google.com/compute/docs/reference/latest/forwardingRules
However, we do not have it in ex_create_forwarding_rule method in GCE driver.

### DIFF
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -3126,7 +3126,8 @@ class GCENodeDriver(NodeDriver):
     def ex_create_forwarding_rule(self, name, target=None, region=None,
                                   protocol='tcp', port_range=None,
                                   address=None, description=None,
-                                  global_rule=False, targetpool=None):
+                                  global_rule=False, targetpool=None,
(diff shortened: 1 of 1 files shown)


sT331h0rs3:
I could not understand, why @coveralls complains about [this line](https://coveralls.io/builds/12098097/source?filename=libcloud%2Fcompute%2Fdrivers%2Fgce.py#L3207). Could anyone please help me with it?
However, the Travis CI build passed, and [coverage part](https://travis-ci.org/apache/libcloud/jobs/246077550) also passed.

tonybaloney:
@sT331h0rs3 ideally this would include a test (or another assertion) with the additional argument to the method. Then coveralls would be happy

tonybaloney:
in this case, I can see this is a harmless change, so other than slightly hurting our coverage % it's welcome 😄 

👍

sT331h0rs3:
Thanks a lot!

PR #1080, opened 2017-06-22
Fix issues found with lgtm

jhelie:
### Description

This fixes more issues flagged by lgtm.com code queries. Most were minor/style but a few were errors meaning the code did not have the intended effected (e.g. ```if x == a or b``` instead of ```if (x == a) or (x == b)```

The standard lgtm deep code analysis highlights dependencies, vulnerabilities ([3 found](https://lgtm.com/projects/g/apache/libcloud/vulnerabilities)) and currently flag [155 alerts](https://lgtm.com/projects/g/apache/libcloud/alerts/) - but you can investigate further by writing [your own custom queries](https://lgtm.com/docs/ql/about-ql) to explore your code base.

I personally find lgtm alerts most useful when using [PR integration](https://lgtm.com/docs/lgtm/config/pr-integration) as it allows to check and deal with potential issues before they find their way into master - for instance I haven't looked into it yet but it seems that [2a450e4](https://lgtm.com/projects/g/apache/libcloud/rev/2a450e4072c0468440877ec3decf3b21eaf369fa) may have introduced a few potential issues).

### Status

- done, ready for review

- [x ] [Code linting](http://libcloud.readthedocs.org/en/latest/development.html#code-style-guide) (required, can be done after the PR checks)
- [ ] Documentation


jhelie:
travis seems stuck, closing and re-opening to re-trigger

libcloud/loadbalancer/drivers/elb.py
```diff
@@ -308,8 +308,7 @@ def ex_create_balancer_listeners(self, name, listeners=None):
             if protocol == 'HTTPS' or protocol == 'SSL':
                 params['Listeners.member.%d.   \
                         SSLCertificateId' % i] = listener[3]
-        else:
-            return False
+        return False
```
pquentin: This one is wrong, I think. You don't want to always return False, but to do so only if `listeners` is empty.

pquentin (changes requested):
All changes are great except the last one, I think.

libcloud/loadbalancer/drivers/elb.py
jhelie: The suggested change preserves the current code behaviour - since there is no `break` statement in the for loop the `else` block will always be executed - see the [help of the corresponding lgtm code query](https://lgtm.com/rules/5980098/) and [here](http://python-notes.curiousefficiency.org/en/latest/python_concepts/break_else.html) for more details.

personally I do find `for` `else` always a bit confusing as their behaviour is a language dependent convention rather than a logical one - and I'm not very familiar with javascript but I think it has the behaviour you're referring to (else triggered if loop iterable empty)

to return `False` if `listeners` is empty I'd suggest something along the lines of:
```
index = -1
for index, listener in enumerate(listeners):
...
if index == -1:
    return False
```

happy to push that update if that makes sense!

jhelie (commented):
@pquentin thanks for having a look at it, I've replied to your review

libcloud/loadbalancer/drivers/elb.py
pquentin: `for` `else` is definitely confusing. The method is currently broken. The corresponding test is the part of the three test that start with `def text_...` instead of `def test_...`. I think we should ignore this in the PR and fix it in another PR.

jhelie: do you want me to revert that change or are you happy to leave it in since it doesn't change the current behaviour?

jhelie: actually I just noticed 2 bugs I had forgotten to fix so will push these and take that opportunity to revert the `for` `else` change here to avoid confusion

pquentin (approved):

pquentin:
Awesome! I think this is ready to be merged. I'm only a contributor (just like you), so I can't do this myself. @tonybaloney?

tonybaloney:
Merging. Thanks for your contribution @jhelie

jhelie:
You're welcome - I'm obviously biased as one of the devs but if you found these useful you might want to give a go at [PR integration](https://lgtm.com/docs/lgtm/config/pr-integration) :)

PR #1084, opened 2017-07-05
[WIP] Reduce boilerplate in tests using pytest

pquentin:
### Description

At first it was about using `pytest.raises`, but now I'm actually trying to convert the whole test suite to pytest to ensure I can use any pytest features (monkeypatch, fixtures, etc.) to improve test coverage and readability.

### Status

- work in progress


tonybaloney:
that's a lot cleaner. I was also moving some tests to use assert instead of `self.assertTrue` etc.

pquentin:
Yeah. The assertTrue -> assert changes can be done using unittest2pytest. But I need to remove the unittest compatibility first.

pquentin:
Closing for now as it will take some time to be ready.

PR #1085, opened 2017-07-12
Scrape prices for Australia region in GCE compute driver

fjros:
### Description

Update `contrib/update_google_prices.py` to scrape the price of GCE sizes in the Australia (Sydney) region.

### Status

- done, ready for review

- [ ] Documentation

### DIFF
--- a/contrib/update_google_prices.py
+++ b/contrib/update_google_prices.py
@@ -53,7 +53,9 @@ def main(argv):
         'eu': 'europe',  # alias for 'europe'
         'europe': 'europe',
         'apac': 'asia',  # alias for 'asia'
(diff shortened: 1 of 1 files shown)


pquentin (approved):
Thanks for this. Tested locally, works as expected. Maybe we should have a commit to update pricing.json itself too?

fjros:
Thanks @pquentin! I wondered the same... Ideally that should be part of either the build pipeline or the release process, but have no idea if that's the case in practice. If not, just let me know and I'll be happy to update `pricing.json` :-)

tonybaloney:
👍  thanks @fjros

PR #1086, opened 2017-07-20
Add missing test_requirement "requests_mock" to setup.py

felixonmars:
### Description

requests_mock is needed and defined in test_requirements.txt, but is missing from setup.py. Let's keep them consistent.

### Status
done, ready for review

- [x] Documentation

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -59,6 +59,7 @@ SUPPORTED_VERSIONS = ['2.6', '2.7', 'PyPy', '3.x']
 TEST_REQUIREMENTS = [
     'mock',
     'requests',
+    'requests_mock',
     'pytest',
     'pytest-runner'
 ]


pquentin:
👍 

@tonybaloney There are other inconsistencies, should they be fixed too?

tonybaloney:
well. yes there are inconsistencies, but they shouldn't necessarily be fixed. This a disagreement in the Python community, should `python setup.py install` install the dependencies for the package to execute _only_ or a full development environment.

I think they should be separate because you're probably installing from source and I don't want all of our users to have to install the dependencies we need for development (such as test runners and mockers).

felixonmars:
But we are looking into test_requires here, which only applies to the test suite. `requests_mock` in the PR is being imported explicitly in multiple files and the test suite will just break without it installed.

tonybaloney:
ah ok yes that should be updated. @pquentin which other inconsistencies?

pquentin:
@tonybaloney I only noticed that test_requires (TEST_REQUIREMENTS) is different from requirements-tests.txt.

tonybaloney:
👍  merging

PR #1088, opened 2017-08-07
Fix Linode DNS records to have 'priority' in extra

mete0r:
### Description

Linode DNS Record does not expose 'priority' value on its `extra` field, which is used in `DNSDriver._get_bind_record_line()`. It's required for MX/SRV records to be exported with `DNSDriver.export_to_bind_format()`.

Linode DNS resources already have 'PRIORITY' field (see https://www.linode.com/api/dns/domain.resource.list and https://github.com/apache/libcloud/blob/trunk/libcloud/test/dns/fixtures/linode/resource_list.json), we can just copy it into the `extra` field.

### Status

- done, ready for review

- [ ] Documentation

### DIFF
--- a/libcloud/dns/drivers/linode.py
+++ b/libcloud/dns/drivers/linode.py
@@ -265,7 +265,8 @@ class LinodeDNSDriver(DNSDriver):
         Build a Record object from the item dictionary.
         """
         extra = {'protocol': item['PROTOCOL'], 'ttl_sec': item['TTL_SEC'],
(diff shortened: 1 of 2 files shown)


tonybaloney:
Thanks for the contribution @mete0r 

Makes sense. The extra dict should contain priority for MX and SRV record, as per the docs (but who reads docs!) http://libcloud.readthedocs.io/en/latest/dns/index.html#terminology

👍

PR #1093, opened 2017-08-20
LIBCLOUD-875 fix route 53 DNS quote encapsulation

tonybaloney:
## fix route 53 DNS quote encapsulation

Amazon's API stipulates that SPF and TXT records have quotes http://docs.aws.amazon.com/Route53/latest/DeveloperGuide/ResourceRecordTypes.html#TXTFormat
The driver does not have any special casing for these records. I assume that existing users either:
- Get an error - see https://github.com/twisted/txacme/issues/120
- Add quotes before calling the methods, in which case this behaviour should be supported

section of our documentation.

### Status

- work in progress

- [ ] Documentation

### DIFF
--- a/libcloud/dns/drivers/route53.py
+++ b/libcloud/dns/drivers/route53.py
@@ -188,6 +188,8 @@ class Route53DNSDriver(DNSDriver):
         return response.status in [httplib.OK]
 
     def create_record(self, name, zone, type, data, extra=None):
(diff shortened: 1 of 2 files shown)


tonybaloney:
Finished. Validated against - 
- TXT with no quotes
- SPF with no quotes
- TXT that already had quotes
- SPF that already had quotes
- TXT that had no quotes, but a quote within it (so it gets escaped).

PR #1094, opened 2017-08-20
Fix build

tonybaloney:
Lint error and issue with building docs on the travis environment (I added support for Python 3 last month).


pquentin:
It's nice to see more tox targets use Python 3. 👍

tonybaloney:
merging locally..

PR #1095, opened 2017-08-28
Need to call `page` when using `GCEList` for pagination to work

sayap:
Otherwise, the `if self.gce_params:` condition will evaluate to false as the dict remains empty.

This is a follow-up fix to https://github.com/apache/libcloud/pull/939

### DIFF
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -7584,10 +7584,10 @@ class GCENodeDriver(NodeDriver):
                   if no matching image is found.
         :rtype:   :class:`GCENodeImage` or ``None``
         """
-        project_images_pages = self.ex_list(
+        project_images_list = self.ex_list(
             self.list_images, ex_project=project, ex_include_deprecated=True)
         partial_match = []
-        for page in project_images_pages:
+        for page in project_images_list.page():
             for image in page:
                 if image.name == partial_name:
                     return image


tonybaloney:
Thanks for the contribution @sayap please can you resolve the linting error https://travis-ci.org/apache/libcloud/jobs/269134539

sayap:
@tonybaloney I have got rid of the linting error with slightly better variable naming :)

pquentin:
Note that `GCEList` defines an `__iter__` method, so it would be simpler to write `for image in project_images_list:` and remove the `for image in page:`.

But your fix works too and is more explicit about the pagination! 👍

sayap:
@pquentin Actually, pagination only works after calling the `page` method.

We need this:

https://github.com/apache/libcloud/blob/8bb7e0c4161b3d3478ee2c739a126b3da7f68f8b/libcloud/compute/drivers/gce.py#L305

so that this evaluates to true and we get pagination:

https://github.com/apache/libcloud/blob/8bb7e0c4161b3d3478ee2c739a126b3da7f68f8b/libcloud/compute/drivers/gce.py#L125

I can't speak for the original author, but it seems like that's how it was designed to work in 1ed50204

pquentin:
Oh, you're right, sorry! Got confused by the `return self` hack. I now understand that `__iter__` return sublists.

pquentin:
@tonybaloney @Kami this pull request is ready to be merged.

pquentin:
@sayap Thank you for your contribution! It has just been merged in trunk.

PR #1096, opened 2017-08-29
test: check that secrets.py is up-to-date

pquentin:
### Description

This reintroduces a check that was removed when switching to pytest.

### Status

- done, ready for review


tonybaloney:
:+1: makes sense. I can't see any side effects in the test run to validate it but according the docs its ok.

pquentin:
Thank you!

PR #1098, opened 2017-08-29
LIBCLOUD-926 - Propagated changes to azure_arm.py from pull 1051

SElsharkawy:
### Description

This suggested update fixes the condition that checks for parameter "location" being None. If the parameter is not None, the original code would still exit and raise ValueError exception.

In the file libcloud/compute/drivers/azure_arm.py.
There are some functions as ex_create_public_ip function that it needs a location parameter, when this parameter is send, a error happen: "Location is required".
Looking at the code I found with the following:

       if location is None and self.default_location:
            location = self.default_location
        else:
            raise ValueError("location is required.")
This does not evaluate a possible parameter, only if the value is None.

I suggested the following change:

        if location is None:
            if self.default_location:
                location = self.default_location
            else:
                raise ValueError("location is required.")
The affected methods are:

   ex_create_network_security_group
   ex_delete_network_security_group
   ex_create_public_ip
The same code has been used in several other places.

### Status

done, ready for review

- [ ] Documentation

### DIFF
--- a/libcloud/compute/drivers/azure_arm.py
+++ b/libcloud/compute/drivers/azure_arm.py
@@ -628,7 +628,7 @@ class AzureNodeDriver(NodeDriver):
 
         if isinstance(auth, NodeAuthSSHKey):
             data["properties"]["osProfile"]["adminPassword"] = \
-                binascii.hexlify(os.urandom(20))
+                binascii.hexlify(os.urandom(20)).decode("utf-8")
             data["properties"]["osProfile"]["linuxConfiguration"] = {
                 "disablePasswordAuthentication": "true",
                 "ssh": {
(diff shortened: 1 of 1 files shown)


pquentin:
There's an unrelated `.decode("utf-8")` change.

The location fix looks good. Note that you can also write:

```
location = location or self.default_location
if location is None:
    raise ValueError("location is required.")
```

pquentin:
I almost forgot: Thank you for your contribution! It's great that you took the time to report this to libcloud. Even better: you included a fix! ✨  I'm not a maintainer, just leaving notes to get this merged faster. It's terse and to the point, but does not mean that the contribution is not good or not appreciated. :)

SElsharkawy:
Hi pquentin,

The " unrelated .decode("utf-8") change" is part of another pull request that fixes a problem in

SElsharkawy:
Hi pquentin,

The " unrelated .decode("utf-8") change" is part of another pull request that fixes another problem that was in the original pull request (1051). 

The error was: When using SSH key authentication, the randomly generated HEX password for admin account is encoded as bytes which results in a JSON serialization error.

The fix: Changed the encoding to UTF-8.

SElsharkawy:
Closed by mistake.

pquentin:
Thank you @SElsharkawy, I integrated your changes into trunk.

Next time, please:

 1. Open one pull request per change, or at least change the title of your pull request to explain that you added changes
 1. Don't open a new pull request, but change the existing one, rebasing your commits if necessary

But again, thank you for your contribution! 👍

PR #1104, opened 2017-09-06
[LIBCLOUD-943] Add the option to Azure ARM ex_create_network_interfac…

aki-k:
### Description

https://issues.apache.org/jira/browse/LIBCLOUD-943

Azure ARM driver's ex_create_network_interface is missing the option to create
a static IP address. "privateIPAllocationMethod" is set to "Dynamic" in the "data" 
dictionary. This change adds the option to create the network interface with a 
static IP address.

The value for the static IP address is defined with private_ip_address="n.n.n.n" in the 
ex_create_network_interface call.

### Status

- done, ready for review

- [x] Documentation

### DIFF
--- a/libcloud/compute/drivers/azure_arm.py
+++ b/libcloud/compute/drivers/azure_arm.py
@@ -1731,9 +1731,9 @@ class AzureNodeDriver(NodeDriver):
         if (private_ip_allocation_method == "Static" and
             private_ip_address is not None):
             data["properties"]["ipConfigurations"][0]\
-                ["properties"]["privateIPAllocationMethod"] = "Static"
+            ["properties"]["privateIPAllocationMethod"] = "Static"
             data["properties"]["ipConfigurations"][0]\
-                ["properties"]["privateIPAddress"] = private_ip_address
+            ["properties"]["privateIPAddress"] = private_ip_address
 
         r = self.connection.request(target,
                                     params={"api-version": "2015-06-15"},
--- a/libcloud/compute/drivers/azure_arm.py
+++ b/libcloud/compute/drivers/azure_arm.py
@@ -1730,10 +1730,9 @@ class AzureNodeDriver(NodeDriver):
 
         if (private_ip_allocation_method == "Static" and
             private_ip_address is not None):
-            data["properties"]["ipConfigurations"][0]\
-["properties"]["privateIPAllocationMethod"] = "Static"
-            data["properties"]["ipConfigurations"][0]\
-["properties"]["privateIPAddress"] = private_ip_address
+            ip_config = data["properties"]["ipConfigurations"][0]
+            ip_config["properties"]["privateIPAllocationMethod"] = "Static"
+            ip_config["properties"]["privateIPAddress"] = private_ip_address
 
         r = self.connection.request(target,
                                     params={"api-version": "2015-06-15"},
--- a/libcloud/compute/drivers/azure_arm.py
+++ b/libcloud/compute/drivers/azure_arm.py
@@ -1729,7 +1729,7 @@ class AzureNodeDriver(NodeDriver):
             }
 
         if (private_ip_allocation_method == "Static" and
-            private_ip_address is not None):
+                private_ip_address is not None):
             ip_config = data["properties"]["ipConfigurations"][0]
             ip_config["properties"]["privateIPAllocationMethod"] = "Static"
             ip_config["properties"]["privateIPAddress"] = private_ip_address
--- a/libcloud/compute/drivers/azure_arm.py
+++ b/libcloud/compute/drivers/azure_arm.py
@@ -1731,9 +1731,9 @@ class AzureNodeDriver(NodeDriver):
         if (private_ip_allocation_method == "Static" and
             private_ip_address is not None):
             data["properties"]["ipConfigurations"][0]\
-            ["properties"]["privateIPAllocationMethod"] = "Static"
+["properties"]["privateIPAllocationMethod"] = "Static"
             data["properties"]["ipConfigurations"][0]\
-            ["properties"]["privateIPAddress"] = private_ip_address
+["properties"]["privateIPAddress"] = private_ip_address
 
         r = self.connection.request(target,
                                     params={"api-version": "2015-06-15"},
--- a/libcloud/compute/drivers/azure_arm.py
+++ b/libcloud/compute/drivers/azure_arm.py
@@ -1659,7 +1659,9 @@ class AzureNodeDriver(NodeDriver):
         return self._to_ip_address(r.object)
 
     def ex_create_network_interface(self, name, subnet, resource_group,
-                                    location=None, public_ip=None):
+                                    location=None, public_ip=None,
+                                    private_ip_allocation_method=None,
+                                    private_ip_address=None):
         """
         Create a virtual network interface (NIC).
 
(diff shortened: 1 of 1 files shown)


aki-k:
> libcloud/compute/drivers/azure_arm.py:1732:13: E129 visually indented line with same indent as next logical line

I read in PEP8 that:

Acceptable options in this situation include, but are not limited to:

```
# No extra indentation.
if (this_is_one_thing and
    that_is_another_thing):
    do_something()
```

pquentin:
You're right: PEP8 accepts the E129 error, but the [pycodestyle linter does not by default](https://github.com/PyCQA/pycodestyle/issues/126), and it's probably a good thing. Thanks for updating your code!

The change make sense, and is documented: thanks for a great first-time contribution. I believe there are a few issues left.

 * Renaming the `private_ip_address` parameter to `ip_address` would be more consistent with most of the drivers. Not sure about prefixing `private_ip_allocation_method` with `_ex`.
 * Azure ARM appears to accept IPv6 addresses, so documenting the address to be IPv4 is too restricting.

Admittedly, those are nitpicks. But a more important issue: we need a test before we can merge this feature! `ex_create_network_interface` is not tested directly, but through  `test_create_node`. There's a fixture at `fixtures/azure_arm/_subscriptions_99999999_9999_9999_9999_999999999999_resourceGroups_000000_providers_Microsoft_Network_networkInterfaces_test_node_1_nic.json` that you should adapt to test using a static private IP.

Thanks!

pquentin:
@aki-k Any progress on this?

aki-k:
I'll ask my friend to help with the test case creation. Is there some guidance available online from libcloud?

pquentin:
@aki-k You can look at http://libcloud.readthedocs.io/en/latest/testing.html which explains how to run tests. You can look at how the other tests work.

And as I was saying: `ex_create_network_interface` is not tested directly, but through  `test_create_node`. There's a fixture at `fixtures/azure_arm/_subscriptions_99999999_9999_9999_9999_999999999999_resourceGroups_000000_providers_Microsoft_Network_networkInterfaces_test_node_1_nic.json` (maybe the name changed) that you should adapt to test using a static private IP.

pquentin:
@aki-k Anything I can help with?

aki-k:
I'm sorry I haven't written the test case. I'm not so experienced with python.
If you could help me with these questions:

- How do I see the Azure REST API reply with libcloud?
- Should I create a new fixture in fixtures/azure_arm/ for this?
- In the existing fixture, I see that there's: "privateIPAddress": "10.0.0.1".
  It's after "provisioningState": "Succeeded" so I guess it's a reply from Azure REST API.
  And in test_azure_arm.py, there's "self.assertEqual(node.private_ips, ['10.0.0.1'])" so
  I don't understand if this functionality already exists somewhere.

pquentin:
- Turns out I never had to write fixtures myself, so that's a good question! I guess a good start is printing the response as JSON in libcloud.http.LibcloudConnection.prepared_request
- I think it would be nice to create a new fixture, yes, so that the new test is independent from the other ones
- The `"self.assertEqual(node.private_ips, ['10.0.0.1'])"` is only here to check that we did read the response from the fixture, I think.

Does that help?

aki-k:
@pquentin When I pprint the object that I get from Azure when creating a network interface with a static public ip address and a static private ip address, I get the following:

https://gist.github.com/aki-k/45c059b2c87a37c9a6549a926e87f8b3

Is that the format that I will use in the fixture? I don't see that "extra" dictionary in the other fixtures.

aki-k:
I was able to get the proper Azure response by printing r.body in ex_create_network_interface() after r = self.connection.request().

aki-k:
Not sure what the problem is but I'm getting "libcloud.common.exceptions.RateLimitReachedError: 429 Rate limit exceeded" for driver.create_node(). It runs for 3.5 minutes and then outputs:

https://gist.github.com/aki-k/3fdb37a794cd25fd0b4a613db652622d

Edit: In Azure portal Activity Log I can see really many ListKeys entries for the time period.

pquentin:
I'm not familiar with Azure so I won't be able to help here.

@ldipenti @tetron Any ideas? Thanks!

VojtechBartos:
@aki-k looks like your changes are on old libcloud version. There is actually infinite loop in `create_node` which is trying to delete old VHD OS disks and create new one. It has been fixed (or at least reduced) in latest versions. If you get trunk updated and rebase you should be okay.

This is the infinite loop https://github.com/aki-k/libcloud/blob/LIBCLOUD-943-ex_create_network_interface-private_ip_allocation_method-static/libcloud/compute/drivers/azure_arm.py#L537

aki-k:
@VojtechBartos 

I updated to libcloud 2.3.0. I can't figure out how the VHD handling works in libcloud. I have a self created VHD which I can get with driver.get_image() providing the whole https URL. 

I don't understand the logic in _get_instance_vhd(). Why does it try to remove the VHD in the while loop? 

Is the 'image' param in create_node() the source or the destination VHD? If it's the source VHD, how do I specify the target VHD? 

Currently when trying to create_node(), I'm running into the _get_instance_vhd() looping for 10 times and get the error message "libcloud.common.types.LibcloudError: <LibcloudError in None "Unable to find a name for a VHD to use for instance in 10 tries, errors were:\n  - <LibcloudError in 
<libcloud.storage.drivers.azure_blobs.AzureBlobsStorageDriver object at 0x7f91539ed198> 'Unexpected status code: 400'>"

VojtechBartos:
@aki-k 

I am not author of the code, i know what you know, so stuff what we can read from the code and deduction ;)


> I don't understand the logic in _get_instance_vhd(). Why does it try to remove the VHD in the while loop?

As far as i understand. When you spin up new VM, OS drive is saved as VHD file, but when you destroy it, it won't delete it, so you need to make sure that file doesn't exists in case you wanna do create machine A -> destroy machine A -> create machine A, cause you wouldn't get a fresh drive. The loop i don't understand as well.

> Is the 'image' param in create_node() the source or the destination VHD? If it's the source VHD, how do I specify the target VHD?

`image` param is the source VHD to from you wanna boot. You wont specify a target VHD, it get's generated for you from the storage account and blob container parameters.

To the error actually you getting i would expect you got wrong storage account or blob container, do you have them created ? `ex_storage_account` and `ex_blob_container` which is default `vhds`

aki-k:
@VojtechBartos 

I created a new storage account in the resource group I'm creating this new VM. Microsoft support confirmed to my question that I can also use a storage account in another resource group, but I'm not sure how to setup libcloud to do it.

If I skip the while loop (and just define instance_vhd and return it) in 

https://github.com/apache/libcloud/blob/trunk/libcloud/compute/drivers/azure_arm.py#L2109

that calls _ex_delete_old_vhd(), I'm able to run create_node() successfully. My self created VHD resides in the same resource group, in the same storage account that I'm trying to create the new VM. This was documented in 

http://libcloud.readthedocs.io/en/latest/compute/drivers/azure_arm.html

"Note: when booting from a user image (AzureVhdImage) the source image and the node image must use the same storage account."

aki-k:
Running the script with environment variables LIBCLOUD_DEBUG and LIBCLOUD_DEBUG_PRETTY_PRINT_RESPONSE=1 defined I get this:

    # -------- begin 139741872798688 request ----------
    curl -i --head -H 'x-ms-version: 2012-02-12' -H 'x-ms-date: Wed, 18 Apr 2018 18:31:54 GMT' -H 'X-LC-Request-ID: 139741872798688' -H 'Authorization: SharedKey STORAGE_ACCOUNT:HASH' -H 'Accept-Encoding: gzip,deflate' -H 'User-Agent: libcloud/2.3.0 (Microsoft Azure (blobs)) ' --compress 'https://STORAGE_ACCOUNT.blob.core.windows.net/vhds?restype=container'
    # -------- begin 139741872798688:139741872738144 response ----------
    11 400 The value for one of the HTTP headers is not in the correct format.
    Vary: Origin
    Server: Windows-Azure-Blob/1.0 Microsoft-HTTPAPI/2.0
    X-Ms-Request-Id: GUID
    X-Ms-Version: 2012-02-12
    Date: Wed, 18 Apr 2018 18:31:55 GMT

Edit: Seems to be similar bug than 

https://issues.apache.org/jira/projects/LIBCLOUD/issues/LIBCLOUD-851

aki-k:
Changing API_VERSION from '2012-02-12' to '2014-02-14' in libcloud/common/azure.py fixed the error message "400 The value for one of the HTTP headers is not in the correct format." I received from create_node() which calls _get_instance_vhd() and _ex_delete_old_vhd().

Edit: should I change the API_VERSION only for blob storage requests?

Kami:

PR #1105, opened 2017-09-06
Re-enable strict checks (treat warnings as errors) when building sphinx docs on Travis CI

Kami:
This pull request re-enables strict checks when building sphinx docs on Travis CI.

This way we can automatically catch invalid / broken docstrings and other documentation related issues (invalid syntax) early on.

#1061 disabled this and since then a bunch of broken docstring syntax got in the repo so it will take a while to fix it all so we can merge this.

On a related note - I know that the PR description said "temporarily", but such things are rarely temporary so we need to be careful when merging such things. Strict checking was enabled for a reason and disabling such automatic checks is rarely a good idea - root cause should be found and fixed instead.

/cc @tonybaloney

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -21,6 +21,8 @@ from sphinx.environment import BuildEnvironment
 from sphinx.ext.autodoc import AutoDirective
 from sphinx.ext.autodoc import AutodocReporter
 
+from sphinx.domains.python import PythonDomain
+
 BASE_DIR = os.path.dirname(os.path.abspath(__file__))
 BASE_DIR = os.path.abspath(BASE_DIR)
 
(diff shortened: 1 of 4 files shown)


Kami:
1&1 driver had a lot of issues - 6fcd3487ece4fb0b2cbee535b4a20599ab2a591a.

We need to be more careful about the docstring syntax during code reviews and also look if we can build more automation around detecting invalid / broken docstring syntax. In this case, docstrings we syntactically correct, but not semantically so sphinx didn't detect more issues.

On a related note - the driver is also missing documentation / explanation for a lot of extension (ex_) arguments which should be mandatory.

Extension arguments / methods without documentation are basically useless - driver user still needs to go check the provider API docs, etc. so kinda defeats the purposes of ease of use.

Kami:
@tonybaloney can you please look into updating coveralls setting so it only posts a comment once per PR instead of posting once after every push (aka update the existing comment instead of posting a new one).

If that's not possible with coveralls, we should consider switching to codecov.io which supports that functionality (we use it in StackStorm/st2). Right now the comments are too spammy - they don't add much value, just make it harder to see actual comments which might provide some actual value to the discussion.

Kami:
The docs build is passed now so I will go ahead and merge this into trunk.

PR #1107, opened 2017-09-07
[LIBCLOUD-944] Add support for private IP addresses in GCE instance creation

gmacf:
### Description

GCE driver doesn't currently allow private IPs to be specified when creating instances. Updated gce.py driver to allow the optional parameter internal_ip which can be used when creating GCE instances.  

### Status

- done, ready for review

- [x] Documentation


pquentin:
Thank you @vongazman for your contribution. Great first-time GitHub contribution! The changes make perfect sense! 👍

However, this pull requests lacks tests. I realize that the existing `external_ip` parameter is not tested, but I believe you should add tests at least for `internal_ip`. You can look at the DimensionData tests to see how you can update the fixtures and test code to do that.

libcloud/test/compute/test_gce.py
```diff
@@ -2900,6 +2911,16 @@ def _regions_us_central1_addresses_lcaddress(self, method, url, body,
                 'regions_us-central1_addresses_lcaddress.json')
         return (httplib.OK, body, self.json_hdr, httplib.responses[httplib.OK])
 
+    def _regions_us_central1_addresses_testaddress(self, method, url, body,
+                                                   headers):
+        if method == 'DELETE':
+            body = self.fixtures.load(
+                'regions_us-central1_addresses_testaddress_delete.json')
```
pquentin: It looks like you had started to test the DELETE path but ended up not to? Either add the file or remove the reference to it.

pquentin (changes requested):
Great, thank you for adding tests! There's one issue left with DELETE. Thanks.

libcloud/test/compute/test_gce.py
gmacf: Thanks, I removed the reference to it.

pquentin:
Thanks @vongazman! ✨ I've included this in trunk.

(Reworked the commits to remove the intermediate ones that would have made bisecting difficult, and signed them off as part of the standard procedure.)

PR #1109, opened 2017-09-10
[WIP] Replace coveralls with codecov.io for code coverage reports

Kami:
This pull request proposes to replace coveralls coverage reports with codecov.io coverage reports.

My main reasons for the proposed change of service are:

* codecov.io has a much nicer and better UI.
* Right now we have coverage comments enabled which means coveralls posts a comment on the pull request on every push. Imo, this adds a lot of noise, but very little value. It discourages people to push often (which is a good practice) and adds noise to the pull request which makes actual valuable review feedback harder to see / find.

Codecov.io can be configured to add only a single comment which is then updated when code and coverage changes. Imo, that's much better than comment spam on every push.


Kami:
@tonybaloney what do you think?

tonybaloney:
+1 to the change. You might need to ask the infrastructure team to make the switch. Agree that the constant update on coverage is annoying

pquentin:
👍

Kami:
@tonybaloney Nice thing about codecov also is that you don't need to configure hooks manually on Github.

It works "out of the box" for public repos which use Travis CI or a similar service.

PR #1111, opened 2017-09-19
Make sure we normalize header types for test mock connection class

Kami:
This pull request fixes the issue reported in https://issues.apache.org/jira/browse/LIBCLOUD-945.

We normalized header types inside the actual Connection class, but not inside the Connection class used by the tests. That's why some of the tests failed with newer version of `requests` which doesn't automatically cast header values to strings anymore.

### DIFF
--- a/libcloud/storage/drivers/azure_blobs.py
+++ b/libcloud/storage/drivers/azure_blobs.py
@@ -634,7 +634,7 @@ class AzureBlobsStorageDriver(StorageDriver):
             chunk_hash = base64.b64encode(b(chunk_hash.digest()))
 
             headers['Content-MD5'] = chunk_hash.decode('utf-8')
-            headers['Content-Length'] = content_length
+            headers['Content-Length'] = str(content_length)
 
             if blob_type == 'BlockBlob':
                 # Block id can be any unique string that is base64 encoded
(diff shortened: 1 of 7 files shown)


libcloud/storage/drivers/azure_blobs.py
```diff
@@ -634,7 +634,7 @@ def _upload_in_chunks(self, response, data, iterator, object_path,
             chunk_hash = base64.b64encode(b(chunk_hash.digest()))
```
Kami, line 634: Changes in those files are not needed anymore because we now cast header values inside the base connection class, but they don't hurt either.

libcloud/storage/drivers/azure_blobs.py
```diff
@@ -861,10 +861,10 @@ def _prepare_upload_headers(self, object_name, object_size,
         self._update_metadata(headers, meta_data)
 
         if object_size is not None:
-            headers['Content-Length'] = object_size
+            headers['Content-Length'] = str(object_size)
 
         if blob_type == 'PageBlob':
-            headers['Content-Length'] = 0
+            headers['Content-Length'] = str('0')
```
pquentin, line 867: Just wondering, isn't `'0'` enough?

PR #1112, opened 2017-09-19
Make sure we don't have any files with names longer than 143 characters

Kami:
It looks like 2.2.0 broke installation on some systems, because we now ship files with names longer than 143 characters.

Most modern operating systems and file systems support file names with lengths up to 255 characters, but it looks like using an encrypted file system (layering ecryptfs on top of ext3/4) drops this limit to 143 characters (see https://stackoverflow.com/questions/34503540/why-does-python-give-oserror-errno-36-file-name-too-long-for-filename-short).

This pull request fixes that and add a CI check which will catch issues like that in the future.

Resolves https://issues.apache.org/jira/browse/LIBCLOUD-946.


Kami:
I think we should do v2.2.1 after the fix is confirmed.

@tonybaloney wdyt?

Kami:
Fix confirmed - https://issues.apache.org/jira/browse/LIBCLOUD-946?focusedCommentId=16172071&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16172071.

It looks like Travis is having some issues though. Build has been stuck for a long time now.

PR #1113, opened 2017-09-20
certifi where() return a str and not a list

llazzaro:
### Description
When using certifi and other libs assumes that you will get a list on CA_CERTS_PATH it will raise error when doing appends.

certifi.where() return a string instead of a list.

### Status

- done, ready for review

- [ X] [Code linting](http://libcloud.readthedocs.org/en/latest/development.html#code-style-guide) (required, can be done after the PR checks)

### DIFF
--- a/libcloud/security.py
+++ b/libcloud/security.py
@@ -55,7 +55,7 @@ else:
 
 if has_certifi and USE_CERTIFI:
     certifi_ca_bundle_path = certifi.where()
-    CA_CERTS_PATH = certifi_ca_bundle_path
+    CA_CERTS_PATH = [certifi_ca_bundle_path]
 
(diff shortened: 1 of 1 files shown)


pquentin:
Thank you for your first contribution! Greatly appreciated.

As is documented in http://libcloud.readthedocs.io/en/latest/other/changes_in_2_0.html#providing-a-list-of-ca-trusts-is-no-longer-supported, providing a list is no longer supported. While it could be supported, it allowing to append to `libcloud.security.CA_CERTS_PATH` would be misleading. I agree that there were a few places where this was still confusing in the code, so I opened #1114 to this effect. Can you look at it and tell me if it makes sense? Thanks!

llazzaro:
@pquentin great!

PR #1114, opened 2017-09-20
Remove final mentions of CA_CERTS_PATHS as list

pquentin:
## Remove final mentions of CA_CERTS_PATHS as list

Since libcloud 2.0, CA_CERTS_PATHS is no longer a list, but there were a
few mentions left in the code, which was confusing.

Thanks to @llazzaro for the inital report in #1113.

done, ready for review

### Checklist

- [x] Documentation

### DIFF
--- a/libcloud/security.py
+++ b/libcloud/security.py
@@ -20,7 +20,7 @@ Usage:
     libcloud.security.VERIFY_SSL_CERT = True
 
     # Optional.
-    libcloud.security.CA_CERTS_PATH.append('/path/to/cacert.txt')
+    libcloud.security.CA_CERTS_PATH = '/path/to/certfile'
 """
 
 import os
@@ -71,7 +71,7 @@ if environment_cert_file is not None:
 
     # If a provided file exists we ignore other common paths because we
     # don't want to fall-back to a potentially less restrictive bundle
-    CA_CERTS_PATH = [environment_cert_file]
+    CA_CERTS_PATH = environment_cert_file
 
 CA_CERTS_UNAVAILABLE_ERROR_MSG = (
     'No CA Certificates were found in CA_CERTS_PATH. For information on '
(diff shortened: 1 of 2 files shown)


libcloud/security.py
```diff
@@ -71,7 +71,7 @@
 
     # If a provided file exists we ignore other common paths because we
     # don't want to fall-back to a potentially less restrictive bundle
-    CA_CERTS_PATH = [environment_cert_file]
```
Kami, line 74: Hm, so this didn't work correctly since 2.0.0?

Kami:
I grepped code for `CA_CERTS_PATH` and it looks like there are some other places which also need to be updated - notably the docs.

On a related note - we should also add some code which checks if this variable is not a string (e.g. a list) and throws a user-friendly exception which tells user it needs to be a string and points them to the docs.

libcloud/security.py
pquentin: This was working because [`http.LibcloudBaseConnection.setup_ca_cert()`](https://github.com/apache/libcloud/blob/trunk/libcloud/http.py#L147) handles lists by taking the first element and issuing a warning. The only thing that was not working was appending to `libcloud.security.CA_CERTS_PATH`. But defining it as a list is still working.

pquentin:
@Kami I believe the only places that mention `CA_CERTS_PATH` as a list are either the fallback code in libcloud.http or `docs/other/ssl-certificate-validation.rst` which is specifically about libcloud <v2.0. Did I miss something else?

About the related note: do you think we should change the existing warning in libcloud.http to an exception now that 2.0 is a few months old?

Kami:
LGTM :+1:

Kami:
Not related to your change, but I would prefer this message to be a bit more explicit and user-friendly - https://github.com/apache/libcloud/blob/trunk/libcloud/http.py#L153.

And perhaps also point user to the corresponding documentation section.

And eventually, we should change this warning to an actual exception.

PR #1120, opened 2017-10-04
On Azure ARM, destroy_node() do VHD cleanup even when NIC cleanup fails

ldipenti:
### Description

When calling destroy_node() on Azure ARM, several cloud calls happen. After successfully destroying the VM, the method tries to remove all NICs related to that VM. If one of these attempts fail, the VHD's life is spared.
The updates on this PR make destroy_node() continue with its task after a NIC cleanup fail, trying to remove the remaining NICs and the assigned VHD.
Also added tests for destroy_node(), modified the test suite a little to be able to simulate different HTTP response sequences.

### Status

- done, ready for review

- [ ] Documentation

### DIFF
--- a/libcloud/compute/drivers/azure_arm.py
+++ b/libcloud/compute/drivers/azure_arm.py
@@ -744,7 +744,8 @@ class AzureNodeDriver(NodeDriver):
                         if h.code == 400 and inuse:
                             time.sleep(10)
                         else:
-                            return False
+                            # NIC cleanup failed, try cleaning up the VHD.
+                            break
 
         # Optionally clean up OS disk VHD.
         vhd = node.extra["properties"]["storageProfile"]["osDisk"].get("vhd")
--- a/libcloud/compute/drivers/azure_arm.py
+++ b/libcloud/compute/drivers/azure_arm.py
@@ -695,12 +695,11 @@ class AzureNodeDriver(NodeDriver):
         this node (default True).
         :type node: ``bool``
 
-        :return: True if the destroy was successful, False otherwise.
+        :return: True if the destroy was successful, raises exception otherwise.
         :rtype: ``bool``
         """
 
         do_node_polling = True
-        success = True
 
         # This returns a 202 (Accepted) which means that the delete happens
         # asynchronously.
@@ -718,7 +717,7 @@ class AzureNodeDriver(NodeDriver):
                 # No need to ask again, node already down.
                 do_node_polling = False
             else:
-                return False
+                raise
 
         # Need to poll until the node actually goes away.
         while do_node_polling:
@@ -731,7 +730,7 @@ class AzureNodeDriver(NodeDriver):
                 if h.code == 404:
                     break
                 else:
-                    return False
+                    raise
 
         # Optionally clean up the network
         # interfaces that were attached to this node.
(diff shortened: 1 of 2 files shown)


pquentin:
Thanks for this! I'm glad that you added tests too. ✨ 

Don't you think we should return False when NIC cleanup fails but VHD cleanup succeeds?

tetron:
@pquentin The abstract interface for a libcloud `Node` just says "True on success and False on failure" but it is not obvious what that should mean in this case.  Perhaps:

* `True` means all resources were either successfully deleted, or returned 404 (in particular, trying to destroy a VM resource that doesn't exist would be considered a success, not a failure)
* `False` means one or more resource deletion actions failed for some other 4xx or 5xx

This would allow a caller to safely invoke destroy_node() in a retry loop until it returns True (assuming the failures are cloud weather that will eventually resolve themselves).  This would allow us to remove the retry loops from destroy_node as well (and document that the caller is responsible for performing retries).

pquentin:
@tetron As I don't use Azure ARM, I would be happy to trust you on the True/False choice. What do you think is the best choice?

Regarding the retry loops, I would prefer not change the API, but maybe improving the way the loops work if we believe they could be improved.

pquentin:
Okay, reading your answer more carefully, you seem to agree that we should only return True when everything was removed with success. @ldipenti Do you think you could change this part of your PR? Thanks!

ldipenti:
@pquentin of course!

libcloud/compute/drivers/azure_arm.py
```diff
@@ -766,9 +776,10 @@ def destroy_node(self, node, ex_destroy_nic=True, ex_destroy_vhd=True):
                         # LibcloudError.  Wait a bit and try again.
                         time.sleep(10)
                     else:
-                        raise
+                        success = False
+                        break
```
pquentin: I don't think swallowing the exception is a good idea here. It's OK if this raises. Maybe another PR could ensure the number of retries is fixed and returns False when we got "LeaseIdMissing", say, 10 times, but it's out of scope here.

What do you think?

pquentin (changes requested):
Thanks, that's much better! We're nearly there.

libcloud/compute/drivers/azure_arm.py
ldipenti: @pquentin how about cases when in a previous call the VHD was successfully deleted but some NIC wasn't? Should we check for 404s and ignore them like we do with NICs and VMs?

pquentin: I had not noticed that we were swallowing an exception in the NIC and VM cases too. I don't really know. Do you think returning False without exposing the exception is the correct choice?

ldipenti: I suppose we could catch & re-raise specific exceptions, for example the ones related to throttling/rate limits, that would make re-trying something dangerous to do, and the rest log the message and return False, maybe @tetron has another idea.

tetron: I'm fine with re-raising if we get an exception from NIC or VHD delete.  The key change here is catch 404 and treat it as succces.

tetron: And actually, the meaning of "False" when returned by destroy_node() is ambiguous, so it is probably better to raise an exception anyway.

pquentin:
Thank you @ldipenti, merged in trunk! ✨ 

For the next PR, can you try splitting your commit messages in two? One short summary on the first line, followed by more details (http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html). Thanks!

PR #1121, opened 2017-10-05
Scaleway Rebased

danhunsaker:
### Description

Revisiting #777 by rebasing it against the latest `trunk`. From its description:

> This is a simple/pure compute driver for https://www.scaleway.com/

See also https://issues.apache.org/jira/browse/LIBCLOUD-956

### Status

done, ready for review

- [x] Documentation


pquentin:
Thanks @danhunsaker!

Did you plan to add tests? I think this was the main issue of #777.

danhunsaker:
After reviewing the original PR, that's absolutely part of my plan. Will have a contributor agreement in place by week's end. Phone doesn't want to cooperate at the moment.

pquentin:
I'm happy to see regular progress here. Thanks!

danhunsaker:
ICLA sent, tests in place, ready for review.

libcloud/compute/drivers/scaleway.py
```diff
@@ -0,0 +1,615 @@
...
+SCALEWAY_API_HOSTS = {
+    'default': 'api.scaleway.com',
+    'account': 'account.scaleway.com',
+    'par1': 'cp-par1.scaleway.com',
+    'ams1': 'cp-ams1.scaleway.com',
+}
+
+# The API doesn't currently expose all of the required values for libcloud,
+# so we simply list what's available right now, along with all of the various
+# attributes that are needed by libcloud.
```
pquentin: How did you retrieve these data? The ideal scenario would be to have a script in the `contrib` directory to scrape these prices. We would then run this script before every release.

libcloud/compute/drivers/scaleway.py
```diff
@@ -0,0 +1,615 @@
...
+
+def _to_lib_size(size):
+    return int(size / 1000 / 1000 / 1000)
+
+
+def _to_api_size(size):
+    return int(size * 1000 * 1000 * 1000)
+
+
+class ScalewayNodeDriver(NodeDriver):
```
pquentin, line 130: This is part of the public API, all methods need to be documented, see https://libcloud.readthedocs.io/en/latest/development.html#docstring-conventions and included in the documentation, see https://github.com/apache/libcloud/blob/trunk/libcloud/compute/drivers/gce.py#L1673 for example.

libcloud/compute/drivers/scaleway.py
```diff
@@ -0,0 +1,615 @@
...
+    """
+    Scaleway NodeDriver
+    """
+
+    type = Provider.SCALEWAY
+    connectionCls = ScalewayConnection
+    name = 'Scaleway'
+    website = 'https://www.scaleway.com/'
+
+    SNAPSHOT_STATE_MAP = {  # TODO map all states
```
pquentin: I see a few TODOs here. Is this something that you plan to do next? Or just a warning for the next person that something is missing here?

pquentin (changes requested):
Thank you, this driver looks great! It's very nice to see that the coverage is good now. The main thing that is missing is code documentation with docstrings. I hope we'll be able to merge this soon!

libcloud/compute/drivers/scaleway.py
danhunsaker: This was filled manually, unfortunately. It might be scrapable, except for disk sizes, which it doesn't report anywhere I've seen. I'm going to be contacting them about exposing the hard-coded info in their API instead, so we don't even need to have a scraper hack in place.

pquentin: Okay, thank you! This should not block this PR, however.

pquentin:
We received your ICLA, thanks!

libcloud/compute/drivers/scaleway.py
danhunsaker: Could have sworn I'd replied to this one. This particular TODO is retained from the original PR. I haven't been able to find any other states that it might actually report, so i'm not sure it's even something that _can_ be done, but I've opted to leave it just in case someone else knows better and can fill it in.

As to the others, I'm planning to handle that soon, though hopefully in a separate PR rather than this one, as it will require a fair amount of back-and-forth with Scaleway to work out the details. ~~Ideally, that PR would remove the hard-coded sizes entirely, replacing them with API calls to Scaleway instead.~~

libcloud/compute/drivers/scaleway.py
danhunsaker: I'll handle this some time this week. Sorry for the delay!

pquentin:
Scaleway said they would like to see this driver happen - https://issues.apache.org/jira/browse/LIBCLOUD-987.

@danhunsaker What is the status of this work?

pquentin:
Can you please fix the lint?

I think we'll need an ICLA from [Nándor István Krácser](https://github.com/bonifaido) too? cc @tonybaloney @Kami

danhunsaker:
Sorry, still in progress, here. Still needs docs, among other things.

libcloud/compute/drivers/scaleway.py
remyleone: @danhunsaker In case you need any support from Scaleway about our APIs just give me sign :)

costela:
hi @danhunsaker, are you still working on this? Don't get me wrong, I don't mean to pressure. It's just that, if you're not, than I might try to pick up where you left off. But I of course don't want to duplicate work if you're still on it :wink:
Thanks anyway for the great work so far! :)

danhunsaker:
@pquentin There we go. Docs added, lint test repaired, rebased on `trunk`... Should be good to go!

@costela Sorry about the long delays on this. Work's been pulling me in a very different direction, lately, so I had to carve out some free time to wrangle the last bits. Room for improvements, of course, but I think they'll be coming in later PRs, rather than updates to this one.

@sieben I'll probably reach out to you sometime soon to go over some of the improvements I have in mind. Not sure when I'll get to that point, though; work demands are pretty intense right now.

pquentin:
@danhunsaker Great!

@costela Could you please review that work?

costela:
@danhunsaker wow, that was fast! Thanks again for all the work! :)

@pquentin LGTM, for a first version. Though I'm not a contributor, so take my review with a pinch of salt! :)
My only reservation is the hard-coded `SCALEWAY_INSTANCE_TYPES`, but since this apparently cannot be solved in a more programmatic way, I'd personally be OK with it as an interim solution.

@sieben it would be really nice if scaleway would add an endpoint for node types, so we can get rid of the above mentioned hard-coded `SCALEWAY_INSTANCE_TYPES` in a subsequent PR. Any chance of that happening in the not-so-distant future?

remyleone:
@costela it's already available you can use the following URL to get information about our offers:

`https://cp-ams1.scaleway.com/products/servers` to get the different offers for the ams1 zone
`https://cp-ams1.scaleway.com/products/servers/availability`to get the availability within the ams1 zone.

Same goes for `par1` zone.
For the availability, it goes to `scarce` when it hits a first low_stock threshold (that is different for every instance) and `shortage` when it hits the lower second threshold.

danhunsaker:
@sieben Ok, so the documentation is outdated, then, because I'm not seeing that endpoint among the options there. I wonder what else I'm not seeing for that reason?

~~@pquentin Do you want that updated before merge, or would that be alright in a second PR?~~

remyleone:
@danhunsaker I'm working on adding documentation on the developer's documentation https://github.com/scaleway/api.scaleway.com/pull/56 I will warn you once it is merged. 

@pquentin Don't hesitate if you need additional support or if you have questions about Scaleway's services.

danhunsaker:
OK, we're pulling from the API, now, and I added a check for the actual size of the base image's volume, so we can be sure to create enough volumes to deploy properly. This should be it for this PR.

libcloud/compute/drivers/scaleway.py
danhunsaker: Fixed in b6356bb. :slightly_smiling_face:

libcloud/compute/drivers/scaleway.py
danhunsaker: Fixed in 29fe802.

libcloud/compute/drivers/scaleway.py
danhunsaker: @sieben Just need two things to resolve the remaining TODOs, at the moment:

One, what are all the valid snapshot states we might see from the API?

And two, will the root volume of a server always be listed as volume `0` in the API response? Or is there a better way to determine that?

remyleone:
Perfect so @costela @pquentin what could we do to make it get merged?

pquentin:
We'll need an ICLA from @bonifaido and a review from a libcloud committer, for example me if I find some time to do it.

@sieben I see that you authored two commits, are you happy with the current code?

remyleone:
Indeed I am :) Let me know if I can do something else to move it forward.

Le mer. 9 mai 2018 à 11:44, Quentin Pradet <notifications@github.com> a
écrit :

bonifaido:
Hi, what is an ICLA ? :)

pquentin:
@bonifaido http://libcloud.readthedocs.io/en/latest/development.html#contributing-bigger-changes :)

remyleone:
@bonifaido Did you signed it?

bonifaido:
Finally, I have sent the ICLA to the Apache Secretary email address. :)

danhunsaker:
Awesome. Thanks for laying the ground work on this one!

pquentin:
@bonifaido We received your ICLA, thanks!

bonifaido:
Great to hear that!

remyleone:
Would some squashing be useful before the merge?

pquentin:
Thanks for asking.

What would be really useful would be a rebase on top of trunk! If doing it involes a squash, then that's fine.

danhunsaker:
I can squash or not in the rebase. I've kept it rebased regularly thus far, so it'll be a matter of a moment to rebase and push again.

pquentin:
Oh, but if the last rebase is recent, then there's no problem, I'll do this myself, there won't be any conflicts.

danhunsaker:
Yeah, looks like trunk hasn't moved since the last rebase, so it should be good to go. 😊

danhunsaker:
Here have a squash. :slightly_smiling_face:

remyleone:
@pquentin how could I help to make it merged? I got a tiny change to provide but I would prefer to provide it against `trunk` rather than delaying this MR longer.

pquentin:
Well I need to put in some work now. :)

But since you're asking: one thing you could do is review other pull requests.

libcloud/compute/drivers/scaleway.py
```diff
@@ -0,0 +1,665 @@
...
+                            httplib.CREATED, httplib.NO_CONTENT]
+
+    def parse_body(self):
+        body = super(ScalewayResponse, self).parse_body()
+
+        links = self.connection.connection.getresponse().links
+        if links and 'next' in links:
+            response = self.connection.request(links['next']['url'],
+                                               data=self.connection.data,
+                                               method=self.connection.method)
```
pquentin: This is quite surprising: a connection request while parsing the body? Did you have another option? Also it looks it only fetches the next link while we could have more urls to fetch?

libcloud/compute/drivers/scaleway.py
pquentin: @sieben Can you please give an answer here? Thanks!

pquentin (commented):
Looks good! I added two comments.

libcloud/compute/drivers/scaleway.py
danhunsaker: It recurses, yes. It's impossible to know how many pages will be in a collection until the initial request is made, so multi-page results need to be processed afterward, but there aren't a lot of points where these further requests could be made without having to place the pagination logic outside the response itself, which seems to miss the purpose of putting responses in their own class to begin with. So it's done in the body parsing section, which is actually a bonus anyway, as the response will then always contain all the pages in a given collection, with the first page appending the recursive result of the next to its own. (And, given the nature of recursion, the second appends the result of the third to its own, so that each level of recursion returns its own results plus any others until the end of the list is reached.)

I'm open to other suggestions, if you have any insights into the design that I've missed. (Which seems likely.)

remyleone:
@cedriccabessa can confirm but the workflow is `available` -> `snapshotting` -> `available`

libcloud/compute/drivers/scaleway.py
```diff
@@ -0,0 +1,665 @@
...
+        :type region: :class:`.NodeLocation`
+
+        :return: the newly created image object
+        :rtype: :class:`.NodeImage`
+        """
+        data = {
+            'organization': self.key,
+            'name': name,
+            'arch': node.extra['arch'],
+            'root_volume': node.extra['volumes']['0']['id']  # TODO check this
```
pquentin: Oh and we have another TODO here? What does it mean? Is someone really going to do this?

libcloud/compute/drivers/scaleway.py
pquentin: I'll look at the solutions adopted in other drivers. (Another solution that I don't like is GCE which makes pagination explicit.)

libcloud/compute/drivers/scaleway.py
danhunsaker: This would be the second question I asked @sieben above - is the root volume always given as volume 0?

danhunsaker:
@sieben So, `available` twice? What does `available` actually communicate to us, then? What would it make the most sense to be mapped to?

remyleone:
Available means that the volume is available for a compute instance. This is the running/normal state. When you are doing a snapshot, the volume gets in the snapshotting state. Once the snapshot is finished the state go back to available. 
So volume-wise I think they are the 2 only states so far.

danhunsaker:
Hmm. That sounds like the volume state, rather than the snapshot state.

remyleone:
@CedricCabessa can you confirm the snapshot states?

CedricCabessa:
Hi @danhunsaker 
Here is how snapshot works:
When you create a snapshot, the snapshot object is immediately created but it is not yet usable: the state of the snapshot object is **snapshotting**. In order to avoid to poweroff a VM while a snapshot is in progress, the volume is also in state **snapshotting**
After a few seconds/minute, the snapshot is finished: the volume and the snapshot are both in state **available**, and you can use your snasphot to create a new instance.

Sometime, the snapshot can fail. In this case the snapshot will be in state **error** and the volume will be set to **available**

As @sieben said; for volume you have **available** and **snapshotting**, and for snapshot you have **available**, **snapshotting** and **error**

Also note that this is for hotsnapshot, for coldsnapshot (when the VM is powered off), everything is simpler and snapshot are done instantaneously so the volume and snapshot are always **available**

Feel free to ping me if you have more questions :-)

danhunsaker:
Perfect. That's exactly what I needed for this one. Any info on the other question, about the root volume always being volume zero for the server?

CedricCabessa:
@danhunsaker yes, the root volume is always volume 0

danhunsaker:
Great, thanks!

libcloud/compute/drivers/scaleway.py
pquentin: I think the ECS approach is better, which makes pagination explicit. Do you think it would be easy to do? If it's more than say one hour or two of work, it's probably not worth it as the current approach does work even if it mixes concerns and is very surprising.

pquentin:
@danhunsaker My understanding of the current state is that you now have enough information to resolve the two TODOs and you need to answer the pagination question above: is it too much work to support pagination explicitly in each method?

danhunsaker:
Right. Sorry for the delay.

It feels like making the user explicitly handle pagination on their own removes some of the benefit of abstraction, but it's not my call, really, as I'm just contributing, here. I do think something consistent across all providers would be the most beneficial, in any case. It sounds like no such standard approach exists, though.

Pushing a commit now with the TODOs resolved (one just needed the comment removed). Will investigate the proposed pagination approach and see what I can figure out, there.

pquentin:
OK, let's not block this PR for the pagination issue. I'll review your change when I get the chance. (In that case a new commit would have been better.)

Thanks!

danhunsaker:
Oops. Actually just pushed another squashed commit right before you posted that. This time, with a different pagination approach.

It's somewhere between the previous approach and what ECS is doing - it's an alternate `request` method on the `ScalewayConnection` class. It pulls the logic out of the response class and into the actual request handling, which should be much more clear about what's going on and how.

pquentin:
Oh I see you already made the recursion change. Thanks, I think that's much better.

danhunsaker:
I take too long to type things sometimes. :grin: Pretty sure that's everything requested on this one.

danhunsaker:
There we go. Both changes as their own commits. Hope that helps!

libcloud/compute/drivers/scaleway.py
```diff
@@ -0,0 +1,662 @@
...
+            'internal_bandwidth': int(
+                (size['network']['sum_internal_bandwidth'] or 0) /
+                (1024 * 1024)),
+            'ipv6': size['network']['ipv6_support'],
+            'alt_names': size['alt_names'],
+        }
+
+        return NodeSize(id=name,
+                        name=name,
+                        ram=int((size['ram'] or 0) / (1024 * 1024)),
```
pquentin, line 203: Let's not worry about this, but I think this is going to return an integer in Python 2, and a more precise float in Python 3, since we only have ints here. Using 1024.0 would have avoided the issue.

pquentin:
Thanks @danhunsaker @bonifaido @sieben, finally merged in trunk! ✨

libcloud/compute/drivers/scaleway.py
danhunsaker: Pretty sure the `int()` wrapping it will return an `int` in any case?

pquentin: Right sorry I was thinking the int() was on `size['ram']`, not the whole expression. I guess using // could have been clearer to me but that's quite an obscure operator so you made the right choice, thanks!

The-Loeki:
Thanks for the great work! I'm toying with it now. I notice that the API docs list `location=` as arg for a location, but you've coded `region=`.

Is that a misunderstanding on my part?

danhunsaker:
The variable names used in Python don't have to match the parameter names sent to the API... I tried to be as consistent with other drivers as I could, though some naming was already in place before I got involved.

The-Loeki:
I'm not sure I'm following; 
http://libcloud.readthedocs.io/en/latest/compute/api.html says to use `location=`, so as an avid RTFM'er I would expect that.

However, I had to use e.g.
```python
cloud = get_driver(Provider.SCALEWAY)(access_key, token)
cloud.list_images(region='ams1')
```
which tripped me over ;)

danhunsaker:
Right. Because Libcloud is a library for several cloud providers, it has several established conventions for certain concepts, such as using `region` as the name for that argument where most drivers need it. Its name in the Python code doesn't have to be the same name passed to the API endpoint on Scaleway's end, so it isn't. All the Scaleway folks are documenting is how to talk to their servers; the Libcloud adapter has its own documentation for its own conventions.

PR #1126, opened 2017-10-09
Add ex_fetch_power_state option to list_nodes().

tetron:
### Description

Allows caller to skip (potentially expensive) requests for extended node
power state status.

Also set timeout on connection object used to fetch token.

### Status

Ready for review

- [x] Documentation

### DIFF
--- a/libcloud/compute/drivers/azure_arm.py
+++ b/libcloud/compute/drivers/azure_arm.py
@@ -360,7 +360,7 @@ class AzureNodeDriver(NodeDriver):
                                  ex_offer, ex_sku, ex_version)
             return i[0] if i else None
 
-    def list_nodes(self, ex_resource_group=None, ex_fetch_nic=True):
+    def list_nodes(self, ex_resource_group=None, ex_fetch_nic=True, ex_fetch_status=True):
         """
         List all nodes.
 
@@ -369,6 +369,9 @@ class AzureNodeDriver(NodeDriver):
 
         :param ex_fetch_nic: Fetch NIC resources in order to get
         IP address information for nodes (requires extra API calls).
+        :type ex_urn: ``bool``
+
+        :param ex_fetch_status: Fetch node instance status (requires extra API calls).
         :type ex_urn: ``bool``
 
         :return:  list of node objects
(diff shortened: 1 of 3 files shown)


libcloud/compute/drivers/azure_arm.py
```diff
@@ -371,6 +371,9 @@ def list_nodes(self, ex_resource_group=None, ex_fetch_nic=True):
         IP address information for nodes (requires extra API calls).
         :type ex_urn: ``bool``
 
+        :param ex_fetch_status: Fetch node instance status (requires extra API calls).
```
tfmorris: Suggest this be phrased as "requires an extra API call per node"

libcloud/compute/drivers/azure_arm.py
```diff
@@ -371,6 +371,9 @@ def list_nodes(self, ex_resource_group=None, ex_fetch_nic=True):
         IP address information for nodes (requires extra API calls).
```
tfmorris: Suggest this be phrased as "requires an extra API call for each NIC"

libcloud/compute/drivers/azure_arm.py
```diff
@@ -1942,39 +1945,40 @@ def _to_node(self, data, fetch_nic=True):
...
+                        state = NodeState.TERMINATED
+                        break
+                    elif status["code"].startswith("ProvisioningState/failed"):
+                        state = NodeState.ERROR
+                        break
+                    elif status["code"] == "ProvisioningState/updating":
+                        state = NodeState.UPDATING
+                        break
+                    elif status["code"] == "ProvisioningState/succeeded":
+                        pass
```
tfmorris: The ProvisioningState is available from the virtual machines list resource group response: https://docs.microsoft.com/en-us/rest/api/compute/virtualmachines/virtualmachines-list-resource-group
without having to make an extra request.

I think `"provisioningState": "succeeded"` is close enough to return NodeState.RUNNING even without the additional detail of PowerState.

pquentin:
@tfmorris Thank you for the review!

@tetron Can you please fix the lint, possibly by moving the status fetch in another function?

tetron:
It's been 20 minute since I pushed and Travis hasn't started running tests yet.  I don't know if it is a rate limit thing or Travis is having trouble (but currently no downtime is reported on their status page).  Guess I'll try to be patient.

pquentin:
@tetron Heh. It did take some time to run, and there's one lint issue remaining. You can use `tox -e lint` to get results faster!

pquentin:
@tetron Thanks! We just need tests for the `ex_fetch_power_state=False` branch to ensure coverage does not decrease.

pquentin:
Thanks @tetron @ldipenti this is now merged in trunk.

ldipenti:
Great, thanks @pquentin !

PR #1127, opened 2017-10-11
[LIBCLOUD-954] ec2: Allow cn-north-1 even without pricing

pquentin:
### Description

See https://issues.apache.org/jira/browse/LIBCLOUD-954 and https://github.com/apache/libcloud/commit/1e1d77f320a2314e96c79011ab62ee028a835503

@fjros and @duanshiqiang, what do you think?

### Status

- done, ready for review

- [x] Documentation

### DIFF
--- a/docs/compute/drivers/ec2.rst
+++ b/docs/compute/drivers/ec2.rst
@@ -10,23 +10,29 @@ platform, Amazon Web Services (AWS).
     :width: 300
     :target: https://aws.amazon.com/ec2/
 
-It allows users to rent virtual servers in more than 9 regions such as:
+It allows users to rent virtual servers in more than 15 regions such as:
 
 * US East (Northern Virginia) Region
 * US East (Ohio) Region
 * US West (Oregon) Region
 * US West (Northern California) Region
+* GovCloud (US) Region
+* Canada (Central) Region
 * EU West (Ireland) Region
+* EU West (London) Region
 * EU Central (Frankfurt) Region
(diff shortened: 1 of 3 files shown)


fjros:
Thanks @pquentin, a nice trade-off :-)

Maybe this should be clearly stated in the docs, since in practice there's at least one api method (I think) not supported for the region.

duanshiqiang:
I think this is good :)

pquentin:
@fjros Thanks. I tried a better approach: setting price to None when it's not available as with cn-north-1. Also improved the docs. Please tell me what you think. I highly appreciate your timely feedback!

fjros:
@pquentin Great!! As far as I'm concerned, this fixes the issue.

Just a minor note... Maybe `_get_size_price()` should directly return None instead of raising an exception. In that way, the behavior would be consistent across the codebase. I mean, just moving the `try..except` block from `list_sizes()` to `_get_size_price()`. What do you think?

pquentin:
Ah, yes, that would be better! I'll do that.

pquentin:
Thinking a bit more about it, that would affect other drivers. I would not want to return None silently in those cases because we would not be alerted of an issue in tests. The change as it is currently only affects EC2.

pquentin:
Merged in trunk.

@fjros Please fell free to open a new PR or JIRA issue if you think of a better way. Thanks again.
@duanshiqiang Can you please close your JIRA issue if that fixes your problem? Thanks.

PR #1130, opened 2017-10-15
add ex_enable_ipv6 in DigitalOcean_v2 driver

vdloo:
#### So ipv6 can be enabled on a Droplet using the API

> To enable IPv6 networking on an existing Droplet (within a region that has IPv6 available), send
> a POST request to /v2/droplets/$DROPLET_ID/actions. Set the "type" attribute to enable_ipv6.

https://developers.digitalocean.com/documentation/v2/#enable-ipv6

### DIFF
--- a/libcloud/compute/drivers/digitalocean.py
+++ b/libcloud/compute/drivers/digitalocean.py
@@ -250,6 +250,12 @@ class DigitalOcean_v2_NodeDriver(DigitalOcean_v2_BaseDriver,
                                       data=json.dumps(attr), method='POST')
         return res.status == httplib.CREATED
 
+    def ex_enable_ipv6(self, node):
+        attr = {'type': 'enable_ipv6'}
+        res = self.connection.request('/v2/droplets/%s/actions' % (node.id),
+                                      data=json.dumps(attr), method='POST')
+        return res.status == httplib.CREATED
+
     def ex_rename_node(self, node, name):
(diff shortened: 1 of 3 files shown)


libcloud/test/compute/fixtures/digitalocean/ex_enable_ipv6.json
```diff
@@ -0,0 +1,13 @@
+{
```
pquentin: Since your change only applies to the v2 driver, this file is not needed right? Can you please remove it? Thanks!

pquentin (changes requested):
Thanks, the changes look good! I only have one question about a v1 fixture.

libcloud/test/compute/fixtures/digitalocean/ex_enable_ipv6.json
vdloo: yup, done!

pquentin:
Thank you, this has been merged in trunk! ✨

libcloud/test/compute/fixtures/digitalocean/ex_enable_ipv6.json
vdloo: @pquentin also see https://github.com/apache/libcloud/pull/1131, those two fixtures also are not needed I think because those methods only apply to the v2 driver as well

PR #1132, opened 2017-10-16
Ensure RawResponse downloads don't consume RAM

pquentin:
## Ensure RawResponse downloads don't consume RAM

The fix is general but the cause is very specific: since libcloud 2.0, `download_object()` in the S3 and GCS drivers first places the item in RAM, then writes it to a file. The main reason it broke is that accessing `body` in a requests Response consumes it entirely and loads it in RAM. To fix this, I moved `body` to a property: it is still accessibly, but loaded only if requested.

Tested this change by downloading a 38M file using this code:

```
from libcloud.storage.drivers.google_storage import GoogleStorageDriver

driver = GoogleStorageDriver(key=..., secret=...)
obj = driver.get_object('container_name', 'object_name')
driver.download_object(obj, '/path/to/file')
```

The RAM usage before:

before

The RAM usage after:

after

### Status

- done, ready for review

- [ ] Documentation

### DIFF
--- a/libcloud/common/base.py
+++ b/libcloud/common/base.py
@@ -278,10 +278,13 @@ class RawResponse(Response):
         if not self._response:
             response = self.connection.connection.getresponse()
             self._response = HttpLibResponseProxy(response)
-            self.body = response.content
             if not self.success():
                 self.parse_error()
         return self._response
+
+    @property
+    def body(self):
+        return self.response.body
 
     @property
     def reason(self):
@@ -587,6 +590,7 @@ class Connection(object):
                     url=url,
                     body=data,
                     headers=headers,
+                    raw=raw,
                     stream=stream)
             else:
                 if retry_enabled:
--- a/libcloud/http.py
+++ b/libcloud/http.py
@@ -319,3 +319,7 @@ class HttpLibResponseProxy(object):
     def version(self):
         # requests doesn't expose this
         return '11'
+
+    @property
+    def body(self):
+        return self._response.content


libcloud/common/base.py
```diff
@@ -587,6 +590,7 @@ def request(self, action, params=None, data=None, headers=None,
                     url=url,
                     body=data,
                     headers=headers,
+                    raw=raw,
```
jcsalterego, line 593: Why was this added?

pquentin: This is because [`prepared_request` uses raw instead of stream](https://github.com/apache/libcloud/blob/c2e71ecb4c2b97d858eebac8cae384438338f268/libcloud/http.py#L224-L238). This probably needs a refactoring to do since the stream parameter is unused, but I wanted to keep my change minimal and avoid disruption because changes here affect every driver.

jcsalterego: Ah, ok.

Kami:
Great catch, will look a bit later.

If possible we should also add a test case to prevent similar regressions in the future.

libcloud/common/base.py
```diff
@@ -278,12 +278,15 @@ def response(self):
         if not self._response:
             response = self.connection.connection.getresponse()
             self._response = HttpLibResponseProxy(response)
-            self.body = response.content
             if not self.success():
                 self.parse_error()
         return self._response
 
     @property
+    def body(self):
```
Kami, line 286: Would be good to add a comment on why we do this.

And a test case would also be great, besides that, LGTM.

Kami (approved):

pquentin:
Again, I don't know how to test this does not happen again. Any ideas?

libcloud/common/base.py
pquentin: Ah, right, will do.

Kami:
Same here - should be able to mock that attribute and verify that it has been accessed inside the test case.

If you need help, let me know (same with other PR) and I can have a look into adding a test case.

pquentin:
@Kami I do need help!

The issue here is that we don't want to check that the body property has been accessed, but that it has *not* been accessed. I was thinking that asserting that the property was not accessed was not robust enough, because I don't think it will help avoid a regression: many other methods are not called during the test.

What would have been interesting would be to get access to the requests Response object and check that `_content_consumed` (https://github.com/requests/requests/blob/775cde0914265d7dab2bc2501ed7abe6b85c4bae/requests/models.py#L811) is False (_content_consumed is private but really old and unlikely to change at this point).

In the context of libcloud.test.storage.test_s3.test_download_object_success I would only need to get access to the response variable defined in download_object, but I don't know how to do that.

Thoughts?

Kami:
Sorry for the delay - I was just looking into that.

You are correct, writing a test case for it is indeed quite hard - we would need to mock a lot of things, notably `HttpLibResponseProxy` and `Response`, then set `body` to `PropertyMock` and assert `call_count` attribute is 0.

I wanted to do that, but it gets really messy and error prone, because there is a lot of indirection.

For now I will go ahead and merge it into trunk and hopefully we will come up with a better way to test it in the future.

Kami:
Re `_content_consumed` sadly asserting on that wont work since that variable is also set to True once content is consumed via iterator - https://github.com/requests/requests/blob/775cde0914265d7dab2bc2501ed7abe6b85c4bae/requests/models.py#L724.

libcloud/common/base.py
```diff
@@ -278,11 +278,14 @@ def response(self):
...
             response = self.connection.connection.getresponse()
             self._response = HttpLibResponseProxy(response)
-            self.body = response.content
             if not self.success():
                 self.parse_error()
         return self._response
 
+    @property
+    def body(self):
+        return self.response.body
```
Kami, line 287: I'm a bit confused, still looking at the test case.

Shouldn't this be `self.response.content` so it results in the same behavior as in the past? Either that, or `self._response.body`, right?

Kami:
Managed to come up with a test case which doesn't include too much mocking and it doesn't look too convoluted and horrible - 98a07ac377a58076e248673a55a9428c9e8dfe21.

It's still not ideal, but it's the best we can do right now and better than no tests. You can confirm it works as intended, because it will fail if you revert change / fix from this PR.

libcloud/common/base.py
pquentin: I should have explained this since November, 1 2017 and should not have waited for a nice test case, sorry.

Note that `self.response` and `self._response` should be roughly equivalent, except that `self.response` is going to populate `self._response` if it does not exist yet and then return that, so that sounds more robust.

Anyway, `self.response` returns a `HttpLibResponseProxy`, and since I'm trying to follow the law of Demeter I've added a `body` property to `HttpLibResponseProxy` (see below in http.py) to make sure that `RawResponse` does not know about the innards of `HttpLibResponseProxy`. Withouth that property, I would have used `self.response._response.content`. (Note that the variable `response` in the `RawResponse.response` property is a `HttpLibResponseProxy`.)

So that's quite confusing and quite subtle, and this explains why this has been missed in the past!

Hope that helps

pquentin:
Thank you @Kami, this is great! 💯 👏

PR #1133, opened 2017-10-16
Fix improper getresponse() uses

pquentin:
### Description

During the migration from httplib to requests, a lot of driver code did not change because they use the high-level libcloud API which hid the change from httplib to requests with the HttpLibResponseProxy adapter. However, in some cases, driver code needs to access the actual response, and does this by calling getresponse().

While this used to return an httlib.HTTPResponse instance, this now returns a requests.Response instance. All code covered by tests was fixed, but OVH and Brightbox common code is not fully covered. I don't have a subscription to either service and can't test the change, but the resulting code can only be more correct since it fixes the `getresponse() result comes from httplib` assumption.

### Status

- done, ready for review

- [ ] Documentation

### DIFF
--- a/libcloud/common/brightbox.py
+++ b/libcloud/common/brightbox.py
@@ -78,12 +78,11 @@ class BrightboxConnection(ConnectionUserAndKey):
         response = self.connection.request(method='POST', url='/token',
                                            body=body, headers=headers)
 
-        response = self.connection.getresponse()
-
         if response.status == httplib.OK:
             return json.loads(response.read())['access_token']
         else:
-            responseCls = BrightboxResponse(response=response, connection=self)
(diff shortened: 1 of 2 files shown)


Kami:
Good catch :+1: 

Is it possible to add some tests for this (aka the regression tests which would catch the original issue)?

pquentin:
@Kami I think these two bugs would have been caught by standard tests, but those lines of code are not covered at all. And it's difficult/impossible to write tests when you do not have a subscription to the service.

Kami (approved):

pquentin:
@Kami Thanks for the review, merged in trunk.

Kami:
Thank you for the original PR and fix :)

PR #1134, opened 2017-10-16
Limit number of retries in Azure ARM destroy_node.

tetron:
Limit number of retries in Azure ARM destroy_node.

### Description

Azure ARM has several retry loops related to deleting the NIC and VHD as part of node cleanup.  This bounds those loops so they don't retry forever.

### Status

- done, ready for review

- [ ] Documentation


libcloud/compute/drivers/azure_arm.py
```diff
@@ -706,6 +709,10 @@ def destroy_node(self, node, ex_destroy_nic=True, ex_destroy_vhd=True):
         this node (default True).
         :type node: ``bool``
 
+        :param ex_retries: Number of times to retry checking if the node is gone,
+        destroying the NIC or destroying the VHD.
```
pquentin: Please mention the default value.

libcloud/compute/drivers/azure_arm.py
```diff
@@ -733,12 +740,14 @@ def destroy_node(self, node, ex_destroy_nic=True, ex_destroy_vhd=True):
 
         # Poll until the node actually goes away (otherwise attempt to delete
         # NIC and VHD will fail with "resource in use" errors).
-        while do_node_polling:
+        retries = ex_retries
+        while do_node_polling and retries > 0:
             try:
                 time.sleep(10)
```
pquentin: I was wondering: in your experience, how much time does it take for a node to go away?

pquentin (changes requested):
Thanks, the change make sense! There's a small docstring change. Also, the lint does not pass.

A possible future improvement is to use libcloud.utils.misc.retry which would reduce verbosity here and allow to add exponential back-off.

libcloud/compute/drivers/azure_arm.py
pquentin: Now that I think more about it, I think `ex_retries` is misleading because it suggests that we retry destroying, but we only check if it's gone multiple times. Do you have an idea for a better name than `ex_retries` that would convey this?

libcloud/compute/drivers/azure_arm.py
tetron: A couple of minutes.  The initial delete request is asynchronous, it responds "202 Accepted", which is why we have an idle loop to wait for it to really go away, because we can't delete the other resources (the NIC and VHD) until the leases are released.

pquentin: Sure. Thanks for the answer. In this case, the initial sleep of 10 seconds is acceptable, otherwise it would have made more sense to poll before sleeping.

pquentin:
@tetron Any progress?

ldipenti:
@pquentin I've renamed ex_retries as requested, added a new param to control the polling delay.

pquentin:
@ldipenti Do you think you could test the retry behavior and fix the conflict from #1137 that I just merged? Thank you!

pquentin:
Thank you! Merged in trunk. ✨

ldipenti:
Great! thanks.

PR #1135, opened 2017-10-18
storage: Fix hash computation performance on upload

pquentin:
### Description

The storage base driver computes the hash of uploaded files: individual drivers use it to ensure the reported hash is correct. Before libcloud 2.x, this was done efficiently: the file was only read once, and we were using `hash.update()` to avoid keeping the whole file in memory. With the switch to the requests module, both of these optimizations were removed inadvertently. It turns out the important one is using `hash.update()`: computing the hash on the whole file in memory is orders of magnitude slower.

### Status

- done, ready for review

- [ ] Documentation

### DIFF
--- a/libcloud/storage/base.py
+++ b/libcloud/storage/base.py
@@ -643,9 +643,9 @@ class StorageDriver(BaseDriver):
     def _hash_buffered_stream(self, stream, hasher, blocksize=65536):
         total_len = 0
         if hasattr(stream, '__next__'):
-            data = libcloud.utils.files.exhaust_iterator(iterator=stream)
-            hasher.update(b(data))
-            total_len = len(data)
+            for chunk in libcloud.utils.files.read_in_chunks(iterator=stream):
+                hasher.update(b(chunk))
+                total_len += len(chunk)
             return (hasher.hexdigest(), total_len)
         if not hasattr(stream, '__exit__'):
             for s in stream:


jcsalterego:
Makes sense to me.

Kami:
Another great catch.

As mentioned in the other PR - can you please add a test case for it so we avoid such regressions in the future?

pquentin:
How would you do that? I don't know how to write a test to ensure there is no performance regression.

Kami:
One way to test it would be to mock `libcloud.utils.files.read_in_chunks`, then call `upload_object` or whatever the method uses this underlying method and verify that `read_in_chunks` has indeed been called.

You will probably also need to mock some other things as well. Mocking is not 100% bullet proof, but it's better than nothing, imo.

Kami (approved):

Kami:
Added a test case in 0710d23de52a1c2c4841937e4c7c17b81950b9ae and merged it into trunk.

I also included small optimization in 484e02a6395f3f05989a90b88bdb4c9b583ab4d7. Afaik, there should be no need to fall back to the `while` loop at the end of the method when iterator object has `next()` method since `next()` and `__next__()` should be equivalent.

Good catch and thanks again.

pquentin:
Oh, thanks @Kami! This one was easier, but for some reason I first tried the other one. I would have added the test at some point, sorry that you had to do it yourself. Your tests and changes are great, thanks.

Kami:
With all those PRs now merged, we should also start thinking about a new release (2.3.0) soon :)

pquentin:
Sure! We have two new drivers ready for review, did you have the time to take a look?

https://github.com/apache/libcloud/pull/1121
https://github.com/apache/libcloud/pull/1154

PR #1139, opened 2017-11-02
Retry-After header handling fixes

ldipenti:
### Description

* Pass response headers to exception_from_message() when receiving an error response.
* Pass Retry-After header to BaseHTTPException, some other errors than 429 may include it. (Example: 503 Service Unavailable)
* Retry-After header may include an HTTP-date value, so check if that's the case and if so, translate it to a delay seconds value.
* Added tests.

### Status

- done, ready for review

- [ ] Documentation

### DIFF
--- a/libcloud/common/exceptions.py
+++ b/libcloud/common/exceptions.py
@@ -13,6 +13,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import time
+
+from email.utils import parsedate_tz, mktime_tz
+
 __all__ = [
     'BaseHTTPError',
     'RateLimitReachedError',
@@ -47,7 +51,7 @@ class RateLimitReachedError(BaseHTTPError):
     message = '%s Rate limit exceeded' % (code)
 
     def __init__(self, *args, **kwargs):
-        self.retry_after = int(kwargs.pop('retry_after', 0))
+        self.retry_after = int(kwargs.pop('headers', {}).get('retry-after', 0))
 
 
 _error_classes = [RateLimitReachedError]
@@ -68,7 +72,11 @@ def exception_from_message(code, message, headers=None):
         'headers': headers
     }
 
-    if headers and 'retry_after' in headers:
-        kwargs['retry_after'] = headers['retry_after']
+    if headers and 'retry-after' in headers:
+        http_date = parsedate_tz(headers['retry-after'])
+        if http_date is not None:
+            # Convert HTTP-date to delay-seconds
+            delay = max(0, int(mktime_tz(http_date) - time.time()))
+            headers['retry-after'] = str(delay)
     cls = _code_map.get(code, BaseHTTPError)
     return cls(**kwargs)
--- a/libcloud/common/exceptions.py
+++ b/libcloud/common/exceptions.py
@@ -51,7 +51,14 @@ class RateLimitReachedError(BaseHTTPError):
     message = '%s Rate limit exceeded' % (code)
 
     def __init__(self, *args, **kwargs):
-        self.retry_after = int(kwargs.pop('headers', {}).get('retry-after', 0))
+        headers = kwargs.pop('headers', None)
+        super(RateLimitReachedError, self).__init__(self.code,
+                                                    self.message,
+                                                    headers)
+        if self.headers is not None:
+            self.retry_after = int(self.headers.get('retry-after', 0))
+        else:
+            self.retry_after = 0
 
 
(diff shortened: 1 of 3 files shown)


libcloud/common/exceptions.py
```diff
@@ -47,7 +51,7 @@ class RateLimitReachedError(BaseHTTPError):
     message = '%s Rate limit exceeded' % (code)
 
     def __init__(self, *args, **kwargs):
```
tetron, line 53: Can we make sure that `code`, `message` and `headers` attributes are set on `RateLimitReachedError` ?  The fact that `RateLimitReachedError` extends `BaseHTTPError` but doesn't set these fields has caused real problems for us in our exception handlers in the past.

libcloud/common/exceptions.py
```diff
@@ -68,7 +72,11 @@ def exception_from_message(code, message, headers=None):
         'headers': headers
     }
 
-    if headers and 'retry_after' in headers:
-        kwargs['retry_after'] = headers['retry_after']
+    if headers and 'retry-after' in headers:
+        http_date = parsedate_tz(headers['retry-after'])
```
tetron, line 96: Have you confirmed this returns `None` on parse error instead of throwing an exception?

libcloud/common/exceptions.py
```diff
@@ -68,7 +72,11 @@ def exception_from_message(code, message, headers=None):
...
+        if http_date is not None:
+            # Convert HTTP-date to delay-seconds
+            delay = max(0, int(mktime_tz(http_date) - time.time()))
+            headers['retry-after'] = str(delay)
```
tetron, line 100: This header munging behavior should be documented somewhere, starting with the documentation string for the function.

tetron (changes requested):

libcloud/common/exceptions.py
ldipenti: Yes, as described on the docs, parsedate_tz() returns None or a tuple.

libcloud/common/exceptions.py
ldipenti: Sure, will add a superclass constructor call to it.

libcloud/common/exceptions.py
ldipenti: Updated docstring with examples.

libcloud/common/exceptions.py
```diff
@@ -47,7 +51,14 @@ class RateLimitReachedError(BaseHTTPError):
     message = '%s Rate limit exceeded' % (code)
 
     def __init__(self, *args, **kwargs):
-        self.retry_after = int(kwargs.pop('retry_after', 0))
+        headers = kwargs.pop('headers', None)
+        super(RateLimitReachedError, self).__init__(self.code,
+                                                    self.message,
+                                                    headers)
```
tetron, line 57: This doesn't look right, I don't think `self.code` and `self.message` are bound yet?  That's the point of calling the superclass constructor...

tetron: And if that's really a bug, the test coverage needs to catch it.

libcloud/common/exceptions.py
tetron: Thanks!

tetron (changes requested):

libcloud/common/exceptions.py
tetron: I see now, they are class variables, so that's okay.  Nevermind.

tetron (approved):

tetron:
@ldipenti this LGTM

pquentin:
@ldipenti Great! But there's still the RateLimitReachedError issue mentioned by @tetron?

ldipenti:
@pquentin class vars are accessible by __init__() and are passed to the superclass' constructor, right? Maybe I'm missing something.

pquentin:
Oh, right, class vars, sorry.

ldipenti:
@pquentin do you have any other suggestions for this PR?

pquentin:
Not really, it sounds good! I'll probably merge it tomorrow or Friday unless someone does it first. 

Unfortunately the process we have means merging takes some work, with signoffs and updates to CHANGES.rst. I also like to use fixups when rebasing in order to minimize the number of commits that don't build, which helps with bisecting.

ldipenti:
Cool, thanks!

pquentin:
Thanks! Merged in trunk. ✨

PR #1146, opened 2017-11-29
[LIBCLOUD-964] Tencent COS storage provider

itamaro:
### Description

Tencent COS (https://cloud.tencent.com/product/cos) is the object storage product of Tencent Cloud (https://cloud.tencent.com/).

### Status

This is mostly done & ready, with some tests still missing.

Not sure what's needed exactly w.r.t. documentation.

- [ ] Documentation

### DIFF
--- a/libcloud/common/base.py
+++ b/libcloud/common/base.py
@@ -912,24 +912,6 @@ class ConnectionUserAndKey(ConnectionKey):
         self.user_id = user_id
 
 
-class ConnectionAppIdAndUserAndKey(ConnectionKey):
-    """
-    Base connection class which accepts an ``app_id``, ``user_id`` and
-    ``key`` argument.
-    """
-
-    user_id = None
-
-    def __init__(self, app_id, user_id, key, secure=True, host=None, port=None,
-                 url=None, timeout=None, proxy_url=None,
-                 backoff=None, retry_delay=None):
-        super(ConnectionAppIdAndUserAndKey, self).__init__(
-            user_id, key, secure=secure, host=host, port=port, url=url,
-            timeout=timeout, backoff=backoff, retry_delay=retry_delay,
-            proxy_url=proxy_url)
-        self.app_id = app_id
-
-
 class BaseDriver(object):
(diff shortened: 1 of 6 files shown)


pquentin:
Thanks for the PR, it's the best way to get feedback. Sorry that you did not get answers on the mailing lists and on JIRA, we don't have enough contributors to help with all of this. I don't know if I'll be able to review your PR myself and would appreciate some help: there's no need to be a commiter to review a pull request!

With regards to documentation, you already have docstrings, which is great. You will also need a page specific to the storage driver, see docs/storage/drivers for examples.

pquentin:
Also make sure that the continuous integration passes.

itamaro:
> Also make sure that the continuous integration passes.

Yeah, I was wondering about how to tackle that.

My implementation relies on the Tencent Python SDK as a 3rd party package, but it's probably not desirable to have it as a "global" dependency for users that don't want to use this storage provider.

What's the recommended approach for such external dependencies?

pquentin:
Oh, I had not noticed that. As you have seen, libcloud does not use external dependencies for individual drivers, you would have to reimplement the useful parts of qcloud_cos for this. While this means we have to reinvent the wheel, it's often not that bad because using the HTTP API is usually not much harder than using a Python wrapper around it. And having minimal dependencies in requests is an important feature of the project.

Wrapping qcloud_cos to make it look libcloud-like is valuable, but it's just not something that should be integrated in libcloud. Sorry about that.

To answer to the question more directly: we use fixtures and mocks for tests, look at the other drivers to see how it's done.

itamaro:
Thanks for pointing me at the right direction!

> To answer to the question more directly: we use fixtures and mocks for tests, look at the other drivers to see how it's done.

Yeah, I noticed that. My tests are based on patching out the actual qcloud_cos client (returning canned responses), but it still requires importing stuff from that package.

I understand I should reimplement the relevant parts from qcloud_cos instead of depending on it, thanks.

Kami:

PR #1150, opened 2017-12-06
Fixed Pytest on Python 2.6 & 3.3

ZuluPro:
### Description

CI was failing with  Python 2.6 & 3.3 because new version of Pytest doesn't support.
I make Python 2.6 & 3.3 use the last compatible Pytest version.

https://docs.pytest.org/en/latest/changelog.html#pytest-3-3-0-2017-11-23

### Status

Done

- [x] Documentation

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -6,8 +6,8 @@ passenv = TRAVIS TRAVIS_JOB_ID TRAVIS_BRANCH
 deps =
     -r{toxinidir}/requirements-tests.txt
     lockfile
-    py{2.6,2.7}: paramiko
     py{2.6}: unittest2
+
 commands = cp libcloud/test/secrets.py-dist libcloud/test/secrets.py
            python setup.py test
 basepython =
--- a/requirements-tests.txt
+++ b/requirements-tests.txt
@@ -7,4 +7,3 @@ codecov
 coverage<4.0
 requests
 requests_mock
-pytest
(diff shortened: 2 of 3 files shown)


tox.ini
```diff
@@ -6,8 +6,8 @@ passenv = TRAVIS TRAVIS_JOB_ID TRAVIS_BRANCH
 deps =
     -r{toxinidir}/requirements-tests.txt
     lockfile
-    py{2.6,2.7}: paramiko
```
pquentin, line 9: I think we need to investigate the paramiko breakage instead of stop testing it. But having a broken CI is indeed bad.

pquentin (commented):
I would prefer dropping support for 2.6 & 3.3. I asked on the dev@ mailing list if anyone is opposed to do this.

tox.ini
ZuluPro: As the code coverage didn't change, and paramiko is installed only with Python 2,
I think this statement was just useless and if it really test something, it should be in all versions.

pquentin:
Okay, no oppositions on the mailing list, and we'll always have time to add support again if the release is blocked because of this during the vote.

@ZuluPro Do you think you coud open a new PR to remove support for 2.6 and 3.3? Let's not remove paramiko just yet to give us some time to investigate. Thanks for your really helpful work on this!

tonybaloney:
Hi Quentin,

I don't know if anyone can check, what happens when you try and install it
from 2.6? Does it just install an older version? I've heard solutions that
would support this kind of setup

pquentin:
@tonybaloney You mean installing pytest from 2.6 or 3.3? It fails, which is why the CI was failing and we noticed the issue. But that's exactly what @ZuluPro was proposing. Do you think keeping 2.6 or 3.3 is valuable?

iler:
Would be nice to see this getting solved so that the CI would be fully functional and the PR's would be processed in normal way. So could we fix this issue and then separately discuss and decide on should the support for 2.6 and 3.3 be dropped?

pquentin:
@iler The problem is that nobody is reviewing the pull requests, including this one.

PR #1152, opened 2017-12-11
[LIBCLOUD-969] Add pricing information to UpCloud's driver list_sizes

mlackman:
### Description

UpCloud's driver were missing the pricing information from list_sizes return values. Added the information to NodeSize.extra data, because pricing zone specific. 

### Status

- done, ready for review

- [x] Documentation

### DIFF
--- a/libcloud/compute/drivers/upcloud.py
+++ b/libcloud/compute/drivers/upcloud.py
@@ -106,20 +106,22 @@ class UpcloudDriver(NodeDriver):
         response = self.connection.request('1.2/zone')
         return self._to_node_locations(response.object['zones']['zone'])
 
-    def list_sizes(self):
+    def list_sizes(self, location=None):
         """
         List available plans
 
-        Note: Node.price will be always None, because the pricing depends,
-        where the Node is hosted. Node.extra['zones'] will contain
-        pricing for different hosting zones.
+        :param location: Location of the deployement. Price depends on
+        location. lf location is not given or price not found for
+        location, price will be None (optional)
+        :type location: :class:`.NodeLocation`
 
         :rtype: ``list`` of :class:`NodeSize`
         """
         prices_response = self.connection.request('1.2/price')
         response = self.connection.request('1.2/plan')
         return self._to_node_sizes(response.object['plans']['plan'],
-                                   prices_response.object['prices']['zone'])
+                                   prices_response.object['prices']['zone'],
+                                   location)
 
     def list_images(self):
         """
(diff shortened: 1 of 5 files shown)


danhunsaker (commented):
This looks good, but it's a bit inconsistent with other drivers, which weakens some of the advantage of using Libcloud in the first place. There are other drivers with similar inconsistencies, and each should be addressed as well, but they are well beyond the scope of this PR, obviously.

libcloud/compute/drivers/upcloud.py
```diff
@@ -263,11 +270,13 @@ def _parse_country(self, zone_id):
         Zone_id format [country]_[city][number], like fi_hel1"""
         return zone_id.split('-')[0].upper()
 
-    def _to_node_sizes(self, plans):
-        return [self._construct_node_size(plan) for plan in plans]
+    def _to_node_sizes(self, plans, prices):
+        plan_price = PlanPrice(prices)
+        return [self._construct_node_size(plan, plan_price) for plan in plans]
```
danhunsaker: Curious about the naming, here. Why `_construct_node_size` instead of the more common (across Libcloud compute drivers) `_to_node_size`? I know it's an internal method, and this name was already in use in prior commits, but it still strikes me as unusual.

libcloud/compute/drivers/upcloud.py
```diff
@@ -109,10 +110,16 @@ def list_sizes(self):
         """
         List available plans
 
+        Note: Node.price will be always None, because the pricing depends,
+        where the Node is hosted. Node.extra['zones'] will contain
```
danhunsaker: This makes some sense, but I'll note that other drivers have handled this by requiring a NodeLocation as an argument, and setting `NodeSize.price` accordingly. Consider switching to this approach, for consistency.

mlackman: Yeah it is better that way, I think the parameter must be named ex_location. I will change this.

danhunsaker: On the contrary, the parameter is just `location` in the drivers I've worked with so far. At the least, both Azure ARM and Amazon EC2 use `location=None` as the argument definition.

mlackman: Ok, I'll go with the plain location. If it is approved its ok.

pquentin:
@danhunsaker Thanks for the review!

mlackman:
Now implemented those changes. CI still broken, but those seems to be problem also with other merge requests

danhunsaker (approved):
Looks great from here!

pquentin:
Thank you @mlackman and @danhunsaker, merged in trunk! ✨

pquentin:
I’d like to thank you again for your contribution to the libcloud project. As you might have noticed, we receive more pull requests that we can handle. If you would like to continue to help the project, reviewing one open pull request would be a great way to do so. But there’s no pressure: life is too short to spend volunteer time on things that you don’t find fulfilling.

PR #1157, opened 2018-01-03
ECS Aliyun: support modify_security_group_attributes

zym-ustc:
### Description

- [describe_security_group_attributes.xml](https://github.com/apache/libcloud/blob/trunk/libcloud/test/compute/fixtures/ecs/describe_security_group_attributes.xml) doesn't match its title.
- Can't modify security group name or description via ECS driver currently

Reference: 
[Aliyun API Doc - DescribeSecurityGroupAttribute](https://help.aliyun.com/document_detail/25555.html?spm=5176.doc25556.6.893.9yP6oY)
[Aliyun API Doc - ModifySecurityGroupAttribute](https://help.aliyun.com/document_detail/25559.html?spm=5176.doc25555.6.897.J4hUYN)

### Status

- done, ready for review

- [x] Documentation

### DIFF
--- a/libcloud/test/compute/test_ecs.py
+++ b/libcloud/test/compute/test_ecs.py
@@ -517,6 +517,13 @@ class ECSDriverTestCase(LibcloudTestCase):
         image = self.driver.copy_image(self.region, self.fake_image, None)
         self.assertIsNotNone(image)
 
+    def test_ex_create_security_group(self):
+        self.sg_description = 'description'
+        self.client_token = 'client-token'
+        sg_id = self.driver.ex_create_security_group(
+            description=self.sg_description, client_token=self.client_token)
+        self.assertEqual('sg-F876FF7BA', sg_id)
+
     def test_ex_list_security_groups(self):
         sgs = self.driver.ex_list_security_groups()
         self.assertEqual(1, len(sgs))
@@ -539,12 +546,35 @@ class ECSDriverTestCase(LibcloudTestCase):
                                                      group_id=ex_security_group_id_value)
         self.assertTrue(result)
 
+    def test_ex_delete_security_group_by_id(self):
+        ex_security_group_id_value = "sg-28ou0f3xa"
+        result = self.driver.ex_delete_security_group_by_id(
+            group_id=ex_security_group_id_value)
+        self.assertTrue(result)
+
+    def test_ex_modify_security_group_by_id(self):
+        ex_security_group_id_value = "sg-28ou0f3xa"
+        result = self.driver.ex_modify_security_group_by_id(
+            group_id=ex_security_group_id_value)
+        self.assertTrue(result)
+
     def test_ex_list_security_groups_with_ex_filters(self):
         ECSMockHttp.type = 'list_sgs_filters'
         self.vpc_id = 'vpc1'
         ex_filters = {'VpcId': self.vpc_id}
         sgs = self.driver.ex_list_security_groups(ex_filters=ex_filters)
         self.assertEqual(1, len(sgs))
+
+    def test_ex_list_security_group_attributes(self):
+        ex_security_group_id_value = 'sg-28ou0f3xa'
+        sgas = self.driver.ex_list_security_group_attributes(
+            group_id=ex_security_group_id_value)
+        self.assertEqual(1, len(sgas))
+        sga = sgas[0]
+        self.assertEqual('ALL', sga.ip_protocol)
+        self.assertEqual('-1/-1', sga.port_range)
+        self.assertEqual('Accept', sga.policy)
+        self.assertEqual('internet', sga.nic_type)
 
     def test_ex_list_zones(self):
         zones = self.driver.ex_list_zones()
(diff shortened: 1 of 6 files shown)


pquentin:
You have a lint error left: `libcloud/compute/drivers/ecs.py:851:1: W293 blank line contains whitespace`. Please check with `tox -e lint`.

libcloud/test/compute/test_ecs.py
```diff
@@ -943,6 +943,15 @@ def _delete_sg_by_id_DeleteSecurityGroup(self, method, url, body, headers):
...
 
+    def _modify_sg_by_id_ModifySecurityGroup(self, method, url, body, headers):
+        params = {'RegionId': self.test.region,
+                  'SecurityGroupId': 'sg-fakeSecurityGroupId',
+                  'SecurityGroupName': 'sg-fakeSecurityName',
+                  'Description': 'sg-fakeSecurityDescription'}
+        self.assertUrlContainsQueryParams(url, params)
+        resp_body = self.fixtures.load('modify_security_group_by_id.xml')
+        return (httplib.OK, resp_body, {}, httplib.responses[httplib.OK])
+
```
pquentin: This is only a fixture, not a test: test methods start with `test_`. You need to ensure such `ex_modify_security_group_by_id` is called from the tests. codecov should have telled us this, I don't know why it's not part of the checks here.

pquentin:
@zym-ustc I agree that those fixtures are confusing. What I can say:

 * only functions of ECSDriverTestCase that start with `test_` are considered as tests, such as test_list_nodes
 * in `ECSDriverTestCase.setUp`, the connection class of the `ECSDriver` instance is defined as `ECSMockHttp`, because we don't want to make real requests in tests, but use mocking instead
* `ECSMockHttp`, defined lower in the file, is a subclass of `MockHttp` which is defined in `libcloud/test/__init__.py` and this is where all the magic happens: when `ECSDriver` makes a request, `MockHttp._get_requests` will try to find a method in `MockHttp` to call.

For `test_list_nodes`, for example, the endpoint is `DescribeInstances`, but instead `ESMockHttp._DescribeInstances` will be called, and it only reads from the `describe_instances.xml` fixture, which is in `libcloud/test/compute/fixtures/ecs/describe_instances.xml` and this is from this file that things like `iZ28n7dkvovZ` come from.

Always make sure that your test fails without your code, and consider using a debugger such as pudb to understand what happens. So you could add `import pudb; pudb_set_trace()` in`test_list_nodes` then run `pytest -k ecs -s` and try to follow what happens. Or you can use `print()`.

Does that help?

libcloud/test/compute/test_ecs.py
```diff
@@ -539,13 +546,36 @@ def test_ex_leave_security_group(self):
...
+        result = self.driver.ex_delete_security_group_by_id(
+            group_id=ex_security_group_id_value)
+        self.assertTrue(result)
+
+    def test_ex_modify_security_group_by_id(self):
+        ex_security_group_id_value = "sg-28ou0f3xa"
+        result = self.driver.ex_modify_security_group_by_id(
+            group_id=ex_security_group_id_value)
+        self.assertTrue(result)
+
```
pquentin: It would be awesome if you could specify a name and description in those tests to improve coverage!

libcloud/test/compute/test_ecs.py
```diff
@@ -535,27 +535,29 @@ def test_ex_list_security_groups(self):
         self.assertEqual('2015-06-26T08:35:30Z', sg.creation_time)
 
     def test_ex_join_security_group(self):
-        ex_security_group_id_value = 'sg-28ou0f3xa'
+        self.sg_id = "sg-28ou0f3xa"
         result = self.driver.ex_join_security_group(self.fake_node,
-                                                    group_id=ex_security_group_id_value)
+                                                    group_id=self.sg_id)
```
pquentin: Can you please explain why you switched to storing `sg_id` inside `self` ?

zym-ustc: You can see, ```sg_id``` is needed by five functions (maybe more in the future) of ```ecs.py```.  We should set it in ```test_``` functions and check it in class ```ECSMockHttp```. Storing ```sg_id``` inside ```self``` make it easier for us to get the value in ```ECSMockHttp``` and keep consistent in other future ```test_``` functions.

I think it's better to treat it as a attribute of class ```security_group```, just like ```self.fate_sg.id```, but ```libcloud/compute/base.py``` doesn't include that class. It is also okay to leave it as ```ex_security_group_id_value``` and check it with the actual value.

pquentin: I see! But you need to do that in the `setUp` method. It's clearer and means the tests still pass if you run them with a different order with say `pytest --failed-first` or if you run only one of them with `pytest -k`.

zym-ustc: Great idea.

pquentin:
You have a few lint issues left: https://travis-ci.org/apache/libcloud/jobs/330377961

libcloud/test/compute/test_ecs.py
```diff
@@ -528,15 +536,27 @@ def test_ex_list_security_groups(self):
...
+        self.assertTrue(result)
+
+    def test_ex_delete_security_group_by_id(self):
+        result = self.driver.ex_delete_security_group_by_id(
+            group_id=self.fake_security_group_id)
+        self.assertTrue(result)
+
+    def test_ex_modify_security_group_by_id(self):
+        self.sg_name = 'name'
+        self.sg_description = 'description'
```
pquentin, line 555: I think I would have preferred to keep the hardcoded values 'name' and 'description' because that's less state to keep track of if at some point we want to parallelize the tests. But that's minor and I will stop pestering you :)

pquentin (approved):
Looks good. Thank you for your tenacity! :)

pquentin:
Thanks, merged in trunk. ✨

libcloud/test/compute/test_ecs.py
zym-ustc: Thanks for your patient guidance  :)

pquentin: (Regarding your function before the edit, yes, I meant just use 'description' and 'name' hardcoded without any variable, in both functions.)

pquentin:
I’d like to thank you again for your contribution to the libcloud project. As you might have noticed, we receive more pull requests that we can handle. If you would like to continue to help the project, reviewing one open pull request would be a great way to do so. But there’s no pressure: life is too short to spend volunteer time on things that you don’t find fulfilling.

zym-ustc:
I 'm willing to help the project. I'll try to review pull requests on familiar drivers at first.

pquentin:
Thanks @zym-ustc! Sometimes the first feedback is to say that lint does not pass (like I did on your PR) or that tests are needed. In those cases, you don't need to be familiar with the drivers.

PR #1160, opened 2018-01-08
Fix read_in_chunks on Python 3.7

pquentin:
According to https://www.python.org/dev/peps/pep-0479/#id41, this will stop working in Python 3.7: fix it now.

### Status

- done, ready for review

- [ ] Documentation

### DIFF
--- a/libcloud/utils/files.py
+++ b/libcloud/utils/files.py
@@ -83,7 +83,7 @@ def read_in_chunks(iterator, chunk_size=None, fill_size=False,
             if empty and yield_empty:
                 yield b('')
 
-            raise StopIteration
+            return
 
         if fill_size:
             if empty or len(data) >= chunk_size:


pquentin:
The failure is unrelated to the PR: `pip install tox` failed.

Kami:
Looks good to me, is it possible to add some kind of test case for it?

pquentin:
It was raising a warning on Python 3.6, it no longer is. read_in_chunks is tested by a number of tests already, the idea is that the behavior should not change. We might add 3.7-dev to the travis build matrix, but it's not feature-frozen yet so we would have to allow failures.

tonybaloney:
@pquentin I've heard it's still pretty slow on travis but it might be worth giving it a go

pquentin:
Since @Kami said that it looked good, I decided to merge this. Let's not slow down travis right now.

PR #1161, opened 2018-01-09
Reduce `tox -e lint` verbosity

pquentin:
### Description

By default, `python -m json.tool` prints to stdout, which makes `tox -e lint` really verbose. This change keeps the errors, if any, but discards the result. There is no automated test to add, but I did check that we were seeing the errors when the input file was not well formed. 

`/dev/null` won't work on Windows, but we already use /bin/bash for `tox -e docs`. Is Windows supported by libcloud? If it is, we should add Appveyor CI. At that point, I'll stop using json.tool for `tox -e lint` but will write a simple Python scripts that uses the `json`module.

Thoughts?

### Status

- done, ready for review

- [ ] Documentation

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -83,7 +83,7 @@ commands = flake8 --ignore=E402 --exclude="test" libcloud/
            flake8 --ignore=E402 integration/
            flake8 --ignore=E402,E902 docs/examples/
            flake8 --ignore=E402,E902 --max-line-length=160 contrib/
-           python -mjson.tool libcloud/data/pricing.json
+           python -mjson.tool libcloud/data/pricing.json /dev/null
 
 [testenv:checks]
 commands = bash ./scripts/check_file_names.sh


pquentin:
(Travis is currently broken because the apt command is not compatible with the Ubuntu version: --force-yes is no longer supported.)

Kami:
I believe build failure is related to upstream error in libvirt package - https://travis-ci.org/apache/libcloud/jobs/326787986#L572

```Errors were encountered while processing:
 libvirt-bin
W: --force-yes is deprecated, use one of the options starting with --allow instead.
E: Sub-process /usr/bin/dpkg returned an error code (1)
```

`--force-yes` being deprecated is just a warning message and not a fatal error.
  
In short - the failure is probably related to same Travis environment / base image change.

tox.ini
```diff
@@ -82,7 +82,7 @@ commands = flake8 --ignore=E402 --exclude="test" libcloud/
            flake8 --ignore=E402 integration/
            flake8 --ignore=E402,E902 docs/examples/
            flake8 --ignore=E402,E902 --max-line-length=160 contrib/
-           python -mjson.tool libcloud/data/pricing.json
+           python -mjson.tool libcloud/data/pricing.json /dev/null
```
Kami, line 86: Is this line missing `>` for stdout redirect aka `> /dev/null`?

Kami: Or does this syntax also work and second argument is file where to print / save the output?

Kami:
Might be related to this - https://twitter.com/travisci/status/950481155740127232

tox.ini
pquentin: @Kami the second option! `>` does not appear to be supported by tox commands. See outfile in https://docs.python.org/3/library/json.html#json-commandline

pquentin:
@Kami ah yes, you're right. And yes, it's definitely related, during the migration only some of the build failed, they now all fail.

pquentin:
@Kami I fixed the build failure in trunk and rebased this PR. Please take a look. Thanks!

(Still pending here, but passed at https://travis-ci.org/pquentin/libcloud/builds/330317350)

PR #1165, opened 2018-01-18
Allow shared vpc in gce

bchazalet:
### Description

It's currently not possible to create an GCE instance within a [shared VPC](https://cloud.google.com/vpc/docs/shared-vpc) and a shared subnet via ansible although it's possible from the google compute console (see https://github.com/ansible/ansible/issues/30759).

In the code, the problem seems to be that even if a fully qualified resource URL is passed for a network (or subnetwork), it is parsed into parts and then reconstructed assuming the resource is under the configured gcloud project. But since we're using a shared VPC, the resource is under a different project.

 For example, say, I have two projects, `my-main-project` which hosts a shared vpc named `sharedvpc` and `my-other-project` which has shared access to `sharedvpc`. What I want to do is to create a GCE instance under `my-other-project` and within the `sharedvpc` network.

So in the gce module, I pass the resource URLs:
```yml
  - name: Create instance
      gce:
        instance_names: "{{ instance_name }}"
        machine_type: "{{ machine_type }}"
        image: "{{ image_name }}"
        network: https://www.googleapis.com/compute/v1/projects/my-main-project/global/networks/sharedvpc
        subnetwork: https://www.googleapis.com/compute/v1/projects/my-main-project/regions/europe-west2/subnetworks/default"
        zone: "{{ zone }}"
        tags: app-servers
        state: present
```

But, in the module's code, the resource URLs will be parsed and transform respectively to `/global/networks/sharedvpc` and `/regions/europe-west2/subnetworks/default` which implicitly refer to the project configured by glcoud, i.e. `my-other-project`. As a result, I get a `ResourceNotFound` error.

Instead, if a resource (network or subnetwork) is fully qualified via a resource URL (i.e. it starts with `https://`), we can simply use it in the underlying google request. This is what the new code does.

### Status

Done, ready for review. I'm new to libcloud, so I'll happily take feedback and amend the PR.

- [ ] Documentation

### DIFF
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -7073,9 +7073,10 @@ class GCENodeDriver(NodeDriver):
         """
         region_name = None
         if name.startswith('https://'):
-            parts = self._get_components_from_path(name)
+            #parts = self._get_components_from_path(name)
             #name = parts['name']
-            region_name = parts['region']
+            #region_name = parts['region']
+            request = name
         else:
             if isinstance(region, GCERegion):
                 region_name = region.name
@@ -7085,15 +7086,16 @@ class GCENodeDriver(NodeDriver):
                 else:
                     region_name = region
 
-        if not region_name:
-            region = self._set_region(region)
-            if not region:
-                raise ValueError("Could not determine region for subnetwork.")
-            else:
-                region_name = region.name
+            if not region_name:
+                region = self._set_region(region)
+                if not region:
+                    raise ValueError("Could not determine region for subnetwork.")
+                else:
+                    region_name = region.name
 
-        #request = '/regions/%s/subnetworks/%s' % (region_name, name)
-        request = '%s' % (name)
+            request = '/regions/%s/subnetworks/%s' % (region_name, name)
+
+        # request = '/regions/%s/subnetworks/%s' % (region_name, name)
         response = self.connection.request(request, method='GET').object
         return self._to_subnetwork(response)
 
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -7074,7 +7074,7 @@ class GCENodeDriver(NodeDriver):
         region_name = None
         if name.startswith('https://'):
             parts = self._get_components_from_path(name)
-            name = parts['name']
+            #name = parts['name']
             region_name = parts['region']
         else:
             if isinstance(region, GCERegion):
@@ -7092,7 +7092,8 @@ class GCENodeDriver(NodeDriver):
             else:
                 region_name = region.name
 
-        request = '/regions/%s/subnetworks/%s' % (region_name, name)
+        #request = '/regions/%s/subnetworks/%s' % (region_name, name)
+        request = '%s' % (name)
         response = self.connection.request(request, method='GET').object
         return self._to_subnetwork(response)
 
@@ -7106,7 +7107,10 @@ class GCENodeDriver(NodeDriver):
         :return:  A Network object for the network
         :rtype:   :class:`GCENetwork`
         """
-        request = '/global/networks/%s' % (name)
+        if name.startswith('https://'):
+          request = name
+        else:
+          request = '/global/networks/%s' % (name)
         response = self.connection.request(request, method='GET').object
         return self._to_network(response)
 
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -7100,7 +7100,7 @@ class GCENodeDriver(NodeDriver):
         """
         Return a Network object based on a network name.
 
-        :param  name: The name of the network
+        :param  name: The name or URL of the network
         :type   name: ``str``
 
         :return:  A Network object for the network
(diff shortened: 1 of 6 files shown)


libcloud/compute/drivers/gce.py
```diff
@@ -7073,9 +7073,10 @@ def ex_get_subnetwork(self, name, region=None):
         """
         region_name = None
         if name.startswith('https://'):
-            parts = self._get_components_from_path(name)
-            name = parts['name']
-            region_name = parts['region']
+            #parts = self._get_components_from_path(name)
+            #name = parts['name']
+            #region_name = parts['region']
```
pquentin: Some initial feedback: please remove code instead of commenting it. Same comment for `# request = '/regions/%s/subnetworks/%s' % (region_name, name)` below. Thanks!

bchazalet: done!

pquentin:
Thanks for the contribution! The detailed explanation explains really clearly what is going on. ✨  To avoid a regression in the future, do you think you could add a test? See `test_ex_get_subnetwork` in `libcloud/test/compute/test_gce.py`.

bchazalet:
sure, I'll do my best to add a test for the changes.

pquentin:
Please tell me if you don't make progress at all on the test front after 15 minutes.

Oh, and you will also need to document that the `name` parameter in `ex_get_network` can also be an URL, as it is done in the `ex_get_subnetwork` docstring.

Finally, the change itself looks good. Do you have any idea why the URL was parsed and generated again? To have an error for wrong URLs? Or something else?

bchazalet:
I am really struggling with tests, it looks like I need to fix my python setup in order to be able to run them.

On the actual content of the tests, how do they work? They don't make real calls to the cloud, do they? How do the assert works then?

In `ex_get_subnetwork`, I think it parsed the url to be consistent with the else branch and have only one `request =` declaration. But that's just a guess.

pquentin:
Tests use fixtures, see https://github.com/apache/libcloud/pull/1157#issuecomment-355883772 for an explanation which uses another driver for context. For GCE, these are files like libcloud/test/compute/fixtures/gce/regions_us-central1_subnetworks_cf_972cf02e6ad49112.json.

Did you install the test requirements? Can you paste the log of `pytest -k gce` ?

bchazalet:
Thanks @pquentin. The fixtures bit make sense. 

The requirements is where I am struggling (I have a problem with setuptools it seems), but I'll get it to work.

pquentin:
Are you working in a virtual environment? These instructions should work:

```
python3 -m venv myvenv
source myvenv/bin/activate
pip install -r requirements-tests.txt
pytest -k gce
```

The last command should print this:

```
========================= test session starts ==========================
platform darwin -- Python 3.6.3, pytest-3.3.1, py-1.5.2, pluggy-0.6.0
rootdir: /Users/quentin/Projects/libcloud, inifile: pytest.ini
collected 8995 items

libcloud/test/compute/test_gce.py .............................. [ 15%]
................................................................ [ 48%]
................................................................ [ 82%]
..................                                               [ 91%]
libcloud/test/loadbalancer/test_gce.py ................          [100%]

======================== 8803 tests deselected =========================
============= 192 passed, 8803 deselected in 22.08 seconds =============
```

bchazalet:
Thanks a lot for that! It worked, unlike all the other solutions so far.

bchazalet:
thanks again @pquentin for the advice, I got the tests working and added simple tests for URLs passing in `get_network` and `get_subnetwork`. Let me know what you think, and if you had more in mind.

I also updated the doc of `get_network`.

pquentin:
Thanks! Those tests are useful. However you're only testing with the correct project name, that is you're only testing the old behavior. This is nice! We did not have such a test until now. But we need to test the new behavior too, by using an URL with a different project name. I'll guess you'll need a new fixture. Maybe something like:

```
url = 'https://www.googleapis.com/compute/v1/projects/other_name/regions/us-central1/subnetworks/cf-972cf02e6ad49112'
subnetwork = self.driver.ex_get_subnetwork(url)
self.assertEqual(subnetwork.name, name)
self.assertEqual(subnetwork.region.name, region_name)
```
and
```
def _projects_other_name_regions_us_central1_subnetworks_cf_972cf02e6ad49112(self, method, url, body, headers):
    body = self.fixtures.load(
        'regions_us-central1_subnetworks_cf_972cf02e6ad49112.json')
    return (httplib.OK, body, self.json_hdr, httplib.responses[httplib.OK])
```

(You need `_projects_other_name` in the method name, see GCEMockHttp._get_method_name in libcloud/test/compute/test_gce.py.)

And you will need to do the same thing for networks.

bchazalet:
you're right, I need to cover the shared network situation. I'll update the tests.

bchazalet:
I've added tests with networks and subnetworks that are declared under a different project, assuming that the 'loaded' project always is `project_name`.

pquentin:
Thanks! Merged in trunk. ✨

bchazalet:
thanks a lot @pquentin. Out of curiosity, when can we expect a new release that would include the changes?

pquentin:
@bchazalet I have no idea. What I know is that we have enough changes to warrant a new release. I can't start the release process myself, though.

On an unrelated note, I’d like to thank you again for your contribution to the libcloud project. As you might have noticed, we receive more pull requests that we can handle. If you would like to continue to help the project, reviewing one open pull request would be a great way to do so. But there’s no pressure: life is too short to spend volunteer time on things that you don’t find fulfilling.

PR #1167, opened 2018-01-25
Made pytest-runner optional

vaygr:
### Description

This fixes what was introduced by b4ce936: a forced dependency of the [pytest-runner](https://github.com/pytest-dev/pytest-runner) for any `setup.py` target, including `build` and `install`, which is not really appropriate, since it should be required only for testing, when you invoke `python setup.py test` or its aliases.

Such a forced dependency makes distribution packaging more complicated, since pytest-runner pulls other dependencies (like setuptools_scm) which may have other dependencies or Python version requirements for no reason.

It also breaks libcloud installation on older distros (e.g. Debian Wheezy), where libcloud **2.0.0** worked just perfectly, but anything above it couldn't because pytest-runner and/or its dependencies cannot be installed.

Plus, testing should always be optional.

Thankfully, there's an [official solution](https://github.com/pytest-dev/pytest-runner#conditional-requirement) to make sure pytest-runner is invoked when it's really needed.

### Status

- done, ready for review

- [x] Code linting

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -125,6 +125,9 @@ install_requires = ['requests']
 if PY2_pre_279:
     install_requires.append('backports.ssl_match_hostname')
 
+needs_pytest = {'pytest', 'test', 'ptr'}.intersection(sys.argv)
+pytest_runner = ['pytest-runner'] if needs_pytest else []
+
 setup(
     name='apache-libcloud',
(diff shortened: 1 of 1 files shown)


pquentin:

vaygr:
Appreciated.

PR #1170, opened 2018-02-02
Fix tests with pytest 3.4

pquentin:
### Description

The 3.4 pytest release introduced a breaking change: pytest no longer changes log levels: the default log level is now Python's default: WARNING. This commit ensures the test no longer depends on external log levels.

pytest 3.4 also adds a .pytest_cache folder: ignore it.

### Status

- done, ready for review

- [ ] Documentation

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,4 @@ include/
 lib/
 pip-selfcheck.json
 report.html
+.pytest_cache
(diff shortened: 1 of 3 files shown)


pquentin:
Note that the coverage is going down because we incorrectly compute coverage on only one Python 3 run instead of running codecov for every tox interpreter. But that should be fixed in another PR.

tonybaloney:
@pquentin pytest version should also be pinned in this PR

pquentin:
Thanks @tonybaloney, fixed. I'll update the pins of the other test requirements in another pull request if you don't mind.

tonybaloney:
:+1: Merging

PR #1171, opened 2018-02-07
Fix IndexError in dimensiondata list_images

tintoy:
### Description

This fix involves filtering out images from locations not returned by the list_locations API.

This is required because the CloudControl API returns all images in the target geographic region (even ones in datacenters the user's organisation does not have access to).

We therefore need to filter out those images (since we can't get a NodeLocation for them).

### Status

Done, ready for review.

- [ ] Documentation

### DIFF
--- a/libcloud/compute/drivers/dimensiondata.py
+++ b/libcloud/compute/drivers/dimensiondata.py
@@ -4181,7 +4181,7 @@ class DimensionDataNodeDriver(NodeDriver):
         if locations is None:
             locations = self.list_locations(location_id)
 
-        location = filter(lambda x: x.id == location_id, locations)[0]
+        location = [match_location for match_location in locations if match_location.id == location_id][0]
         cpu_spec = self._to_cpu_spec(element.find(fixxpath('cpu', TYPES_URN)))
 
         if LooseVersion(self.connection.active_api_version) > LooseVersion(
--- a/libcloud/compute/drivers/dimensiondata.py
+++ b/libcloud/compute/drivers/dimensiondata.py
@@ -4161,8 +4161,18 @@ class DimensionDataNodeDriver(NodeDriver):
         images = []
         locations = self.list_locations()
 
+        # The CloudControl API will return all images
+        # in the current geographic region (even ones in
+        # datacenters the user's organisation does not have access to)
+        #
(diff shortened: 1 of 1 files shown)


tintoy:
@tonybaloney, would you mind having a look over this to see if it looks correct to you? I've tested it locally, but another pair of eyes would be appreciated :)

tintoy:
More generally, it may be worth refactoring `_to_image` to take a specific location, rather than a list of locations, and for the driver to internally cache the list of locations for situations where up-to-date locations are not an issue (e.g. unless you need to know datacenter status). But I figured that's a more extensive PR :)

tonybaloney:
@tintoy the `filter` method returns a list in Python 2 and an iterator in Python 3 (IList vs IEnumerable) and so cannot be indexed https://travis-ci.org/apache/libcloud/jobs/338753055#L979

Workarounds are: Don't use filter. Use a list comprehension
OR, convert to list around the filter response, e.g. list(filter(x))

tintoy:
I considered list comprehensions but I was worried the double-square-brackets would look weird. But yeah that makes sense; will fix now.

tonybaloney:
the filter method is out of fashion nowadays. I recommend the list comprehension. Plate du jour

tintoy:
Done :)

tintoy:
Personally I only target Python 3.x if I can help it, so when I'm contributing to projects that also have to target 2.x I sometimes forget this stuff :-D

tintoy:
I kinda wish Python had an easily-accessible form of `FirstOrDefault()` - you can use something like `next(xxx)` but then you have to deal with `StopIteration` :-/

tintoy:
Oops, sorry, should have noticed the line length thing; I had a wisdom tooth removed this morning and am still a little out of it. Third time's the charm :)

tonybaloney:
@tintoy tonybaloney/retox :-)

tintoy:
Heh - yeah sorry I normally run the tests (and I think retox is super cool in this regard), just a little vague this arvo and forgot :blush:

tonybaloney (approved):
👍 thanks Adam

tintoy:
Thanks Ant 😀

pquentin:
Thanks! Merged in trunk. ✨

PR #1179, opened 2018-03-08
Allow shared vpc in managed instance group creation

bchazalet:
### Description

It's currently not possible to create a Managed Instance Group with an instance template that creates instances in a network and subnetwork that are shared from a different project.

The problem seems to be that, when parsing the json representing the instance group (in `_to_instancegroup`), the name part of the network (or subnetwork) is extracted from a fully qualified URL, i.e. implictly assuming that the network is under the same project that we are in. But if we're using a shared VPC, the network resource will be under a different project and we must stick to using fully qualified URL.

This PR modifies the `_to_instancegroup` method so that the network and subnetwork values contained in the json representation are used as they are and passed directly to `ex_get_network` and `ex_get_subnetwork` which know how to handle fully qualified resource URLs.

### Status

Done, ready for review. To be honest I am not sure the test I've added is useful, but I am not too sure about how to test this: I'll happily receive feedback.

- [ ] Documentation

### DIFF
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -9025,13 +9025,11 @@ class GCENodeDriver(NodeDriver):
         # does not contain instances.
         network = instancegroup.get('network', None)
         if network:
-            obj_name = self._get_components_from_path(network)['name']
-            network = self.ex_get_network(obj_name)
+            network = self.ex_get_network(network)
 
         subnetwork = instancegroup.get('subnetwork', None)
         if subnetwork:
-            parts = self._get_components_from_path(subnetwork)
-            subnetwork = self.ex_get_subnetwork(parts['name'], parts['region'])
+            subnetwork = self.ex_get_subnetwork(subnetwork)
 
         return GCEInstanceGroup(
             id=instancegroup['id'], name=instancegroup['name'], zone=zone,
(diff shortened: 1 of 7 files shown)


vdloo:
There's a linting issue here:
```
libcloud/test/compute/test_gce.py:3565:57: E128 continuation line under-indented for visual indent
```

The other failing tests are unrelated, see https://github.com/apache/libcloud/pull/1180

pquentin:
@bchazalet Thanks! It would be nice to fix the lint and rebase on top of trunk, it would help me see the coverage changes.

Also, thanks for the test. Do you think you could write a test that would fail without your changes?

bchazalet:
I've modified the test so that it would fail without the fix. Although, since we're mocking the network bit it'd fail with a mock error, e.g. `AttributeError: 'GCEMockHttp' object has no attribute '_global_networks_shared_network_for_mig'`.

In reality, and for future reference, it'd fail because it can't find the resource (because it's looking for it in the wrong project), e.g. :

`An exception occurred during task execution. To see the full traceback, use -vvv. The error was: libcloud.common.google.ResourceNotFoundError: {u'domain': u'global', u'message': u"The resource 'projects/project_name/regions/europe-west2/subnetworks/shared_network' was not found", u'reason': u'notFound'}`

pquentin:
Thanks! Merged in trunk. ✨

PR #1180, opened 2018-03-08
pin libvirt-python to 4.0.0

vdloo:
version was not constrained and the latest update broke the build because the libvirtmod extension failed to compile.

> According to https://bugs.launchpad.net/openstack-requirements/+bug/1753539, pinning libvirt-python to 4.0.0 should fix our issue.
> Do you think you could open a pull request wich replaces all libvirt-python occurrences to libvirt-python==4.0.0 in tox.ini? That would be really helpful. Thanks

\- @pquentin


pquentin:
Thanks! Merged into trunk. ✨

PR #1181, opened 2018-03-13
[LIBCLOUD-985] Add coverage for newer GCE network and subnet options.

maxlip:
### Description

Addressing [LIBCLOUD-985](https://issues.apache.org/jira/browse/LIBCLOUD-985), PR adds the ability to handle a few newer GCE network and subnet options upon creation.
For networks, supports BGP routing mode
For subnetworks, supports secondary IP Ranges, Private IP access to GCP APIs

See [](https://cloud.google.com/compute/docs/reference/rest/v1/networks/insert) and [https://cloud.google.com/compute/docs/reference/rest/v1/subnetworks/insert](url)

### Status

- done, ready for review

- [x] Documentation

### DIFF
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -3775,8 +3775,7 @@ class GCENodeDriver(NodeDriver):
         :param  secondaryipranges: List of dicts of secondary or "alias" IP
                                    ranges for this subnetwork in
                                    [{"rangeName": "second1",
-                                   "ipCidrRange": "192.168.168.0/24"},
-                                   {k:v, k:v}] format.
+                                   "ipCidrRange": "192.168.168.0/24"} format.
         :type   secondaryipranges: ``list`` of ``dict`` or ``None``
 
         :return:  Subnetwork object
@@ -3870,13 +3869,14 @@ class GCENodeDriver(NodeDriver):
         else:
             network_data['autoCreateSubnetworks'] = (mode.lower() == 'auto')
 
-        if routing_mode.lower() not in ['regional', 'global']:
-            raise ValueError("Invalid Routing Mode: '%s'. Must be 'REGIONAL', "
-                             "or 'GLOBAL'." % routing_mode)
-        else:
-            network_data['routingConfig'] = {
-                'routingMode': routing_mode.upper()
-            }
+        if routing_mode:
+            if routing_mode.upper() not in ['REGIONAL', 'GLOBAL']:
+                raise ValueError("Invalid Routing Mode: '%s'. Must be "
+                                 "'REGIONAL' or 'GLOBAL'." % routing_mode)
+            else:
+                network_data['routingConfig'] = {
+                    'routingMode': routing_mode.upper()
+                }
 
         request = '/global/networks'
 
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -3775,7 +3775,8 @@ class GCENodeDriver(NodeDriver):
         :param  secondaryipranges: List of dicts of secondary or "alias" IP
                                    ranges for this subnetwork in
                                    [{"rangeName": "second1",
-                                   "ipCidrRange": "192.168.168.0/24"} format.
+                                   "ipCidrRange": "192.168.168.0/24"},
+                                   {k:v, k:v}] format.
         :type   secondaryipranges: ``list`` of ``dict`` or ``None``
 
         :return:  Subnetwork object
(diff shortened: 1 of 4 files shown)


libcloud/compute/drivers/gce.py
```diff
@@ -3843,15 +3863,23 @@ def ex_create_network(self, name, cidr, description=None, mode="legacy"):
...
-        request = '/global/networks'
-
         if mode == 'legacy':
             if not cidr:
                 raise ValueError("Must specify IPv4Range with 'legacy' mode.")
             network_data['IPv4Range'] = cidr
         else:
             network_data['autoCreateSubnetworks'] = (mode.lower() == 'auto')
 
+        if routing_mode.lower() not in ['regional', 'global']:
```
pquentin: I think this will fail when `routing_mode` is `None` (default value)?

libcloud/test/compute/test_gce.py
```diff
@@ -983,10 +991,12 @@ def test_ex_create_route(self):
     def test_ex_create_network(self):
         network_name = 'lcnetwork'
         cidr = '10.11.0.0/16'
-        network = self.driver.ex_create_network(network_name, cidr)
+        routing_mode = 'REGIONAL'
+        network = self.driver.ex_create_network(network_name, cidr, routing_mode='regional')
```
pquentin: Maybe we should still test the default value too? This would have avoided the bug above.

libcloud/compute/drivers/gce.py
```diff
@@ -3767,6 +3768,17 @@ def ex_create_subnetwork(self, name, cidr=None, network=None, region=None,
...
 
+        :param  privateipgoogleaccess: Allow access to Google services without
+                                       assigned external IP addresses.
+        :type   privateipgoogleaccess: ``bool` or ``None``
+
+        :param  secondaryipranges: List of dicts of secondary or "alias" IP
+                                   ranges for this subnetwork in
+                                   [{"rangeName": "second1",
+                                   "ipCidrRange": "192.168.168.0/24"},
+                                   {k:v, k:v}] format.
```
pquentin: Do we need to specify `{k:v, k:v}`? Just making sure it's intentional.

pquentin (changes requested):
I think there's a bug when routing_mode is None, but otherwise this looks good!

libcloud/test/compute/test_gce.py
maxlip: Maybe test both (realized I can also add test coverage for "description" field while I'm at it:
```python
    def test_ex_create_network(self):
        network_name = 'lcnetwork'
        cidr = '10.11.0.0/16'
        description = 'A custom network'
        routing_mode = 'REGIONAL'

        # Test defaults
        network = self.driver.ex_create_network(network_name, cidr)
        self.assertTrue(isinstance(network, GCENetwork))
        self.assertEqual(network.name, network_name)
        self.assertEqual(network.cidr, cidr)

        # Test using more options
        network = self.driver.ex_create_network(network_name, cidr,
                                                description=description,
                                                routing_mode=routing_mode)
        self.assertEqual(network.description, description)
        self.assertEqual(network.extra['routingConfig']['routingMode'], routing_mode)
```

Just saying list of dicts is enough for the subnetwork docstring, I guess.

libcloud/compute/drivers/gce.py
maxlip: I didn't get failures in testing w/o a value (would end up passing blank routingConfig{} data), but how about I just rejigger this a bit by only triggering logic if routing_mode is specified?
```python
        if routing_mode:
            if routing_mode.lower() not in ['regional', 'global']:
                raise ValueError("Invalid Routing Mode: '%s'. Must be "
                                 "'REGIONAL' or 'GLOBAL'." % routing_mode)
            else:
                network_data['routingConfig'] = {
                    'routingMode': routing_mode.upper()
                }
```

libcloud/test/compute/test_gce.py
pquentin: Sounds good!

libcloud/compute/drivers/gce.py
pquentin: The change looks good, but you're saying it's unnecessary? I don't understand, if routing_mode is None, `routing_mode.lower()` is going to raise an exception, right? Am I missing something?

libcloud/compute/drivers/gce.py
```diff
@@ -3775,8 +3775,7 @@ def ex_create_subnetwork(self, name, cidr=None, network=None, region=None,
         :param  secondaryipranges: List of dicts of secondary or "alias" IP
                                    ranges for this subnetwork in
                                    [{"rangeName": "second1",
-                                   "ipCidrRange": "192.168.168.0/24"},
-                                   {k:v, k:v}] format.
+                                   "ipCidrRange": "192.168.168.0/24"} format.
```
pquentin: Sorry if I was not clear, I was not asking you to change this!

pquentin (approved):
Thanks! Will merge when I get the chance.

pquentin:
Thanks! Merged in trunk. ✨

PR #1191, opened 2018-03-19
implement attach/detach floating IPs for DO

vdloo:
add methods to the DigitalOcean_v2_NodeDriver to attach and detach floating IPs from instances uniform to the `ex_attach_floating_ip_to_node` and `ex_detach_floating_ip_from_node` methods from the [OpenStack_1_1_NodeDriver](https://github.com/apache/libcloud/blob/2b298b7cd8d17b4b1c0db20292fdb7867a512ac6/libcloud/compute/drivers/openstack.py#L2394)

see https://developers.digitalocean.com/documentation/v2/#assign-a-floating-ip-to-a-droplet and https://developers.digitalocean.com/documentation/v2/#unassign-a-floating-ip

cc @AlexanderGrooff


libcloud/test/compute/test_digitalocean_v2.py
```diff
@@ -518,6 +534,20 @@ def _v2_floating_ips_167_138_123_111(self, method, url, body, headers):
...
+        else:
+            raise NotImplementedError()
+
+    def _v2_floating_ips_154_138_103_175_actions(self, method, url, body, headers):
+        if method == 'POST':
+            body = self.fixtures.load('detach_floating_ip.json')
+            return (httplib.CREATED, body, {}, httplib.responses[httplib.CREATED])
+        else:
+            raise NotImplementedError()
+
```
pquentin, line 550: Out of scope for this pull request, but I'm not a huge fan of this style because the raise path is never exercised and lowers the coverage. I think I'd prefer something like `assert method in ['GET', 'POST']` at the beginning of the function. Not sure it's worth the trouble, and for now consistency is better!

pquentin:
Thanks, merged in trunk! ✨

PR #1193, opened 2018-03-21
LIBCLOUD-986: Add prefix option to azure blob list and iterate containers

jhawkinson:
### Description
While testing out the libcloud storage abstraction I've come across a curious omission in driver support. It seems that for many drivers there is additional support for the list_container_objects() and iterate_container_objects() to support a prefix (or ex_prefix). 

This functionality is not present in the Azure driver for some reason.   I've gone and checked the list_blobs API and it also supports this functionality as seen here:

https://docs.microsoft.com/en-us/rest/api/storageservices/list-blobs

I've added the support to the AzureBlobsStorageDriver with these changes.

### Status
- done, ready for review

### Checklist
- [ ] Documentation

### DIFF
--- a/libcloud/storage/drivers/azure_blobs.py
+++ b/libcloud/storage/drivers/azure_blobs.py
@@ -381,11 +381,12 @@ class AzureBlobsStorageDriver(StorageDriver):
             if not params['marker']:
                 break
 
-    def iterate_container_objects(self, container):
+    def iterate_container_objects(self, container, ex_prefix=None):
         """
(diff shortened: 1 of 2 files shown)


pquentin:
Thank you for your contribution @jhawkinson! Merged in trunk, this will be part of the next release. ✨

pquentin:
Can you please close the https://issues.apache.org/jira/browse/LIBCLOUD-986? I unfortunately can't do that myself.

PR #1195, opened 2018-03-27
Guess s3 multipart upload content type

iurisilvio:
### Description

I added mimetype guessing to S3 multipart upload.

https://issues.apache.org/jira/browse/LIBCLOUD-958

### Status

Done, ready for merge.

- [ ] Documentation


libcloud/test/storage/test_s3.py
```diff
@@ -905,6 +907,29 @@ def test_upload_big_object_via_stream(self):
         self.assertEqual(obj.name, object_name)
         self.assertEqual(obj.size, CHUNK_SIZE * 3)
 
+    def test_upload_object_via_stream_guess_file_mime_type(self):
+        def dummy_content_type(name):
+            return 'application/zip', None
+
+        old_func = libcloud.utils.files.guess_file_mime_type
+        libcloud.utils.files.guess_file_mime_type = dummy_content_type
```
pquentin: Can you please use monkeypatch.setattr from pytest? It's more robust: for example, if the test fails before entering the `try` block, we will have modified `libcloud.utils.files.guess_file_mime_type` for the other tests too.

libcloud/test/storage/test_s3.py
```diff
@@ -905,6 +907,29 @@ def test_upload_big_object_via_stream(self):
...
+            self.mock_response_klass.type = None
+
+        container = Container(name='foo_bar_container', extra={},
+                              driver=self.driver)
+        object_name = 'foo_test_stream_data'
+        iterator = BytesIO(b('234'))
+        try:
+            self.driver.upload_object_via_stream(container=container,
+                                                 object_name=object_name,
+                                                 iterator=iterator)
```
pquentin: non blocking nitpick: Is there a way to check that the correct content type was used during the upload?

pquentin (changes requested):
Looks good, the only change to make is the `libcloud.utils.files.guess_file_mime_type` monkeypatch.

libcloud/test/storage/test_s3.py
iurisilvio: I tried, but `upload_object_via_stream` don't return anything useful to validate it. I'd have to mock `_initiate_multipart` to assert headers or extract all header generation to another method.

libcloud/test/storage/test_s3.py
iurisilvio: Changed to use Python `mock.patch`.

libcloud/test/storage/test_s3.py
iurisilvio: I changed this test to use a mock. I added a check to assert if my mock function is called. Not perfect, but at least we know the code is trying to guess the mimetype.

pquentin: Great, thanks! Another non-blocking nitpick that really you don't have to follow: can we try to be more resilient to future typos here? See https://engineeringblog.yelp.com/2015/02/assert_called_once-threat-or-menace.html. I think using autospec would work in this case?

pquentin (approved):
Thanks! Will merge when I get the chance.

libcloud/test/storage/test_s3.py
iurisilvio: Great article. I had mock assert issues in the past, I know how it is painful to debug. Added `autospec=True` and it works fine. :+1:

pquentin:
I appreciate that you took the time to improve your patch multiple times. Thanks! Merged in trunk. ✨

PR #1201, opened 2018-04-16
Update docstring for storage provider class

c-w:
### Description

The docstring listing all the class variables in the storage provider class went out of sync with the actual class variables. This commit fixes that by removing docstring entries for class variables that no longer exist and by adding docstring entries for class variables that have been added.

### Status

Done, ready for review.

- [x] Documentation

### DIFF
--- a/libcloud/storage/types.py
+++ b/libcloud/storage/types.py
@@ -40,19 +40,33 @@ class Provider(object):
     :cvar DUMMY: Example provider
     :cvar ALIYUN_OSS: Aliyun OSS storage driver
     :cvar AURORAOBJECTS: AuroraObjects storage driver
+    :cvar AZURE_BLOBS: Azure Blob Storage driver
+    :cvar BACKBLAZE_B2: Backblaze B2 Cloud Storage driver
     :cvar CLOUDFILES: CloudFiles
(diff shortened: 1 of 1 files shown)


pquentin:
Thanks! Do you know if this list appears in the documentation? I think the important list to maintain is https://libcloud.readthedocs.io/en/latest/storage/supported_providers.html, and we should try to avoid others if possible?

c-w:
Hi @pquentin, thanks for the reply. That list indeed looks up to date and in sync with the changes suggested in this pull request.

I believe there still might be value in also having the class docstring since this enables getting context about all the supported providers in an IDE such as PyCharm via auto-completion and go-to-definition.

If you wish to avoid maintaining two lists of supported providers, maybe it would be worth removing the listing in the docstring and instead add a link to the official documentation instead?

pquentin:
@c-w I don't use AWS storage, so that's your call! I'd be happy to merge this, or replace by a link. Please tell me.

c-w:
@pquentin If we could merge it as is, that would be great. Would really help with discoverability of all the storage providers in IDEs (e.g. I was just exploring the library and auto-completion didn't tell me that there's support for Azure Blobs). Thanks!

pquentin:
Thanks! Merged in trunk. ✨

PR #1202, opened 2018-04-16
Fix hash value in azure blob store

daviskirk:
### Description

* Upgrade to minimum api version allowed by azure (see )

* Return actual file (md5) hash instead of etag ('"0xFFFFFFFFFFFFFFF"')

The current trunk version of libcloud throws a Libcloud error also described in (https://issues.apache.org/jira/browse/LIBCLOUD-851)

```
  File "...", line 12, in <module>
    c = d.list_containers()[1]
  File "libcloud/libcloud/storage/base.py", line 209, in list_containers
    return list(self.iterate_containers())
  File "libcloud/libcloud/storage/drivers/azure_blobs.py", line 371, in iterate_containers
    (response.status), driver=self)
libcloud.common.types.LibcloudError: <LibcloudError in <libcloud.storage.drivers.azure_blobs.AzureBlobsStorageDriver object at 0x7f7389bf1080> 'Unexpected status code: 400'>
```
and the corresponing response body of the failing request is:
```
response.body
'<?xml version="1.0" encoding="utf-8"?><Error><Code>InvalidHeaderValue</Code><Message>The value for one of the HTTP headers is not in the correct format.\nRequestId:dd262993-501e-0030-6cd1-d58682000000\nTime:2018-04-16T22:24:39.6566784Z</Message><HeaderName>x-ms-version</HeaderName><HeaderValue>2012-02-12</HeaderValue></Error>'
```

From looking at:
https://docs.microsoft.com/en-us/rest/api/storageservices/versioning-for-the-azure-storage-services#for-blob-storage-accounts

the easiest fix is to upgrade the api version to the minumum supported version (2014-02-12).

In addition, the other cloud drivers all provide the actual md5 hash when uploading or getting an object.  This allows the user to check if a file has changed online compared to locally.  The current master version provides the etag value as the "hash" attribute (interestingly with literal quotes around the tag: `'"0xFFFFFFFFFFFFFFF"'`) which is not compatible with the other drivers.

Replace this with the PR description (mention the changes you have made, why
you have made them, provide some background and any references to the provider
documentation if needed, etc.).

I am assuming that these are both bugs and not design decisions (if so feel free to close this request).

### Status

- done, ready for review

- [ ] Documentation

### DIFF
--- a/libcloud/common/azure.py
+++ b/libcloud/common/azure.py
@@ -32,7 +32,7 @@ from libcloud.common.base import CertificateConnection
 from libcloud.common.base import XmlResponse
 
 # Azure API version
-API_VERSION = '2014-02-14'
+API_VERSION = '2017-07-29'
 
 # The time format for headers in Azure requests
 AZURE_TIME_FORMAT = '%a, %d %b %Y %H:%M:%S GMT'
--- a/libcloud/common/azure.py
+++ b/libcloud/common/azure.py
@@ -32,7 +32,7 @@ from libcloud.common.base import CertificateConnection
 from libcloud.common.base import XmlResponse
 
 # Azure API version
-API_VERSION = '2012-02-12'
+API_VERSION = '2014-02-14'
 
 # The time format for headers in Azure requests
 AZURE_TIME_FORMAT = '%a, %d %b %Y %H:%M:%S GMT'
--- a/libcloud/storage/drivers/azure_blobs.py
+++ b/libcloud/storage/drivers/azure_blobs.py
@@ -301,7 +301,7 @@ class AzureBlobsStorageDriver(StorageDriver):
         for meta in metadata.getchildren():
             meta_data[meta.tag] = meta.text
 
-        return Object(name=name, size=size, hash=etag, meta_data=meta_data,
+        return Object(name=name, size=size, hash=extra['md5_hash'], meta_data=meta_data,
                       extra=extra, container=container, driver=self)
 
     def _response_to_object(self, object_name, container, response):
@@ -353,7 +353,7 @@ class AzureBlobsStorageDriver(StorageDriver):
                 key = key.split('x-ms-meta-')[1]
                 meta_data[key] = value
 
-        return Object(name=object_name, size=size, hash=etag, extra=extra,
+        return Object(name=object_name, size=size, hash=extra['md5_hash'], extra=extra,
                       meta_data=meta_data, container=container, driver=self)
 
     def iterate_containers(self):
@@ -939,7 +939,7 @@ class AzureBlobsStorageDriver(StorageDriver):
                 object_name=object_name, driver=self)
 
         return Object(name=object_name, size=bytes_transferred,
-                      hash=headers['etag'], extra=None,
+                      hash=server_hash, extra=None,
                       meta_data=meta_data, container=container,
                       driver=self)
 
--- a/libcloud/storage/drivers/azure_blobs.py
+++ b/libcloud/storage/drivers/azure_blobs.py
@@ -301,8 +301,9 @@ class AzureBlobsStorageDriver(StorageDriver):
         for meta in metadata.getchildren():
             meta_data[meta.tag] = meta.text
 
-        return Object(name=name, size=size, hash=extra['md5_hash'], meta_data=meta_data,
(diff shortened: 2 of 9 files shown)


pquentin:
Thanks for this!

Did you test that the other methods continue to work with this API version? Do you think upgrading to a more recent version would work too?

daviskirk:
@pquentin no, I'm relying on the tests for that.  Although seeing that the tests rely on saved responses I see your point.  Is there any sort of integration tests that are triggerable that run the "real" api.
Also is there any documentation as to how to generate the responses that are saved and used in the test?  I can imagine that some have them have changed over the years (similar to the object storage api here where all tests passed before this pull request but the real world api broke anyway).

daviskirk:
I now upgraded to the most recent api version and checked the object storage tests against actual api calls which all work as expected (I've also upgraded the response fixtures to reflect the response structures I found running against the live api).  However, I have not tested the compute azure / azure arm drivers with real api calls.

pquentin:
Thanks! So you tested the storage driver, but not the compute driver, right?

To trigger the real API, there's the secrets.py file, but it does not appear to be supported by the storage tests. I don't know if there's a way to automatically save the actual answers, but I don't think there is.

@tetron @ldipenti @mermoldy Do you think you could help us to make sure that the Azure compute driver still works with the new API version? Thanks!

daviskirk:
@pquentin yes, that is exactly right.

tetron:
The Azure compute driver specifies explicit API version for each call, since API versions are by component and not uniform across all of Azure.  So a change in the API version for storage shouldn't affect the Azure ARM driver.  I can't speak for the older service manager driver, however.

tetron:
We should go test it, though.

Kami:
@c-w Maybe you or one of your colleagues can help review (& verify) this change :)

c-w:
@michaelperel @jmspring @cicorias Could one of you please review this PR and verify the fix? I can also take a look towards mid next week.

c-w:
@daviskirk I'm seeing some authentication errors with your branch running against a live storage account:

```
  File "/home/vsts/work/1/s/tests/test_storage.py", line 120, in test_objects_stream_io
    self._test_objects(do_upload, do_download)
  File "/home/vsts/work/1/s/tests/test_storage.py", line 67, in _test_objects
    container = self.driver.create_container(_random_container_name())
  File "/opt/hostedtoolcache/Python/2.7.15/x64/lib/python2.7/site-packages/libcloud/storage/drivers/azure_blobs.py", line 516, in create_container
    method='PUT')
  File "/opt/hostedtoolcache/Python/2.7.15/x64/lib/python2.7/site-packages/libcloud/common/base.py", line 637, in request
  File "/opt/hostedtoolcache/Python/2.7.15/x64/lib/python2.7/site-packages/libcloud/common/base.py", line 152, in __init__
    message=self.parse_error(),
  File "/opt/hostedtoolcache/Python/2.7.15/x64/lib/python2.7/site-packages/libcloud/common/azure.py", line 84, in parse_error
    raise InvalidCredsError(error_msg)
InvalidCredsError: 'AuthenticationFailed: Server failed to authenticate the request. Make sure the value of Authorization header is formed correctly including the signature.'
```

You can see the full test run on [Azure Pipelines](https://clewolff.visualstudio.com/libcloud-tests/_build/results?buildId=40) and the code for the tests at [c-w/libcloud-tests](https://github.com/c-w/libcloud-tests).

c-w:
Doing a bit more investigation into this: libcloud `2.5.0` works with storage accounts where the `kind` field is set to `Storage` or `StorageV2` but not `BlobStorage`. I'm hypothesizing that the changes in this pull request fix the `BlobStorage` scenario but may affect the `Storage` and `StorageV2` scenarios.

c-w:
In order to reduce the risk of this pull request, I've pulled out the changes related to the API version update into a separate PR (see https://github.com/apache/libcloud/pull/1294). I've also modified the code such that the API version update only affects the storage driver so that we can be sure not to affect other Azure-related drivers.

Once the API version update is merged, this pull request can be reviewed and merged independently for the hash value fix.

Kami:
This has been merged into trunk via #1294.

@daviskirk thanks again for your contribution and sorry for the delay.

PR #1205, opened 2018-05-01
[ec2] Add subnet modification method

danhunsaker:
What it says on the tin...

### DIFF
--- a/libcloud/compute/drivers/ec2.py
+++ b/libcloud/compute/drivers/ec2.py
@@ -2796,6 +2796,41 @@ class BaseEC2NodeDriver(NodeDriver):
 
         return subnet
 
+    def ex_modify_subnet_attribute(self, subnet, attribute='auto_public_ip',
+                                   value=False):
+        """
+        Modifies a subnet attribute.
+        You can only modify one attribute at a time.
+
+        :param      subnet: The subnet to delete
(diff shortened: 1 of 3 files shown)


pquentin:
Thanks! This looks good. We're going to need a test, do you think you could add one?

danhunsaker:
Oops. Totally forgot to write one. I'll get one in as soon as I have a chance.

libcloud/compute/drivers/ec2.py
```diff
@@ -2796,6 +2796,41 @@ def ex_create_subnet(self, vpc_id, cidr_block,
...
+        :rtype:     ``bool``
+        """
+        params = {'Action': 'ModifySubnetAttribute', 'SubnetId': subnet.id}
+
+        if attribute == 'auto_public_ip':
+            params['MapPublicIpOnLaunch.Value'] = value
+        elif attribute == 'auto_ipv6':
+            params['AssignIpv6AddressOnCreation.Value'] = value
+        else:
+            return False
```
Kami, line 2828: I think it would be better to throw in this case - e.g. ``raise ValueError('Unsupported attribute: %s' % (attribute))``.

Kami:
Thanks for adding the test case and sorry for the delay.

I made a small change (2085972af239fc91bfc3d7e06898038e5b0b6bd6) and merged changes into trunk.

PR #1206, opened 2018-05-01
Fix: DigitalOcean volume list can be None

jdesfossez:
The returned structure seems to have changed for DigitalOcean and the
data can now be None, which means we have to check before using the
result in the map().

Signed-off-by: Julien Desfossez <jdesfossez@digitalocean.com>

### DIFF
--- a/libcloud/compute/drivers/digitalocean.py
+++ b/libcloud/compute/drivers/digitalocean.py
@@ -125,6 +125,8 @@ class DigitalOcean_v2_NodeDriver(DigitalOcean_v2_BaseDriver,
 
     def list_volumes(self):
         data = self._paginated_request('/v2/volumes', 'volumes')
+        if data is None:
+            return []
         return list(map(self._to_volume, data))
 
     def create_node(self, name, size, image, location, ex_create_attr=None,


vdloo:
Hi @jdesfossez, thanks for the fix. 

We noticed that our garbage collector started crashing on this since the 26th of April. I could not find any mention of this change in the DigitalOcean changelog. Do you know if we can expect more unannounced (potentially breaking) changes in the DigitalOcean API?

before:
```
In [2]: conn = get_connection('digitalocean')

In [3]: conn.list_volumes()
TypeError                                 Traceback (most recent call last)
<ipython-input-3-f087464ec811> in <module>()
----> 1 conn.list_volumes()

/usr/local/venv/hypernode-control/src/apache-libcloud/libcloud/compute/drivers/digitalocean.pyc in list_volumes(self)
    126     def list_volumes(self):
    127         data = self._paginated_request('/v2/volumes', 'volumes')
--> 128         return list(map(self._to_volume, data))
    129 
    130     def create_node(self, name, size, image, location, ex_create_attr=None,

TypeError: argument 2 to map() must support iteration
```

after:
```
In [1]: from hypernode.provisioning.node_operations import get_connection

In [2]: conn = get_connection('digitalocean')

In [3]: conn.list_volumes()
Out[3]: []
```

Could you add a test to the test-suite to reflect this new behavior? You could add something like this to test_digitalocean_v2.py:

after test_list_volumes_none:
```
    def test_list_volumes_empty(self):
        volumes = self.driver.list_volumes()
        self.assertEqual(len(volumes), 0)
```

and after _v2_volumes_NONE add _v2_volumes_NONE with a fixture like:
```
{
  "volumes": None,
  "links": {},
  "meta": {
    "total": 0
  }
}
```

andrewsomething:
Hi all,

This was brought to my attention internally at DigitalOcean, and I want to apologize for any inconvenience. This was an inadvertent change to our API that has now been reverted. We're committed to providing a stable API. If there ever arises a need for a breaking change, it will be communicated in advance as part of a deprecation process. Our API changelog can be found here: https://developers.digitalocean.com/documentation/changelog/

I'm a contributor to libcloud myself, so please feel free to tag me on any questions related to our API.

Sorry again for the disruption to your work.

pquentin:
@jdesfossez @vdloo Can you please confirm than the DigitalOcean fix work for you?

pquentin:
@andrewsomething Thanks! It's great to see such a fast response from DigitalOcean.

jdesfossez:
Yes, I confirm it works !
Thanks @andrewsomething !

vdloo:
Yes, it works. Thanks for the quick fix @andrewsomething 

```
In [2]: conn = get_connection('digitalocean')

In [3]: conn.list_volumes()
Out[3]: []
```

pquentin:
Great, thanks. @jdesfossez, please close the pull request when you get the chance. :)

PR #1207, opened 2018-05-01
Support for AWS EC2 Spot Instances

ldipenti:
### Description

Spot instances on AWS are a way to get cheaper nodes taking advantage of Amazon's over capacity. The downside of using this kind of instances is that whenever AWS need the allocated spot instances to deliver normal (on-demand) ones (at full price), it will shutdown them after a 2 minute period, so this has to be taken into account when designing systems so they can recover correctly.

Recently, AWS was updated so that requesting Spot instances is easier than ever, the client just to add a couple of parameters to the usual RunInstances call, immediately getting an instance id from the API that can be treated as a normal Node.

For testing purposes I added a new example script that creates t2.micro spot instances.

Ref:
* https://aws.amazon.com/blogs/compute/new-amazon-ec2-spot-pricing/
* https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_InstanceMarketOptionsRequest.html

### Status

- done, ready for review

- [X] Documentation

### DIFF
--- a/example_compute_ec2spot.py
+++ b/example_compute_ec2spot.py
@@ -1,0 +1,52 @@
+import os
+import time
+from libcloud.compute.types import Provider
+from libcloud.compute.providers import get_driver
+from libcloud.compute.base import NodeImage
+from libcloud.compute.types import NodeState
+
+SIZE_ID = 't2.micro'
+AMI_ID = 'ami-4e79ed36'
+REGION = 'us-west-2'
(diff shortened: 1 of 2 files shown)


cure:
Ping, can we get this merged?

yukw777:
We can close this as https://github.com/apache/libcloud/pull/1398 was merged.

Kami:
Thanks for the contribution. Closing since #1398 was merged.

PR #1208, opened 2018-05-09
Update Azure Blobs Storage driver documentation

c-w:
### Description

The current screenshots in the Azure Blobs Storage driver documentation show an old version of the Azure portal. Some of the terminology and where to find access keys has changed, so this commit updates the screenshots and the wording in the documentation.

Additionally, Azure now also offers a new type of storage account called "Blob storage" which is not supported by the current libcloud driver implementation. This change therefore adds wording to avoid user confusion when creating a storage account for use with libcloud.

### Status

- done, ready for review

### Checklist

- [x] Documentation


pquentin:
Thanks! Merged in trunk. ✨

PR #1209, opened 2018-05-10
added AWS c5.* m5.* instance types to all appropriate regions

yanowitz:
### Description

Since the original c5.* and m5.* were added to libcloud, they have become available in more regions.

There were some tests that asserted the length of various lists of instance types. In general, I've found these kinds of tests are not helpful because they simply test that the code says what the code says (and hence are quite brittle). I'm not sure there is a good test (beyond successful compilation) that a static data structure is correctly written.

(I'm also not clear on why the library maintains these static lists instead of querying AWS for the valid values, but I assume there's a good reason for that).

### Status

- done, ready for review

- [ ] Documentation


pquentin:
@ZuluPro can you please review this change?

ZuluPro:
Hello @pquentin @yanowitz 
You've seen I'm on the case ;)

To avoid to update listing manually, I created this package https://github.com/ZuluPro/python-ec2-sizes.
It parses the AWS pricing JSON (>200MB) and stores it as pickles, `dict`s and JSONs.

Do you think we could integrate it to libcloud (as third package or in contrib) ?

pquentin:
@ZuluPro It would make sense to do this before each release, and allow users to do it manually if they want to, yes. Is this what you were thinking about?

ZuluPro:
Hey @pquentin 
Yep good idea !

ZuluPro:
@pquentin @alex Any news ?

pquentin:
@yanowitz We appreciate the pull request! But we're going to merge #1211 instead of this so that you never have to do this kind of pull requests again. :) Thanks.

yanowitz:
That's a much better approach!

PR #1211, opened 2018-05-20
Added script to get automatically EC2 instance sizes

ZuluPro:
Added script to get automatically EC2 instance sizes

### Description

- I created a script to download & parse the big JSON from AWS
This script outputs Python constants in the same format than in `ec2.py`
I placed the output in libcloud/compute/constants.py (maybe bad?)
- I cleaned `ec2.py` from the big dicts configuring sizes and region
I import from constants and just add things not related to AWS (nimbus)
- Updated tests
Tests were too static, so I deleted some.
I also disable lint on the constants file as it is autogenerated and prettified
I remove some lint config from tox.ini to setup.cfg.

### Status

Done

- [x] Documentation


ZuluPro:
@pquentin @alex @yanowitz
What do you think about it ?

It is related to the #1209
And all the ones I did for upgrade AWS instance types & regions.

pquentin:
Sounds good, however I did not take the time to review this in depth yet, sorry. Not sure when it will happen. Other libcloud committers should feel free to look at this!

ZuluPro:
@tonybaloney What do you think about ?

contrib/scrap-ec2-sizes.py
```diff
@@ -0,0 +1,271 @@
...
+Use it as following:
+    $ python scrap-ec2-sizes.py > ../libcloud/compute/constants.py
+"""
+
+import re
+import os
+import json
+
+import requests
+import ijson
```
pquentin: Note that by default ijson uses the Python backend which does not stream memory: https://github.com/isagalaev/ijson/issues/60. Since you did not notice it, I think using json is acceptable here. :)

contrib/scrap-ec2-sizes.py
```diff
@@ -0,0 +1,271 @@
...
+    # Set vars
+    sizes = {}
+    regions = {r['id']: r for r in REGION_DETAILS.values()}
+    for region_id in regions:
+        regions[region_id]['instance_types'] = []
+    # Parse
+    json_file = get_json()
+    products_data = ijson.items(json_file, 'products')
+    products_data = next(products_data)
+    for sku in products_data:
```
pquentin, line 217: Can you make sure that the output is deterministic since we're committing the result to git? It looks like using `sorted(products_data)` is enough.

contrib/scrap-ec2-sizes.py
```diff
@@ -0,0 +1,271 @@
...
+#  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+#  KIND, either express or implied.  See the License for the
+#  specific language governing permissions and limitations
+#  under the License.
+"""
+This script downloads and parses AWS EC2 from https://pricing.us-east-1.amazonaws.com/offers/v1.0/aws/AmazonEC2/current/index.json.
+It writes a Python module with constants about EC2's sizes and regions.
+
+Use it as following:
+    $ python scrap-ec2-sizes.py > ../libcloud/compute/constants.py
```
pquentin: It would be better to document this as `python contrib/scrap-ec2-sizes.py > libcloud/compute/constants.py`.

We also need to update "Making a release" in docs/committer_guide.rst. I suggest adding a "0. Update committed files" section with the CHANGES and libcloud/__init__.py lines from the "1. Pre-release check list" section.

pquentin (changes requested):
Thanks! This looks good. Only two small changes left.

contrib/scrap-ec2-sizes.py
ZuluPro: It's a bit difficult to sort at this level. I use ijson because it's a >250MB JSON.
So to use `sorted` here will try to parse and put that big block in RAM.

contrib/scrap-ec2-sizes.py
ZuluPro: Sorry, I'm not sure to understand, could you rephrase it ?

contrib/scrap-ec2-sizes.py
pquentin: Running mprof run on the scrap-ec2-sizes, I don't really see a difference between the default mode:

default

and the sorted() mode:

sorted

And 200MB is acceptable for a script that gets run once in a while, since having a deterministic output is very important to have useable diffs.

contrib/scrap-ec2-sizes.py
pquentin: Something like this:

```
diff --git a/docs/committer_guide.rst b/docs/committer_guide.rst
index 0f998088..4128e962 100644
--- a/docs/committer_guide.rst
+++ b/docs/committer_guide.rst
@@ -91,12 +91,17 @@ Making a release (for release managers)
 This section contains information a release manager should follow when
 preparing a release.

-1. Pre-release check list
+0. Update committed files
 ~~~~~~~~~~~~~~~~~~~~~~~~~

-* Make sure tests pass on all the supported Python versions (``tox``)
 * Make sure ``CHANGES`` file is up to date
 * Make sure ``__version__`` string in ``libcloud/__init__.py`` is up to date
+* Update constants. ``python contrib/scrap-ec2-sizes.py > libcloud/compute/constants.py``
+
+1. Pre-release check list
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+* Make sure tests pass on all the supported Python versions (``tox``)
 * Remove the ``tox`` directory with ``rm -rf .tox``
 * Remove the _secrets_ file with ``rm libcloud/test/secrets.py``
 * Remove leftover builds from previous releases. ``rm -f dist/apache*``
```

pquentin:
@ZuluPro Would you mind if I apply the two changes I mentioned a week ago? Specifically, https://github.com/apache/libcloud/pull/1211#discussion_r194523398 and https://github.com/apache/libcloud/pull/1211#discussion_r194523577. I'd like to release a new libcloud version with this pull request.

ZuluPro:
Hey @pquentin , Sorry for my latency, yes of course, you can continue this PR. (I need it too :D )

I'm sorry, I'm very busy in the meantimes, and I have to deal with a lot of things to find 300MB of RAM on my VM 😄

pquentin:
@ZuluPro Thanks! I have one more question.

The 'capacitystatus', 'ebsOptimized', 'operation', 'licenseModel', 'preInstalledSw', 'tenancy', 'usagetype' fields in extra are not deterministic because they actually vary across regions. Should we leave them out?

pquentin:
@ZuluPro I ended up leaving them out. See https://github.com/apache/libcloud/commit/384e42336e073594acccb5ad1c0e873830c9a06c

Merci Anthony ! Merged in trunk. ✨

PR #1214, opened 2018-05-23
[LIBCLOUD-996] Support GCP service_account authentication from JSON obj/str

bverschueren:
### Description

Currently GCP service_account authentication credentials can be provided using the JSON/PEM file acquired from GCP console or gcloud utility. It would be convenient if the JSON content could be directly provided as an object or serialized JSON string to eliminate the need of having the file locally on the calling host.

In this PR:
- add logic to support JSON dict and str handling
- rearrange PEM/JSON related GCE_PARAMS_* constant names in test_google.py for better clarity
- add tests for JSON content as a dict and as a str

### Status

- done, ready for review

- [x] Documentation


Kami:
Thanks for the contribution.

This looks like a reasonable enough change to me. @bverschueren can you please update documentation for the GCE drivers to also document this authentication option?

Kami:
And ideally we would also have some more robust test which also verify that the actual used / parsed ``key`` value is correct.

bverschueren:
@Kami: rebased and updated the docs with this authentication option.
Also added a more thorough verification of the PEM key and a test with a non-valid key. For the tests to succeed, this requires however that the test key is a valid one, so I borrowed the test key from the cryptography lib (https://github.com/pyca/cryptography/blob/3ce8883fd62043419b225790cd524b7619b4bb72/vectors/cryptography_vectors/asymmetric/PKCS8/unenc-rsa-pkcs8.pem).

Kami:
@bverschueren I'm sorry for a (very long) delay.

I'm cleaning up old issues and PRs and I noticed that this one never made it in.

I synced it with the latest trunk, re-formatted code with black, made a small change, added some additional test cases for edge case scenarios and merged it into trunk.

PR #1215, opened 2018-05-25
Fix LIBCLOUD-997, implements LIBCLOUD-998 and LIBCLOUD-604

micafer:
#### LIBCLOUD-604:

Add equivalent function for returning listed subnets in OpenStack driver. Similar to `ex_list_subnet` for EC2 endpoint providers.

#### LIBCLOUD-997:

The OpenStack_2_NodeDriver uses two connections

the base class connection attribute and a new one: image_connection.

In this case if the user has specified the ex_force_base_url one of them will fail, as both uses the same base url, but there are different services with different urls.  If the user set the compute base url (the standard case) the image one will not work a viceversa

#### LIBCLOUD-998:
OpenStack_2_NodeDriver class uses the deprecated path "/os-networks" of the compute API instead of using the "/v2.0/networks" of the neutron API that i the current one.

### Status
- done, ready for review

- [x] Documentation


pquentin:
@vdloo do you think you can look into this as our local OpenStack expert? :) Thanks!

vdloo:
Definitely, I'll try to find some time soon

pquentin:
@vdloo Now if you are on holiday this can wait!

libcloud/test/compute/test_openstack.py
```diff
@@ -1697,6 +1679,38 @@ def test_ex_accept_image_member(self):
         self.assertEqual(image_member.extra['updated'], '2018-03-02T14:20:37Z')
         self.assertEqual(image_member.extra['schema'], '/v2/schemas/member')
 
+    def test_ex_list_networks(self):
+        networks = self.driver.ex_list_networks()
+        network = networks[0]
+
+        self.assertEqual(len(networks), 2)
+        self.assertEqual(network.name, 'net1')
```
vdloo, line 1687: noticed that you left out the `self.assertEqual(network.cidr` that was here before, doesn't really matter but not sure if that was intended

vdloo:
perfect!

before:
```
In [13]: conn.ex_list_networks()
Out[13]: [<OpenStackNetwork id="12348a8c-1234-4e8f-a805-2035365f4d43" name="L2_NETWORK" cidr="None">]

In [16]: conn.ex_list_networks()[0].extra
Out[16]: {}

In [14]: conn.ex_list_subnets()
AttributeError                            Traceback (most recent call last)
<ipython-input-14-f6ae1a76cfd5> in <module>()
----> 1 conn.ex_list_subnets()
```

after:
```
In [21]: conn.ex_list_networks()
Out[21]: [<OpenStackNetwork id="12348a8c-1234-4e8f-a805-2035365f4d43" name="L2_NETWORK" cidr="None">]

In [30]: conn.ex_list_networks()[0].extra
Out[30]: {'subnets': [u'2f31a16a-1234-4329-1234-358ea86a7577']}

In [22]: conn.ex_list_subnets()
Out[22]: [<OpenStack_2_SubNet id="2f31a16a-1234-4329-1234-358ea86a7577" name="L2_SUBNET" cidr="43.117.12.0/22">]
```

libcloud/test/compute/test_openstack.py
micafer: I remove it as it is allways None in real cases so, I think that it have no sense to set a value in the tests that will never appear. But If you think that is interesting we can add it again,

pquentin:
Thanks @micafer @vdloo!

PR #1217, opened 2018-06-05
[gce] ex_create_network can have NoneType routing_mode

danhunsaker:
Check for that before trying string operations on it.

### DIFF
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -3892,13 +3892,14 @@ class GCENodeDriver(NodeDriver):
         else:
             network_data['autoCreateSubnetworks'] = (mode.lower() == 'auto')
 
-        if routing_mode.lower() not in ['regional', 'global']:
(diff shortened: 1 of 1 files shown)


pquentin:
The docstring says:

        :param  routing_mode: Create network with 'Global' or 'Regional'
                              routing mode for BGP advertisements.
                              Defaults to 'Regional'
        :type   routing_mode: ``str`` or ``None``

Maybe the correct fix is to set routing_mode to `Regional` if it's None?

danhunsaker:
That default happens on Google's end rather than ours, and they seem to prefer not sending a value at all unless you have an explicit preference for one or the other. But that's certainly an option as well. I was just opting for actually supporting `None`, as the docstring also indicates is valid.

pquentin:
Thanks, that makes sense, I'll merge this when I have the chance.

pquentin:
Thanks! Merged in trunk. ✨

PR #1220, opened 2018-06-30
[scaleway] Fix the pagination

danhunsaker:
Pagination was broken in the original PR. The `links` variable wasn't being updated after loading each new page, so it kept requesting page 2 in an infinite loop. This hotfix resolves that issue.


pquentin (approved):
Nice catch, thanks @danhunsaker! Will merge when I get the chance during the next week.

pquentin:
Thanks! Merged in trunk. ✨

PR #1221, opened 2018-07-01
add ex_get_node_details for digitalocean_v2 driver

vdloo:
### Description

Adds ex_get_node_details for the digitalocean_v2 driver. This method already
exists for other drivers like [OpenStack](https://github.com/apache/libcloud/blob/e2e17414dd9e8ba3542e181885ec242b820bfc48/libcloud/compute/drivers/openstack.py#L317).

Right now to list one node object using the DigitalOcean driver you can
run .list_nodes and then filter out the one you need by ID. This works
fine, but once start using that on accounts with many instances (thousands)
that will becomes slow and the chance of encountering API exceptions goes
up and you end up with frequent errors like:

```
return connection.list_nodes()
File "/usr/local/venv/project/src/apache-libcloud/libcloud/compute/drivers/digitalocean.py", line 349, in list_nodes
data = self._paginated_request('/v2/droplets', 'droplets')
File "/usr/local/venv/project/src/apache-libcloud/libcloud/common/digitalocean.py", line 242, in _paginated_request
new_data = self.connection.request(url, params=params)
File "/usr/local/venv/project/src/apache-libcloud/libcloud/common/base.py", line 862, in request
File "/usr/local/venv/project/src/apache-libcloud/libcloud/common/base.py", line 179, in __init__
headers=self.headers)
BaseHTTPError: Server was unable to give you a response. (code: 500)
```

Or also very often:
```
MalformedResponseError: <MalformedResponseException in <project.provisioning.driver.digitalocean.DigitalOcean_v2_NodeDriver object at 0x4f1c290> 'Failed to parse JSON'>: '<!DOCTYPE html>\n<html>\n<head>\n  <title>DigitalOcean - Something went wrong!</
```

Using .ex_get_node_details instead of .list_nodes when you need to retrieve a
node object by droplet ID will make that action faster and less likely to fail.

Example:
```
In [13]: conn.ex_get_node_details('7772242')
Out[13]: <Node: uuid=2aa2a2834970788c565aaf3ccd0b148dd51ce777, name=testnode, state=RUNNING, public_ips=[u'37.123.0.123'], private_ips=[], provider=DigitalOcean ...>
```

### Status

- done, ready for review

### DIFF
--- a/libcloud/compute/drivers/digitalocean.py
+++ b/libcloud/compute/drivers/digitalocean.py
@@ -495,6 +495,20 @@ class DigitalOcean_v2_NodeDriver(DigitalOcean_v2_BaseDriver,
                                       method='DELETE')
         return res.status == httplib.NO_CONTENT
 
+    def ex_get_node_details(self, node_id):
+        """
+        Lists details of the specified server.
+
+        :param       node_id: ID of the node which should be used
+        :type        node_id: ``str``
+
+        :rtype: :class:`Node`
+        """
+        data = self._paginated_request(
+            '/v2/droplets/{}'.format(node_id), 'droplet'
+        )
+        return self._to_node(data)
+
     def ex_create_floating_ip(self, location):
         """
         Create new floating IP reserved to a region.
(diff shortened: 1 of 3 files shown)


pquentin (approved):
Thanks @vdloo! This looks good. Thank you for explaining the motivation, it was interesting to read!

I'll merge this when I get the chance during the next week, but feel free to do it yourself if you want to.

PR #1223, opened 2018-07-03
[LIBCLOUD-1002] Add instance template support to GCE create_node

rojopolis:
### Allows Nodes to be created from GCE Instance Templates

This change allows Nodes to be created from predefined Instance Templates in GCE.  
This is useful in a number of scenarios:
* Organizational standards are implemented with Templates.
* etc.

The attributes specified in the call to create_node will take precedence over those defined in the template (per provider documentation).

#### Provider docs: 
- https://cloud.google.com/compute/docs/reference/rest/v1/instances/insert
- https://cloud.google.com/compute/docs/instances/create-vm-from-instance-template

### Status
- done, ready for review

- [x] Documentation

### DIFF
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -28,6 +28,7 @@ from libcloud.common.google import GoogleBaseConnection
 from libcloud.common.google import GoogleBaseError
 from libcloud.common.google import ResourceNotFoundError
 from libcloud.common.google import ResourceExistsError
+from libcloud.common.google import InvalidRequestError
 from libcloud.common.types import ProviderError
 
 from libcloud.compute.base import Node, NodeDriver, NodeImage, NodeLocation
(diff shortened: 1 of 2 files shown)


rojopolis:
Perhaps the signature for create_node should be modified to accept a generic `template` parameter rather than the provider specific `ex_source_instancetemplate` and `ex_launch_template_*` ( #1224 ).  It would actually be helpful for the specific project I'm working on, but I'm not sure how this project feels about changes to signatures like that.

Kami:
> Perhaps the signature for create_node should be modified to accept a generic template parameter rather than the provider specific ex_source_instancetemplate and ex_launch_template_* ( #1224 ). It would actually be helpful for the specific project I'm working on, but I'm not sure how this project feels about changes to signatures like that.

If we can abstract template functionality in a standard base class, I would be fine with that.

Kami:
Thanks for your contribution.

PR #1224, opened 2018-07-03
[LIBCLOUD-1004] Add launch template support to EC2 create_node

rojopolis:
### Allows Nodes to be created from EC2 Launch Templates

This change allows Nodes to be created from predefined Launch Templates in EC2.  
This is useful in a number of scenarios:
* Organizational standards are implemented with Templates.
* etc.

The attributes specified in the call to create_node will take precedence over those defined in the template (per provider documentation).

#### Provider docs: 
- https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_RunInstances.html
- https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_LaunchTemplateSpecification.html

### Status
- done, ready for review

- [x] Documentation

### DIFF
--- a/libcloud/compute/drivers/ec2.py
+++ b/libcloud/compute/drivers/ec2.py
@@ -1862,16 +1862,46 @@ class BaseEC2NodeDriver(NodeDriver):
                                               the operating systems command
                                               for system shutdown.
         :type       ex_terminate_on_shutdown: ``bool``
+
+        :keyword    ex_launch_template_name: Template to create the instance
+                                             from.
+        :type       ex_launch_template_name: ``str`` or ``None``
+
+        :keyword    ex_launch_template_id:   Template ID to create instance
+                                             from.
+        :type       ex_launch_template_id:   ``str`` or ``None``
+
+        :keyword    ex_launch_template_version: Template version to create
+                                                instance from.
+        :type       ex_launch_template_version: ``str``, ``int`` or ``None``
         """
-        image = kwargs["image"]
-        size = kwargs["size"]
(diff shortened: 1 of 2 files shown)


libcloud/compute/drivers/ec2.py
```diff
@@ -1862,17 +1862,47 @@ def create_node(self, **kwargs):
...
+
         params = {
             'Action': 'RunInstances',
-            'ImageId': image.id,
             'MinCount': str(kwargs.get('ex_mincount', '1')),
             'MaxCount': str(kwargs.get('ex_maxcount', '1')),
-            'InstanceType': size.id
         }
 
+        if 'ex_launch_template_name' in kwargs:
```
Kami, line 1885: How do those arguments interact with each other? Are they mutually exclusive aka only a single one can be provided or you can mix them and provide more than one?

libcloud/test/compute/test_ec2.py
```diff
@@ -126,6 +126,14 @@ def test_create_node(self):
         self.assertEqual(node.extra['tags']['Name'], 'foo')
         self.assertEqual(len(node.extra['tags']), 1)
 
+        def test_create_node_with_launch_template(self):
```
Kami, line 129: Overall the change looks good to me, but it would be good to add some additional tests which verify that the correct parameters are provided in the body when ``ex_launch_template_*`` keyword arguments are used.

This would need to happen inside the Mock response class.

Kami:
Thanks for your contribution.

PR #1227, opened 2018-07-17
Add Python 3.7 to tox and Travis.

andrewsomething:
### Description

This adds Python 3.7 to tox and Travis. I do not expect this to get merged in its current state. There are a number of failures. My aim is to get more visibility into those failures so that we can start fixing them.

### Status

- work in progress

- [ ] Documentation

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ python:
     - 3.4
     - 3.5
     - 3.6
+    - 3.7
     - pypy
     - pypy3
 os:
--- a/setup.py
+++ b/setup.py
@@ -166,6 +166,7 @@ setup(
         'Programming Language :: Python :: 3.4',
         'Programming Language :: Python :: 3.5',
         'Programming Language :: Python :: 3.6',
+        'Programming Language :: Python :: 3.7',
         'Programming Language :: Python :: Implementation :: CPython',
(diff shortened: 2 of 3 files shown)


andrewsomething:
Well that didn't work! :smile_cat:  Looks like Travis is still working on 3.7 support. So I guess I'll just close this for the time being.

Locally, I'm seeing quite a few failures. A good potion of those seem to be due to the use of `async` as a kwarg as it is now a now reserved keyword. https://docs.python.org/3/whatsnew/3.7.html

pquentin:
@andrewsomething Thanks!

Things to do:

* Travis: use dist: xenial and sudo: required to get Python 3.7 support, see https://github.com/python-trio/trio/pull/556 for an example.
* Delete the async variable in the Azure compute driver 
* Stop raising StopIteration
* Stop assuming that urlquote("~abc") will give '%7Eabc': it [gives '~abc' on Python 3.7](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.quote).

The full list of backward-incompatible changes: https://docs.python.org/3/whatsnew/3.7.html#porting-to-python-37

pquentin:
Our trunk now supports 3.7, thanks @andrewsomething!

PR #1245, opened 2018-10-12
Bug fixes for GCENodeDriver.ex_create_firewall()

paulmibus:
Remove incorrect firewall direction check for allow/deny rules.

Only default source_ranges to 0.0.0.0/0 if direction is INGRESS.

### Description

Fixed two small bugs in GCENodeDriver.ex_create_firewall().

The first issue was that a direction check was being applied to determine whether allowed or denied would be passed to the GCE API call. Both ingress and egress rules can have allowed or denied blocks so this shouldn't be conditional on the direction.

The second issue was that source_ranges was automatically set to [0.0.0.0/0] if no other source filters were applied. Since source filters are invalid for egress rules, I added a check for the INGRESS direction prior to setting a default source range.

### Status

done, ready for review

- [ ] Documentation

### DIFF
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -3268,12 +3268,13 @@ class GCENodeDriver(NodeDriver):
         firewall_data['direction'] = direction
         firewall_data['priority'] = priority
         firewall_data['description'] = description
-        if direction == 'INGRESS':
+        if allowed is not None:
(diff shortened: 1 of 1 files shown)


Kami:
Is this still relevant?

If it is, can you please add corresponding test cases?

Kami:
Thanks for your contribution.

PR #1247, opened 2018-10-13
Fix trigger_rtd_build.py

pquentin:
### Description

I think this was broken when switching more tox.ini entries to Python 3. See https://readthedocs.org/projects/libcloud/builds/: the only builds more recent than 6 months are my tests. 

### Status

- work in progress

- [x] Documentation

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,7 @@ matrix:
             - libvirt-dev
       # Trigger ReadTheDocs build on docs builder
       after_success:
-        - ./contrib/trigger_rtd_build.py 8284
+        - python3 ./contrib/trigger_rtd_build.py 8284
 
 install:
   - pip install --upgrade "pip==9.0.1"
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,6 +46,7 @@ matrix:
             - libvirt-dev
       # Trigger ReadTheDocs build on docs builder
       after_success:
+        - pip3 install requests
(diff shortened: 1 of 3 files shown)


vdloo (approved):

vdloo:
:+1:

pquentin:
That did not work, the requests package was not found, I need to fix that.

Click on the arrow before "python3 ./contrib/trigger_rtd_build.py 8284" in https://travis-ci.org/apache/libcloud/jobs/440923637 to see the issue.

vdloo:
Ah I see:
```
0.03s$ python3 ./contrib/trigger_rtd_build.py 8284
  File "./contrib/trigger_rtd_build.py", line 18, in <module>
    import requests
ImportError: No module named 'requests'
```
I like requests but what's wrong with using urllib2 in this case btw?

pquentin:
Because urllib2 is Python 2-only and it was easier for me to make the requests version work than to use urllib.request, but I'll try again with urllib.request.

Kami:
Good catch :+1:

pquentin:
Thanks @Kami :)

Ok, now I remember why I preferred to use requests instead of urllib.request: it's way easier. I don't think it's possible to have a POST request without a Request object, and with requests I don't have to worry about certificates.

I fixed the requests installation, we can now see a recent build in https://readthedocs.org/projects/libcloud/builds/. Please take another look.

pquentin:
Merged.

PR #1249, opened 2018-10-22
openstack: handle missing user email

ktdreyer:
### Description

Keystone may not always return an "email" entry for a user account.
Prior to this change, we could crash if we queried a user that lacked an
email record. Gracefully handle this case by setting email to None.

### Status

- done, ready for review

### Checklist

- [x] Documentation


vdloo:
Hi @ktdreyer! Perhaps you can add a unit test for this somewhere in `libcloud/test/common/test_openstack_identity.py`? Maybe if you edit `libcloud/test/compute/fixtures/openstack_identity/v3/v3_users.json` you can change one of the users there to not have an email record?

ktdreyer:
Good idea @vdloo . I've added a new unit test for this. Before the modification to `libcloud/common/openstack_identity.py`, the test raises `KeyError`, and now it passes.

vdloo:

PR #1252, opened 2018-11-10
[LIBCLOUD-1022] Enhance digitalocean dns driver with `ttl` parameter

kiniou:
### Description

DigitalOcean provides a parameter to update dns records with `ttl` but the current driver's code actually doesn't support it.

For reference:

- https://developers.digitalocean.com/documentation/v2/#update-a-domain-record
- https://developers.digitalocean.com/documentation/v2/#create-a-new-domain-record

### Status

ready for review

- [ ] Documentation


kiniou:
Hi 👋

I was not sure if the `ttl` parameters should go directly in methods signature or in `extra` arguments.

There seems to be some inconsistence between some drivers (eg. [powerdns](https://github.com/apache/libcloud/blob/trunk/libcloud/dns/drivers/powerdns.py#L189)) and [documentation](https://libcloud.readthedocs.io/en/latest/dns/examples.html#create-a-record-with-a-custom-ttl).

Just let me know if i need to do the switch 😉.

tonybaloney:
thanks for the contribution @kiniou 
the `ttl` parameter should go in the method signature for zones, and in the `extra` dict for records. I appreciate that makes little sense!

libcloud/dns/drivers/digitalocean.py
```diff
@@ -178,6 +178,10 @@ def create_record(self, name, zone, type, data, extra=None):
                 params['weight'] = extra['weight']
             except KeyError:
                 params['weight'] = 'null'
+            try:
+                params['ttl'] = extra['ttl']
```
tonybaloney, line 183: `params['ttl'] = extra.get('ttl', None)` is a little cleaner

tonybaloney (changes requested):

libcloud/dns/drivers/digitalocean.py
kiniou: :ok_hand: I also thought it would be cleaner to do `extra.get` but I thought the actual try-catch based code was following some coding rules i should follow. Maybe it would be more readable to do this on the others parameters. 

Anyway, I'll edit PR with your suggestion :wink:

kiniou: According to the relevant DigitalOcean API documentations, the `ttl` can not be nulled and I've just tested the change to `extra.get('ttl', None)` and the API response fails with an error 500.

I've looked for a better syntactic way but i think the try-catch structure is the only way to not include this parameter if it does not exists.

vdloo: you could do something like this:
```
if 'ttl' in extra:
    params['ttl'] = extra['ttl']
```

kiniou: Yes ! :man_facepalming:
Thank you !

Kami:
Re ``ttl`` in the ``extra`` attribute - it looks like a lot of the existing drivers support TTL attribute for records so I think it would be time to "promote" this attribute from ``extra`` to ``ttl`` argument in create / update record methods in the next major release.

For backward compatibility reasons we should still support ``ttl`` in extra, with ``ttl` having precedence over extra attribute.

Kami (approved):

Kami:
Sorry for the delay.

I made a small change (ec563f995c5080acba0087c20408a13df58b3d51) and merged changes into trunk. Thanks again for your contribution.

PR #1260, opened 2018-11-28
remove unused variable in parse_error

vdloo:
This variable is not used, `text` is overwritten in any of the possible
cases (except in the path where it throws the KeyPairDoesNotExistError
but in that case it does not matter).


Kami (approved):

vdloo:
Hey @Kami what's the new merge procedure now? Before the whole apache 'migration of git repos' thing we couldn't just merge to master on github. Is that OK now? Do we then push to master afterwards to edit the changes file? Or should that be part of the feature branch before merging?

Kami:
If you have Github and Apache accounts linked you can now simply merge directly on Github via merge button.

As far as editing changelog goes - I'm fine either way.

PR #1261, opened 2018-11-28
openstack: handle missing user "enabled" attribute

ktdreyer:
### Description

Keystone may not always return an "enabled" attribute for a user account. Prior to this change, we could crash if we queried a user that lacked an "enabled" attribute. Gracefully handle this case by setting "enabled" to None.

### Status

- done, ready for review

- [x] Documentation

### DIFF
--- a/libcloud/test/common/test_openstack_identity.py
+++ b/libcloud/test/common/test_openstack_identity.py
@@ -379,6 +379,12 @@ class OpenStackIdentity_3_0_ConnectionTests(unittest.TestCase):
         self.assertEqual(user.name, 'userwithoutemail')
         self.assertEqual(user.email, None)
 
+    def test_get_user_without_enabled(self):
+        user = self.auth_instance.get_user(user_id='c')
+        self.assertEqual(user.id, 'c')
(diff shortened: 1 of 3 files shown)


libcloud/test/common/test_openstack_identity.py
```diff
@@ -379,6 +379,12 @@ def test_get_user_without_email(self):
         self.assertEqual(user.name, 'userwithoutemail')
         self.assertEqual(user.email, None)
 
+    def test_get_user_without_enabled(self):
+        user = self.auth_instance.get_user(user_id='c')
+        self.assertEqual(user.id, 'c')
+        self.assertEqual(user.name, 'userwithoutenabled')
+        self.assertEqual(user.enabled, None)
```
vdloo, line 386: you could have done `self.assertIsNone` here

vdloo:

libcloud/test/common/test_openstack_identity.py
ktdreyer: @vdloo thank you for catching this!

I've globally switched this out for `assertIsNone` in https://github.com/apache/libcloud/pull/1264

PR #1265, opened 2018-12-04
Override the auth header for the VCloud 5.5 Connection

cheald:
### Description

Newer VCD installs (I'm working against API level 30.0, for reference) enforce a minimum API level on authentication. This simply causes the authentication payload to send a version 5.5 header rather than a version 1.5 header, so our auth request isn't rejected.

### Status

done, ready for review

- [ ] Documentation

### DIFF
--- a/libcloud/compute/drivers/vcloud.py
+++ b/libcloud/compute/drivers/vcloud.py
@@ -863,6 +863,15 @@ class VCloud_1_5_Connection(VCloudConnection):
 
 
 class VCloud_5_5_Connection(VCloud_1_5_Connection):
+
+    def _get_auth_headers(self):
+        """Compatibility for using v5.5+ APIs"""
+        return {
+            'Authorization': "Basic %s" % base64.b64encode(
+                b('%s:%s' % (self.user_id, self.key))).decode('utf-8'),
+            'Accept': 'application/*+xml;version=5.5'
+        }
+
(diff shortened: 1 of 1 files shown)


vdloo:
Hi @cheald, could you add a simple unit test to https://github.com/apache/libcloud/blob/da9ae84e7c20d8921f573a880732da7b44f28fa6/test/compute/test_vcloud.py#L84? Perhaps just calling _get_auth_headers and checking if `version=5.5` is in the Accept key of the dict would be sufficient. Unit tests both serve to document behavior and prevent stuff from accidentally being changed later on, so even though that would be a pretty trivial test I think it would still hold value.

Also I have a question here as I am not that familiar with the vcloud APIs. Would this break anything for not 'Newer VCD installs' or would that also keep working as intended?

cheald:
Sure, I'll see about wiring that up. I didn't include one mostly out of ignorance of where/what to test. :)

I've only started looking into using libcloud with our new hosting provider - they provide a vCloud Director install with the version 30.0 API. I'm still quite new to the whole vCloud ecosystem, but logically, I don't believe this should break anything for "not newer installs" - this is the 5_5_Connection subclass, which only applies when the requested version is 5.5. Earlier versions use the 1_5_Connection.

Kami:
Thanks for your contribution.

PR #1266, opened 2018-12-05
Dummy Node ID doesn't allow 0

utam0k:
### Description

If Dummy Node ID is `0`, it sets `None` when Node init.
Refs: https://github.com/apache/libcloud/blob/v2.4.0/libcloud/compute/base.py#L205
I think this is not the expected behavior.
So, I did not to allow it is `0`.

Before
```
[(None, 'dummy-0'), ('1', 'dummy-1'), ('2', 'dummy-2')]
```
After
```
[('1', 'dummy-1'), ('2', 'dummy-2')]
```

### Status
- done, ready for review

- [x] Documentation

### DIFF
--- a/libcloud/compute/drivers/dummy.py
+++ b/libcloud/compute/drivers/dummy.py
@@ -84,16 +84,17 @@ class DummyNodeDriver(NodeDriver):
             self.nl = []
             startip = _ip_to_int('127.0.0.1')
             for i in range(num):
-                ip = _int_to_ip(startip + i)
-                self.nl.append(
-                    Node(id=i,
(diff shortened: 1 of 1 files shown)


vdloo:
Hi @utam0k, I agree that getting 0 when you pass in None is kind of unexpected. But this change however would change established behavior which a lot of users probably depend on. For that reason I would vote against changing this as it would make it harder to upgrade libcloud to the latest version. I'm not sure what the policy is in regards to breaking changes in libcloud but I personally would prefer if we keep it as easy as possible to stay up to date with the library.

PR #1242, opened 2018-09-26
Implement new versions of OpenStack API calls [LIBCLOUD-874]

micafer:
### Description

This PR overwrite the usage of the deprecated nova API methods:
- /os-volumes
- /os-snapshots
- /os-security-groups
- /os-security-group-rules
- /os-floating-ips
- /os-floating-ip-pools

Replacing them with corresponding calls to neutron and cinder.

Also other minor fixes has been made as fixing the function parse_error to avoid error if "code" is not in the response message or fixing _to_port function to make it work with old OpenStack versions where some fields as: port_security_enabled or project_id are not returned by the API.

### Status

- done, ready for review

### Checklist (tick everything that applies)

- [x] [Code linting](http://libcloud.readthedocs.org/en/latest/development.html#code-style-guide) (required, can be done after the PR checks)
- [x] Documentation
- [x] [Tests](http://libcloud.readthedocs.org/en/latest/testing.html)
- [ ] [ICLA](http://libcloud.readthedocs.org/en/latest/development.html#contributing-bigger-changes) (required for bigger changes)

PR #1267, opened 2018-12-10
fix OpenStack_2_NodeDriver.detach_volume

vdloo:
Since https://github.com/apache/libcloud/pull/1242 the OpenStack_2_NodeDriver uses a volume object returned via volumev2_connection. This object will not contain serverId but server_id. This PR makes the detach_volume method compatible with v2 volumes.

Fixes:
```
return connection.detach_volume(volume)
File "/usr/local/venv/project/src/apache-libcloud/libcloud/compute/drivers/openstack.py", line 328, in detach_volume
(attachment['serverId'], attachment['id']),
KeyError: 'serverId'
```

### DIFF
--- a/libcloud/compute/drivers/openstack.py
+++ b/libcloud/compute/drivers/openstack.py
@@ -322,14 +322,19 @@ class OpenStackNodeDriver(NodeDriver, OpenStackDriverMixin):
         # when ex_node is not provided, volume is detached from all nodes
         failed_nodes = []
         for attachment in volume.extra['attachments']:
-            if not ex_node or ex_node.id == attachment['serverId']:
+            if not ex_node or ex_node.id in filter(None, (attachment.get(
+                'serverId'
+            ), attachment.get('server_id'))):
                 response = self.connection.request(
                     '/servers/%s/os-volume_attachments/%s' %
-                    (attachment['serverId'], attachment['id']),
+                    (attachment.get('serverId') or attachment['server_id'],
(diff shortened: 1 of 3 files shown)


vdloo:
@micafer could you take a look at this one?

libcloud/compute/drivers/openstack.py
```diff
@@ -322,14 +322,19 @@ def detach_volume(self, volume, ex_node=None):
         # when ex_node is not provided, volume is detached from all nodes
         failed_nodes = []
         for attachment in volume.extra['attachments']:
-            if not ex_node or ex_node.id == attachment['serverId']:
+            if not ex_node or ex_node.id == attachment.get(
+                'serverId'
+            ) or attachment['server_id']:
```
micafer: This comparison is not correct.
In case of passing an ex_node, if attachment['server_id'] is not None, the attachment is deleted using any node.
It should be something like:
```
if not ex_node or ex_node.id in [attachment.get('serverId'), attachment['server_id']]
```

vdloo: oops, good catch! I'll fix that

vdloo:
failing tests because of https://github.com/apache/libcloud/pull/1271

edit: fixed in https://github.com/apache/libcloud/pull/1271

tonybaloney:
@micafer please could you review the changes again

micafer (approved):
LGTM

PR #1272, opened 2018-12-28
Handle Google Storage Interoperability access keys of more than 20 characters

MonsieurV:
### Description

Google Cloud Storage currently emits access keys of more than 20 characters.

For e.g. an access key generated on my account (revoked since): `GOOGDF5OVRRGU4APFNSTVCXI` (24 characters)

So we can't rely anymore [on the doc](https://cloud.google.com/storage/docs/migrating?hl=fr#keys), when it is stating:
> An access key is a 20 character alphanumeric string

### Status

- done, ready for review

Should we add tests?

- [X] Documentation

### DIFF
--- a/libcloud/common/google.py
+++ b/libcloud/common/google.py
@@ -610,9 +610,12 @@ class GoogleAuthType(object):
     @staticmethod
     def _is_gcs_s3(user_id):
         """
-        Checks S3 key format: 20 alphanumeric chars starting with GOOG.
+        Checks S3 key format: alphanumeric chars starting with GOOG.
         """
-        return len(user_id) == 20 and user_id.startswith('GOOG')
+        return (
+            len(user_id) >= 20 and len(user_id) < 30 and user_id
+            .startswith('GOOG')
+        )
 
     @staticmethod
     def _is_sa(user_id):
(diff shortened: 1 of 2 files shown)


vdloo:
> Should we add tests?

yes please, maybe [you can do something with this](https://github.com/apache/libcloud/blob/5c962166e9fa9fc5e0dd31b7f265ac8ba1137855/libcloud/test/common/test_google.py#L63). Also is there a max amount of chars this can be that we should also check for?

MonsieurV:
@vdloo I've added a simple test.

For the upper limit of characters, good point. I have no idea.

I've found no specification on that in the documentation (which currently is out of step with reality, as it states that _an access key is a 20 character alphanumeric string_).

When I generate and revoke several access keys on my console, I always get 24 characters. I presume they no longer generate 20 characters keys, but that it would be safe to put an upper limit of say 30 (this is arbitrary).

We could try to contact Google to have the doc updated?

PS. Tell me when it's good for you, and I squash the commits.

vdloo:
yeah looks good, if you squash this I'll merge it

MonsieurV:
Done :ok_hand:

vdloo:

PR #1277, opened 2019-01-28
[LIBCLOUD-1034] fix scheme issue in https case

hldh214:
### Description

Hello everyone, first apologize my poor english

im beginner and i have a question about http proxy(https://libcloud.readthedocs.io/en/latest/other/using-http-proxy.html)

from here: https://github.com/apache/libcloud/blob/trunk/libcloud/http.py#L117
```python
if parsed.scheme != 'http':
    raise ValueError('Only http proxies are supported')
```

i guess this code means we cant use proxy on https requests, sounds not cool for nowadays

from here: https://github.com/apache/libcloud/blob/trunk/libcloud/http.py#L102
```python
self.session.proxies = {
    # requests assume `self.proxy_scheme` as http/https protocol
    # so if we want proxy our https request
    # we need 2 add `'https': 'http://127.0.0.1:1080'` instead of `http` k/v pair
    self.proxy_scheme: proxy_url
}
```

i dont know if this `self.proxy_scheme` has other meaning or not
please let me know if i have any misunderstand or mistake, Thanks <3

### Status

- done, ready for review

- [ ] Documentation


Kami:
Thanks for the contribution.

In the past, HTTPs proxies weren't supported and didn't work and that's why that check is there.

Maybe this has changed with our migration to ``requests`` - someone would need to check and confirm it indeed works with HTTPs proxies.

hldh214:
thanks for the reply
ill update unit test later

Kami:
@hldh214 Did you happen to have time yet to check and verify it with an HTTPS proxy?

Kami:
I started to look into this today. Sadly the change will be much more involved than I was hoping for.

Part of the issue is related to "set_http_proxy" driver method which doesn't work correctly because of the way how connection class is instantiated. This was either broken when moving to the requests library or it never worked correctly.

Kami:
Superseded by #1314.

Thanks again for your contribution. I will get it finished inside #1314.

PR #1279, opened 2019-03-08
Make tests pass after 2031

bmwiedemann:
Background:
As part of my work on reproducible builds for openSUSE, I check that software still gives identical build results in the future.
The usual offset is +15 years, because that is how long I expect some software will be used in some places.
This showed up failing tests in our python-apache-libcloud package build.

With this patch, tests completed successfully on i586 and x86_64 in 2034

See https://reproducible-builds.org/ for the overall topic - only slightly relevant here.

- done, ready for review

- [ ] Documentation

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,6 +18,13 @@ General
   
 - Updating command line arguments to current version in Azure examples (GITHUB-1273) [mitar]
 
+- [GCE, SoftLayer] Update GCE and Softlayer drivers to utilize crypto
+  primitives from the ``cryptography`` library instead of deprecated and
+  unmaintained ``PyCrypto`` library.
+  
+  (GITHUB-1280)
+  [Ryan Petrello]
+
 Common
 ~~~~~~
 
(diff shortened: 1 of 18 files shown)


vdloo:
cool :+1:

bmwiedemann:
See also 2 related issues of 2038-01-19:
https://github.com/pytest-dev/pytest/pull/4903
https://issues.apache.org/jira/browse/LIBCLOUD-1038

Kami:
Thanks, the change looks good to me.

It looks like https://issues.apache.org/jira/browse/LIBCLOUD-1038 is related to an actual limitation / issue in the xmlrpclib from Python stdlib? And probably nothing we can do about it.

I would imagine that probably already got changed / fixed in Python 3.x?

Kami:
Some related Python issues:

* https://bugs.python.org/issue2985
* https://mail.python.org/pipermail/python-dev/2009-July/090649.html

It looks like a solution / workaround for VCL test failure would be to send string integer of an integer when number is larger than 32 bits (and assume VCL API handles that correctly).

Kami:
I pushed a change which I believe will fix the test failure mentioned in https://issues.apache.org/jira/browse/LIBCLOUD-1038.

It looks like the method already expects a string for ``start`` argument (per docstring), so it should be safe and correct thing to do that.

Kami:
Merged, thanks!

PR #1281, opened 2019-03-21
Add router management functions in OpenStack

micafer:
### Description

* Add functions to list, create, destroy and manage interfaces in network routers in OpenStack driver.
Also the functions to create and delete network has been added in the OpenStack_2_NodeDriver to make it work with the neutron API.
* Add the network_id to the subnet object.
* Add the ex_update_subnet function.
* Add the ex_update_port function.
* Add extra kwargs parameters has been added in the subnet creation to enable to specify optional parameters.

### Status

- done, ready for review

- [ ] Documentation


libcloud/compute/drivers/openstack.py
```diff
@@ -2979,7 +3012,7 @@ def ex_list_subnets(self):
             self._subnets_url_prefix).object
         return self._to_subnets(response)
 
-    def ex_create_subnet(self, name, network, cidr, ip_version=4):
+    def ex_create_subnet(self, name, network, cidr, ip_version=4, **kwargs):
```
Kami: I would prefer to explicitly declare support arguments. Using ``**kwargs`` makes it impossible / hard to see which arguments method takes without checking the code. It also makes automatic code introspection impossible.

https://libcloud.readthedocs.io/en/latest/development.html#don-t-abuse-kwargs

Kami:
Thanks for the contribution.

From a high level the code change looks good to me (besides some concerns of using ``**kwargs``. Since this functionality is available by multiple providers, we should eventually see if we can introduce a new networking API.

Something along the lines of #345.

libcloud/compute/drivers/openstack.py
```diff
@@ -3416,6 +3486,148 @@ def ex_list_floating_ip_pools(self):
...
+        :param port: Port object to be added to the router
+        :type port: :class:`OpenStack_2_PortInterface`
+
+        :rtype: ``bool``
+        """
+        return self._manage_router_interface(router, 'add', port=port)
+
+    def ex_del_router_port(self, router, port):
+        """
+        Remove port to a router
```
Kami: to a -> from a :)

Kami:
Merged, thanks.

PR #987, opened 2017-02-15
packet.net optimizations

mgogoulos:
### Description

- Use multiprocessing to get nodes without specifying project_id
Function list_nodes can also receive a project_id and return node on this
project, but if called without one it will fetch nodes on all projects

- fix list_sizes to return only server sizes (now it contains other items, as storage)
- provide cloud_init on create_node
- add functions ex_start_node and ex_stop_node
- add extra metadata on list_nodes
- add Project object with information for the Project
- add ex_list_projects function

### Status

done, ready for review

### Checklist (tick everything that applies)

- [x] [Code linting](http://libcloud.readthedocs.org/en/latest/development.html#code-style-guide) (required, can be done after the PR checks)
- [x] Documentation
- [ ] [Tests](http://libcloud.readthedocs.org/en/latest/testing.html)
- [ ] [ICLA](http://libcloud.readthedocs.org/en/latest/development.html#contributing-bigger-changes) (required for bigger changes)

PR #1282, opened 2019-03-22
Packet driver extensions

d-mo:
## Extend Packet driver

This PR extends and supersedes https://github.com/apache/libcloud/pull/987

### Description
* Support pagination & extra metadata in list_nodes
* Support cloud_init & all available params in create_node
* Add block storage support: list_volumes, create_volume, destroy_volume, attach_volume, detach_volume, ex_modify_volume, ex_restore_volume, ex_clone_volume, ex_describe_volume, create_volume_snapshot, destroy_volume_snapshot, list_volume_snapshots
* New node methods: ex_start_node, ex_stop_node, ex_reinstall_node, ex_rescue_node, ex_update_node, ex_get_node_bandwidth, ex_list_ip_assignments_for_node, ex_list_events_for_node, ex_associate_address_with_node
* New project methods: ex_list_projects, ex_list_events_for_project, ex_describe_all_addresses_for_project, list_volumes_for_project
* New ip address methods: ex_describe_all_addresses, ex_describe_address, ex_request_address_reservation, ex_disassociate_address
* New BGP methods: ex_list_bgp_sessions_for_project, ex_create_bgp_session, ex_list_bgp_sessions_for_node, ex_list_bgp_sessions, ex_get_bgp_session, ex_delete_bgp_session
* Use multiprocessing to get nodes for all projects in parallel if project_id not specified
* Fix list_sizes to return only server sizes (it also contained storage items)
* Add Project object with information for the Project
* Return size id instead of name in node.extra['plan']
* Return location code instead of name in node.extra['facility']
* Add tests for new methods

### Status

- done, ready for review

- [ ] Documentation


libcloud/compute/drivers/packet.py
```diff
@@ -88,11 +98,88 @@ class PacketNodeDriver(NodeDriver):
...
+    ]
+    retval = []
+    for future in futures:
+        result = yield from future
+        retval.extend(result)
+    return retval""" % resource_type, glob, loc)
+        loop = asyncio.get_event_loop()
+        return loop.run_until_complete(loc['_list_async'](loc['self']))
+
+    def list_nodes_for_project(self, ex_project_id, include='plan', page=1,
```
Kami: This should be an extension method - ``ex_list_nodes_for_project..

In addition to that, you can also add optional ``ex_project_id`` argument to ``list_nodes`` method and also handle per project filtering there as well.

libcloud/compute/drivers/packet.py
```diff
@@ -88,11 +98,88 @@ class PacketNodeDriver(NodeDriver):
...
+
+    def list_resources_async(self, resource_type):
+        # The _list_nodes function is defined dynamically using exec in
+        # order to prevent a SyntaxError in Python2 due to "yield from".
+        # This cruft can be removed once Python2 support is no longer
+        # required.
+        assert resource_type in ['nodes', 'volumes']
+        glob = globals()
+        loc = locals()
+        exec("""
```
Kami, line 156: Ideally we would find a better cross-driver way of handling that in the future.

Kami:
Thanks for a great contribution :+1: 

Overall, it looks good to me. I only added a small comment about ``list_nodes_for_project`` method which should be an extension method.

Can you please make that change so I can go ahead and merge those changes into trunk?

libcloud/compute/drivers/packet.py
```diff
@@ -88,11 +98,88 @@ class PacketNodeDriver(NodeDriver):
...
-                                       ).object['devices']
+    def __init__(self, key, project=None):
+        # initialize a NodeDriver for Packet using the API token
+        # and optionally the project (name or id)
+        # If project specified we need to be sure this is a valid project
+        # so we create the variable self.project_id
+        super(PacketNodeDriver, self).__init__(key=key, project=None)
+        self.project_name = project
+        self.project_id = None
+        self.projects = self.ex_list_projects()
```
Kami, line 110: Performing any kind of long running operations inside the constructor is not the best thing to do.

One approach would be to run this code lazily during the first HTTP request if ``self.project_id`` or ``self.project_name`` variable is not set yet.

In addition to that, there could be some kind of argument / method which user could use to force check if a specified project is valid. Perhaps ``ex_validate_project`` argument in this method which defaults to False or similar.

libcloud/compute/drivers/packet.py
```diff
@@ -88,11 +98,88 @@ class PacketNodeDriver(NodeDriver):
...
-        data = self.connection.request('/projects/%s/devices' %
-                                       (ex_project_id),
-                                       params={'include': 'plan'}
-                                       ).object['devices']
+    def __init__(self, key, project=None):
+        # initialize a NodeDriver for Packet using the API token
+        # and optionally the project (name or id)
+        # If project specified we need to be sure this is a valid project
+        # so we create the variable self.project_id
+        super(PacketNodeDriver, self).__init__(key=key, project=None)
```
Kami: I assume you don't actually want to pass ``project`` keyword argument to the parent class (``BaseDriver``) constructor, since the parent constructor doesn't support that argument (it will simply ignore it, but it may confuse people reading the code).

d-mo: Fixed

libcloud/compute/drivers/packet.py
d-mo: Fixed

libcloud/compute/drivers/packet.py
d-mo: Agreed. Are there examples of such lazy requests in other drivers?

Kami: Can't recall from top of my head.

Can probably just use @property and internal variable which is set to ``None`` inside the constructor and inside the property method we call the API if that value is none. Something along the lines of:

```python
@property
def projects(self):
    if not self._projects:
        self._projects = self.ex_list_projects()

    return self._projects
```

libcloud/compute/base.py
```diff
@@ -447,7 +447,7 @@ class NodeLocation(object):
     'US'
     """
 
-    def __init__(self, id, name, country, driver):
+    def __init__(self, id, name, country, driver, extra=None):
```
Kami, line 450: This change is breaking GCE driver because it calls parent constructor after setting ``extra`` attribute on the class instance.

I have a fix locally which I will include when merging those changes.

libcloud/compute/drivers/packet.py
Kami: I believe something like this should do it - https://github.com/Kami/libcloud/commit/8e2e9b18d0db61cb16947ce00f227c76e944f13c, but I didn't get a chance to test it with actual Packet.net API yet.

As you can see, I added additional guard condition to avoid infinite loop in case account has no projects.

Kami:
I've merged changes into trunk. Thanks for a great contribution!

When you get a chance, it would great if you can test the change I made in 8e2e9b18d0db61cb16947ce00f227c76e944f13c against the real provider API. All the tests do pass, bit that's not a guarantee it works end to end :)

d-mo:
I can confirm your change works as expected against the real API. Thanks for your help @Kami!

PR #1283, opened 2019-03-27
Fixes LIBCLOUD-1039 - cloudfiles download_object_as_stream

mattseymour:
### Description

Related ticket: https://issues.apache.org/jira/browse/LIBCLOUD-1039?

When downloading an object as a stream `download_object_as_stream`
in storage/drivers/cloudfiles.py a generator object should be
returned. Currently a rerquest/response object is returned
resulting in a HttpLibResponseProxy error.

This commit alters the callback_kwargs to use a
requests/response.iter_content resulting in an actual iterator
being returned. This code change is in other storage/driver 
specific code and appears to have been missed in cloudfiles.

### Status

- done, ready for review

- [ ] Documentation

### DIFF
--- a/libcloud/storage/drivers/cloudfiles.py
+++ b/libcloud/storage/drivers/cloudfiles.py
@@ -418,8 +418,12 @@ class CloudFilesStorageDriver(StorageDriver, OpenStackDriverMixin):
 
         return self._get_object(obj=obj, callback=read_in_chunks,
                                 response=response,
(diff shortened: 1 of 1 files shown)


Kami:
Thanks for the contribution.

The change looks good to me, but it would be good to also have a test case for it.

(I assume this regression was introduced when migrating to requests)

Kami:
I've confirmed this bug fix. I also added a test case and merged this into trunk.

PR #1284, opened 2019-03-29
LIBCLOUD-1040_CloudSigma_provider_ex_list_subscriptions_fails_if_your…

aki-k:
…_subscription_includes_free_tier_resources

## Title

CloudSigma provider ex_list_subscriptions() fails if your subscription includes free tier resources

### Description

When using driver.ex_list_subscriptions(status='all', resources=None) to list your resources, you'll receive the error:

  File "../libcloud/2.4.0/lib/python3.5/site-packages/libcloud/utils/iso8601.py", line 117, in parse_date
    m = ISO8601_REGEX.match(datestring)
TypeError: expected string or bytes-like object

CloudSigma gave some free resources to their customers a few years ago.

CloudSigmaSubscription id=free_mem, resource=mem, amount=1073741824, period=None, object_uuid=None
CloudSigmaSubscription id=free_dssd, resource=dssd, amount=53687091200, period=None, object_uuid=None

I suspect these are the problem resources with no end time.

I've added to iso8601.py on lines 117 and 118:

    if datestring is None:
        datestring = "9999-01-01T12:00:00+00:00"

Not sure if this is an appropriate fix but seems to fix it.

### Status

- done, ready for review

- [ ] Documentation

### DIFF
--- a/libcloud/utils/iso8601.py
+++ b/libcloud/utils/iso8601.py
@@ -117,8 +117,9 @@ def parse_date(datestring, default_timezone=UTC):
     default timezone specified in default_timezone is used. This is UTC by
     default.
     """
-    if datestring is None:
-        datestring = "9999-01-01T12:00:00+00:00"
+    if not datestring:
+        raise ValueError('datestring must be valid date string and not None')
+
     m = ISO8601_REGEX.match(datestring)
     if not m:
         raise ParseError("Unable to parse date string %r" % datestring)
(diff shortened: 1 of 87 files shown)


libcloud/utils/iso8601.py
```diff
@@ -114,6 +114,8 @@ def parse_date(datestring, default_timezone=UTC):
     default timezone specified in default_timezone is used. This is UTC by
     default.
     """
+    if datestring is None:
```
Kami: I think it would be better to handle this differently (returning "random" date is almost never a good idea) - either return None in such scenario or throw an exception if datestring is None and handle that exception in CloudSigma driver code.

Kami:
Thank you for the contribution.

I've added a comment on how I think we could handle this better.

Kami:
I made a necessary change to the code - 41fd9648cc805f496f5f95480a0bb84e078a8d67 and I will go ahead and merge it.

Thanks again for your original contribution :+1:

PR #1287, opened 2019-05-17
[LIBCLOUD-1043] Fix Azure upload_object_via_stream used with iter

c-w:
### Description

As described in [LIBCLOUD-1043](https://issues.apache.org/jira/browse/LIBCLOUD-1043), the Azure Blob Storage driver currently doesn't support use-cases like this:

```py
driver.upload_object_via_stream(iter(something), ...)
```

This is due to the fact that the current implementation of `upload_object_via_stream` assumes that the passed-in iterator always implements `seek`. This pull request adds a fallback for iterators where this is not the case.

### Status

- done, ready for review

- [x] Documentation **Bugfix only**
- [x] [Tests](http://libcloud.readthedocs.org/en/latest/testing.html) **Added new unit test**
- [ ] [ICLA](http://libcloud.readthedocs.org/en/latest/development.html#contributing-bigger-changes) (required for bigger changes) **Code change is small**

### DIFF
--- a/libcloud/storage/drivers/azure_blobs.py
+++ b/libcloud/storage/drivers/azure_blobs.py
@@ -18,6 +18,7 @@ from __future__ import with_statement
 import base64
 import os
 import binascii
+from io import BytesIO
 
 from libcloud.utils.py3 import ET
 from libcloud.utils.py3 import httplib
@@ -799,6 +800,9 @@ class AzureBlobsStorageDriver(StorageDriver):
         """
         @inherits: :class:`StorageDriver.upload_object_via_stream`
 
+        Note that if ``iterator`` does not support ``seek``, the
+        entire generator will be buffered in memory.
+
         :param ex_blob_type: Storage class
         :type ex_blob_type: ``str``
 
@@ -825,7 +829,12 @@ class AzureBlobsStorageDriver(StorageDriver):
         """
         self._check_values(ex_blob_type, ex_page_blob_size)
         if ex_blob_type == "BlockBlob":
-            iterator.seek(0, os.SEEK_END)
+            try:
+                iterator.seek(0, os.SEEK_END)
+            except AttributeError:
+                buffer = BytesIO()
+                buffer.writelines(iterator)
+                iterator = buffer
             blob_size = iterator.tell()
             iterator.seek(0)
         else:
--- a/libcloud/test/storage/test_azure_blobs.py
+++ b/libcloud/test/storage/test_azure_blobs.py
@@ -874,6 +874,24 @@ class AzureBlobsTests(unittest.TestCase):
         self.assertEqual(obj.size, 3)
         self.mock_response_klass.use_param = None
 
+    def test_upload_blob_object_via_stream_from_iterable(self):
+        self.mock_response_klass.use_param = 'comp'
+        container = Container(name='foo_bar_container', extra={},
+                              driver=self.driver)
+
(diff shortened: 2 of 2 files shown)


c-w:
Note that the build fails but as of today and commit a1f9897 this is also the case for trunk ([sample build](https://travis-ci.org/CatalystCode/libcloud/builds/533938363)).

The failing tests are unrelated to the changes in this pull request, namely:

```
ElasticContainerDriverTestCase.test_create_cluster
ElasticContainerDriverTestCase.test_deploy_container
ElasticContainerDriverTestCase.test_destroy_cluster
ElasticContainerDriverTestCase.test_ex_create_service
ElasticContainerDriverTestCase.test_get_container
ElasticContainerDriverTestCase.test_list_clusters
ElasticContainerDriverTestCase.test_list_containers
ElasticContainerDriverTestCase.test_restart_container
ElasticContainerDriverTestCase.test_start_container
ElasticContainerDriverTestCase.test_stop_container
```

c-w:
Tagging @michaelperel since he last modified this code.

libcloud/storage/drivers/azure_blobs.py
```diff
@@ -825,7 +826,12 @@ def upload_object_via_stream(self, iterator, container, object_name,
         """
         self._check_values(ex_blob_type, ex_page_blob_size)
         if ex_blob_type == "BlockBlob":
-            iterator.seek(0, os.SEEK_END)
+            try:
+                iterator.seek(0, os.SEEK_END)
+            except AttributeError:
+                buffer = BytesIO()
+                buffer.writelines(iterator)
```
Kami, line 836: How does this work - does it need to exhaust and buffer the whole iterator in memory (inside buffer object) or does it read the iterator as it iterates over it and writes it as part of a request?

c-w: Yes, this does buffer the whole iterator in memory unfortunately.

Given that [the content-size header is required](https://docs.microsoft.com/en-us/rest/api/storageservices/put-blob#request-headers-all-blob-types) we have to find the size of the iterator before making the request so I don't see a way to avoid this for the general case. E.g. using [tee](https://docs.python.org/3/library/itertools.html#itertools.tee) followed by something like [ilen](https://more-itertools.readthedocs.io/en/stable/_modules/more_itertools/more.html#ilen) will still copy the iterator in memory.

One potential work-around could be to upload chunks from the iterator via individual [put block](https://docs.microsoft.com/en-us/rest/api/storageservices/put-block) requests followed with a [put block list](https://docs.microsoft.com/en-us/rest/api/storageservices/put-block-list) request, but that would be a much more invasive change to the codebase. (Note that I also haven't proved out this approach in code yet so for now it's just a hypothesis from reading the docs.)

Do you have any suggestions to improve this?

Kami: I'm fine with this for now, but at the very least, we should add a note to the documentation (and docstring) which points out that using that method will buffer the whole object in memory.

I do think that using put block would be a good future improvement (especially for very large objects). IIRC, one of the existing storage drivers already utilizes tsimilar approach, but I can't recall which one (it does require a bit more work though to track all the chunks, etc.).

c-w: Added a note to the docstring for now.

Kami (approved):

Kami:
Merged, thanks.

PR #1289, opened 2019-05-26
Add a workaround for Python 3.7 build failure on Travis

Kami:
It looks like Python 3.7 build fails on Travis because an old version of setuptools library is used.

This pull request should fix the build failure.


Kami:
It looks like it indeed fixed the issue so I went ahead and merged it.

PR #1290, opened 2019-05-28
Various code fixes and improvements

Kami:
This pull request fixes various issues in the code which were mostly identified when running tests under newer versions of Python 3 (newer versions print various deprecation warning and also report open file handle leakage, etc.).

1. Fix various tests and drivers which didn't correctly cleanup and close open files
2. Fix ``libcloud.enable_debug`` debug function so it correctly closes the file handle on exit
3. Update test code to use non-deprecated assertion methods on newer versions of Python 3
4. Various other deprecation errors
5. Ignore user warnings when running tests using tox (this way we silence non-important warnings)
6. Update various tests to use smaller sleep / poll intervals to speed tests up (tests now finish ~40 seconds or so faster)
7. Update py.test config so it prints 10 slowest tests at the end of the test run report. This will allow us to identify and fix slow tests in the future.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,7 +56,7 @@ install:
   - TOX_ENV=py$TRAVIS_PYTHON_VERSION
 
 script:
-  - tox -vvv -e $TOX_ENV
+  - tox -e $TOX_ENV
 
 cache:
   apt: true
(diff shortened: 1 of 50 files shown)


Kami:
Tests run time is now down from ~160 to ~110 seconds. Those small sleep optimizations added up.

PR #1291, opened 2019-05-28
Re-enable PyPy builds

Kami:
This pull request will try to re-enable builds under PyPy. It looks like they were disabled at some point.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,8 +20,8 @@ matrix:
     - python: 3.7
       dist: xenial
       sudo: required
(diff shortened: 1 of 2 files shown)


Kami:
Confirmed it's working - https://travis-ci.org/apache/libcloud/jobs/538387079, https://travis-ci.org/apache/libcloud/jobs/538387080. It does increase overall / aggregate build time, but it's still shorter than it was in the past because of the timing improvements in #1290.

Will go ahead and merge it.

PR #1292, opened 2019-05-30
[LIBCLOUD-1001] Update Cloudflare DNS driver to API v4

c-w:
### Description

As discussed in [LIBCLOUD-1001](https://issues.apache.org/jira/browse/LIBCLOUD-1001), Cloudflare turned off [the old JSON API](https://www.cloudflare.com/api_json.html) and replaced it with [the new v4 API](https://api.cloudflare.com/). This pull request updates the Cloudflare DNS driver to use the new API.

Note that in order to keep the implementation of the v4 driver streamlined, none of the existing `ex_` methods were migrated to the new API.

Note that [LIBCLOUD-994](https://issues.apache.org/jira/browse/LIBCLOUD-994) is also fixed by the new driver implementation.

### Status

- done, ready for review
- [CI passed](https://travis-ci.org/CatalystCode/libcloud/builds/539373218)

- [x] Documentation


c-w:
Tagging @Kami as the original author of the Cloudflare DNS driver.

Kami:
@c-w Thanks for the contribution.

Since it's a bigger change, signed ICLA is required (http://libcloud.readthedocs.org/en/latest/development.html#contributing-bigger-changes).

c-w:
Thanks for the patience @Kami. The ICLA is now signed.

libcloud/dns/drivers/cloudflare.py
```diff
@@ -141,290 +159,252 @@ class CloudFlareDNSDriver(DNSDriver):
         RecordType.URL: 'LOC'
     }
 
+    PAGE_SIZE = 20
```
Kami: Does CloudFlare API support larger page size? If so, using something larger (e.g. 100) might be more efficient in some scenarios for accounts with a lot of domains / records.

libcloud/dns/drivers/cloudflare.py
```diff
@@ -141,290 +159,252 @@ class CloudFlareDNSDriver(DNSDriver):
...
+    def _paginate(self, get_page):
+        for page in itertools.count(start=1):
+            params = {'page': page, 'per_page': self.PAGE_SIZE}
+
+            items = get_page(params)
+
+            for item in items:
+                yield item
+
+            if len(items) < self.PAGE_SIZE:
```
Kami: Does CloudFlare API return some kind of indication in the response headers which could let us know if there are more results / next page?

If so, this could help us avoid unnecessary HTTP request where we are on the last page, but the last page contains ``PAGE_SIZE`` number of items. That is assuming that assuming that the API doesn't return some kind of error if you request a page which doesn't exist.

Kami: EDIT:

Yeah, it looks like it should be contain in the optional ``result_info`` attribute in the response body - https://api.cloudflare.com/

libcloud/dns/drivers/cloudflare.py
```diff
@@ -141,290 +159,252 @@ class CloudFlareDNSDriver(DNSDriver):
...
-        """
-        params = {'z': zone.domain, 'v': 0}
-        self.connection.set_context({'zone_domain': zone.domain})
-        resp = self.connection.request(action='devmode', params=params)
-        result = resp.object
-        return result.get('result', None) == 'success'
+        item = response.object['result']
+        zone = self._to_zone(item)
 
-    def ex_purge_cached_files(self, zone):
```
Kami: It looks like some of those extension methods haven't been implemented in the v4 driver.

Not a big deal for now, but eventually it would be good to implement the ones which are still supported by v4 API.

libcloud/dns/drivers/cloudflare.py
c-w: Good idea. I've increased the page size to the API maximum (instead of the previous API default value) in 9970385.

libcloud/dns/drivers/cloudflare.py
c-w: Good catch. Used the `result_info` object if it's present to avoid the extra HTTP request in 828b471.

libcloud/dns/drivers/cloudflare.py
c-w: I've added placeholders for the extension methods in 48330b7 so that we don't loose track of them. Given that currently the driver is fully non-functional and given that there are quite a few of the extension methods, would you be comfortable with getting the basic API merged and add the extension methods as they are requested by users and/or more development bandwidth frees up?

Kami: Yeah, I agree, not a blocker.

Kami (approved):
LGTM.

Kami:
I think it would be good to release v2.6.0 in the near future once this PR and a couple of other ones are merged.

Do you want to try the release process for v2.6.0?

c-w:
Thanks for the review @Kami and yes, I'd be happy to go through the release flow for `v2.6.0`.

PR #1295, opened 2019-06-05
Fix error getting floating ip instance_id

micafer:
### Description

In some cases the `device_owner` of the port can be `compute:None`. In this cases the float ip does not return the correct instace_id where the float ip is attached.

### Status

- done, ready for review

- [ ] Documentation


Kami:
Thanks for the contribution.

I will go ahead and merge it into master, but it would be great if you could also add a corresponding test case for it.

PR #1300, opened 2019-06-30
ec2 list_volumes should accept more filters

palashgandhi:
### Description

EC2 API requests are throttled which sometimes requires the API requests to be more fine tuned to avoid having to get back large amounts of responses. For more information, see this [the AWS API reference](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/query-api-troubleshooting.html#api-request-rate).

This change adds the ability to pass more filters to the EC2 DescribeVolumes call, the filter needs to be a dictionary of filter name and value pairs as described [here](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeVolumes.html).

I had 2 options here:
1. Remove the `node` parameter from this call and ensure that the caller passes in the node ID as part of the `filters` dict.
2. Add a new parameter `filters` and leave the `node` parameter as it is in order to preserve the original method signature.

I chose the latter only for the reason that it preserves the original method signature.

### Status

- done, ready for review

- [x] Documentation

### DIFF
--- a/libcloud/compute/drivers/ec2.py
+++ b/libcloud/compute/drivers/ec2.py
@@ -1783,13 +1783,29 @@ class BaseEC2NodeDriver(NodeDriver):
             )
         return locations
 
-    def list_volumes(self, node=None):
+    def list_volumes(self, node=None, ex_filters=None):
+        """
+        List volumes that are attached to a node, if specified and those that
+        satisfy the filters, if specified.
+
+        :param node: The node to which the volumes are attached.
+        :type node: :class:`Node`
+
+        :param ex_filters: The dictionary of additional filters.
+        :type ex_filters: ``dict``
+
+        :return: The list of volumes that match the criteria.
+        :rtype: ``list`` of :class:`StorageVolume`
+        """
         params = {
             'Action': 'DescribeVolumes',
         }
+        if not ex_filters:
+            ex_filters = {}
         if node:
-            filters = {'attachment.instance-id': node.id}
-            params.update(self._build_filters(filters))
+            ex_filters['attachment.instance-id'] = node.id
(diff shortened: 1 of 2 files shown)


libcloud/compute/drivers/ec2.py
```diff
@@ -1783,12 +1783,28 @@ def list_locations(self):
             )
         return locations
 
-    def list_volumes(self, node=None):
+    def list_volumes(self, node=None, filters=None):
```
Kami: Since it's an extension argument please rename it to ``ex_filters``.

palashgandhi: Fixed.

libcloud/test/compute/test_ec2.py
```diff
@@ -1548,6 +1553,26 @@ def _DescribeVolumes(self, method, url, body, headers):
...
+            'Filter.1.Value.1': 'i-d334b4b3',
+        }
+
+        expected_params_2 = {
+            'Filter.1.Name': 'status',
+            'Filter.1.Value.1': 'available'
+        }
+
+        try:
+            self.assertUrlContainsQueryParams(url, expected_params_1)
```
Kami, line 1568: Eventually we should improve this utility method so the order doesn't matter :)

Kami:
Merged, thanks for the contribution.

libcloud/test/compute/test_ec2.py
Kami: I just checked and see the method already supports that.

It's just the way tests are structured. This should do it - f8deceb04c41e36bc3002b3fd9c4f0d6e752b383.

PR #1305, opened 2019-07-05
adding gridscale driver to libcloud

PrinceSydney:
### Description

This change brings in the driver for the cloud provider gridscale.
This pull request will contribute the necessary changes in libcloud, for gridscale to support a libcloud implementation.

### Status

- done, ready for review

### Checklist

- [x] Documentation


Kami:
Thank you for your contribution.

We will review it when we get a chance.

libcloud/compute/drivers/gridscale.py
```diff
@@ -0,0 +1,977 @@
...
+
+        :return: List of node objects
+        :rtype: ``list`` of :class:`.Node`
+        """
+        result = self._sync_request(data=None, endpoint='objects/servers/')
+        nodes = []
+        for key, value in self._get_response_dict(result).items():
+            node = self._to_node(value)
+            nodes.append(node)
+            continue
```
Kami: This continue here is redundant, but not a blocker.

libcloud/compute/drivers/gridscale.py
```diff
@@ -0,0 +1,977 @@
...
+        :type size: :class:`.NodeSize`
+
+        :param image: OS image to attach to the storage.
+        :type image: :class:`.GridscaleTemplate`
+
+        :param location: The data center to create a node in.
+        :type location: :class:`.NodeLocation`
+
+        :param auth: sshkey uuid.
+        :type auth: :class:`.NodeAuthSSHKey`
```
Kami: It looks like the ``auth`` argument is actually a string which contains an ID (UUID) of a SSH key which needs to be available on the account?

libcloud/compute/drivers/gridscale.py
```diff
@@ -0,0 +1,977 @@
...
+        :type snapshot: :class:`.VolumeSnapshot`
+
+        :return: Newly created StorageVolume.
+        :rtype: :class:`.StorageVolume`
+        """
+
+        return self._create_volume_from_template(size, name, location)
+
+    def _create_volume_from_template(self, size, name, location=None,
+                                     template={}):
```
Kami: Mutable defaults are not really safe in Python - https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

Please default it to ``None`` and do something along the lines of ``template = template or {}`` inside the method body.

libcloud/compute/drivers/gridscale.py
```diff
@@ -0,0 +1,977 @@
...
+        :return: True if the destroy was successful, otherwise False
+        :rtype: ``bool``
+
+        """
+        result = self._sync_request(endpoint='objects/templates/{}'
+                                    .format(node_image.id),
+                                    method='DELETE')
+        return result.status == 204
+
+    def patch_node(self, node, name=None):
```
Kami: This method and ``patch_storage`` is not part of the standard Libcloud API, so it should be prefixed with ``ex_`` (``ex_patch_node``, ``ex_patch_storage``).

Actually, it looks like all those two methods do is rename a node / storage volume, so they should probably be called something along the lines of ``ex_rename_node`` and ``ex_rename_volume``.

libcloud/compute/drivers/gridscale.py
```diff
@@ -0,0 +1,977 @@
...
+        :return: ``True`` or ``False``
+        :rtype: ``bool``
+        """
+        result = self._sync_request(data=json.dumps({'name': name}),
+                                    endpoint='objects/servers/{}'
+                                    .format(storage.id),
+                                    method='PATCH')
+        return result.status == 204
+
+    def ex_patch_networks(self, network, name=None):
```
Kami: Same here, ``ex_rename_network`` is probably a better and more explicit name.

libcloud/compute/drivers/gridscale.py
```diff
@@ -0,0 +1,977 @@
...
+        :type storage: ``object``
+
+        :param name: rename storage
+        :type name: ``str``
+
+        :return: ``True`` or ``False``
+        :rtype: ``bool``
+        """
+        result = self._sync_request(data=json.dumps({'name': name}),
+                                    endpoint='objects/servers/{}'
```
Kami: I assume that's a typo and should be `endpoint='objects/storages/{}'`?

Having a test case which also verifies that a particular driver method hits the correct API endpoint also wouldn't hurt :)

libcloud/compute/drivers/gridscale.py
```diff
@@ -0,0 +1,977 @@
...
+
+        """
+
+        if node.extra['power'] is True:
+            data = dict({'power': False})
+            self._sync_request(data=json.dumps(data),
+                               endpoint='objects/servers/{}/power'
+                               .format(node.id),
+                               method='PATCH')
+            time.sleep(3)
```
Kami: Perhaps we should make this value user-configurable - e.g. ``ex_sleep_interval=3``.

Also, I assume this perform a "hard" power off and not a soft one? For a soft one, I would imagine 3 seconds is way too low.

libcloud/compute/drivers/gridscale.py
Kami: If that's indeed the case, this method should take ``ex_ssh_key_ids`` argument instead.

Right now it looks like it actually takes ssh key id, but it incorrectly uses ``NodeAuthSSHKey`` class which expects a public key which is to be installed on the server.

Here is an example of using ``ex_ssh_key_ids`` argument - https://github.com/apache/libcloud/blob/trunk/libcloud/compute/drivers/digitalocean.py#L132.

Eventually, we could also add new ``NodeAuthSSHKeyId`` class (or similar) and make it part of a standard API and compatible with "auth" argument.

libcloud/compute/drivers/gridscale.py
```diff
@@ -0,0 +1,977 @@
...
+
+        extra = self._extract_values_to_dict(data=data, keys=extra_keys)
+        ips = []
+
+        for diction in data['relations']['public_ips']:
+            ips.append(diction['ip'])
+
+        state = ''
+
+        if data['power'] is True:
```
Kami, line 844: Hm, interesting, does Gridscale only support two server states?

Usually providers support a lot more intermediate states such as "pending", "powering on", etc.

Kami:
@PrinceSydney Thanks again for the contribution. I finally had a chance to review it today.

Overall, the PR looks like a great start, but there are some small things which still need to be improved or fixed.

Once those comments are addressed, please also make sure the build passes (it looks like it's failing now).

libcloud/compute/drivers/gridscale.py
PrinceSydney: The ssh key UUID is required to be able to attach an image to a node, while in the process of creating this node.

Kami: Yes, but you should utilize ``ex_ssk_key_ids`` argument for that and not ``auth`` one.

Kami: To clarify - ``ex_ssk_key_ids`` argument contains a list of key ids which are to be attached to the node upon creation.

On the other hand, ``NodeAuthSSHKey`` instance and the ``auth`` argument should contain a raw public key which is to be installed on the node.

PrinceSydney: I've noticed it when i looked over the example you provided, this was very understandable.
thanks for the short summary :) .

Kami: Also, I had a look at the API docs and I couldn't find anything about ``shkeyids`` attribute for the image templates.

I assume that's implementation details and that's how images are implemented?

Kami:
I've made some fixes and changes and merged PR into master (please see a list of changes I made here - https://github.com/apache/libcloud/commits/trunk).

In theory ``deploy_node`` should also work now, but I didn't test it. Something along those lines:

```python
ssh_key = driver.list_key_pairs()[0]
ssh_key_uuid = ssh_key.fingerprint

auth = NodeAuthSSHKey(ssh_key.pubkey)

node = conn.deploy_node(name='test', image=image], size=size, auth=auth,
                        deploy=..., ex_ssh_key_ids=[ssh_key_uuid])
```

Thanks again for the contribution.

libcloud/compute/drivers/gridscale.py
PrinceSydney: Thats right. There are a few things the API doesn't show just yet but will be in the future.

PR #1306, opened 2019-07-09
Add Python type annotation for base Compute API

Kami:
The goal of this pull request is to add type annotations to the base APIs in an incremental manner.

Adding type annotations to an existing project is much harder (and takes much more time) compared to adding it as you develop a new project and that's why I will take an incremental approach.

I plan to start with the base APIs and see how it goes. It's likely that those type annotations will require various issues in the driver code which will be hard and time consuming to fix (and that's also the reason why I plan to start with the base APIs).


Kami:
It turns out this will be quite a challenging and painful task.

Existing type annotations I added already uncovered some issues with sub-classes not handling parent types correctly, etc.

We don't have much choice here, but to use a type annotation work around in such scenarios (supporting multiple types aka Union), because we can't break the API.

In addition to that, it will also be hard to come up with good type annotations for our dynamic ``get_driver`` and other factory methods.

Kami:
In addition to my comment above - I also decided I will start with an API per pull request to make sure the PRs don't grow too large.

In this PR I will work on the compute API and other common code which is needed to get that to work.

As far as type annotations for our dynamically returned provider drivers go - I still haven't been able to figure this out.

We may just need to go with returning a reference to the base NodeDriver. It's not ideal since it won't know about any provider class specific methods and arguments, but I'm not even sure yet if we can correctly annotate our code which does our dynamic driver class loading (we may need a MyPy plugin or it may even not be possible).

Kami:
@tonybaloney ^ maybe you have an idea if we can get type annotations to work for our dynamically loaded and returned driver classes.

For now, I have this - f5a201637f517c86128ef049314996e3bc97e195. As mentioned above, it's not ideal since it won't be aware of any provider driver class specific methods and arguments.

I did quite a lot of research and experimentation and I couldn't get it to work (aka annotate ``get_driver`` method that it returns a reference to provider driver class which is dependent on the value of provided ``provider`` argument constant).

I believe we may only be able to achieve that using a custom MyPy plugin (but I could be wrong).

Kami:
This is "mostly" working with limitations mentioned above (get_driver always returns reference to the base driver).

I still need to fix enum handling (I needed to change our custom ENUM type so it works correctly with MyPy), then I can hopefully merge it into trunk.

Kami:
I finally managed to get all the tests to pass.

Sadly it wasn't that straightforward. One of the main offenders was our custom "enum" type and making sure it works with type annotations and it's also fully backward compatible. For that, I needed to implement a bunch of "glue" code - 9604610fdd3bc9c86a5c7cc5ca9d42bf2d100be8

Kami:
@c-w maybe you have an idea on how (if possible) to handle that in mypy^ - https://github.com/apache/libcloud/pull/1306#issuecomment-509980658

c-w:
@Kami I'm not sure this is currently possible in mypy since we only know the type of the returned driver value at runtime based on the provider argument.

The current approach looks good to me and may even encourage more users to stick to the methods defined in the base API (a win in my mind since it'll make code more easily cross-cloud portable). We do lose mypy coverage for any `ex_` methods, but if a user has code that relies on those, they can always work around it like this:

```py
driver: MySpecificDriver = get_driver(...)
```

Kami:
Thanks.

And yes, I believe what you said should work (I will test it and document it).

I plan to release release a candidate / beta version with those changes (probably after the next release) to make sure it doesn't break anything. Only thing I'm really worried about is the enum changes I made.

Kami:
I'll work on a patch release first (``2.6.1``) and after that's out, merge this into trunk and prepare release candidate / beta releases with those changes for people to test them out.

Kami:
@c-w That's what I came up with - 428c3b63.

For some reason, mypy doesn't recognize driver classes (e.g. ``EC2NodeDriver`` to be a subclass of ``NodeDriver``), so the following doesn't work:

```python
EC2 = get_driver(Provider.EC2)  # type: Type[EC2NodeDriver]
Rackspace = get_driver(Provider.RACKSPACE)  # type: Type[RackspaceNodeDriver]
```

The approach I used is not ideal, but I confirmed it works.

I'm not sure why it doesn't recognize it (I didn't dig in too much yet), I assume it's something to do with scoping.

Might even need to open mypy issue / question about it, but I assume they will come back to us with that our code already relies on too much "magic" and meta programming :D

Kami:
It turns out this PR inadvertently introduced dependency on ``requests`` when running ``setup.py`` install and we didn't have any install tests which would have caught that.

I believe I fixed that in 4e42ea3c5a7c4cc83dae53958cdc72f9b5df2750, 5159babc482eb4cf8c87215d53dfe71f828a3a0c and added a check in 75f07bd2c75ae6b82ccc07090bdefaa07b9938b8.

EDIT: Due to libcloud now also depending on enum and typing package, this needed more work - 21a78a17929f0633817c337208ab2a21dc0639f9.

This approach is much safer anyway. We shouldn't have ``setup.py`` importing any code from libcloud package which is to be installed...

This should also fix read the docs build.

Kami:
Just a heads up - the plan is to add ``py.typed`` file which declares Python packages includes type annotations once type annotations are included for all the base APIs.

Kami:
@c-w I just found out today that casting to a driver type doesn't work as expected - https://github.com/python/mypy/issues/3903 :/

MyPy won't inherit type annotations from the base class if subclass overrides the method (it doesn't matter if it preserves the parent method signature).

Pretty much only option we have at this point is to add type annotations to all the subclassed methods.

This is obviously far from ideal :/

---

I think only reasonable option we have right now is to simply rely on type annotations from the base driver class.

Going forward, we will probably need to use something like this - https://github.com/python/mypy/pull/7548 and also add type annotations to any subclass methods where signature differs from the parent / base class method.

Issue #1309, opened 2019-07-11
Error in resize function in OpenStack driver


micafer:
When calling the "ex_resize" function of the OpenStack driver I get this error:
```
BaseHTTPError: 400 Bad Request Invalid input for field/attribute resize. Value: {u'flavorRef': u'8', 
 u'personality': [], u'name': u'userimage-156283439906', u'imageRef': u'0de96743-4a12-4470-b8b2-
 6dc260977a40', u'metadata': {}}. Additional properties are not allowed (u'metadata', u'name', 
 u'imageRef', u'personality' were unexpected)
```

PR #1311, opened 2019-07-12
Fix error in resize function in OpenStack driver #1309

micafer:
### Description

Fix the error in resize function in OpenStack driver described in issue #1309

### Status

- done, ready for review

- [ ] Documentation

### DIFF
--- a/libcloud/compute/drivers/openstack.py
+++ b/libcloud/compute/drivers/openstack.py
@@ -1588,7 +1588,7 @@ class OpenStack_1_1_NodeDriver(OpenStackNodeDriver):
 
         :rtype: ``bool``
         """
-        server_params = self._create_args_to_params(node, size=size)
+        server_params = {'flavorRef': size.id}
         resp = self._node_action(node, 'resize', **server_params)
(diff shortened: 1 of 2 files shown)


Kami (approved):
LGTM. Thanks.

libcloud/compute/drivers/openstack.py
```diff
@@ -1588,7 +1588,7 @@ def ex_resize(self, node, size):
 
         :rtype: ``bool``
         """
-        server_params = self._create_args_to_params(node, size=size)
+        server_params = {'flavorRef': size.id}
```
Kami, line 1591: P.S. Next time it would help if you can link to the corresponding API documentation (e.g. https://developer.openstack.org/api-ref/compute/?expanded=resize-server-resize-action-detail in this case).

Kami, line 1591: For consistency, we should also rename this to ``ex_resize_node``. I will make that change in trunk and leave ``ex_resize`` there for backward compatibility reasons.

PR #1315, opened 2019-07-17
Various Gridscale driver and paramiko SSH client improvements

Kami:
While testing Gridscale drivers I noticed some issues which I originally missed during the code review.

This pull request fixes some of those issues and implements some new functionality for the completeness sake.

1. Add missing ``encode_data()`` method on the Connection class and rely on that for encoding request body (data) instead of doing it manually instead the driver methods.

   This way it follows Libcloud API which means it works correctly with logging connection and other abstractions.

2. Implement ``import_key_pair_from_string`` / ``import_key_pair_from_file`` method.

3. Add new ``ex_list_volumes_for_node`` and ``ex_list_ips_for_node`` methods and ability to delete associated resources when destroying a node

4. Support for ``deploy_node()`` functionality. Needed various improvements in the non-driver code.

5. Add some tests which were missing and make some of the existing tests more robust / strict by also checking HTTP method name inside the mock response class.

6. Some small bug fixes in the driver.


Kami:
@PrinceSydney just a heads up - a5a79a6, b545797.

I missed that during the review.

Encoding data inside the driver class would work as far as drive functionality is concerned, but it doesn't follow the Libcloud way of serializing request data and as such, it would break things like the debug functionality - https://libcloud.readthedocs.io/en/latest/troubleshooting.html#debugging

Kami:
I tried to test and get deployment to work, but it turned out it a huge time sink.

Main problem is that paramiko only supports key files in old PKCS#1 PEM format. New style OpenSSH keys or key files in PEM format without `---BEGIN <type> PRIVATE KEY` header are not supported.

I updated the code to handle such scenario better:

1. Propagate "invalid key file" exception to the user instead of re-trying the deployment.
2. If user passes in a key file which contains private key in a PEM format, but without a valid header we try to manually convert into into a format which paramiko accepts.

We also need to add a note to our deployment documentation which clearly says what kind of keys paramiko supports so users won't waste as much time as I did on this.

Kami:
@PrinceSydney here is an example of ``deploy_node`` functionality - 
73d1d04.

Issue #1317, opened 2019-07-18
azure_arm: delete public IP address (ex_delete_public_ip)


redha1419:
## Feature Request

This feature is just a function with in the azure_arm driver. Lib cloud users can now delete Azure ARM public ip address resources. the ex_delete_resource() method could not be used as  `RESOURCE_API_VERSION` is too old to support the publicIpAddress resource.

PR #1318, opened 2019-07-18
added azure_arm ex_delete_public_ip

redha1419:
### Description

This feature is just a function with in the azure_arm driver. Lib cloud users can now delete Azure ARM public ip address resources. the ex_delete_resource() method could not be used as RESOURCE_API_VERSION is too old to support the publicIpAddress resource.

### Status
- testing
- done, ready for review

### Checklist

- [x] Documentation

### DIFF
--- a/libcloud/compute/drivers/azure_arm.py
+++ b/libcloud/compute/drivers/azure_arm.py
@@ -1873,7 +1873,7 @@ class AzureNodeDriver(NodeDriver):
         :param public_ip: Public ip address resource to delete
         :type public_ip: `.AzureIPAddress`
         """
-        resource = resource.id
+        resource = public_ip.id
         self.connection.request(
             resource,
             method='DELETE',
--- a/libcloud/compute/drivers/azure_arm.py
+++ b/libcloud/compute/drivers/azure_arm.py
@@ -1866,6 +1866,22 @@ class AzureNodeDriver(NodeDriver):
                                     method='PUT')
         return self._to_ip_address(r.object)
 
+    def ex_delete_public_ip(self, public_ip):
+        """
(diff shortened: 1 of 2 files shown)


redha1419:
https://github.com/apache/libcloud/issues/1317

Kami:
@redha1419 Thanks for the contribution.

The change looks good to me, but can you please add a test case for it? You can find other example test cases in ``libcloud/test/compute/test_azure_arm.py``

Kami:
@redha1419 Thanks for working on the tests. I helped you a bit and added missing fixtures - d63de53644aad61fdddaa312fd7343cee2f552d9.

In addition to that, it looks like this method is not really needed per say since there already is an existing ``ex_delete_resource`` method which should work with arbitrary resource, including IP address.

So in a sense, this new method is really just an alias for ``ex_delete_resource`` method - c83b61d8d9fd8214469122d71b73c8b4da2700f7.

Kami:
EDIT: I was wrong - about ``ex_resource_delete`` working for IP addresses.

It looks like IP address functionality is part of the new API version (https://docs.microsoft.com/en-us/rest/api/virtualnetwork/publicipaddress(preview)/delete) and code change in c83b61d won't actually work.

I'll revert my change in c83b61d and leave your original code change which uses newer API version in place.

EDIT: Here we go - 52195a6e9d15a31d0668d90111542435d638b62b. Next time would also be good to add a code comment (like I did) to make it clear why we aren't directly calling ``ex_delete_resource`` method.

redha1419:
Yes I should have left a comment about the API version. Thanks so much for looking into it though. @Kami.

PR #1323, opened 2019-07-25
Add Gandi LiveDNS driver

zepheiryan:
## Add Gandi LiveDNS driver

[Gandi](https://gandi.net) released an API for their DNS service completely different from their previous XMLRPC implementation.  This contribution adds a driver to cover the new service, per [LIBCLOUD-949](https://issues.apache.org/jira/browse/LIBCLOUD-949).

### Description

This is a new driver and does not touch on any of the previous Gandi XMLRPC driver code.  The implementation is similar to other providers that use HTTP and JSON.

Gandi's [documentation](https://doc.livedns.gandi.net/) is the overall basis for the implementation though there are some minor discrepancies in it that don't match up to usage experience.  They also left out any documentation of errors.

One point of possible confusion is that zones and domains are separate classes in this service, which enables reuse of one zone file for multiple domains but becomes somewhat difficult to shoehorn into Libcloud's classes, especially since what's referred to as a Libcloud zone corresponds to a Gandi domain, and nothing really corresponds to a Gandi zone.  There is precedence since this is also true of the existing Gandi XMLRPC implementation, which appears to me to have chosen to go with ignoring Gandi zones as much as possible in favor of domains.  The new Gandi API does allow for this (actions taken on a domain will apply to the zone and all related domains behind the scenes), but it means things like `delete_zone` and `update_zone` have no real meaning - one cannot delete a Gandi domain from the API, and it is basically immutable if ignoring Gandi zones - so I didn't implement those methods.

### Status

Done, ready for review

- [x] Documentation


libcloud/common/gandi_live.py
```diff
@@ -0,0 +1,250 @@
...
+            httplib.CREATED,
+            httplib.NO_CONTENT
+        ]
+        if self.status in valid_http_codes:
+            if json_error:
+                raise JsonParseError(body, self.status)
+            else:
+                return body
+        elif self.status == httplib.NOT_FOUND:
+            if (not json_error) and ('cause' in body):
```
Kami: Parenthesis are redundant here, but also not a blocker.

libcloud/common/gandi_live.py
```diff
@@ -0,0 +1,250 @@
...
+                message = body
+            raise InvalidRequestError(message, self.status)
+        elif self.status == httplib.CONFLICT:
+            if (not json_error) and ('cause' in body):
+                message = self._get_error(body)
+            else:
+                message = body
+            raise ResourceConflictError(message, self.status)
+        else:
+            if (not json_error) and ('cause' in body):
```
Kami: This logic seems to be repeated 4 times, perhaps it could be refactored inside a separate method (maybe even the existing ``_get_error`` method)?

libcloud/dns/drivers/gandi_live.py
```diff
@@ -0,0 +1,347 @@
...
+    """
+    def create_zone(self, domain, type='master', ttl=None, extra=None):
+        if extra and 'name' in extra:
+            zone_name = extra['name']
+        else:
+            zone_name = '%s zone' % domain
+        raw_zone_data = {
+            'name': zone_name,
+        }
+        post_zone_data = json.dumps(raw_zone_data)
```
Kami: This encoding should happens inside the connection class - please add ``encode_data`` to the connection class and remove data encoding inside the driver methods.

For details and context, please see https://github.com/apache/libcloud/pull/1315#issuecomment-512226526

libcloud/test/dns/test_gandi_live.py
```diff
@@ -0,0 +1,174 @@
...
+        self.assertEqual(zone.domain, 'example.net')
+        self.assertIsNone(zone.ttl)
+
+    def test_create_zone(self):
+        zone = self.driver.create_zone('example.org',
+                                       extra={'name': 'Example'})
+        self.assertEqual(zone.id, 'example.org')
+        self.assertEqual(zone.domain, 'example.org')
+
+    def test_get_zone(self):
```
Kami, line 80: Nice work on the tests :+1: 

It would also be good to add some tests for failure scenarios / edge cases (aka record doesn't exist, zone doesn't exist, record validation failure, etc.).

libcloud/test/dns/test_gandi_live.py
```diff
@@ -0,0 +1,174 @@
...
+        self.assertIsNone(zone.ttl)
+
+    def test_list_records(self):
+        records = self.driver.list_records(self.test_zone)
+        self.assertEqual(len(records), 3)
+        record = records[0]
+        self.assertEqual(record.id, 'A:@')
+        self.assertEqual(record.name, '@')
+        self.assertEqual(record.type, RecordType.A)
+        self.assertEqual(record.data, ['127.0.0.1'])
```
Kami: Hm, can single record contain multiple values?

I believe we handle this inside other drivers by creating multiple ``Record`` class instances (all of them share the same name, but have a different ``value`` which is a string), but I'm not 100% and I would need to double check.

Kami:
Thanks for the contribution.

Overall it looks like a great start. There are just some minor things which need to be sorted out.

As far as ``Record.data`` field goes - I checked and I see that not all of the drivers follow base API which says ``Record.data`` field should be a string (that's not great since it breaks some of the portability aspect, we probably missed some of that during code reviews).

I still think it would be better to return multiple ``Record`` objects in such scenario, to comply with the base API.

Here is an example of how a similar scenario is handled in the AWS Route 53 driver - https://github.com/apache/libcloud/blob/trunk/libcloud/dns/drivers/route53.py#L447.

I know that means more work in the driver, but that's kinda the whole idea behind the abstraction - we should try to expose the same interface everywhere where possible.

libcloud/common/gandi_live.py
zepheiryan: I originally grabbed most of this from the Google common lib.  Fixed this and the previous one - does it matter who clicks the 'resolve conversation' button (if it's used)?

libcloud/dns/drivers/gandi_live.py
zepheiryan: Ah, didn't realize that was a thing.  Fixed.

zepheiryan:
> I know that means more work in the driver, but that's kinda the whole idea behind the abstraction - we should try to expose the same interface everywhere where possible.

Sure.  Yeah, that complicates it since the API itself is returning a JSON array there.  It might be a bit until I get to adding error tests and putting this more in line with the base Libcloud API.

libcloud/test/dns/test_gandi_live.py
zepheiryan: More tests added that increase coverage.

zepheiryan:
I've made a change to mimic Route53 in part, but this is another conceptual mapping problem - a record in Gandi encompasses all values, whereas in Route53 each individual value constitutes a record, and multiple values are a convenience.  So it's possible there to delete one value (record), whereas with Gandi you can only delete all values.  If you want to change the number of values associated with a name-type, you basically have to delete and recreate the record instead of updating it (whereas if you could pass in the new set of values directly, you could just call update).  This `update_record` is still mostly useful just because most records are unlikely to have multiple values, but beyond that you'd have to know something about driver specifics.

Kami:
Thanks for making those changes - I will have a look.

And yes, edge cases like that can be challenging. In theory (from end user friendliness and standard libcloud API perspective) in such scenarios, Libcloud driver should handle this transparently - aka if user updates such record, existing one is deleted and new one is created.

This is obviously more complex thought and also introduces more (partial) failure scenarios since single method call now results in multiple API calls.

Kami:
Merged, thanks.

PR #1324, opened 2019-07-26
Fix proxy url

gvengel:
### Description

I ran into a couple issues when trying to use the cloudfiles driver behind an HTTP proxy. 

* libcloud was not configuring the requests library proxy settings correctly.
* The cloudfiles driver was not passing the proxy into the identity connection.
* The proxy documentation was referencing the wrong argument name.

This PR closes these issues, but there may be other related issues in other drivers.

### Status

Done, ready for review. This could probably use some additional tests, and a review of other code for similar bugs. It looks like 2.5.1 has other proxy changes, so I wanted to get the PR in sooner rather than later and kept it simple.

### Checklist

- [x] Documentation
- [] [ICLA](http://libcloud.readthedocs.org/en/latest/development.html#contributing-bigger-changes) (required for bigger changes)

### DIFF
--- a/docs/other/using-http-proxy.rst
+++ b/docs/other/using-http-proxy.rst
@@ -73,10 +73,10 @@ With basic auth authentication (http proxy):
 
     http_proxy=http://<username>:<password>@<proxy hostname>:<proxy port> python my_script.py
 
-2. Passing ``http_proxy`` argument to the connection class constructor
+2. Passing ``proxy_url`` argument to the connection class constructor
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-By passing ``http_proxy`` argument to the
+By passing ``proxy_url`` argument to the
 :class:`libcloud.common.base.Connection` class constructor, you can specify
 which proxy to use for a particular connection.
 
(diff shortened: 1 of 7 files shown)


Kami:
Thanks for the contribution. I will look into and test the changes.

There were indeed some bugs with handling of HTTP proxies which were supposed to be fixed in - #1314.

Biggest problem is that not all the drivers correctly pass all the constructor arguments to the parent driver and connection class and as such, proxy is not set when using constructor argument approach. I believe that's also the issue you hit with the OpenStack driver.

In such scenario, it's recommended to use either environment variable or ``driver.connection.connection.set_http_proxy`` method approach.

I will also make sure to add a note about that to the documentation.

docs/other/using-http-proxy.rst
```diff
@@ -73,10 +73,10 @@ With basic auth authentication (http proxy):
 
     http_proxy=http://<username>:<password>@<proxy hostname>:<proxy port> python my_script.py
 
-2. Passing ``http_proxy`` argument to the connection class constructor
```
Kami, line 76: Good catch. This was indeed a typo :)

libcloud/http.py
```diff
@@ -102,7 +102,8 @@ def set_http_proxy(self, proxy_url):
         self.http_proxy_used = True
 
         self.session.proxies = {
-            self.proxy_scheme: proxy_url
+            'http': proxy_url,
```
Kami, line 105: I tested it and it works fine with just a single attribute set, but it also shouldn't hurt to set both (http and https).

Kami:
I've added a note to the documentation on support drivers not supporting constructor method (62c52c027823f23db1b03286f40b4d39f2c5468d) and will go ahead and merge this into trunk.

Kami:
Merged, thanks.

libcloud/http.py
gvengel: This was definitely broken for me. I tested with various versions of requests just in case. Perhaps you have other mitigating factors, like setting the http_proxy and https_proxy environment variables, or only making http requests and not https, or vice versa? 

Anyway, like you say it can't hurt to set both, and the code's been merged, so yay!

If I have some spare time, I'll looking into writing some integration tests for this.

PR #1326, opened 2019-07-26
Fix hash validation with stream uploads

gvengel:
### Description

Bug fix for cloudfiles upload_object_via_stream() erroneously throwing a hash mismatch when using a file handle as the iterator. Example:

```python
from libcloud.storage.types import ObjectDoesNotExistError

driver = get_driver('cloudfiles')
driver = driver('user', 'key')
container = driver.get_container(container_name='test')
with open('/tmp/test.file', 'w') as fh:
    fh.write('test')
with open('/tmp/test.file') as fh:
    driver.upload_object_via_stream(iterator=fh, container=container, object_name='test.file')
```

This would result in the exception:

```
libcloud.storage.types.ObjectHashMismatchError: <ObjectHashMismatchError in <libcloud.storage.drivers.cloudfiles.CloudFilesStorageDriver object at 0x103f43d68>, value=MD5 hash checksum does not match (expected=d41d8cd98f00b204e9800998ecf8427e, actual=098f6bcd4621d373cade4e832627b4f6), object = test.file>
```

d41d8cd98f00b204e9800998ecf8427e is the hash for an empty file. The problem is the upload happens before the hash is calculated and the file handle is read to the EOF. The fix is to seek the file handle to the beginning before feeding it into the hash function.

### Status

done, ready for review

- [ ] Documentation

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -164,6 +164,11 @@ Storage
   ``2012-02-12`` to ``2014-02-14'``. (LIBCLOUD-851, GITHUB-1202, GITHUB-1294)
   [Clemens Wolff - @c-w, Davis Kirkendall - @daviskirk]
 
+- [Common, CloudFiles] Fix ``upload_object_via_stream`` via stream method and
+  ensure we rewind to the beginning of the provided iterator in case iterator
+  has already been iterated or is not positioned at the begining. (GITHUB-1326)
+  [Gabe Van Engel - @gvengel, Tomaz Muraus]
+
 DNS
 ~~~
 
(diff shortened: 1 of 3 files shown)


Kami:
Thanks for the contribution.

Can you please also add a corresponding test case?

Kami:
I've added a test case (858d9e6e16d0f1a2160537d9acac1ca0f897ef70) and will go ahead and merge this into trunk.

It's also worth pointing out that you made change in the base driver class so this change potentially affects all the drivers. I think the change is the correct one, but we should still keep an eye on potential issues once a new version is released.

gvengel:
Sorry, I've been busy lately, but did intend to add tests. Thanks for doing it for me, @Kami!

PR #1328, opened 2019-07-31
powerdns driver adapted to work with pdns 3 and 4

biggosh:
### Description

Added support for PowerDNS API version 1 maintaining the backward compatibility with experimental API version.

### DIFF
--- a/libcloud/dns/drivers/powerdns.py
+++ b/libcloud/dns/drivers/powerdns.py
@@ -23,6 +23,7 @@ from libcloud.dns.base import DNSDriver, Zone, Record
 from libcloud.dns.types import ZoneDoesNotExistError, ZoneAlreadyExistsError
 from libcloud.dns.types import Provider, RecordType
 from libcloud.utils.py3 import httplib
+import pprint
 
 __all__ = [
     'PowerDNSDriver',
@@ -143,7 +144,7 @@ class PowerDNSDriver(DNSDriver):
                                              host=host, port=port,
                                              **kwargs)
 
-    def create_record(self, name, zone, type, data, extra=None):
+    def create_record(self, name, zone, type, data, extra=None, disabled=False, comment=None):
         """
         Create a new record.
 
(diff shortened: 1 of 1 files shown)


biggosh:
Hi. I've checked the error, but I can't understand the reason.

libcloud/dns/drivers/powerdns.py
```diff
@@ -143,7 +144,7 @@ def __init__(self, key, secret=None, secure=False, host=None, port=None,
                                              host=host, port=port,
                                              **kwargs)
 
-    def create_record(self, name, zone, type, data, extra=None):
+    def create_record(self, name, zone, type, data, extra=None, disabled=False, comment=None):
```
Kami: Those are extension arguments (arguments which are not supported by all the providers), please prefix them with ``ex_``.

So ``ex_disabled`` and ``ex_comment``.

libcloud/dns/drivers/powerdns.py
```diff
@@ -368,7 +392,7 @@ def list_zones(self):
         response = self.connection.request(action=action, method='GET')
         return self._to_zones(response)
 
-    def update_record(self, record, name, type, data, extra=None):
+    def update_record(self, record, name, type, data, extra=None, disabled=False, comment=None):
```
Kami: Same here.

Kami:
Thanks for the contribution.

It would also be good to add some test cases which test both API version - 3 and 4.

libcloud/dns/drivers/powerdns.py
```diff
@@ -439,13 +485,27 @@ def _to_zones(self, items):
...
+            for item in items.object['records']:
+                records.append(self._to_record(item, zone))
+        if self._pdns_version() == 4:
+            for item in items.object['rrsets']:
+                for record in item['records']:
+                    records.append(self._to_record(item, zone, record))
         return records
+		
+    def _pdns_version(self):
+        if self.api_root == '': return 3
```
Kami, line 515: How is version currently declared? To which version does the driver currently default to?

Kami: EDIT: I see it's declared here - https://github.com/apache/libcloud/blob/9cab4689ac005fed2de0761174cf13981d525128/libcloud/dns/drivers/powerdns.py#L133

libcloud/dns/drivers/powerdns.py
biggosh: The new push has been modified: disabled and comment has been added to extra parameter already declared

biggosh:
I cannot add tests because I don't know how to add (I'm a very old school programmer) them and I'm programming also in free time, just not to forgot my first IT love. Sorry

Kami:
I made some styling changes to get build to pass and merged this into trunk.

Thanks again for your contribution. In the future, if you get a chance, it would be good to add some tests for this functionality.

PR #1332, opened 2019-08-02
[GH-1322] Detect broken `yum install python-requests`

### LINKED ISSUE #1322: list_container_objects doesn't work for Azure blob storage under python 2.7.5
## Summary

`list_container_objects` doesn't work for Azure blob storage under python 2.7.5

## Detailed Information

###libcloud version:
2.5.0/2.4.0

###Python version:
2.7.5

###Operating System:
RHEL 7.3

###Steps which are needed to reproduce it:
Using the following code to list objects in Azure blob container:
```

cls = get_driver(Provider.AZURE_BLOBS)
driver = cls(key='XXX', secret='XXXXXXXX')
container = driver.get_container('myContainer')
objs = driver.list_container_objects(container, '123')

print objs
```

It will throw the following exception at line `objs = driver.list_container_objects(container, '123')`:
```
>>> objs = driver.list_container_objects(container, '123')
  File "/usr/lib/python2.7/site-packages/libcloud/storage/drivers/azure_blobs.py", line 437, in list_container_objects
    ex_prefix=ex_prefix))
  File "/usr/lib/python2.7/site-packages/libcloud/storage/drivers/azure_blobs.py", line 401, in iterate_container_objects
    params=params)
  File "/usr/lib/python2.7/site-packages/libcloud/common/base.py", line 636, in request
  File "/usr/lib/python2.7/site-packages/libcloud/common/base.py", line 155, in __init__
    self.object = self.parse_body()
  File "/usr/lib/python2.7/site-packages/libcloud/common/azure.py", line 97, in parse_body
    return super(AzureResponse, self).parse_body()
  File "/usr/lib/python2.7/site-packages/libcloud/common/base.py", line 233, in parse_body
    driver=self.connection.driver)
libcloud.common.types.MalformedResponseError: <MalformedResponseException in <libcloud.storage.drivers.azure_blobs.AzureBlobsStorageDriver object at 0x7f9f08838f90> 'Failed to parse XML'>: u'\u010f\u0165\u017c<?xml version="1.0" encoding="utf-8"?><EnumerationResults ContainerName="https://.....
```

The response XML seems to be valid except it begins with `u'\u010f\u0165\u017c`.
I tried in python 2.7.15 and it works fine. Maybe this issue is related to python 2.7.5 only?
Python 2.7.5 is the default version for RHEL 7.3, so please help me. Thanks!

c-w:
### Description

As discussed in https://github.com/apache/libcloud/issues/1322, libcloud breaks when used on RHEL 7.6 with requests installed from yum instead of PyPI. In order to prevent user confusion, this pull request introduces a check for this environment state and provides guidance for the user on how to correctly set up their system for libcloud to function.

Resolves https://github.com/apache/libcloud/issues/1322

### Status

- done, ready for review

- [x] [Code linting](http://libcloud.readthedocs.org/en/latest/development.html#code-style-guide) ([build passed](https://travis-ci.org/CatalystCode/libcloud/builds/567156431))
- [x] Documentation (updated changelog)
- [x] [Tests](http://libcloud.readthedocs.org/en/latest/testing.html) (added unit test)

### DIFF
--- a/libcloud/__init__.py
+++ b/libcloud/__init__.py
@@ -35,6 +35,12 @@ try:
 except ImportError:
     have_paramiko = False
 
+try:
+    import requests  # NOQA
+    have_requests = True
+except ImportError:
+    have_requests = False
+
 __all__ = [
     '__version__',
     'enable_debug'
@@ -72,6 +78,8 @@ def _init_once():
 
     This checks for the LIBCLOUD_DEBUG environment variable, which if it exists
     is where we will log debug information about the provider transports.
+
+    This also checks for known environment/dependency incompatibilities.
     """
     path = os.getenv('LIBCLOUD_DEBUG')
     if path:
@@ -91,6 +99,18 @@ def _init_once():
 
         if have_paramiko and hasattr(paramiko.util, 'log_to_file'):
             paramiko.util.log_to_file(filename=path, level=logging.DEBUG)
+
+    # check for broken `yum install python-requests`
+    if have_requests and requests.__version__ == '2.6.0':
+        chardet_version = requests.packages.chardet.__version__
+        required_chardet_version = '2.3.0'
+        assert chardet_version == required_chardet_version, (
+            'Known bad version of requests detected! This can happen when '
+            'requests was installed from a source other than PyPI, e.g. via '
+            'a package manager such as yum. Please either install requests '
+            'from PyPI or run `pip install chardet==%s` to resolve this '
+            'issue.' % required_chardet_version
+        )
 
 
 _init_once()
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -202,6 +202,11 @@ Storage
   responses. (GITHUB-1325, GITHUB-1322)
   [Clemens Wolff - @c-w]
 
+- [Azure Blobs] Detect bad version of requests that leads to errors in parsing
+  Azure Storage responses. This scenario is known to happen on RHEL 7.6 when
+  requests was installed via yum. (GITHUB-1332, GITHUB-1322)
+  [Clemens Wolff - @c-w]
+
 - [Common, CloudFiles] Fix ``upload_object_via_stream`` and ensure we start
   from the beginning when calculating hash for the provided iterator. This way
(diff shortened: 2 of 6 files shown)


Kami:
Can you please also add a test case for it?

On a related note - we should also advise uses to use virtualenv with no system packages in such scenario (that's a good practice in general).

libcloud/__init__.py
```diff
@@ -24,6 +24,8 @@
 import codecs
 import atexit
 
+import requests
```
Kami: We import this file in ``setup.py`` to read ``__version__`` attribute from it which means ``setup.py`` calls could fail if ``requests`` is not installed on the system (aka when setup.py is called during installations, it's unlikely these days, but still) - https://github.com/apache/libcloud/blob/trunk/setup.py#L76. It's generally a bad practice to do that, but that code has probably been there for a long time.

I think it would be safer and better for us to simply parse this file as string and use regex to match ``__version__`` attribute.

Something along those lines - https://github.com/StackStorm/st2/blob/master/scripts/dist_utils.py#L134

c-w: In order to simplify the check and to stick with the existing pattern used for the paramiko library, I've changed the code to use a try/catch import. During `setup.py`, we don't particularly care about executing the requests version check (this can be delayed until first run of the library) so the functionality should be equivalent to what was there before. Let me know if this is okay.

c-w:
@Kami I believe I addressed all the comments. I also verified that `python setup.py install` works on a clean virtualenv. Could you please take another look and let me know if this is ready for merge? Thanks!

Kami (approved):

Kami:
Sorry for the delay. I pushed a small change (8d179ed9d5402988e4e5138aa3db51befc634274), besides that it looks good to me :+1:

Kami:
@c-w a lot of changes have accumulated in trunk since the last release.

Do you want to start the work needed to get the 2.6.0 release out - https://libcloud.readthedocs.io/en/latest/committer_guide.html#making-a-release-for-release-managers?

PR #1339, opened 2019-08-07
Various fixes for the Backblaze B2 driver

### LINKED ISSUE #1338: Backblaze B2: AUTH_API_HOST not picking up the default
Running the demo code in the docs:

```

account_id = 'XXXXXX'
application_key = 'YYYYYY'

cls = get_driver(Provider.BACKBLAZE_B2)
driver = cls(account_id, application_key)
driver.list_containers()
```

Produces the error:

```
requests.exceptions.ConnectionError: HTTPSConnectionPool(host='none', port=443): Max retries exceeded with url: /b2api/v1/b2_list_buckets?accountId=xxxxxxxxxx (Caused by NewConnectionError('<urllib3.connection.VerifiedHTTPSConnection object at 0x7f7366797610>: Failed to establish a new connection: [Errno -5] No address associated with hostname'))

```
If the host is supplied to the constructor, then it works as intended:
```
driver = cls(account_id, application_key, host='api.backblaze.com')
```

I can replicate this on Python 2 and Python 3. I am using libcloud via django-storages, so there isn't a way for me to pass the hostname via the constructor. 

Looking at the backblaze provider the default host is provided and correct, but somehow not making it into the connection object before request is called.

I can take another look at this this afternoon, but a patch didn't jump out at me immediately.

Kami:
This pull request fixes an issue in the Backblaze B2 driver as reported in #1338.

While working on this, I noticed ``download_object_as_stream`` is also broken, so I fixed that as well.

It looks like both of those regressions were inadvertently introduced when we moved to the ``requests`` library.

It's likely that there are more issues similar to this one hiding in other drivers. Sadly our functional tests didn't catch those issues.

Resolves #1338.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -61,6 +61,10 @@ General
 - Fix HTTP(s) proxy support in the OpenStack drivers. (GITHUB-1324)
   [Gabe Van Engel - @gvengel]
 
+- Fix logging connection class so it also works when data type is ``bytearray``
+  or ``bytes``. (GITHUB-1339)
+  [Tomaz Muraus]
+
 Compute
 ~~~~~~~
 
@@ -198,6 +202,13 @@ Storage
   we avoid hash mismatch errors in scenario where provided iterator is already
   iterated / seeked upon before calculating the hash. (GITHUB-1326)
   [Gabe Van Engel - @gvengel, Tomaz Muraus]
+
+- [Backblaze B2] Fix a bug with driver not working correctly due to a
+  regression which was inadvertently introduced in one of the previous
+  releases. (GITHUB-1338, GITHUB-1339)
+
+  Reported by Shawn Nock - @nocko.
+  [Tomaz Muraus]
 
 DNS
 ~~~
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -210,6 +210,9 @@ Storage
   Reported by Shawn Nock - @nocko.
   [Tomaz Muraus]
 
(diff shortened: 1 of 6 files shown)


Kami:
@tonybaloney I noticed more storage drivers use ``response.response`` in ``download_object_as_stream`` method.

This attribute is an instance of ``HttpLibResponseProxy`` which is not compatible with the pre-requests migration response format so those methods will fail / throw an exception:

```bash
  File "example_storage.py", line 42, in <module>
    for chunk in iterator:
  File "/home/kami/w/lc/libcloud/libcloud/utils/files.py", line 74, in read_in_chunks
    chunk = b(get_data(*args))
  File "/home/kami/w/lc/libcloud/libcloud/utils/py3.py", line 203, in next
    return iterator.next()
AttributeError: 'HttpLibResponseProxy' object has no attribute 'next'
```

Or am I missing something?

If that's is indeed the issue and it's broken, one option to make it work out of the box (and also be backward compatible) is to add ``next()`` method to ``HttpLibResponseProxy`` class.

Something along the lines of:

```python

    def next(self):
        for chunk in self._response.iter_content():
            return chunk
```

I confirmed that works, at least with the Backblaze driver.

It's not the most ideal thing though. Alternative would be to update all those places to use ``response.iter_content()``.

EDIT: If we go with the second approach, do we even still need ``HttpLibResponseProxy`` class at that point? Or maybe some other code still depends on it?

Kami:
Just a heads up - I will merge those Backblaze driver fixes and work on the requests upload_object_via_stream regression (?) in a separate PR.

PR #1341, opened 2019-08-13
[WIP] Try using Github Actions for CI/CD

Kami:
Just playing with using Github actions instead of Travis for CI/CD.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,43 +1,17 @@
-
 name: CI
 
 on: [push]
 
 jobs:
   unit_tests:
-    name: Run Build
+    name: Run Unit Tests
     runs-on: ${{ matrix.os }}
 
     strategy:
       matrix:
-        include:
-        - python_version: 2.7
-          os: ubuntu-latest
-          tox_env: py2.7
-        - python_version: 3.4
-          os: ubuntu-latest
-          tox_env: py3.4
-        - python_version: 3.5
-          os: ubuntu-latest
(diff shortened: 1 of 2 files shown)


Kami:
At the moment, performance seems to be much better compared to Travis CI, but that's mostly related to small / non existent build queue.

Once Github Actions CI/CD goes out of beta and more open-source projects start using it, it will likely encounter same issues with large build queues and delayed builds as Travis and Circle CI do for open source projects.

c-w (commented):
Overall the Github Actions look good to me. I've also had good experiences with Azure Devops CI (which based on the syntax looks like it may be the backend of Github Actions) on other projects (e.g. [libcloud-tests](https://github.com/c-w/libcloud-tests)).

If there's a desire to not tie the project too closely to any particular CI provider such as Github Actions or Travis, I would recommend to implement the CI steps as a Docker container. This is slightly more work at the start since all the project setup steps must be made explicit in the Dockerfile, but has the benefit of making it trivial to run the CI locally (just build and run the Docker image) and ensures that all the CI steps are reproducible across environments (since the CI environment is self-contained to the Docker container). This approach reduces the CI provider's job to "build this Dockerfile and run it in parallel with a bunch of environment variables set" which essentially moves vendor lock-in to zero.

I've used the pattern of Dockerizing the CI on many projects (e.g. [gutenberg](https://github.com/c-w/gutenberg)) and it always worked well for me. If there's desire to try this for libcloud, I can make a proposal branch. Just let me know!

PR #1090, opened 2017-08-08
[LIBCLOUD-870] Implement create methods for AWS ALB driver

lrvan:
### Description

1) Implemented AWS ALB driver methods to create application load balancer or its parts:
- create_balancer
- ex_create_balancer
- ex_create_target_group
- ex_register_targets
- ex_create_listener
- ex_create_listener_rule

2) Fixed parameter/field names/types in some methods to be uniform across the driver.
3) Extended \_to\_#object_type# methods to fetch and return additional data about the object from AWS response.

### Status
- done, ready for review

### Checklist (tick everything that applies)

- [ ] [Code linting](http://libcloud.readthedocs.org/en/latest/development.html#code-style-guide) (required, can be done after the PR checks)
- [ ] Documentation
- [ ] [Tests](http://libcloud.readthedocs.org/en/latest/testing.html)
- [ ] [ICLA](http://libcloud.readthedocs.org/en/latest/development.html#contributing-bigger-changes) (required for bigger changes)

PR #1349, opened 2019-09-04
Implemented create methods for AWS ALB driver

lrvan:
### Description

This is rebased against trunk version of previous PR:
https://github.com/apache/libcloud/pull/1090

1) Implemented AWS ALB driver methods to create application load balancer or its parts:
- create_balancer
- ex_create_balancer
- ex_create_target_group
- ex_register_targets
- ex_create_listener
- ex_create_listener_rule

2) Fixed parameter/field names/types in some methods to be uniform across the driver.
3) Extended \_to\_#object_type# methods to fetch and return additional data about the object from AWS response.

### Status
- done, ready for review

- [ ] Documentation


lrvan:
@tonybaloney @Kami PR has been updated against trunk, tests are green. Please take a look.

tonybaloney (approved):
LGTM, very well structured and written contribution. thanks!

libcloud/loadbalancer/drivers/alb.py
```diff
@@ -60,26 +211,496 @@ def __init__(self, access_id, secret, region, token=None):
...
-            'LoadBalancerNames.member.1': balancer_id
+            'LoadBalancerArns.member.1': balancer_id
         }
         data = self.connection.request(ROOT, params=params).object
         return self._to_balancers(data)[0]
 
-    def ex_balancer_list_listeners(self, balancer):
-        return balancer.extra.get('listeners', [])
+    def create_balancer(self, name, port, protocol, algorithm, members,
+                        ex_scheme="", ex_security_groups=[], ex_subnets=[],
```
Kami: Please don't use mutable default arguments - https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

So instead of ``ex_security_groups=[], ex_tags={}``, etc. do something like:

```python
def(...ex_security_groups=None, eg_tags=None):
   ex_security_groups = ex_security_groups or []
   ex_tags = ex_tags or {}
```

libcloud/loadbalancer/drivers/alb.py
```diff
@@ -60,26 +211,496 @@ def __init__(self, access_id, secret, region, token=None):
...
+        listener = self.ex_create_listener(balancer, port, protocol,
+                                           target_group,
+                                           ssl_cert_arn=ex_ssl_cert_arn)
+
+        balancer.extra['listener'] = listener
+
+        return balancer
+
+    def ex_create_balancer(self, name, addr_type="ipv4",
+                           scheme="internet-facing", security_groups=[],
```
Kami: Same here (re mutable default arguments).

Kami (changes requested):
Thanks for the contribution.

I've added a couple of small comments, besides that, LGTM.

It also looks like we don't have any documentation for AWS ALB load-balancer driver, so some docs would be great as well :)

Kami:
Also, since this is a bigger change, please sign an ICLA, if you haven't already.

libcloud/loadbalancer/drivers/alb.py
lrvan: good catch :) I've corrected it in the last commit. Thanks for pointing me out to this nuance.

Regarding the docs - let's publish them as a separate change (not sure if I can find time for them asap). So lets incorporate this particular functionality into dev (since it already has some documentation in doc strings) in order to make it available for the community.

I've sent signed ICLA just today.

Kami:
Merged, thanks.

Please follow up with docs when you get a chance.

PR #1351, opened 2019-10-02
Fix list_sizes on Packet when using project API tokens

d-mo:
### Description

The `list_sizes` method of the Packet driver was invoking the `/plans` API endpoint, which is only available when using personal API tokens. Project specific API tokens should invoke `/project/<project_id>/plans` instead. This PR uses the latter endpoint when `project_id` is set.

### Status

- done, ready for review

- [ ] Documentation


Kami:
Thanks for the contribution.

I fixed the test case and merged changes into trunk. Next time please verify that tests pass after submitting a change :)

PR #1354, opened 2019-10-04
[Bug fix] GCE Error setting/adding project metadata

aitorzabala:
### Description
A bug was detected when setting new metadata values into GCE project while using the GCE provider from libcloud "trunk" branch. Browsing through the code I noticed that arguments are not passed correctly to the gce driver (set common metadata), causing gce driver to return an error when setting metadata into a GCE project.

Error:
- ValueError: Unsupported metadata format.
Solution:
- Removed unnecessary "self" argument to ex_set_common_instance_metadata and included force argument, which was currently missing.

### Status
- done, ready for review

- [X] Documentation

### DIFF
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -899,7 +899,7 @@ class GCEProject(UuidMixin):
         :return: True if successful
         :rtype:  ``bool``
         """
-        return self.driver.ex_set_common_instance_metadata(self, metadata)
+        return self.driver.ex_set_common_instance_metadata(metadata, force)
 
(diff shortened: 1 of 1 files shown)


Kami:
Thanks for contributing this bug fix.

I noticed the same issue exists in ``GCEProject.set_usage_export_bucket`` method so I also fixed that. In addition to that, I also added corresponding test cases and updated methods to use keyword arguments to be on the safe side - 87247f31aeb43f30981ead86be2022a5ede17905.

It's worth noting that this bug only affected methods on the ``GCEProject`` class. So you could still use a working version by accessing ``ex_`` methods directly on the driver instance class.

Kami (approved):

Issue #1356, opened 2019-10-09
GCE node deletion times out - allow an async option


zimventures:
## Feature Request
Currently the GCE `node_delete()` function uses the `async_request()` call (which ironically isn't actually async). Deleting a basic n1-standard-1 node running Centos in us-central times out every time. This isn't an uncommon occurrence. Searching the intertubes there are other references of other projects having the same issue:

https://github.com/kubevirt/cloud-image-builder/issues/72
https://github.com/ansible/ansible/issues/29597

The current way to avoid the timeout is to use the `ex_destroy_multiple_nodes()` function with a gigantic timeout value. An option, for sure, but not flexible enough. 

An easy solution in this case would be to offer both a sync and async option for deletion. Applications can poll, if desired. Thankfully, the async option is fairly straightforward, defaulting to the current sync mode of operation:

```python
        if sync:
            self.connection.async_request(request, method='DELETE')
        else:
            self.connection.request(request, method='DELETE')
```

Kami:
I think that's a reasonable approach although it would be good to have a consistent API for all the operations which are async (aka involve polling).

So ideally all of those methods would take ``sync=False`` / ``ex_sync=False`` argument.

Kami:
Closing since this has been implemented as part of #1357.

PR #1357, opened 2019-10-09
adding a sync/async flag for destroy_node

zimventures:
## Changes GCE node deletion to allow async issuance
Adds the ability for GCE provider to perform node deletion in an async manner. 

### Description

See issue #1356 for full back-story. 
tldr; even the most basic node deletions in GCP will time out when using the `destroy_node()` function as it works today. 

### Status
Covers issue #1356 
- work in progress

- [ ] Documentation

### DIFF
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -6831,7 +6831,7 @@ class GCENodeDriver(NodeDriver):
         self.connection.async_request(request, method='DELETE')
         return True
 
-    def destroy_node(self, node, destroy_boot_disk=False, sync=True):
+    def destroy_node(self, node, destroy_boot_disk=False, ex_sync=True):
         """
         Destroy a node.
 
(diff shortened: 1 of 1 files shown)


Kami:
Thanks for the contribution.

As mentioned in #1356, I'm fine with this change.

And yes, ``async`` is not a great name in this context.

Operation is indeed async in a sense that we need to poll for the result. ``polling_request`` or similar would indeed be a better name :)

libcloud/compute/drivers/gce.py
```diff
@@ -6816,7 +6831,7 @@ def ex_destroy_autoscaler(self, autoscaler):
         self.connection.async_request(request, method='DELETE')
         return True
 
-    def destroy_node(self, node, destroy_boot_disk=False):
+    def destroy_node(self, node, destroy_boot_disk=False, sync=True):
```
Kami: I will go ahead and rename this to ``ex_sync`` since this is a "standard" API method so non-standard arguments should be prefixed with ``ex_`` (there is no ``ex_`` prefix needed with methods which already contain ``ex_`` prefix itself).

Kami:
Merged, thanks.

zimventures:
> Merged, thanks.

Glad to help! I'm working adding some async calls to other methods and will use ex_sync for those as well.

PR #1364, opened 2019-11-03
start and stop methods have been added to Node class.

emakarov:
### Description

Node class has been extended with methods `start` and `stop` for more convenient usage of Node objects

### DIFF
--- a/libcloud/compute/base.py
+++ b/libcloud/compute/base.py
@@ -258,6 +258,34 @@ class Node(UuidMixin):
         """
         return self.driver.destroy_node(self)
 
+    def start(self, sync=True):
+        """
+        Start this node that is stopped.
+
+        This calls the node's driver and starts the node
+        (executes driver.ex_start_node)
+
+        :keyword  sync: If true, do not return until destroyed or timeout
+        :type     sync: ``bool``
+
+        :return: ``bool``
+        """
+        return self.driver.ex_start_node(self, sync=sync)
+
+    def stop(self, sync=True):
+        """
+        Stops this running node.
+
+        :keyword  sync: If true, do not return until destroyed or timeout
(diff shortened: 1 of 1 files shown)


libcloud/compute/base.py
```diff
@@ -258,6 +258,34 @@ def destroy(self):
         """
         return self.driver.destroy_node(self)
 
+    def start(self, sync=True):
```
Kami, line 261: Those are extension methods so please prefix them with ``ex_``.

emakarov: @Kami I was trying to understand the method naming idea but couldn't find out why some methods have prefix `ex_` but others don't have. `start` and `stop` look cleaner than `ex_start` and `ex_stop`. Could you clarify what kind of extension is meant here? I suppose all cloud providers should have ability to start and stop nodes

Kami: In Libcloud world, "extension" methods are the ones which are not part of the standard API (aka are not supported by the majority of the providers and where the implementations might substantially differ across various providers).

Having said that, I think I think it would also be reasonable to promote ``start_node`` and ``stop_node`` methods to the standard compute API in the next major release since they are available on a lot of provider drivers.

This would require some more changes though and we would probably want to keep ``ex_`` version of methods in there for a while for backward compatibiliy / ease of migration reasons.

So maybe best approach is to merge this PR after we make the "promote methods to the standard API" changes, to avoid making changes twice.

emakarov: I see, thanks for clarification.
Please let me know if you still want to change method names or it's ok as it now

Kami:
Thank for the contribution, this has now been superseded and resolved by changes in #1375.

PR #1370, opened 2019-11-19
Support aws eu-north-1 region (Stockholm) #1369

michaelsembwever:
ref: https://github.com/apache/libcloud/issues/1369

### Status

ready for testing and review.

### Checklist (tick everything that applies)

- [ ] [Code linting](http://libcloud.readthedocs.org/en/latest/development.html#code-style-guide) (required, can be done after the PR checks)
- [ ] Documentation
- [x] [Tests](http://libcloud.readthedocs.org/en/latest/testing.html)
- [x] [ICLA](http://libcloud.readthedocs.org/en/latest/development.html#contributing-bigger-changes) (required for bigger changes)

Issue #1366, opened 2019-11-06
S3 get_driver using region name


maingoh:
For some reason it seems there is one driver for each S3 region.
Thus we cannot dynamically get a s3 driver depending on the region string.

For example something like this:
```
cls = get_driver(Provider.S3, region_name='eu-west-1')
``` 

Or maybe it would be better to have only one S3StorageDriver class where we can specify the region during construction.

```
driver = cls(access_key_id, secret_acces_key, region=region_name, region_name=region_name)
```

Though it seems there is two region attribute which should match. `region_name` from `S3StorageDriver` and `region` from `AWSDriver`.

Kami:
The best approach would probably be to support ``region`` argument in the driver constructor.

That's something we already do in other drivers and similar to what you suggested in option 2).

Kami:
@maingoh #1371 should resolve this issue.

It would be great if you can check / verify it if you get a chance.

maingoh:
Great ! I will try :)

maingoh:
It is working, just waiting for a new version :)

Kami:
Thanks for confirming :)

PR #1371, opened 2019-11-19
Move S3 driver to single driver class + "region" constructor argument model

Kami:
This pull request moves S3 driver from "driver class per region" model to "single driver class with region constructor argument" model.

This way it's consistent with other multi region providers and more pleasant for the end users.

The plan is to deprecate and remove those region specific class in a future major release.

In addition to that change, it also includes some other small changes:

1. Add all the missing and supported regions to the S3 driver
2. Incorporate changes from #1370 
3. Update exception which is thrown if user is using driver for region X, but the bucket they wish to upload the object to is in region Y
4. Fix and improve provider feature matrix generation script so it's more robust and excludes S3 class per region entries
5. Add new ``provider-tables`` tox target for generating those provider tables in a consistent environment with all the dependencies

## TODO

- [x] Tests
- [x] Upgrade notes entry

Resolves #1370 #1366


Kami:
I wanted to also introduce Python 3.8 support as part of this PR, but I had some issues with getting tests to pass on Travis so I will merge this PR into trunk and tackle that in a separate PR.

PR #1375, opened 2019-11-26
Promote start node and stop node methods to standard Libcloud compute API

Kami:
This pull request adds new ``start_node(self, node)`` and ``stop_node(self, node)`` methods to the base compute ``NodeDriver`` class.

This means those two methods have been "promoted" and are now part of the standard Libcloud compute API.

### Background, Reasoning

A big chunk of the existing compute drivers (18) already implemented this functionality as part of ``ex_start_node`` and ``ex_stop_node`` method so it only makes sense to promote those methods to be part of the standard compute API.

For backward compatibility reasons, existing extension methods are left in place and they continue to behave as they did in the previous releases. Those methods will be removed as part of a cleanup in a future major release.

### TODO

- [x] Changelog entry
- [x] Upgrade notes entry


erjohnso:
Great idea! I skimmed through and it looks good to me @Kami.

PR #1377, opened 2019-11-30
Drop support for Python 2.7 and Python 3 < 3.5.0

Kami:
This pull request updates the code and drops support for Python 2.7 and Python 3 < 3.5.0 (as discussed on the mailing list).

The plan is to release Libcloud v3.0.0 with those changes once Python 2.7 is officially EOL (January 1, 2020).

For people who still need to use Python 2.7 or Python 3 < 3.5.0, they can install older Libcloud release which still supports that version.

The idea is to first only drop support in ``setup.py`` and stop testing Libcloud with those versions, but in the future we also plan to refactor the code and get rid of all the ``libcloud.utils.py3`` and other Python 2 and Python 3 compatibility / glue code which won't be needed anymore.

Since we will only support Python >= 3.5.0 it means we will also be able to utilize native type hinting syntax (no more comments).

## TODO

- [x] Changelog entry
- [x] Update notes entry
- [x] Blog post that announces that change
- [x] Update ReadTheDocs configuration to build docs with Python 3.7
- [ ] Add documentation section on Python versions we support
- [x] Update website section on supported Python versions
- [x] Remove drivers which don't support Python 3
  - [x] vSphere driver relies on pysphere which doesn't work with Python 3 (~~perhaps try using https://github.com/machalekj/pysphere/tree/2to3?~~ - that fork is old and unmaintained)


Kami:
I created new ``2.8.x`` branch which we can use if we need to do another (bug fix) release before v3.0.0 so I will go ahead and merge those changes unto trunk.

PR #1384, opened 2019-12-13
Only guess host from region map if not explicitly set

### LINKED ISSUE #1383: host parameter of S3StorageDriver does nothing in 2.7.0
## Summary

We noticed that connecting to a local minio server failed after upgrading to libcloud 2.7.0

## Detailed Information

To use the S3 interface of a local server we need to set the host explicitly. In the 2.7.0 sources (see https://github.com/apache/libcloud/blob/trunk/libcloud/storage/drivers/s3.py) we have this:

```python
class S3StorageDriver(AWSDriver, BaseS3StorageDriver):
    name = 'Amazon S3'
    connectionCls = S3SignatureV4Connection
    region_name = 'us-east-1'

    def __init__(self, key, secret=None, secure=True, host=None, port=None,
                 region=None, token=None, **kwargs):
        # Here for backward compatibility for old and deprecated driver class
        # per region approach
        if hasattr(self, 'region_name') and not region:
            region = self.region_name  # pylint: disable=no-member

        self.region_name = region

        if region and region not in REGION_TO_HOST_MAP.keys():
            raise ValueError('Invalid or unsupported region: %s' % (region))

        self.name = 'Amazon S3 (%s)' % (region)

        host = REGION_TO_HOST_MAP[region]
        super(S3StorageDriver, self).__init__(key=key, secret=secret,
                                              secure=secure, host=host,
                                              port=port,
                                              region=region, token=token,
                                              **kwargs)
```

Note that the `host = REGION_TO_HOST_MAP[region]` line overrides the host parameter

gluap:
### Description

Overwriting host parameter is inconsistent with function signature and a breaking change compared with previous behavior. This change will respect the host if passed by the user but guess from the region if not explicitly passed.

Reason: If the user passes a host he expects it to be respected, be it because he wants to use a local proxy, test against a local server or any other reason not to use one of the predefined amazon servers.

This fixes #1383 .

### Status

done, ready for review

    should pass (tried to keep pep8 for that one extra line)
- [x] Documentation
    small bug fix restores consistent behaviour -- no change vs documentation
    CI shows unit tests passed.
    a one liner surely doesn't require icla, I'm ticking the check for that reason.

### DIFF
--- a/libcloud/storage/drivers/s3.py
+++ b/libcloud/storage/drivers/s3.py
@@ -1050,7 +1050,9 @@ class S3StorageDriver(AWSDriver, BaseS3StorageDriver):
 
         self.name = 'Amazon S3 (%s)' % (region)
 
-        host = REGION_TO_HOST_MAP[region]
+        if host is None:
+            host = REGION_TO_HOST_MAP[region]
+
         super(S3StorageDriver, self).__init__(key=key, secret=secret,
                                               secure=secure, host=host,
                                               port=port,


Kami:
Thanks for the contribution.

This change looks good, but it would also good to add test case which verify behavior is as expected.

Kami:
I added a test case for it (50f475055733b6d617be1dfd64f45194d24c0cdc) and merged it into trunk.

Thanks for the contribution.

PR #1386, opened 2019-12-13
Support disk_size parameter for create_node

yukw777:
### Description

I've added support for `disk_size` in `create_node()` for GCE. This closely follows the change that added the same for `ex_create_multiple_nodes()`. This is so that we can launch nodes on GCE with different sizes of disks easily.

### Status
- done, ready for review

- [x] Documentation
- [ ] [ICLA](http://libcloud.readthedocs.org/en/latest/development.html#contributing-bigger-changes) (required for bigger changes) => I think this is a small change.

### DIFF
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -8281,6 +8281,10 @@ class GCENodeDriver(NodeDriver):
         :param  ex_accelerator_count: The number of accelerators to associate
                                       with the node.
         :type   ex_accelerator_count: ``int`` or ``None``
+        
+        :keyword  ex_disk_size: Specify size of the boot disk.
+                                Integer in gigabytes.
+        :type     ex_disk_size: ``int`` or ``None``
 
         :return:  A tuple containing a request string and a node_data dict.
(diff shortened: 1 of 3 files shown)


libcloud/compute/drivers/gce.py
```diff
@@ -4201,7 +4205,7 @@ def create_node(
             ex_can_ip_forward, ex_disks_gce_struct, ex_nic_gce_struct,
             ex_on_host_maintenance, ex_automatic_restart, ex_preemptible,
             ex_subnetwork, ex_labels, ex_accelerator_type,
-            ex_accelerator_count)
```
Kami, line 4204: To avoid potential bugs it would probably be safer to use keyword instead of positional argumentshere  in the future, but this also works for now since other arguments are already positional.

Kami:
Merged, thanks for contribution.

libcloud/compute/drivers/gce.py
yukw777: I definitely agree, and actually was thinking about fixing it, but then decided against it so as to keep the scope of this PR small.

PR #1387, opened 2019-12-16
GCENodeDriver's deploy_node() now accepts Deployments as well as all the parameters accepted by create_node()

### LINKED ISSUE #1381: GCENodeDriver's deploy_node() should accept Deployments and other parameters that are accepted by create_node()
## Feature Request

I was trying to use `Deployment` when launching nodes using GCE, but I noticed that the `deploy_node()` of `GCENodeDriver` has a different function signature from other drivers and does not accept `Deployment`. Furthermore, it doesn't accept other essential parameters you can pass in to `create_node()`, making it very limited in its usefulness. Is there a reason why this is so? I fail to see why `GCENodeDriver` has overridden `NodeDriver`'s default `deploy_node()` when there is even this comment: https://github.com/apache/libcloud/blob/d23eeb2862d761cc88973c2e73175b6fed87af66/libcloud/compute/base.py#L968

I think `GCENodeDriver` should accept `Deployment` for its `deploy_node()` method like other drivers. Would this be as simple as removing the overriding method? Thank you for your help in advance!

yukw777:
### Description

I've deleted the overridden `deploy_node()` function from `GCENodeDriver` so that it'd use base `NodeDriver`'s `deploy_node()` instead. This allows users to pass in `Deployment`s and other parameters accepted by `create_node()`. This is a backwards incompatible change, but it's simple to fix by using `ScriptFileDeployment` or setting the metadata with the content of the script manually. I think the gain is worth the breaking change. Closes #1381.

### Status
- done, ready for review

- [x] Documentation
- [x] [Tests] I didn't add new unit tests as I'm merely removing a function, and the base `deploy_node()` should be covered by existing tests. I did manually test with my GCE account.
- [x] [ICLA](http://libcloud.readthedocs.org/en/latest/development.html#contributing-bigger-changes) I think this is a small change.

### DIFF
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -3995,7 +3995,7 @@ class GCENodeDriver(NodeDriver):
             ex_on_host_maintenance=None, ex_automatic_restart=None,
             ex_preemptible=None, ex_image_family=None, ex_labels=None,
             ex_accelerator_type=None, ex_accelerator_count=None,
-            ex_disk_size=None):
+            ex_disk_size=None, **kwargs):
         """
         Create a new node and return a node object for the node.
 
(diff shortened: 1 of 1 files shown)


libcloud/compute/drivers/gce.py
```diff
@@ -3995,7 +3995,7 @@ def create_node(
             ex_on_host_maintenance=None, ex_automatic_restart=None,
             ex_preemptible=None, ex_image_family=None, ex_labels=None,
             ex_accelerator_type=None, ex_accelerator_count=None,
-            ex_disk_size=None):
+            ex_disk_size=None, **kwargs):
```
yukw777, line 3998: I saw that other drivers like `ec2` and `azure` do this, so I followed their pattern.

Kami:
Thanks for the contribution.

I will need to take a look at this in more detail and perhaps also test it to see if we can get rid of ``**kwargs``.

I know it's used in other places, but it's legacy code and in general a bad practice since it makes code very hard to follow and understand (and it also masks various issues since the method just silently accepts arbitrary arguments).

peterAsapp:
@Kami i had to add `**kwargs` b/c `deploy_node()` code calls `create_node()` with all of the argument it received including `deploy`, which causes an unexpected argument error in `create_node()`. That's why I added it. But do let me know if we can get around that!

Kami:
@peterAsapp Yeah, I wonder if we could some how refactor that ``deploy_node()`` behavior.

IIRC, some drivers rely on ``deploy`` argument being passed to ``create_node()`` since they read generated SSH auth password for it, but they shouldn't need access to other arguments.

I will push a change I think which may be a good compromise (although still not ideal) and then we can decide what we think about that approach.

EDIT: The argument name is ``auth``and not deploy.

Kami:
@yukw777 @peterAsapp That's the WIP workaround / mitigation I was thinking about - #1389.

What do you think?

@tonybaloney Maybe you also have an opinion on that.

Some more context is in #1389. Keep in mind though that some of that is legacy and old code and we can't fully fix / improve it without breaking backward compatibility.

yukw777:
@Kami that PR looks good to me. As soon as you merge it, I'll rebase this one and remove `**kwargs`.

Kami:
After merging #1389 I tried to test this change, but it doesn't appears to be working, it appears to be incomplete.

For ``deploy_node()`` to work, driver needs to declare authentication type it supports using ``features`` driver class variable (aka it won't work "magically" if driver doesn't declare support for it and accepts and correctly handles ``auth`` argument).

In addition to that, ``create_node()`` then needs to accept and use ``auth`` argument.

So without those changes, deployment won't work. In short - the actual change will likely end up being much more involved.

---

EDIT: I'm looking into it, I believe it should be relatively simple to get it to work utilizing metadata.

Kami:
^ I managed to get it to work.

Luckily it was relatively trivial and didn't require many changes - ec8157aef5229c8adb14d52f43dcf5e5a65a57b3, e4d2dae6d9849acfebe3400af240cd5a4df803ae, b49ed7e953ecacc35d42953245e7a95ee72b374c.

I also added an example since without it, people won't really know how to use it / get it to work since it relies on GCE metadata functionality to install a public key and without public key, deployment functionality is not of much use (well, unless the key is already baked into the image which is used to launch the VM or GCP project wide SSH key metadata is used).

Kami:
@yukw777 I went ahead and merged all those changes into master. Thanks again for your contribution.

Overall, I think this change is reasonable and good since ``deploy_node()`` now complies with the base compute API.

We could also get ``deploy_node()`` functionality to work without actually needing to SSH into the server and simply utilizing ``startup-script`` metadadata functionality (same as before), but that has limitations and doesn't comply with the Libcloud base API (startup script is limited in size which is true for most providers which utilize cloud init like systems for managing startup scripts).

/cc @erjohnso ^

---

On a somewhat related note, I'm also looking into a change so we can exclude ``auth`` argument from ``create_node()`` method signature where it's not needed (aka as in GCE driver case).

And since this is a backward incompatible change, we also need to add entry to Upgrade notes file.

PR #1389, opened 2019-12-18
Reduce reliance on **kwargs in "create_node()" methods

Kami:
This pull request tries to mitigate / resolve an issue with legacy code / kwargs abuse brought up in #1383.

## Description, Background

Currently a lot of ``create_node()`` method signatures use ``**kwargs`` instead of regular positional + keyword arguments.

That's bad for many reasons (makes code introspection hard / impossible, bad API, hard to follow and understand the code, etc.).

One of the reasons for that is that ``deploy_node()`` method passes all the keyword argument it takes to ``create_node()`` method, even though majority of the arguments it takes are ``deploy_node()`` specific (ssh_username, max_tries, ssh_interface, etc.) and shouldn't be passed / propagated to ``create_node()``.

Only keyword argument which potentially needs to be propagated is ``auth`` argument. That argument is used by provider drivers which support password based SSH authentication (e.g. Linode).

## Proposed Solution

This solution updates ``deploy_node()`` code to only pass non deploy_node specific keyword arguments to the ``create_node()`` method.

If for some reason driver method throws (e.g. that driver still expects some deploy_node arguments which really shouldn't happen, but there might be some legacy code which does that out there), we simply fall back to the old approach and pass all the arguments to ``create_node()`` method.

This will allow drivers to still support ``deploy_node()`` functionality without needing to use ``**kwargs`` in ``create_node()`` method signature. They will be able to normally declare positional and keyword arguments they support in that method.

## TODO

- [x] Changelog entry
- [x] Update some of the existing drivers to explicitly declare supported arguments in ``create_node()`` method instead of using ``**kwargs``

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -40,6 +40,11 @@ Compute
   (GITHUB-1389)
   [Tomaz Muraus]
 
+- Add MyPy type annotations for ``create_node()`` and ``deploy_node()``
+  method.
+  (GITHUB-1389)
+  [Tomaz Muraus]
+
 Storage
 -------
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -28,6 +28,15 @@ Compute
   Because of that, a lot of the compute drivers which support deploy
   functionality needed to use ``**kwargs`` in ``create_node()`` method
   signature which made code hard to read and error prone.
+
+  Also update various affected drivers to explicitly declare supported
+  arguments in the  ``create_node()`` method signature (Dummy, Abiquo,
+  Joyent, Bluebox, OpenStack, Gandy, VCL, vCloud, CloudStack, GoGrid
+  HostVirtual, CloudSigma, ElasticStack, RimuHosting, SoftLayer, Voxel,
+  Vpsnet, KTUcloud, BrightBox, ECP, OpenNebula, UPcloud).
+
+  As part of this change, also various issues with invalid argument names
+  were identified and fixed.
   (GITHUB-1389)
   [Tomaz Muraus]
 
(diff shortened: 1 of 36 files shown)


yukw777:
I think this is a good compromise. It'd be good to track the progress on cleaning this up though.

Kami:
Yeah, and probably even better approach in the future would be for ``deploy_node()`` method to take ``create_node_kwargs={}`` argument with which user would specify exactly which arguments should get passed to ``create_node()`` so we wouldn't need to do this "magic" try to determine which arguments should go where.

This would be a backward incompatible change though. Well, with (again) reliance on ``**kwargs`` in ``deploy_node()`` we could probably make the change backward compatible (aka support both approaches), but that's not ideal either.

Kami:
Here is another change I made - 155c0f8b.

This will allow us to simplify that method even further in the future (and perhaps go with "create_node_kwargs" dict argument approach as documented above.

And here is a change we should eventually make in all the drivers which (ab)use ``**kwargs`` - a846c6d8e5b42e3b559cb0f7c58b9bbc7188f36f.

It makes code much easier to understand, we can now add mypy annotations, etc.

Kami:
I've also changed ``create_node()`` signature on the base driver class and added type annotations for that method + ``deploy_node()`` method - fd2360ef3a10868070ffe3f1d49ab27a2d08d138.

In addition to that, I also updated majority of``create_node()`` methods which misuse ``**kwargs`` to explicitly declare supported keyword arguments - 13f2dec66b738a0cb1ff0eb79f08b33389d4accd.

Let's just say that change wasn't fun (and it took way longer than I would hope), but it already uncovered some inconsistencies between the code and tests, drivers not complying with the standard API (aka non standard arguments were not prefixed with ``ex_``) and actual issues / bugs which were masked because incorrect / unsupported argument names were previously simply ignored.

All the tests pass, let's just hope I didn't inadvertently introduce any regressions.

PR #1393, opened 2019-12-21
Fix wheel METADATA and ensure conditional dependencies are handled correctly

### LINKED ISSUE #1392: The wheel distribution of this package on pypi contains the requirement for enum34
## Summary

The wheel distribution of this package on pypi contains the requirement for enum34

## Detailed Information

enum34 appears to be dynamical included as a requirement in setup.py depending if your Python version is pre 3.4.

```
PY2_or_3_pre_34 = sys.version_info < (3, 4, 0)
# ...
if PY2_or_3_pre_34:
    INSTALL_REQUIREMENTS.append('typing')
    INSTALL_REQUIREMENTS.append('enum34')
```

The wheel package contains the requirement without any special qualifier. (I'm not even sure if it is possible to spec requirements in a wheel according to python version)

From METADATA
```
Requires-Python: >=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, <4
Requires-Dist: requests (>=2.5.0)
Requires-Dist: typing
Requires-Dist: enum34
```

This does create a problem in `zc.buildout` style environments that are running newer python versions.

I think it is probably time to drop support for Python 3.3

Kami:
In a prevision version we added some more conditional dependencies - ``typing`` and ``enum34`` packages which are needed for Python versions older than 3.4.0.

Those versions were dynamically included as part of ``install_requirements`` so the actual ``Reqires-Dist`` metadata entry which is part of the wheel file was not correct and depended on the version of Python which was used to generate the wheel.

## Proposed Solution

This pull request fixes that by utilizing PEP 508 (https://www.python.org/dev/peps/pep-0508/#environment-markers) environment markers notation.

In addition to that, we also use ``extra_requires`` fallback notation for users who install Libcloud using old versions of pip and setuptools (reference: https://hynek.me/articles/conditional-python-dependencies/).

Keep in mind that this change now requires a Libcloud committer which is making a release to generate a wheel using a recent version of setuptools for the METADATA entries to be correct.

I updated ``setup.py`` to fail on ``bdist_wheel`` command if an older version is used. I will also document this under committer docs.

I tested it locally using various Python versions and confirmed it works correctly.

The universal wheel metadata now looks like this:

```bash
...
Requires-Dist: requests (>=2.5.0)
Requires-Dist: backports.ssl-match-hostname ; python_version < "2.7.9"
Requires-Dist: typing ; python_version < "3.4.0"
Requires-Dist: enum34 ; python_version < "3.4.0"
...
```

## Note on supporting Python <= 3.4.0

In the very near future (#1377), we plan to drop support for Python <= 3.4.0.  At that point, this change won't be needed anymore and we will be able to simplify ``setup.py``.

The reason why I made this change is so we can make another release in the ``v2.x`` release series which contains the correct wheel metadata entries. This will be the last release which will support Python <= 3.4.0.

If we planned to support Python 2.7 and Python 3.4.0 for longer, I would probably also work on automated tests for this fix / change, but at this point, I don't think it's worth the effort since we will be dropping support for that versions soon.

Resolves #1392.

### Thanks

Thanks to @arterrey for reporting this issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,14 +14,14 @@ matrix:
   fast_finish: true
   include:
     - python: 2.7
-      before_script: TOX_ENV=py2.7,py2.7-dist
+      before_script: TOX_ENV=py2.7,py2.7-dist,py2.7-dist-wheel
     - python: 3.4
     - python: 3.5
     - python: 3.6
     - python: 3.7
       dist: xenial
       sudo: required
-      before_script: TOX_ENV=py3.7,py3.7-dist
+      before_script: TOX_ENV=py3.7,py3.7-dist,py3.7-dist-wheel
     - python: 3.8
       dist: xenial
       sudo: required
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,7 +5,7 @@ Changes in Apache Libcloud in development
 -----------------------------------------
 
 Common
--------
+------
 
 - Fix a regression with ``get_driver()`` method not working if ``provider``
   argument value was a string (e.g. using ``get_driver('openstack')``
@@ -25,6 +25,13 @@ Common
 
   NOTE: At the moment, type annotations are only available for the base
   compute API.
+  [Tomaz Muraus]
+
+- Fix universal wheel METADATA and ensure conditional dependencies
+  (backports.ssl_match_hostname, typing, enum34) are handled correctly.
+
+  Reported by Adam Terrey (@arterrey).
+  (GITHUB-1392, GITHUB-1393)
   [Tomaz Muraus]
 
 Compute
--- a/docs/committer_guide.rst
+++ b/docs/committer_guide.rst
@@ -95,6 +95,12 @@ preparing a release.
 3. Creating release artifacts
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+.. note::
+
+  It's important that you have the latest versions of ``setuptools``, ``wheel``
+  and ``pip`` installed to ensure the generated wheel files contain correct
+  metadata.
+
 We have a script that runs the required setup.py commands and then hashes
 and signs the files. You will need the latest version of ``pip`` and the ``wheel``
 package. To run it:
--- a/setup.py
+++ b/setup.py
@@ -18,6 +18,7 @@ import sys
 import re
 import fnmatch
 
+import setuptools
 from setuptools import setup
 from distutils.core import Command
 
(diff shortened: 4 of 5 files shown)


Kami:
@arterrey I believe this should fix the issue you reported in #1392.

I had quite some trouble testing it using ``buildout`` since latest version of buildout wheel plugin is broken and doesn't work with the latest version of ``wheel`` packages (so I couldn't test it using the locally build wheel from this branch).

In the end, I got a bit creative and confirmed it's working by building an egg from this branch locally and using this ``buildout.cfg`` (both using Python 2.7 and Python 3):

```ini
[buildout]
parts = libcloud

[libcloud]
recipe = zc.recipe.egg
eggs = apache-libcloud
index =
find-links = .
```

```bash
rm -rf eggs/*
python setup.py bdist_egg
# For Python 3 venv
cp dist/apache_libcloud-2.7.0-py3.6.egg eggs/ 
# For Python 2 venv
cp dist/apache_libcloud-2.7.0-py2.7.egg eggs/
```

I confirmed it now doesn't try to download typing and enum34 packages when using Python >= 3.4.0 (and it tries to download them when using Python 2.7).

Having said that, I would also appreciate if you could confirm the fix as well.

Kami:
@tonybaloney @erjohnso @vdloo Just a heads up^

> Keep in mind that this change now requires a Libcloud committer which is making a release to generate a wheel using a recent version of setuptools for the METADATA entries to be correct.
>
> I updated setup.py to fail on bdist_wheel command if an older version is used. I will also document this under committer docs.

I don't think it's likely that one of you will do another release in the ``v2.x`` series before we drop support for Python < 3.4.0, but in case you do, you need to use latest version of ``setuptools`` to ensure generated wheel contains correct metadata.

And on that note - I plan to do v2.8.0 once @arterrey confirms this fix is working. This will likely be the last major / minor release in the v2.x release series before dropping support for Python < 3.4.0.

Kami:
Ignore my "no tests" comment above.

I decided to improve the existing installation tests (c2d23d5dfb6b4abc7ed5c7008e39fae1e014a2a9) and add new ones which test wheel installation method (6ad129fcda51d91489ab979345939cda199a23de, 165151dcc2787c7ec9975d728cc977338b5aad99).

This means we now have automated tests which test installing the package using ``python setup.py install`` method and directly installing it from a wheel and verify that the correct dependencies are installed.

Those tests won't catch 100% of the issues (there are still some edge cases when using different Python version to build and install the wheel), but it's much better than nothing (and they would likely have caught and original issue reported by @arterrey in #1392).

arterrey:
Yes this is working now. Your awesome @Kami , thank you for looking into and fixing this so quickly. 

---

`bdist_wheel` the repo and placed in test dir

used the following `buildout.cfg`:
```
[buildout]
parts = libcloud

[libcloud]
recipe = zc.recipe.egg
eggs = apache-libcloud
index =
find-links = .
interpreter = py
```
```
$ python3 -venv .
$ bin/pip install -U pip setuptools zc.buildout
...
$ bin/buildout
Creating directory '/home/adam/tmp/libcloud/test/eggs'.
Creating directory '/home/adam/tmp/libcloud/test/parts'.
Creating directory '/home/adam/tmp/libcloud/test/develop-eggs'.
Getting distribution for 'zc.recipe.egg>=2.0.6'.
WARNING: The easy_install command is deprecated and will be removed in a future version.
Got zc.recipe.egg 2.0.7.
Installing libcloud.
Not found: /apache-libcloud/
Not found: /apache-libcloud/
Getting distribution for 'apache-libcloud'.
Got apache-libcloud 2.6.0.
Not found: /requests/
Not found: /requests/
Getting distribution for 'requests>=2.5.0'.
Got requests 2.22.0.
Not found: /urllib3/
Not found: /urllib3/
Getting distribution for 'urllib3!=1.25.0,!=1.25.1,<1.26,>=1.21.1'.
Got urllib3 1.25.7.
Not found: /idna/
Not found: /idna/
Getting distribution for 'idna<2.9,>=2.5'.
Got idna 2.8.
Not found: /chardet/
Not found: /chardet/
Getting distribution for 'chardet<3.1.0,>=3.0.2'.
WARNING: The easy_install command is deprecated and will be removed in a future version.
warning: no files found matching 'requirements.txt'
zip_safe flag not set; analyzing archive contents...
Got chardet 3.0.4.
Not found: /certifi/
Not found: /certifi/
Getting distribution for 'certifi>=2017.4.17'.
Got certifi 2019.9.11.
Generated interpreter '/home/adam/tmp/libcloud/test/bin/py'.
$ bin/py
>>> import enum
>>> enum.__file__
'/usr/lib64/python3.7/enum.py'
```

Kami:
@arterrey Great, thanks for confirming.

I will work on 2.8.0 release this week.

Kami:
@arterrey EDIT: From looking at the output you provided again, it looks like it used an egg from an old release (2.6.0) which you likely had cached locally.

To confirm the fix, you will probably need to delete rest of the libcloud eggs from ``eggs/`` directory, run ``python setup.py bdist_egg`` copy egg from ``dist/`` to ``eggs/`` and then run the buildout command (at least that's how I tested it).

PR #1394, opened 2019-12-23
Add KubeVirt driver & tests

Eis-D-Z:
### Description

Kubevirt driver with initial support for the k8s/KubeVirt add-on.

Can list, start, stop, destroy kubevirt type vm's from a kubernetes cluster.

Can create a Persistent Volume Claim with the assumption that the Persistent Volume will be created dynamically by Kubernetes, so a relevant storage class that allows for such provisioning must be declared. (You can list storage classes to view them)

Can create a vm with a docker image with embedded disk as image, and as many persistent volume claims as desired. 
  - Network support is limited to 'pod' with 'bridge' or 'masquearde' type interfaces.
 - From the supported disk types of KubeVirt, containerDisk is used for the image, while only  persistentVolumeClaims can be added as disks. Disk types can either be "disk" or "lun" and the supported bus are "virtio", "sata" and "scsi".

### Status

- working driver, can be further extended to include more features, needs docstrings
- Note: Many features of the driver are logically more suited to be in the Kubernetes container driver, please keep this in consideration in the review

### Checklist (tick everything that applies)

- [x] [Code linting](http://libcloud.readthedocs.org/en/latest/development.html#code-style-guide) (required, can be done after the PR checks)
- [ ] Documentation
- [x] [Tests](http://libcloud.readthedocs.org/en/latest/testing.html)
- [x] [ICLA](http://libcloud.readthedocs.org/en/latest/development.html#contributing-bigger-changes) (required for bigger changes)

PR #1395, opened 2019-12-23
WIP: Add LXD driver & tests

pockerman:
##WIP: Add support for LXC containers

### Description

Add support for LXC containers.
- Add ```container/drivers/lxd.py``` module with the following classes
- ```LXDAPIException``` simple exception class
-  ```LXDStoragePool``` wrapper of the info of a storage pool
- ```LXDServerInfo``` wrapper of the info of the LXC server
- ```LXDResponse``` represents the response to an LXD API call
- ```LXDConnection``` represents a connection
- ```LXDtlsConnection``` TLS connection
- ```LXDContainerDriver``` main class wrapping the LXD API

### Status

- work in progress

- [ ] Documentation


Kami:
Thanks for the contribution.

Since this is a larger contribution, can you please also work on signing an ICLA (https://libcloud.readthedocs.io/en/latest/development.html#contributing-bigger-changes)?

pockerman:
yes of course

libcloud/container/drivers/lxd.py
```diff
@@ -0,0 +1,1269 @@
...
+def strip_http_prefix(host):
+    # strip the prefix
+    prefixes = ['http://', 'https://']
+    for prefix in prefixes:
+        if host.startswith(prefix):
+            host = host.strip(prefix)
+    return host
+
+
+def check_certificates(key_file, cert_file, **kwargs):
```
Kami, line 63: It appears that some of that logic is similar to the one in #1394 so once those PRs are merged, both of the drivers could perhaps be refactored to use those common utility functions.

libcloud/container/drivers/lxd.py
```diff
@@ -0,0 +1,1269 @@
...
+                                'authenticate with LXD tls. '
+                                'This can be found in the server.')
+
+
+def assert_response(response_dict, status_code=200):
+
+    # if the type of the response is an error
+    if response_dict['type'] == LXD_ERROR_STATUS_RESP:
+        # an error returned
+        raise LXDAPIException(message="response type is error")
```
Kami: For ease of debugging / troubleshooting, would it make sense to also include ``respose_dict`` value in the actual exception error message?

libcloud/container/drivers/lxd.py
```diff
@@ -0,0 +1,1269 @@
...
+            return self.body
+
+        try:
+            # error responses are tricky in Docker. Eg response could be
+            # an error, but response status could still be 200
+            content_type = self.headers.get('content-type', 'application/json')
+            if content_type == 'application/json' or content_type == '':
+                if self.headers.get('transfer-encoding') == 'chunked' and \
+                        'fromImage' in self.request.url:
+                    body = [json.loads(chunk) for chunk in
```
Kami, line 247: Eventually it would be nice if we could refactor some of the transfer-encoding chunked logic to a base ``ChunkedEncodingJsonResponse`` or similar Response class.

libcloud/container/drivers/lxd.py
```diff
@@ -0,0 +1,1269 @@
...
+                    body = json.loads(self.body)
+            else:
+                body = self.body
+        except ValueError:
+            m = re.search('Error: (.+?)"', self.body)
+            if m:
+                error_msg = m.group(1)
+                raise Exception(error_msg)
+            else:
+                raise Exception(
```
Kami: Should we also include original response in the error message body for ease of debugging?

libcloud/container/drivers/lxd.py
```diff
@@ -0,0 +1,1269 @@
...
+
+class LXDtlsConnection(KeyCertificateConnection):
+
+    responseCls = LXDResponse
+
+    def __init__(self, key, secret, secure=True,
+                 host='localhost', port=8443, ca_cert='',
+                 key_file=None, cert_file=None, **kwargs):
+
+        if 'certificate_validator' in kwargs.keys():
```
Kami: Please explicitly declare supported arguments (``certificate_validator``) which are known up front instead of using ``**kwargs`` (I know a lot of the existing drivers did, but we just tried to clean and improve some of that legacy code).

libcloud/container/drivers/lxd.py
```diff
@@ -0,0 +1,1269 @@
...
+        :rtype: :class: .LXDServerInfo
+
+        """
+        response = self.connection.request("/%s" % (self.version))
+        response_dict = response.parse_body()
+        assert_response(response_dict=response_dict, status_code=200)
+        meta = response_dict["metadata"]
+        return LXDServerInfo.build_from_response(metadata=meta)
+
+    def deploy_container(self, name, image, cluster=None,
```
Kami, line 427: Please prefix arguments which are not part of the standard container API with ``ex_`` (profiles, ephemeral, config, etc,).

libcloud/container/drivers/lxd.py
```diff
@@ -0,0 +1,1269 @@
...
+                         parameters=None, start=True,
+                         architecture=default_architecture,
+                         profiles=[default_profiles],
```
Kami: You don't appear to be mutating ``profiles`` inside the function body, but it would still be safer and less error prone to avoid mutable defaults (https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments) and something like ``profiles = None``
and inside the body ``profiles = proxies or [default_profiles]``).

libcloud/container/drivers/lxd.py
```diff
@@ -0,0 +1,1269 @@
...
+        container = self._deploy_container_from_image(name=name, image=image,
+                                                      parameters=parameters,
+                                                      cont_params=c_params)
+
+        if start:
+            container.start()
+
+        return container
+
+    def get_container(self, id, get_ip_addr=True):
```
Kami: Please prefix ``get_ip_addr`` with ``ex_`` since it's not part of the standard API.

libcloud/container/drivers/lxd.py
```diff
@@ -0,0 +1,1269 @@
...
+                # the list of addresses
+                addresses = networks["addresses"]
+
+                for item in addresses:
+                    ips.append(item["address"])
+
+        metadata.update({"ips": ips})
+        return self._to_container(metadata=metadata)
+
+    def start_container(self, container, timeout=default_time_out):
```
Kami: Please prefix ``timeout`` with ``ex_`` since it's not part of the standard API.

libcloud/container/drivers/lxd.py
```diff
@@ -0,0 +1,1269 @@
...
+        :param timeout: Time to wait for the operation to complete
+        :type  timeout: ``int``
+
+        :rtype: :class:`libcloud.container.base.Container`
+        """
+        return self._do_container_action(container=container, action='start',
+                                         timeout=timeout,
+                                         force=True, stateful=True)
+
+    def stop_container(self, container, timeout=default_time_out):
```
Kami: Please prefix ``timeout`` with ``ex_`` since it's not part of the standard API.

libcloud/container/drivers/lxd.py
```diff
@@ -0,0 +1,1269 @@
...
+        :type  timeout: ``int``
+
+        :return: The container refreshed with current data
+        :rtype: :class:`libcloud.container.base.Container
+        """
+        return self._do_container_action(container=container, action='stop',
+                                         timeout=timeout,
+                                         force=True, stateful=True)
+
+    def restart_container(self, container, timeout=default_time_out):
```
Kami: Please prefix ``timeout`` with ``ex_`` since it's not part of the standard API.

libcloud/container/drivers/lxd.py
```diff
@@ -0,0 +1,1269 @@
...
+        :param timeout: Time to wait for the operation to complete
+        :type  timeout: ``int``
+
+        :rtype: :class:`.Container`
+        """
+        return self._do_container_action(container=container, action='restart',
+                                         timeout=timeout,
+                                         force=True, stateful=True)
+
+    def destroy_container(self, container, timeout=default_time_out):
```
Kami: Please prefix ``timeout`` with ``ex_`` since it's not part of the standard API.

libcloud/container/drivers/lxd.py
```diff
@@ -0,0 +1,1269 @@
...
+
+        :param timeout: Time to wait for the operation to complete
+        :type timeout ``int``
+
+        :rtype: :class:`.Container`
+        """
+
+        # if the container is running then we cannot delete
+
+        if container.state == ContainerState.RUNNING:
```
Kami: I'm fine with doing that check here although we don't really have clear rules / guidelines on how to handle such scenarios in Libcloud drivers.

A lot of drivers simply let the API handle that validation instead of doing it in the driver. This way it still works if user doesn't have the latest reference to the ``container`` object locally.

So another option would be to simply remove that check here.

libcloud/container/drivers/lxd.py
```diff
@@ -0,0 +1,1269 @@
...
+
+        # return a dummy container
+        container = Container(driver=self, name=container.name,
+                              id=container.name,
+                              state=ContainerState.TERMINATED,
+                              image=None, ip_addresses=[],
+                              extra=None)
+        return container
+
+    def list_containers(self, image=None, cluster=None, detailed=True):
```
Kami: Please prefix ``detailed`` with ``ex_``.

libcloud/container/drivers/lxd.py
```diff
@@ -0,0 +1,1269 @@
...
+        req = '/%s/images/%s' % (self.version, fingerprint)
+        response = self.connection.request(req)
+
+        #  parse the LXDResponse into a dictionary
+        response_dict = response.parse_body()
+        assert_response(response_dict=response_dict, status_code=200)
+
+        return self._to_image(metadata=response_dict['metadata'])
+
+    def install_image(self, path, **description_dict):
```
Kami: Ideally, all those non standard arguments would be prefixed with ``ex_`` or you would take  a single ``ex_description_dict`` argument which is a dictionary.

Kami:
Nice work and thanks again for the contribution.

I saw some similarities with #1394, are those PRs somewhat related?

pockerman:
both contributions are part of an extension underway for mist.io

libcloud/container/drivers/lxd.py
pockerman: Basically, this is a copy verbatim from the Docker driver which means I should work more on it at some point. PyCharm shows 138 implementations of ```JsonResponse``` we need someone brave :) :)

libcloud/container/drivers/lxd.py
pockerman: basically I am seriously thinking to remove this altogether...I put it there initially to get more informative errors when trying to use TLS. I am happy to so as you suggested but what do you think about removing it completely?

libcloud/container/drivers/lxd.py
pockerman: ok I can remove it then...

Kami:
Thanks for addressing the review comments.

There are still a couple of things which would need some work and improvements (refactoring), but those are not blockers so we can go ahead and merge this as is and improve other things in separate PRs.

(I also pushed two small changes to fix the build)

PR #1396, opened 2019-12-24
Also run unit tests on Windows using Python 3.7

Kami:
Just an experiment to see if we can get unit tests to run under Python 3.7 on Windows.

IIRC, at some point in the past, we did have some unit tests running in Windows, but for one reason or another, they were removed.

Now that #1377 has been merged, this has freed up some Travis CI builders which means that even if we add another builder, in theory, overall test run time should still be the same / shouldn't increase (as long as Windows build doesn't take more than ~2 minutes).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,10 +69,10 @@ matrix:
         - python3 ./contrib/trigger_rtd_build.py
 
 install:
-  - pip install --upgrade "pip==19.3.1"
-  - pip install --upgrade "setuptools==42.0.2"
-  - pip install "virtualenv==16.7.6"
-  - pip install "tox==3.14.2"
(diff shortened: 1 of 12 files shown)


Kami:
Alright, unit tests now pass on Windows.

Only downside is that those tests take ~3 min 40 seconds, whereas most other tests / jobs take maximum of ~2-3 minutes.

The reason for that is that Travis Windows build support is not really great (caching is not really implemented, etc.).

I will go ahead and merge it and if slow tests become an issue, we can always mark those windows unit tests as "allow failure" and set "fail fast" Travis option so the whole build won't wait / block on Windows unit tests job.

I still think it's good to have those tests since in the past we had some small Windows related issues (e.g. maximum file name length limit on Windows).

PR #1397, opened 2019-12-26
Promote `ex_prefix` to top-level storage SDK argument

c-w:
### Description

Most storage drivers implement `ex_prefix` for `list_container_objects` and `iterate_container_objects`. As such, this pull request promotes `ex_prefix` to a top-level argument. For each driver that didn't previously implement `ex_prefix`, a fallback to filter objects client-side is has also been added.

### Status

- done, ready for review

### Checklist

- [x] Code linting ([build passed](https://travis-ci.org/CatalystCode/libcloud/builds/629544380))
- [x] Documentation (updated existing docstrings)
- [x] Tests (updated existing tests)
- [x] ICLA ([committer](http://people.apache.org/phonebook.html?uid=clewolff))


Kami:
Overall, I'm fine with this change, but another option would be to simply name the argument ``prefix`` (aka promote it to be part of the base API).

Then we could also have a fallback, if driver doesn't implement ``_filter_objects_by_prefix`` method (or similar) we would simply fall back to the implementation on the base driver which performs simple local filtering on the whole result set.

Since that method would perform filtering locally, it would probably also be a good idea to document that behavior (perhaps also emit a warning?) to avoid surprises.

c-w:
@Kami Would promoting `ex_prefix` to `prefix` be a breaking API change or would we also keep the `ex_prefix` argument? If we're considering a breaking API change, I'd suggest to first merge this pull request (which is backwards compatible so potentially releasable via a patch version) and consider the argument promotion for a future change.

Kami:
@c-w That's correct, but since it wasn't previously part of the base API, we probably should only update drivers and methods which supported ``ex_prefix`` argument before, to also support it for backward compatibility reasons going forward (at appears those are azure, cloudfiles, oss, dummy and s3 drivers).

And inside those methods we should probably just do something like ``prefix = prefix or ex_prefix`` and document in upgrade notes that the new argument name is ``prefix`` which has precedence over ``ex_prefix``, which was left in place for backward compatibility reasons.

In short - I would just like to avoid adding ``ex_prefix`` to the base API since base API shouldn't really have any ``ex_`` prefixed arguments and methods.

If you won't get a chance to make those changes in the near future, I can also look into it.

c-w:
Thanks for the additional context @Kami. I'll work on the change and update this PR.

libcloud/storage/drivers/atmos.py
```diff
@@ -451,11 +451,35 @@ def _emc_meta(self, response):
...
+
         headers = {'x-emc-include-meta': '1'}
         path = self._namespace_path(container.name) + '/'
         result = self.connection.request(path, headers=headers)
         entries = self._list_objects(result.object, object_type='regular')
-        for entry in entries:
-            metadata = {'object_id': entry['id']}
-            yield Object(entry['name'], 0, '', {}, metadata, container, self)
+        objects = self._entries_to_objects(container, entries)
+        return self._filter_listed_container_objects(objects, prefix)
```
Kami, line 480: I was thinking about doing this a bit differently, but that approach also works.

We just need to make sure that all the drivers which don't implement custom / server side based filtering call that method.

Kami (approved):
Nice work, LGTM :+1: 

When you get a chance it would also be good to add upgrade notes entry (document this new argument and advise users to migrate from ``ex_prefix`` to ``prefix`` if they currently use ``ex_prefix``).

c-w:
Added upgrade note in 16c692f.

PR #1409, opened 2020-01-14
Fix pagination in the "list_nodes()" method in the GCE driver

### LINKED ISSUE #1360: list_nodes on GCE Projects Only Returns 500 Results
## Summary

When running list_nodes against a GCE environment that has greater than 500 instances, only the top 500 results will be returned.

## Detailed Information

Using libcloud 2.6.0, running list_nodes using the GCE Provider will only return the top 500 results. This is similar to issue #1203 where the list_nodes function is running self.connection.request rather than self.connection.request_aggregated_items.

I believe changing lines 19-23 of the list_nodes function to something like this will fix the issue:

```
if zone is None:
        response = self.connection.request_aggregated_items('instances')
    else:
        request = '/zones/%s/instances' % (zone.name)
        response = self.connection.request(request, method='GET').object
```

Kami:
This pull request is an attempt to try to fix an issue reported in #1360.

### Background / Problem Description

Currently ``list_nodes()`` method only returns up to 500 nodes (which is a page size value we use with our list requests and also a max page size limit set by Google).

This is true either when we are requesting nodes for all the zones (aggregated request) or when requesting nodes for a single zone.

### Proposed Fix

This pull request fixes that by utilizing ``request_aggregated_items`` method in both scenarios (when we are requesting nodes in all the zones and also when requesting nodes for a single zone).

I tested the change by setting ``maxResults`` in the ``request_aggregated_items`` method to ``1`` and verified that all the pages are correctly retrieved since I don't have an account with more than 500 instances to test it out.

I tested the following scenarios:

1. List nodes for all the zones
2. List nodes for a particular zones with multiple nodes in that zone
3. List nodes for a particular zone with no nodes in that zones

---

On a somewhat related note -  unless I'm missing something, it appears the same issue may also exist with other "list" API endpoints when retrieving items for a specific zone and that response contains more then max page size number of items (since we don't follow through pagination when requesting items for a single zone).

As you can see in the diff, this change also allowed me to simplify ``list_nodes()`` method. If my assertion / observation above is correct, this would mean we can also simplify other list methods.

@erjohnso Please correct me if I'm wrong or missing something.

Resolves #1360.

### DIFF
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -171,7 +171,6 @@ class GCEConnection(GoogleBaseConnection):
                     # Special case when we are handling pagination for a
                     # specific zone
                     items = response['items']
-                    response['items'] = {}
                     response['items'] = {
                         'zones/%s' % (zone): {
                             api_name: items
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -2586,31 +2586,35 @@ class GCENodeDriver(NodeDriver):
         :return:  List of Node objects
         :rtype:   ``list`` of :class:`Node`
         """
-        list_nodes = []
         zone = self._set_zone(ex_zone)
         response = self.connection.request_aggregated_items('instances',
                                                             zone=zone)
 
-        if 'items' in response:
-            # The aggregated response returns a dict for each zone
-            # Create volume cache now for fast lookups of disk info.
-            self._ex_populate_volume_dict()
-            for v in response['items'].values():
-                for i in v.get('instances', []):
-                    try:
-                        list_nodes.append(
-                            self._to_node(i,
-                                          use_disk_cache=ex_use_disk_cache)
-                        )
-                    # If a GCE node has been deleted between
-                    #   - is was listed by `request('.../instances', 'GET')
-                    #   - it is converted by `self._to_node(i)`
-                    # `_to_node()` will raise a ResourceNotFoundError.
-                    #
-                    # Just ignore that node and return the list of the
-                    # other nodes.
-                    except ResourceNotFoundError:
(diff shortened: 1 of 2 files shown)


libcloud/compute/drivers/gce.py
```diff
@@ -159,6 +166,16 @@ def request_aggregated_items(self, api_name):
             self.gce_params = params
             response = self.request(request_path, method='GET').object
             if 'items' in response:
+                if zone:
+                    # Special case when we are handling pagination for a
+                    # specific zone
+                    items = response['items']
+                    response['items'] = {}
+                    response['items'] = {
+                        'zones/%s' % (zone): {
```
Kami, line 175: In aggregated request response, this key will contain the zone name (e.g. ``zones/europe-west1-b``), but the actual key is not used anywhere so the name is irrelevant and could really contain any value (it's just the structure of the actual value that matters).

libcloud/compute/drivers/gce.py
```diff
@@ -2569,50 +2586,35 @@ def list_nodes(self, ex_zone=None, ex_use_disk_cache=True):
...
         zone = self._set_zone(ex_zone)
-        if zone is None:
-            request = '/aggregated/instances'
-        else:
-            request = '/zones/%s/instances' % (zone.name)
-        response = self.connection.request(request, method='GET').object
+        response = self.connection.request_aggregated_items('instances',
+                                                            zone=zone)
+
+        if not response.get('items', []):
```
Kami, line 2593: This return early change is purely for style purpose so we can make the code a bit easier to read and get rid of one level of indentation on "happy" code path when response contains results.

Kami:
I'll merge this since I'm working on v3.0.0rc1 release so that change can be included as well.

PR #1408, opened 2020-01-10
Implement SAS URL generation for Azure Storage

c-w:
### Description

As discussed in https://github.com/apache/libcloud/issues/1403, this pull request implements `AzureBlobsStorageDriver.get_object_cdn_url` via a [service SAS](https://docs.microsoft.com/en-us/rest/api/storageservices/create-service-sas#construct-a-service-sas).

### Status

- done, ready for review

### Checklist

- [x] Code linting ([build passed](https://travis-ci.org/CatalystCode/libcloud/builds/635449613))
- [x] Documentation ([updated changelog](https://github.com/CatalystCode/libcloud/blob/azure-sas-url/CHANGES.rst#storage))
- [x] Tests ([e2e tests passed](https://clewolff.visualstudio.com/libcloud-tests/_build/results?buildId=682))
- [x] ICLA ([committer](https://people.apache.org/phonebook.html?uid=clewolff))

PR #1410, opened 2020-01-14
Add type annotations to storage API

c-w:
### Description

As discussed in https://github.com/apache/libcloud/pull/1408#discussion_r366147905, this PR adds type annotations to the StorageDriver API.

For now, mypy errors are ignored for the s3, google_storage and cloudfiles drivers since the use of inheritance on the connection classes is causing issues which will require a more thorough strategy to address.

### Status

- done, ready for review

### Checklist

- [x] Code linting ([build passed](https://travis-ci.org/CatalystCode/libcloud/builds/636950988))
- [x] Documentation (updated docstrings)
- [x] Tests ([build passed](https://travis-ci.org/CatalystCode/libcloud/builds/636950988))
- [x] ICLA ([committer](https://people.apache.org/phonebook.html?uid=clewolff))

### DIFF
--- a/libcloud/storage/base.py
+++ b/libcloud/storage/base.py
@@ -20,6 +20,7 @@ Provides base classes for working with storage
 # Backward compatibility for Python 2.5
 from __future__ import with_statement
 
+from typing import Dict
 from typing import Iterator
 from typing import List
 from typing import Optional
@@ -425,7 +426,7 @@ class StorageDriver(BaseDriver):
 
     def upload_object(self, file_path, container, object_name, extra=None,
                       verify_hash=True, headers=None):
-        # type: (str, Container, str, Optional[dict], Optional[bool], Optional[dict]) -> Object  # noqa: E501
+        # type: (str, Container, str, Optional[dict], bool, Optional[Dict[str, str]]) -> Object  # noqa: E501
         """
         Upload an object currently located on a disk.
 
--- a/libcloud/storage/base.py
+++ b/libcloud/storage/base.py
@@ -20,6 +20,9 @@ Provides base classes for working with storage
 # Backward compatibility for Python 2.5
 from __future__ import with_statement
 
+from typing import Iterator
+from typing import List
+from typing import Optional
 from typing import Type
 
 import os.path                          # pylint: disable-msg=W0404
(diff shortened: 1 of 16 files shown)


libcloud/storage/base.py
```diff
@@ -383,28 +408,32 @@ def download_object(self, obj, destination_path, overwrite_existing=False,
...
         :type chunk_size: ``int``
+
+        :rtype: ``iterator`` of ``bytes``
         """
         raise NotImplementedError(
             'download_object_as_stream not implemented for this driver')
 
     def upload_object(self, file_path, container, object_name, extra=None,
                       verify_hash=True, headers=None):
+        # type: (str, Container, str, Optional[dict], Optional[bool], Optional[dict]) -> Object  # noqa: E501
```
Kami: Since ``verify_hash`` has a default value, I don't think it needs ``Optional`` annotation.

Kami: And for ``headers`` we could probably use ``Dict[str, str]``.

Kami:
Nice work :+1: 

---

What's your take on utilizing Python 3 native syntax for type annotations vs using comments?

In theory, using comments, allows the trunk to still work with Python 2.7, but since we don't officially support it any more, I'm not sure if that matters or not since sooner or later we will likely start utilizing other Python 3 only features as well (and we have v2.8.x branch we can base v2.8.x bug fix release on, if needed).

libcloud/storage/base.py
c-w: Done in d80b17c.

c-w:
> What's your take on utilizing Python 3 native syntax for type annotations vs using comments?

@Kami Personally I'm not a huge fan of the type comments for function signatures and would be in favor of eventually switching to the native type annotations to increase readability.

E.g. compare the following two snippets:

```py
def upload_object_via_stream(self, iterator, object_name, extra=None,
                             headers=None):
    # type: (Iterator[bytes], str, Optional[dict], Optional[Dict[str, str]]) -> Object  # noqa: E501
```

```py
def upload_object_via_stream(self,
                             iterator: Iterator[bytes],
                             object_name: str,
                             extra: Optional[dict] = None,
                             headers: Optional[Dict[str, str]] = None,
                             ) -> Object:
```

From my point of view, the former is harder to read since the function arguments must be matched with the type declaration based on index/position as opposed to being attached to each argument explicitly.

c-w:
The type annotations uncovered a bug: most drivers were discarding the base API's `headers` argument in `upload_object` and `upload_object_via_stream`. I fixed this in 3b6cdc9.

Kami (approved):

Kami:
@c-w Good catch on the headers issue :+1: 

As far as comments notation goes - in-line annotations can also be used when using comments, but I also agree that native notation is nicer.

So perhaps now that we don't need to support Python 2.7 anymore, we can just switch to that notation.

PR #1414, opened 2020-01-20
Fix incorrect type annotations in the base compute API

### LINKED ISSUE #1413: mypy issues as of 2.8.0
## Summary
Just wanted to feed back to you all a couple issues we had running mypy on our custom drivers after upgrading to libcloud 2.8.0.

## Detailed Information
- The type of NodeDriver.type is Provider, but our custom drivers assign a string to this field which cause a mypy error.  Perhaps it should be Union[Provider, str]? 
- NodeSize, NodeLocation, and NodeImage have driver typed to be Type[NodeDriver] in their constructors, but this does not match Node, StorageVolume, and others which have it typed as NodeDriver.  We have subclasses of some of these and use self.driver as an instance of a driver, just like the Node class, and not as a driver class.

Kami:
This pull request fixes a couple of incorrect type annotations in the base compute API as reported by #1413.

Resolves #1413.

### DIFF
--- a/libcloud/compute/base.py
+++ b/libcloud/compute/base.py
@@ -350,7 +350,7 @@ class NodeSize(UuidMixin):
                  disk,  # type: int
                  bandwidth,  # type: Optional[int]
                  price,  # type: float
-                 driver,  # type: Type[NodeDriver]
+                 driver,  # type: NodeDriver
                  extra=None  # type: Optional[dict]
                  ):
         """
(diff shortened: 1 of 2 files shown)


libcloud/compute/base.py
```diff
@@ -822,7 +822,7 @@ class NodeDriver(BaseDriver):
     name = None  # type: str
     api_name = None  # type: str
     website = None  # type: str
-    type = None  # type: Provider
+    type = None  # type: Union[Provider,str]
```
Kami, line 825: NOTE: Ideally this would always be ``Provider`` enum, but a lot of the existing code uses ``str`` type so we still need to support it for backward compatibility reasons.

c-w (approved):

PR #1423, opened 2020-02-08
[WIP] Google Cloud Platform auth and GKE driver improvements

Kami:
This pull request includes a couple of GCP auth and GKE driver fixes / improvements:

1. Fix ``list_clusters()`` method in GKE driver and make sure it returns ``ContainerCluster`` object instances. Previously it incorrectly returned raw dict as returned by the API.
2. Make sure ``list_clusters()`` works if ``ex_zone`` argument is not specified (aka listing clusters for all the zones)
3. Make sure we specify correct default auth scope otherwise authentication will fail aka driver won't work.
4. Update ``GoogleOAuth2Credential`` class and make sure we include hashed version of the used authentication scopes in the name of the file which is used to cache the auth token. If we don't do that, this class will try to incorrectly used the old / invalid cached auth token when user changes the value for the ``scopes=`` driver constructor argument.

Keep in mind that this will invalidate any existing cached token on Libcloud upgrade, but this shouldn't be a big deal since those tokens are short lived anyway.

## TODO

- [ ] Tests
- [ ] Changelog entry

### DIFF
--- a/libcloud/common/google.py
+++ b/libcloud/common/google.py
@@ -80,6 +80,7 @@ import datetime
 import os
 import socket
 import sys
+import hashlib
 
 from libcloud.utils.connection import get_response_object
 from libcloud.utils.py3 import b, httplib, urlencode, urlparse, PY3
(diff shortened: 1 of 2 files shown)


jhadjar:
Hi. Any update on this PR?

Kami:
Closing, superseded by #1599.

PR #1426, opened 2020-02-16
Minor bugfix: ec2 driver create_node ex_userdata broken in tag 2.8

Chaffelson:
### Description

Release tag 2.8 reworked the kwargs based implementation of the ec2 driver, introducing this error where the literal string 'ex_userdata' was passed during create_node intead of the actual ex_userdata kwarg
### Status

done, ready for review

- [x] Documentation

Packaged tests are broken, tested manually using 3.0.0rc1
No change to documentation
No material change to code linting status

### DIFF
--- a/libcloud/compute/drivers/ec2.py
+++ b/libcloud/compute/drivers/ec2.py
@@ -1970,7 +1970,7 @@ class BaseEC2NodeDriver(NodeDriver):
             params['KeyName'] = ex_keyname
 
         if ex_userdata:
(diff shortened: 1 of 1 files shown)


Kami:
Thanks and a good catch.

That was a bad bug :/

We also need to make sure we have some tests for this scenario to avoid such regressions in the future.

Kami:
I've added a test case (5593d6cd0c06d1bb0416d157a89a956ba249092d) and merged it into trunk.

Thanks for contribution.

PR #1431, opened 2020-02-20
Add new base storage API methods for partial (range) object downloads

Kami:
This pull request adds two new methods to the base storage API for downloading object range aka part of an object - ``download_object_range`` and ``download_object_range_as_stream``.

The reason why I added new methods and not new arguments to the existing ``download_object`` and ``download_object_as_stream`` methods is because of how "advertisement" of supported methods and functionality works in Libcloud.

If I added a new argument, there would be no easy programatically detectable way for driver to "advertise" that it support range download functionality.

Our option would be to throw an exception saying that this functionality is not supported if user passed in ``start_bytes`` or similar argument to that method, but the problem is that this would require us or user to actually run a method to determine if that functionality is available.

To begin with I implemented it for the following drivers:

- S3 (tested it with AWS S3 and Google Storage)
- Local storage

I also planned to look into adding it to the Azure driver.

## TODO

- [x] Test
- [x] Code refactor to avoid some duplicated code
- [x] Documentation changes
- [x] Upgrade notes entry


Kami:
@c-w @tonybaloney @erjohnso Would also like to hear your feedback on the method vs argument point and method and argument naming.

As far as naming goes,  perhaps``start_offset`` and ``end_offset`` would be better? But I did want to signal that it refers to bytes offset since it appears some providers also support range download units which are not bytes.

Kami:
I've added test cases, etc.

First I used 1-base indexing, but then I decided to use 0 based one, so it's consistent with Python string / list indexing behavior. (aka ``content[start_bytes:end_bytes]``). This way first index is ``0`` and last one is ``len(obj_content)``.

And it appears it also works in that manner with the S3 provider.

If that's not the case for some other provider, we will need to handle that inside that provider driver to expose a consistent API to our users.

c-w (commented):
Design of new methods LGTM.

Given that we're using a [storage API version newer than 2011-08-18](https://github.com/apache/libcloud/blob/62c9c855f4fb8b409307cdc2f8358dda8f36f58a/libcloud/storage/drivers/azure_blobs.py#L197), the range download implementation for Azure Blobs should be fairly similar to S3: see [range header docs](https://docs.microsoft.com/en-us/rest/api/storageservices/Specifying-the-Range-Header-for-Blob-Service-Operations) and [implementation in the first-party SDK](https://github.com/Azure/azure-storage-python/blob/2ee73e7abe3670f5e1e76e9e53b32f90c1f16691/azure-storage-blob/azure/storage/blob/_serialization.py#L85-L97).

If you'd like to validate a future implementation against real Azure Storage accounts, feel free to open a PR on [libcloud-tests](https://github.com/c-w/libcloud-tests), adding a new test case to [SmokeStorageTest](https://github.com/c-w/libcloud-tests/blob/564f24488952ec8d53e20994ec7769a0a423b539/tests/test_storage.py#L18). I didn't find any documentation on using `x-ms-range` with Azurite and IoT Edge Storage, so I'd assume that those storage types should also support range downloads.

Kami:
@c-w Thanks for the response.

I pushed some changes which implement that functionality for Azure Blobs driver (it's pretty much the same as the code for S3 based drivers) - 76cde3007a60ced15ffa7c1eee9f6fafd5baa3c6, 8b16adde57f276b4613a63abb3b17385e0890d4b.

At the moment, I tested it manually using my Azure account.

Kami:
This PR is now more or less ready to be merged.

I just need to decide (and make any code changes, if necessary) if the ``end_bytes`` offset should be inclusive or not. Currently, it is inclusive.

In Python indexing world, end index is not inclusive, but for Range header values, end offset seems to be inclusive for all the providers I tested it with (S3, Google Storage, Azure Blobs).

To put it into perspective - if you specify ``Range: bytes=0-5`` header value and the object content is ``0123456789``, you would get ``012345`` value back.

This would map to ``content[0: 5 + 1]`` in the Python world.

We can hide this behavior from the end user inside the implementation details, we just need to decide which behavior makes the most sense. I'm leaning towards the Python indexing behavior since Libcloud users are working with Python after all.

Kami:
I've also implemented those new methods for the CloudFiles drivers, added more tests and changed end_bytes offset to be non-inclusive so it's consistent with Python indexing.

I will go ahead and merge this into trunk.

PR #1432, opened 2020-02-20
Make id lower for consistent uuid generation

foospidy:
### Description

When a node is created, the resource groups name in the id string is lower case. In subsequent calls, like in list_nodes(), for some reason Azure returns the resource groups name in upper case. This results in a different uuid from when create_node is called. This also breaks wait_until_running(). Forcing lower case in _to_node() seems to resolve this issue.

### Status

- done, ready for review

### Checklist

- [ ] Documentation

### DIFF
--- a/libcloud/compute/drivers/azure_arm.py
+++ b/libcloud/compute/drivers/azure_arm.py
@@ -2235,7 +2235,7 @@ class AzureNodeDriver(NodeDriver):
             elif ps == "succeeded":
                 state = NodeState.RUNNING
 
-        node = Node(data["id"],
+        node = Node(data["id"].lower(),
                     data["name"],
                     state,
                     public_ips,


Kami:
Thanks for the contribution.

Tests are failing (will likely need to rename the fixture or or change how the names are handled on the mock http class) so this will need a bit more work.

We will also need some tests for this change and verify it indeed works correctly end to end.

Kami:
@c-w I assume it it's a good idea to test this with actual Azure deployment, right?

c-w:
@Kami Yes, I agree that we should validate this against a real Azure environment. I have a subscription set up which I'm using for libcloud integration testing which we can use for this.

@foospidy Would it be possible for you to add a test script for the functionality to the [libcloud-tests](https://github.com/c-w/libcloud-tests) repo (e.g. could be as simple as creating a VM and then deleting it again)? The repo executes a suite of integration tests on a schedule so that we get alerted if there are changes in the underlying cloud APIs which cause a breakage in libcloud. Authentication to Azure is handled via a service principal (see the `auth_*` variables in [azure-pipelines-test-azure.yaml](https://github.com/c-w/libcloud-tests/blob/564f24488952ec8d53e20994ec7769a0a423b539/azure-pipelines-test-azure.yaml#L20)).

foospidy:
@c-w Thanks for the tips on setting up a test. I'll try to look into it soon, hoping I have some time free up within the next few weeks.

Kami:
Thanks for your contribution.

PR #1434, opened 2020-03-01
Add type annotations for the base DNS API

Kami:
This pull request adds type annotations for the base DNS API.

### DIFF
--- a/example_compute.py
+++ b/example_compute.py
@@ -21,14 +21,8 @@ from libcloud.compute.drivers.rackspace import RackspaceNodeDriver
 
 from typing import Type, cast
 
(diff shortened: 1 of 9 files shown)


example_compute.py
```diff
@@ -21,14 +21,8 @@
 
 from typing import Type, cast
 
-ec2_cls = get_driver(Provider.EC2)
```
Kami, line 24: Sadly this doesn't work as expected.

This means we need to rely on base / parent class annotations until we have type annotations for all the driver subclasses.

https://github.com/apache/libcloud/pull/1306#issuecomment-593092777

PR #1436, opened 2020-03-03
Fill auth_user_info in v3 connection

lln-ijinus:
### Description

When changing the OS_AUTH_VERSION to '3.x_password' instead of '2.0_password', we lose the user id information. This information is important to link for example snapshots to the user that made them.

This PR offers to restore the auth_user_info  when authenticate with OpenStackIdentity_3_0_Connection

### Status
ready for review

- [ x] [Code linting](http://libcloud.readthedocs.org/en/latest/development.html#code-style-guide) (required, can be done after the PR checks)
- [ x] Documentation
- [ x] [Tests](http://libcloud.readthedocs.org/en/latest/testing.html)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,7 +5,7 @@ Changes in Apache Libcloud in development (3.0.0)
 -------------------------------------------------
 
 Common
-~~~~~~~
+~~~~~~
 
 - Fix ``LIBCLOUD_DEBUG_PRETTY_PRINT_RESPONSE`` functionality and make sure it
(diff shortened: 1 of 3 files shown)


Kami:
Thanks for the contribution.

It would also be good to add a test case for it.

Kami:
I've added some tests for v3.x auth versions.

It looks like they weren't added when 3.x auth version support was initially implemented.

Kami:
Merged change into trunk.

PR #1437, opened 2020-03-03
Implement g8 provider for libcloud

grimpy:
Signed-off-by: Jo De Boeck <jo@gig.tech>

### Description

Implement driver for https://gig.tech/ it's G8

### Status

- done, ready for review

- [ ] Documentation


Kami:
Thanks for the contribution.

In the mean time, while we review it, can you please work on filling your ICLA?

Thanks.

grimpy:
@Kami I filled the ICLA where do I send it to?

Kami:
@grimpy Please email it to secretary@apache.org as per https://www.apache.org/licenses/contributor-agreements.html#submitting

(I need to update ICLA link in the issue template, it looks like the page got moved around)

grimpy:
@Kami ICLA has been send

libcloud/test/compute/test_gig_g8.py
```diff
@@ -0,0 +1,151 @@
...
+from libcloud.test.file_fixtures import ComputeFileFixtures
+
+
+class G8MockHttp(MockHttp):
+    """Fixtures needed for tests related to rating model"""
+    fixtures = ComputeFileFixtures('gig_g8')
+
+    def __getattr__(self, key):
+        def method(method, path, params, headers):
+            response = self.fixtures.load('{}.json'.format(key.lstrip("_")))
```
Kami: I know a lot of existing test cases don't do that, but it's also preferred to check HTTP method as well (perhaps you could add it to the filename, e.g. ``POST_machines_create.json`` or similar.

libcloud/compute/drivers/gig_g8.py
```diff
@@ -0,0 +1,501 @@
...
+
+        """
+        location = self._location["locationCode"]
+
+        sizes = []
+        for size in self._api_request("/sizes/list", {"location": location}):
+            sizes.extend(self._to_size(size))
+        return sizes
+
+    def list_nodes(self, network=None):
```
Kami: ``network`` argument is not part of the base API so please prefix it with ``ex_``.

libcloud/compute/drivers/gig_g8.py
```diff
@@ -0,0 +1,501 @@
...
+    def list_volumes(self):
+        volumes = []
+        for disk in self._api_request("/disks/list",
+                                      {"accountId": self._account_id}):
+            if disk["status"] not in ["ASSIGNED", "CREATED"]:
+                continue
+            volumes.append(self._to_volume(disk))
+        return volumes
+
+    def create_volume(self, size, name, description, disk_type="D"):
```
Kami: ``description`` and ``disky_type`` arguments are not part of the base API so please prefix them with ``ex_``.

libcloud/compute/drivers/gig_g8.py
```diff
@@ -0,0 +1,501 @@
...
+                      'RESIZING': NodeState.RECONFIGURING,
+                      'CREATING_TEMPLATE': NodeState.PENDING,
+                      }
+
+    name = "GiG G8 Node Provider"
+    website = 'https://gig.tech'
+    type = Provider.GIG_G8
+    connectionCls = G8Connection
+
+    def __init__(self, apiurl, token, account_id):
```
Kami: For consistency with the base API, constructor argument should look something like this:

```python
def __init__(self, user_id, key, api_url, ....
```

And in this case, user_id would be account id.

Kami:
If you get a chance it would also be great to add type annotations for all the new methods and classes.

Here is an example of how that's handled on the base driver class - https://github.com/apache/libcloud/blob/trunk/libcloud/compute/base.py#L190.

libcloud/compute/drivers/gig_g8.py
```diff
@@ -269,12 +273,14 @@ def ex_create_network(self, name, private_network="192.168.103.0/24",
                                     {"cloudspaceId": networkid})
         return self._to_network(network)
 
-    def ex_destroy_network(self, network):
+    def ex_destroy_network(self, ex_network):
```
Kami: Sorry, I should have made this more clear in my original comment - ``ex_`` argument name prefix is only needed if the method name itself doesn't contain it.

If the method name already contains the prefix, then argument itself doesn't need to contain it.

I can address this when merging the PR.

Kami: EDIT: I see you just pushed some more changes so perhaps you can address that and let me know when it's ready and I'll go ahead and merge it.

Kami:
Thanks for addressing the feedback.

I will go ahead and merge those changes into trunk.

Issue #1438, opened 2020-03-06
GoogleStorageDriver does not handle service account HMAC credentials


pmezard:
Hello,

GoogleStorageDriver may use S3-compatibility layer credentials to connect to GCS. There are at least two ways to generate such credentials:

- As User Account HMAC keys. This is inconvenient for services configuration because it ties access to the existence of a user account.
- As Service Account HMAC keys. These should be used for long running services.

Unfortunately, GoogleStorageDriver does not identify the second type correctly. The problematic code is here:

  https://github.com/apache/libcloud/blob/trunk/libcloud/common/google.py#L615

It checks the identifier starts with "GOOG" which is true in my case, but my identifier is 61 characters long, way above the 30 characters check.

Callers may force the driver to identify the credentials as GCS_S3, but it requires some digging in the class hierarchy. You need to:

- Override `_ex_connection_class_kwargs` to return `{"auth_type": "GCS_S3"}`, to pass it to connectionCls. But this is not enough.
- Pass `auth_type="GCS_S3"` to GoogleStorageDriver so that jsonConnectionCls is correctly instantiate.

I find the need for both fix a little weird, I would have expected the first one to be enough, but I do not know libcloud internals to have a relevant opinion.

I would just drop the max length check.

Thanks.

pquentin:
The key length was [set to 20 in 2015](https://github.com/apache/libcloud/commit/3849f65fe5289f7d7160b7537afa930c7e2ea8a8) and [bumped to 20-30 in 2018](https://github.com/apache/libcloud/commit/9b9aae720d34698b6284f8b6800fd8ae569ecb82). I agree that dropping the length check altogether would be nice.

cc @crunk1 @erjohnso @MonsieurV @vdloo

MonsieurV:
Yep, I also agree on dropping the length.

It made sense when we had a unique length for the keys - to ensure we were getting the right key format. Now that Google keeps updating the key length, it won't be maintainable as it. Checking the key prefix should be enough.

Does you need a PR contribution for that?

Kami:
@MonsieurV Yeah, a PR with that change (+ tests) would be great.

Kami:
@MonsieurV and if you get a chance to work it in the next couple of days, I'll also try to include in the upcoming v3.0.0 release.

MonsieurV:
@Kami should be good :)

pquentin:
This issue can now be closed. Thanks!

Kami:
[...]

PR #1440, opened 2020-03-10
Only check Google S3 key prefix

MonsieurV:
### Description

Only check Google S3 key prefix (GOOG), as key length is not a reliable hint anymore.

See #1438

[The documentation](https://cloud.google.com/storage/docs/migrating#keys) also now only states:
> Your Google HMAC access ID starts with GOOG.

with no mention of key length anymore.

### Status

Done, ready for review

### DIFF
--- a/libcloud/common/google.py
+++ b/libcloud/common/google.py
@@ -617,10 +617,7 @@ class GoogleAuthType(object):
         """
         Checks S3 key format: alphanumeric chars starting with GOOG.
         """
-        return (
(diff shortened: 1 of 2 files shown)


pquentin:
Thanks, I tested your code and I can confirm that it fixes the issue @pmezard and I were having at work. 👍

pquentin (approved):

Kami:
Merged.

Thanks for contribution and thanks to @pquentin and @pmezard for verifying it.

PR #1443, opened 2020-03-18
Improve location / zone handling in the GCE driver

### LINKED ISSUE #1441: GCE create_node() with location=None (default) throws AttributeError
## Summary

Running with the GCE driver using `create_node(.. , location=None)` (which is a default) throws an `AttributeError`

## Detailed Information

OS: Ubuntu 19.10
libcloud Version: `apache-libcloud==2.8.1`
Python version: `Python 3.7.5`

Code to reproduce:

```python
gce = get_driver(Provider.GCE)(
            f'{user}@{project_id}.iam.gserviceaccount.com',
            api_key_file,
            project=project_id)

gce.create_node(name, size, None,
                             ex_network=network,
                             ex_subnetwork=subnet,
                             ex_tags=tags,
                             ex_metadata=metadata, 
                             external_ip='ephemeral',
                             internal_ip=internal_ip,
                             ex_image_family=image,
                             ex_disk_size=disk_sz)
```

With exception:

```
[ .. snip .. ]
  File "/home/kevin/.local/lib/python3.7/site-packages/proj/gcp/builder.py", line 181, in create_vm
    ex_disk_size=disk_sz)
  File "/home/kevin/.local/lib/python3.7/site-packages/libcloud/compute/drivers/gce.py", line 4169, in create_node
    location = self.ex_get_zone(location)
  File "/home/kevin/.local/lib/python3.7/site-packages/libcloud/compute/drivers/gce.py", line 7811, in ex_get_zone
    if name.startswith('https://'):
AttributeError: 'NoneType' object has no attribute 'startswith'
```

Kami:
## Description

This pull request improves zone / location handling in the GCE driver.

It includes the following changes:

1. ``create_node()`` method now also works if datacenter / location argument is not specified. If zone is not specified, it's inferred from the size object and if that's not possible, we throw a more user-friendly error.

2. Update GCE docs and examples and make it clear that a lot of operations depend on zone being set.

Previously, if user didn't specify a zone to work with (either via ``datacenter`` driver constructor argument or via ``zone`` / ``location`` argument passed to a specific method), a cryptic and non user-friendly error such as "AttributeError: 'NoneType' object has no attribute 'startswith'" was thrown.

Thanks to @kbknapp for reporting this issue.

Resolves #1441.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -32,11 +32,13 @@ Compute
   ``location`` argument to the method).
 
   Reported by Kevin K. - @kbknapp.
+  (GITHUB-1443)
   [Tomaz Muraus]
 
 - [GCE] Update ``ex_get_disktype`` method so it works if ``zone`` argument is
   not set.
   [Tomaz Muraus]
+  (GITHUB-1443)
 
 Storage
 ~~~~~~~
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -23,6 +23,21 @@ Common
   (GITHUB-1439)
   [Miguel Caballer - @micafer]
 
+Compute
+~~~~~~~
+
+- [GCE] Update ``create_node()`` method so it throws an exception if node
+  location can't be inferred and location is not specified by the user (
+  either by passing ``datacenter`` constructor argument or by passing
+  ``location`` argument to the method).
+
+  Reported by Kevin K. - @kbknapp.
+  [Tomaz Muraus]
+
+- [GCE] Update ``ex_get_disktype`` method so it works if ``zone`` argument is
+  not set.
+  [Tomaz Muraus]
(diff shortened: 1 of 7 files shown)


Kami:
@erjohnso WDYT?

I think that ideally we would just make ``datacenter`` constructor argument mandatory, but I assume that may break a bunch of the existing code.

The problem is that a lot of operations depend on location being set - either via constructor argument or explicitly for a method invocation.

One option is to call new ``_verify_zone_is_set`` method in all the places where that is mandatory, but it's likely we will miss some so it's not 100% robust.

An alternative would to make ``datacenter`` constructor argument mandatory, but also default it to some value (e.g. ``us-east-b``). That's not ideal either though.

PR #1444, opened 2020-03-26
Fix auto assigning of volume device in OpenStack

jhesketh:
### Description

When trying to attach a volume to a node with the OpenStack driver, using the 'auto' device name fails with:

libcloud.common.exceptions.BaseHTTPError: 400 Bad Request Invalid input for field/attribute device. Value: auto. u'auto' does not match '(^/dev/x{0,1}[a-z]{0,1}d{0,1})([a-z]+)[0-9]*$'

OpenStack expects a device, or a null/None value[0]

https://docs.openstack.org/api-ref/compute/?expanded=#attach-a-volume-to-an-instance

### Status

- done, ready for review

- [ ] Documentation


Kami:
Thanks for the contribution.

I will look into adding a test case and merging it into trunk.

Kami:
Added tests (1b2c399b06193145e0062667867e0b39ee754e9c) and merged it into trunk.

Thanks again for the contribution.

PR #1358, opened 2019-10-11
add more new gce image families

XN137:
### Description

similar to #1304 we add new image families for `centos-8` and `debian-10` from https://cloud.google.com/compute/docs/images#unshielded-images to prevent similar errors as in #1303

### Status

- done, ready for review

### Checklist (tick everything that applies)

- [x] [Code linting](http://libcloud.readthedocs.org/en/latest/development.html#code-style-guide) (required, can be done after the PR checks)
- [ ] Documentation
- [ ] [Tests](http://libcloud.readthedocs.org/en/latest/testing.html)
- [ ] [ICLA](http://libcloud.readthedocs.org/en/latest/development.html#contributing-bigger-changes) (required for bigger changes)

PR #1449, opened 2020-04-25
add latest ubuntu image families to gce

XN137:
### Description

similar to https://github.com/apache/libcloud/pull/1358 we update the image families with the latest data from https://cloud.google.com/compute/docs/images

still unsure whether ubuntu versions that are now EOL and/or that are no longer listed on that page should be removed - so we just kept them around.

### DIFF
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -1800,6 +1800,7 @@ class GCENodeDriver(NodeDriver):
         "userinfo-email": "userinfo.email"
     }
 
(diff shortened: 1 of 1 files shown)


Kami (approved):
LGTM :+1: 

Thanks for the contribution.

Kami:
Looks like that Python 3.5 tests are failing because of some race, I will try to replicate it locally.

Kami:
Pushed a fix for test failure in abbce640d974049f6c2a2c701e80c2cf8c680f59.

XN137:
thank you for your work!

Issue #1403, opened 2020-01-08
AWS S3 Presigned URL and Azure Blob Storage Shared access signature


rvolykh:
## Feature Request

AWS S3 has the support of [Presigned Object URL](https://docs.aws.amazon.com/AmazonS3/latest/dev/ShareObjectPreSignedURL.html) and Azure Blob Storage has the support of [Shared access signature](https://docs.microsoft.com/en-us/rest/api/storageservices/delegate-access-with-shared-access-signature).

Both of those features allow anyone (with whom shared link was shared) to download or upload a file. This feature is very useful when you want to temporary share a file with some user(s) or to allow them to upload a file, without sharing your credentials (temporary creds, creating an account, etc).

Currently, [libcloud Storage interface (API)](https://libcloud.readthedocs.io/en/stable/storage/api.html#storage-base-api) has two methods which more or less are related to sharing object access:
- [enable_object_cdn](https://libcloud.readthedocs.io/en/stable/storage/api.html#libcloud.storage.base.StorageDriver.enable_object_cdn) - Enable object CDN;
- [get_object_cdn_url](https://libcloud.readthedocs.io/en/stable/storage/api.html#libcloud.storage.base.StorageDriver.get_object_cdn_url) - Return an object CDN URL.

\* Both methods are not implemented either for AWS S3 nor Azure Blob Storage libcloud storage drivers.

My proposal is to extend `get_object_cdn_url` with additional parameter (mode=["read", "write"]) and implement support for AWS S3 and Azure Blob Storage.

c-w:
Great suggestion. I'll take a stab at implementing the Azure Blob Storage part of this.

Kami:
Resolved via #1408.

PR #1453, opened 2020-05-01
Implement s3 get_object_cdn_url using pre-signed urls

avirshup:
### Description

Allows the s3 driver to generate [pre-signed URLs](https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-query-string-auth.html) using the `get_object_cdn_url` method. Anyone in possession of these URLs can _read_ the object until the URL expires. (As suggested in issue #1403)

Notes:
 - I've tried to follow the example of PR #1408 - which implements that Azure version of this function - as much as possible.
 - to prevent code duplication, this calls some internal functions from the `AWSRequestSignerAlgorithmV4` class.
 - many providers use the S3 driver. Because this is tied specifically to AWS's authentication methods, I had to modify a few tests to make sure that this doesn't change behavior elsewhere. 

(Long-time user, first-time PR-opener - happy to make changes as needed)

### Status
Ready for review

- [ ] Documentation

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -422,6 +422,18 @@ Storage
   (GITHUB-1410)
   [Clemens Wolff - @c-w]
 
+- [AWS S3] Implement ``get_object_cdn_url`` for the AWS storage driver.
+
+  The AWS storage driver can now be used to generate temporary URLs that
+  grant clients read access to objects. The URLs expire after a certain
+  period of time, either configured via the ``ex_expiry`` argument or the
+  ``LIBCLOUD_S3_STORAGE_CDN_URL_EXPIRY_HOURS`` environment variable
+  (default: 24 hours).
+
+  Reported by @rvolykh.
+  (GITHUB-1403)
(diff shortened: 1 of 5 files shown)


Kami:
Thanks for the contribution.

The change looks good to me :+1:

Kami:
> many providers use the S3 driver. Because this is tied specifically to AWS's authentication methods, I had to modify a few tests to make sure that this doesn't change behavior elsewhere.

Yeah, IIRC, the class hierarchy is not the best, so some changes like that are not totally straight forward (some of the existing functionality should be implemented on the AWS S3 driver and not on the base S3 one from which other S3 compatible provider drivers inherit from).

Kami:
I've tested the changes using AWS S3 and things look fine :+1: 

The only thing I'm wondering is if we should expose ``ex_expiry`` as seconds.

Currently you can still use a float (e.g. 0.1 hours), but that's not ideal.

Kami:
I went ahead and merged your changes into trunk.

PR #1456, opened 2020-05-02
Fixes #1455, added location filtering for digitalOcean driver

### LINKED ISSUE #1455: method list_sizes() on driver object does not accept location
## Summary

method list_sizes() on driver object does not accept location parameter. The docs states that it should accept a location parameter: https://libcloud.readthedocs.io/en/stable/compute/api.html#libcloud.compute.base.NodeDriver.list_sizes

## Detailed Information

Tried to get all the sizes through the digitalOcean driver, filtering on a certain location. I pass a NodeLocation object (retrieved from driver with function list_locations()) into list_sizes(). After that i get the error:

list_sizes() takes 1 positional argument but 2 were given

Expected to get a list of sizes available for the given location as stated in the documentation. 

## Fix

The api of digitalOcean returns a region key for every returned size, this key holds a list with all regions this size is available. Accept location parameter in list_sizes() and check if the NodeLocation.id is in regions.

RobertH1993:
### Description

Added location filtering for list_sizes() in the digitalOcean driver. Now its compliant with the documentation. See issue #1455 for more info.

### Status

- done, ready for review

### DIFF
--- a/libcloud/compute/drivers/digitalocean.py
+++ b/libcloud/compute/drivers/digitalocean.py
@@ -133,9 +133,12 @@ class DigitalOcean_v2_NodeDriver(DigitalOcean_v2_BaseDriver,
         data = self._paginated_request('/v2/droplets', 'droplets')
         return list(map(self._to_node, data))
 
-    def list_sizes(self):
+    def list_sizes(self, location=None):
         data = self._paginated_request('/v2/sizes', 'sizes')
-        return list(map(self._to_size, data))
+        sizes = list(map(self._to_size, data))
+        if location:
+            sizes = [x for x in sizes if location.id in x.extra['regions']]
+        return sizes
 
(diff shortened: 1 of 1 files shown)


Kami:
Thanks for the contribution.

Yes if driver complies with the base API and implements all the functionality, then it should also implement ``location`` argument for the ``list_sizes()`` method.

If possible, it's preferred to perform filtering on the API size, but if that's not supported by the provider API, doing client side filtering (as in your PR) is fine as well.

Kami:
I've added a test case (47d666f92f08eb3830a3b92a6f1c6e0732a1c5d4, 0aed5581837254ce96e14c1964c93002e17503da) and merged it into trunk.

Thanks again for the contribution and next time please ensure your changes also include corresponding test cases.

PR #1462, opened 2020-05-25
G8driver refreshtoken

grimpy:
### Description

Update g8 driver fixed some small issues when creating portforwards
And adds functionality to update the apitoken

### Status
- done, ready for review

### DIFF
--- a/libcloud/common/gig_g8.py
+++ b/libcloud/common/gig_g8.py
@@ -74,7 +74,7 @@ def is_jwt_expired(jwt):
         claimsdata = claimsdata.decode('utf-8')
     data = json.loads(claimsdata)
     # check if it's about to expire in the next minute
-    return data['exp'] < time.time() - 60
+    return data['exp'] < time.time() + 60
 
 
 def maybe_update_jwt(jwt):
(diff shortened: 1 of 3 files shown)


libcloud/common/gig_g8.py
```diff
@@ -26,6 +35,74 @@ def add_default_headers(self, headers):
...
+    :rtype: str
+    """
+    rem = len(input) % 4
+
+    if rem > 0:
+        input += b'=' * (4 - rem)
+    return base64.urlsafe_b64decode(input)
+
+
+def is_jwt_expired(jwt):
```
Kami, line 60: Can you please add some unit tests for this function (covering both expired and non expired scenario, various forms of input string, etc.).

grimpy: I added a test for expire and none expired scenarios, not sure what to do about 'forms of input string' the method would pretty much crash with split or encoding errors if you pass invalid formats

Kami: Thanks.

This should be fine for now.

With different forms of input strings, I meant thing such as JWT string payload missing some fields, being corrupted, etc. and making sure we handle that correctly.

Kami:
Merged, thanks.

Kami:
This required a bit more work to fix the tests failing in trunk - 4004375ccb9d5e899e70b85b33bb7a018c312eee

PR #1345, opened 2019-08-22
New compute driver OUTSCALE SDK

linaup-m:
### Description

This PR add a new compute driver name OUTSCALE SDK, it is added in file types.py of compute section and there is a new file named outscale_sdk.py. This permits to perform actions of the OUTSCALE API with outscale python sdk (osc sdk). See the documentation link: https://docs-beta.outscale.com/#outscale-api

### Status

- done, ready for review

### Checklist (tick everything that applies)

- [x] [Code linting](http://libcloud.readthedocs.org/en/latest/development.html#code-style-guide) (required, can be done after the PR checks)
- [ ] Documentation (do the documentation refer to the docstring ?)
- [x] [Tests](http://libcloud.readthedocs.org/en/latest/testing.html)
- [x] [ICLA](http://libcloud.readthedocs.org/en/latest/development.html#contributing-bigger-changes) (required for bigger changes)

I am not sure if it's ok to use my library or if I add to use the libcloud library. 
Thanks for responses.

PR #1476, opened 2020-07-21
New compute driver OUTSCALE

Aikoh0:
## Description

This PR add a new compute driver named OUTSCALE, it was added in file types.py of compute section, in a new osc.py file for authentification in common section and in a new file named outscale.py to perform actions of the OUTSCALE API using requests. See the documentation link: https://docs.outscale.com/api

This PR is a reworked version of https://github.com/apache/libcloud/pull/1345 using python requests lib instead of using outscale python sdk.

### Status
- done, ready for review

- [x] Code linting (required, can be done after the PR checks)

- [x] Documentation (do the documentation refer to the docstring ?)

- [x] Tests

- [x] ICLA (required for bigger changes)

Happy to receive thoughs.


Kami:
@tgn-outscale Thanks for the contribution and sorry for the delay. I will have a more detailed look a bit later.

For now, I'm just trying to understand how this fits along with the existing Outscale driver which is based on the EC2 one? Is the EC2 based one still relevant, or should we remove it?

Aikoh0:
Hi, @Kami, thanks for your reply

The Outscale EC2 driver is using our old API, and this implementation is using the new one.
As the old version is still in use and the new version still lacking some features, I think the EC2 version is still relevant for now.

Kami:
@tgn-outscale Thanks for the clarification. It would be a good idea to also point this out in the driver documentation page inside docs/ directory.

libcloud/common/osc.py
```diff
@@ -0,0 +1,151 @@
...
+    'OSCRequestSignerAlgorithmV4',
+]
+
+
+class OSCRequestSigner(object):
+    """
+    Class which handles signing the outgoing AWS requests.
+    """
+
+    def __init__(self, access_key, access_secret, version, connection):
```
Kami: It you get a chance, it would also be good to add type annotations for MyPy.

libcloud/compute/drivers/outscale.py
```diff
@@ -0,0 +1,1109 @@
...
+    def list_locations(self, dry_run=False):
+        """
+        Lists available regions details.
+
+        :return: regions details
+        :rtype: ``dict``
+        """
+        action = "ReadRegions"
+        data = json.dumps({"DryRun": dry_run})
+        signer = OSCRequestSignerAlgorithmV4(access_key=self.key,
```
Kami: Looks like the next couple of lines are duplicated across all the driver methods.

To avoid duplication it would be good to refactor that in some common method.

In addition to that, it looks like ``OSCRequestSignerAlgorithmV4`` could also be instantiated inside the driver constructor since it doesn't seem to depend on any method specific arguments.

libcloud/compute/drivers/outscale.py
```diff
@@ -0,0 +1,1109 @@
...
+        headers = signer.get_request_headers(action=action,
+                                             data=data,
+                                             service_name=self.service_name,
+                                             region=self.region)
+        endpoint = self._get_outscale_endpoint(self.region,
+                                               self.version,
+                                               action)
+        return requests.post(endpoint, data=data, headers=headers)
+
+    def create_public_ip(self, dry_run=False):
```
Kami: Methods which are not part of a standard Libcloud API need to be prefixed with ``ex_`` - so ``ex_create_public_ip``, ``ex_delete_public_ip``, etc.

libcloud/compute/drivers/outscale.py
```diff
@@ -0,0 +1,1109 @@
...
+                                             region=self.region)
+        endpoint = self._get_outscale_endpoint(self.region,
+                                               self.version,
+                                               action)
+        return requests.post(endpoint, data=data, headers=headers)
+
+    def create_node(self,
+                    image_id,
+                    dry_run=False,
+                    block_device_mapping=None,
```
Kami: Argument which are not part of the Libcloud standard API (aka are not defined on the base NodeDriver class) need to be prefixed with ``ex_`` - ex_bsu_optimized, ex_nics, etc.

libcloud/compute/drivers/outscale.py
```diff
@@ -0,0 +1,1109 @@
...
+        headers = signer.get_request_headers(action=action,
+                                             data=data,
+                                             service_name=self.service_name,
+                                             region=self.region)
+        endpoint = self._get_outscale_endpoint(self.region,
+                                               self.version,
+                                               action)
+        return requests.post(endpoint, data=data, headers=headers)
+
+    def reboot_node(self, node_ids):
```
Kami: To comply with the libcloud standard API, this method needs to take a single ``node`` argument which must be ``Node`` class instance.

If you also want to support rebooting multiple notes, you should add a new extension method. For example: ``ex_reboot_nodes(self, node_ids)``.

libcloud/compute/drivers/outscale.py
```diff
@@ -0,0 +1,1109 @@
...
+        headers = signer.get_request_headers(action=action,
+                                             data=data,
+                                             service_name=self.service_name,
+                                             region=self.region)
+        endpoint = self._get_outscale_endpoint(self.region,
+                                               self.version,
+                                               action)
+        return requests.post(endpoint, data=data, headers=headers)
+
+    def list_nodes(self, data="{}"):
```
Kami: To comply with the base Libcloud API, this method needs to return a list of ``Node`` objects.

libcloud/compute/drivers/outscale.py
```diff
@@ -0,0 +1,1109 @@
...
+        headers = signer.get_request_headers(action=action,
+                                             data=data,
+                                             service_name=self.service_name,
+                                             region=self.region)
+        endpoint = self._get_outscale_endpoint(self.region,
+                                               self.version,
+                                               action)
+        return requests.post(endpoint, data=data, headers=headers)
+
+    def delete_node(self, node_ids):
```
Kami: Same here - should take a single ``node`` argument which represents ``Node`` class instance.

libcloud/compute/drivers/outscale.py
```diff
@@ -0,0 +1,1109 @@
...
+        headers = signer.get_request_headers(action=action,
+                                             data=data,
+                                             service_name=self.service_name,
+                                             region=self.region)
+        endpoint = self._get_outscale_endpoint(self.region,
+                                               self.version,
+                                               action)
+        return requests.post(endpoint, data=data, headers=headers)
+
+    def list_images(self, data="{}"):
```
Kami: Needs to return a list of ``NodeImage`` objects. data is a non-standard argument so it needs to be prefixed with ``ex_``.

libcloud/compute/drivers/outscale.py
```diff
@@ -0,0 +1,1109 @@
...
+        headers = signer.get_request_headers(action=action,
+                                             data=data,
+                                             service_name=self.service_name,
+                                             region=self.region)
+        endpoint = self._get_outscale_endpoint(self.region,
+                                               self.version,
+                                               action)
+        return requests.post(endpoint, data=data, headers=headers)
+
+    def delete_image(self, image_id):
```
Kami: Needs to take ``NodeImage`` object.

Kami:
I've added some in-line comments.

Overall, the changes mostly look good, but the code needs a bit more work to reduce duplication and to make sure the methods and method arguments comply with the base Libcloud API - that's one of the main ideas behind Libcloud - all the provider drivers should expose the same API (where possible), so users can easily switch between different providers and non-standard methods and arguments should be prefixed with ``ex_``.

Aikoh0:
Thanks for your review @Kami, I will make these changes and come back to you as soon as possible.

Kami:
You are welcome.

If you need any additional clarification on the extension method and argument naming, please let me know.

Aikoh0:
Hello @Kami,
I have made the changes according to what you said for more compliance with the base driver.
Let me know if I missed or misunderstood anything.

libcloud/compute/drivers/outscale.py
```diff
@@ -0,0 +1,1018 @@
...
+    """
+    Outscale SDK node driver
+    """
+
+    type = Provider.OUTSCALE
+    name = 'Outscale API'
+    website = 'http://www.outscale.com'
+
+    def __init__(self,
+                 key: str = None,
```
Kami, line 47: Thanks for adding type annotations :+1:

libcloud/compute/drivers/outscale.py
```diff
@@ -0,0 +1,1018 @@
...
+        if self._call_api(action, data).status_code == 200:
+            return True
+        return False
+
+    def ex_delete_public_ip(self,
+                            dry_run: bool = False,
+                            public_ip: str = None,
+                            public_ip_id: str = None):
+        """
+        Delete instances.
```
Kami, line 111: Some comments are out of sync with code, but I will update / fix that when merging the pr :)

libcloud/compute/drivers/outscale.py
```diff
@@ -0,0 +1,1018 @@
...
+                    ex_security_groups: [str] = None,
+                    ex_subnet_id: str = None,
+                    ex_user_data: str = None,
+                    ex_vm_initiated_shutdown_behavior: str = None,
+                    ex_vm_type: str = None
+                    ):
+        """
+        Create a new instance.
+
+        :param      ex_image_id: The ID of the OMI used to create the VM.
```
Kami, line 286: That's fine for now, but ideally this method would comply with the standard API in the future - this means take name, image and size arguments.

This would make it much easier for users to switch from other providers to this one without needing to change much code.

Kami:
@tgn-outscale Thanks for addressing the review feedback.

I merged the PR into trunk, but there are still some places which would ideally be improved in the near future:

- Try to make more methods comply with the Libcloud standard API (where possible) - this would make it easier for users to switch to this driver (right now it utilizes a lot of non-standard argument names which means Libcloud code which works with other provider drivers won't work with this one). This includes both - the argument the methods take and method return values - e.g. ``create_node()`` should take name, size and image arguments and return ``Node`` instance class (and not a dictionary). Same goes for some other methods. 
- Need to add some documentation with examples to ``docs/compute/drivers/``

Aikoh0:
Thanks for your help @Kami, I will try to implement those changes when I will find time to do so,
Best regards

PR #1477, opened 2020-07-21
Add expose ports through services for VMs, extra ports and services d…

Eis-D-Z:
### Description
Added methods to create services through which ports can be exposed for VMs, each VM will have one service for each type through which multiple ports can be exposed. 

### Status
- done, ready for review

- [ ] Documentation

### DIFF
--- a/libcloud/compute/drivers/kubevirt.py
+++ b/libcloud/compute/drivers/kubevirt.py
@@ -15,21 +15,18 @@
 
 # NOTE: Re-enable once we add mypy annotations for the base container API
 # type: ignore
-
 """
 kubevirt driver with support for nodes (vms)
 """
-
 import json
 import time
+import hashlib
 
 from datetime import datetime
 
 from libcloud.common.types import LibcloudError
-from libcloud.common.kubernetes import KubernetesBasicAuthConnection
 from libcloud.common.kubernetes import KubernetesDriverMixin
 from libcloud.common.kubernetes import VALID_RESPONSE_CODES
-
 from libcloud.compute.types import Provider, NodeState
 from libcloud.compute.base import NodeDriver, NodeSize, Node
 from libcloud.compute.base import NodeImage, NodeLocation, StorageVolume
@@ -37,7 +34,6 @@ from libcloud.compute.base import NodeImage, NodeLocation, StorageVolume
 __all__ = [
     "KubeVirtNodeDriver"
 ]
-
 ROOT_URL = '/api/v1/'
 KUBEVIRT_URL = '/apis/kubevirt.io/v1alpha3/'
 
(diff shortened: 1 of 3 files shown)


libcloud/compute/drivers/kubevirt.py
```diff
@@ -167,10 +165,17 @@ def destroy_node(self, node):
         """
         namespace = node.extra['namespace']
         name = node.name
-        # stop the vmi first
+        # find and delete services for this VM only
+        services = self.ex_list_services(namespace=namespace, node_name=name)
+        for service in services:
+            service_type = service['spec']['type']
+            self.ex_create_service(node=node, ports=[],
```
Kami: I think it would be better to have another ``ex_delete_service`` method which can call ``ex_create_service`` underneath with those arguments.

Also, from the method docstring it looks like you may need to specify ``override_existing_ports`` argument. Or is it not needed?

libcloud/compute/drivers/kubevirt.py
```diff
@@ -181,7 +186,8 @@ def destroy_node(self, node):
     # only has container disk support atm with no persistency
     def create_node(self, name, image, location=None, ex_memory=128, ex_cpu=1,
                     ex_disks=None, ex_network=None,
-                    ex_termination_grace_period=0):
+                    ex_termination_grace_period=0,
+                    ports={}):
```
Kami: Please don't use mutable defaults - https://florimond.dev/blog/articles/2018/08/python-mutable-defaults-are-the-source-of-all-evil/#:~:text=In%20Python%2C%20when%20passing%20a,or%20even%20a%20class%20instance.

Doing ``ports=None`` and then inside method ``ports = ports or {}`` is better and safer.

libcloud/compute/drivers/kubevirt.py
```diff
@@ -1022,3 +1096,143 @@ def _to_node(self, vm, is_stopped=False):
...
+            if not port_group.get('name', ""):
+                port_group['name'] = 'port-{}'.format(port_group['port'])
+            ports_to_expose.append(
+                {'protocol': port_group.get('protocol', 'TCP'),
+                 'port': int(port_group['port']),
+                 'targetPort': int(port_group['target_port']),
+                 'name': port_group['name']})
+        headers = None
+        data = None
+        if len(service_list) > 0:
```
Kami, line 1188: Per my comment above, I think it would be better to be explicit and have another ``ex_delete_service`` method.

Kami:
Thanks for the contribution.

I added some comments in-line.

libcloud/compute/drivers/kubevirt.py
Eis-D-Z: Hey, 
An empty ports will result in the service being deleted. The `override_existing_ports` argument was added because there is no general standard in case there already are some ports forwarded and a user requests additional ports to be opened. Sometimes he might want to keep the previous, sometimes not. I thought it would be better to spare the user from manually deleting  in the second case (also I was not brilliant enough to have an `ex_delete_service` which makes a lot of sense).

Eis-D-Z:
Thanks for pointing out those issues. They have been fixed!

Kami:
Thanks for addressing the feedback - I merged changes into trunk.

PR #1478, opened 2020-07-23
Update lxd driver with networks, storage pools and further refine exi…

Eis-D-Z:
### Description

This enhances the lxd driver with network and storage pool capability while further refining existing methods, for example changing an image id to the unique fingerprint instead of the image name, etc...

### Status
- done, ready for review

- [ ] Documentation


libcloud/container/drivers/lxd.py
```diff
@@ -476,22 +518,47 @@ def deploy_container(self, name, image, cluster=None,
         to use as basis for limits e.g. "c2.micro"
         :type  ex_instance_type: ``str``
 
+        :param ex_timeout: Timeoute
```
Kami: Typo

libcloud/container/drivers/lxd.py
```diff
@@ -476,22 +518,47 @@ def deploy_container(self, name, image, cluster=None,
...
+        :type  ex_timeout: ``int``
+
         :rtype: :class:`libcloud.container.base.Container`
         """
+
+        if parameters:
+            parameters = json.loads(parameters)
+
+            if "mode" in parameters["source"] and \
```
Kami: You could simplify it and just do ``if parameters["source"].get("mode", None) == "pull": ...``

libcloud/container/drivers/lxd.py
```diff
@@ -662,10 +730,17 @@ def destroy_container(self, container, ex_timeout=default_time_out):
...
-        response_dict = response.parse_body()
-        assert_response(response_dict=response_dict, status_code=100)
+        try:
+            response = self.connection.request(req, method='DELETE')
+
+            response_dict = response.parse_body()
+            assert_response(response_dict=response_dict, status_code=100)
+        except BaseHTTPError as e:
+            # handle the case where the container is running
+            message_list = e.message.split(",")
```
Kami: I see this logic duplicated across multiple methods so it would perhaps be better to refactor it into some common method - e.g. ``_get_error_message_for_exception`` or similar.

And some tests for this method would also be good.

Kami: EDIT: Or perhaps even, ``_get_lxd_ap_exception_for_error`` and returning ``LXDAPIException`` class instance.

libcloud/container/drivers/lxd.py
```diff
@@ -971,6 +1073,25 @@ def list_images(self):
...
 
+    def ex_has_image(self, alias):
+        """
+        Helper function. Returns true is the image with the
+        given alias exits on the host
+
+        :param alias: the image alias
+        :type  alias: ``str``
+
+        :rtype: boolean
```
Kami: Looks like method actually returns a tuple, please update the comment accordingly.

libcloud/container/drivers/lxd.py
```diff
@@ -971,6 +1073,25 @@ def list_images(self):
...
+        """
+
+        # get all the images existing on the host
+        images = self.list_images()
+
+        for img in images:
```
Kami: I assume there is no more efficient way of doing that (aka filter on the server side) and we need to do late filtering locally?

Kami:
Thanks for the contribution - I've added some comments.

libcloud/container/drivers/lxd.py
Eis-D-Z: There is after all, this was overlooked, thank you for bringing it to my attention.

Eis-D-Z:
The issues you point out in your comments have been fixed, I have also done a -f push and a lot of commits appeared. I did this to keep the correct history since most of the code here is not mine. (Earlier I hadn't figured a way to do this)

libcloud/container/drivers/lxd.py
```diff
@@ -971,6 +1066,32 @@ def list_images(self):
             images.append(self.ex_get_image(fingerprint=fingerprint))
         return images
 
+    def ex_has_image(self, alias):
+        """
+        Helper function. Returns true and the image fingerprint
+        if the image with the given alias exits on the host.
```
Kami: exits -> exists

libcloud/container/drivers/lxd.py
```diff
@@ -1155,21 +1278,66 @@ def ex_get_storage_pool_volume(self, pool_id, type, name):
...
+
+            pool_id = pool.split('/')[-1]
+
+            volumes = self.ex_list_storage_pool_volumes(pool_id=pool_id)
+
+            for vol in volumes:
+                if vol.name == name:
+                    return vol
+
+        return self._to_storage_volume(pool_id=None,
```
Kami: Hm, shouldn't we simply return None if a matching volume doesn't exist?

libcloud/container/drivers/lxd.py
```diff
@@ -1225,12 +1444,102 @@ def ex_delete_storage_pool_volume(self, pool_id, type, name):
...
+
+            req = "/%s/storage-pools/%s/volumes/%s/%s" % (self.version,
+                                                          pool_id,
+                                                          type,
+                                                          name)
+            response = self.connection.request(req, method="DELETE")
+
+            response_dict = response.parse_body()
+            assert_response(response_dict=response_dict, status_code=200)
+        except BaseHTTPError as err:
```
Kami, line 1453: We could probably reduce some of the duplication in those driver methods by moving some of that error handling logic to the response class.

libcloud/container/drivers/lxd.py
```diff
@@ -1225,12 +1444,102 @@ def ex_delete_storage_pool_volume(self, pool_id, type, name):
...
+        response = self.connection.request(req)
+
+        response_dict = response.parse_body()
+        assert_response(response_dict=response_dict, status_code=200)
+
+        nets = response_dict["metadata"]
+        networks = []
+        for net in nets:
+            name = net.split('/')[-1]
+            networks.append(self.ex_get_network(name=name))
```
Kami, line 1478: This seems a bit wasteful - do we really need to perform another request for each network returned by /1.0/networks? Aka doesn't the list response contain enough information?

Kami:
Thanks for the contribution.

It mostly looks good. I just had some clarifying questions and some code could perhaps be optimized to avoid some duplication.

libcloud/container/drivers/lxd.py
Eis-D-Z: The /1.0/networks returns just a list of URLs ( [networks endpoint LXD API](https://github.com/lxc/lxd/blob/master/doc/rest-api.md#10networks) ), the previous contributor made a whole LXDNetwork class which needs a bit more data. As it is now an additional request for each network is required. If you think that there's no need to have a whole network class let me know.

libcloud/container/drivers/lxd.py
Eis-D-Z: It is tricky, there are two kinds of operations, sync and async for lxd. The error handling with the timeout which is duplicated a few times is for the async methods. The response class won't have that info though so the code would not get simpler. If you have in mind some way please share.

libcloud/container/drivers/lxd.py
Eis-D-Z: True...

libcloud/container/drivers/lxd.py
Kami: I think classes are a good idea in general since they provide a better and more user-friendly API to the developers (in contrasts of dictionaries which are usually opaque and hard to work with since you need to inspect the docs / similar to see what attributes should be there...).

Kami:
Thanks for addressing the feedback - changes have been merged into trunk.

PR #1479, opened 2020-07-23
vSphere driver

Eis-D-Z:
### Description

With this driver one can manage a vSphere environment, the connection is made with a vCenter server, it might not work with just the ESXi host. This driver has a dependency called pyvmomi that can be installed with `pip install pyvmomi`, it is needed since the REST API of vSphere is not robust enough yet. The REST API driver will work, except list_images, snapshot methods and console method which need the SOAP driver. Images in this driver are considered  VM templates or ovf files. The REST APIdriver also lacks the ability to clone a VM due to insufficient endpoints of the API. The added tests check core functionality of the REST driver only.

### Status
- done, ready for review

- [x] Documentation


Eis-D-Z:
Unfortunately there is one last test that won't pass since the driver has async code. In production environments with VMs in the triple digits, the list_nodes method gets pretty slow, for each VM \_to\_node gets called in a serial manner where probably more API requests are made.., this can take some notable time. I would argue that every driver could benefit from an async list_nodes, in this case it is optional, there is an argument to enable or disable the async_ functionality. I wonder if something can be done in regard to the test that can't be fulfilled.

PR #1481, opened 2020-08-04
V sphere drv

Eis-D-Z:
### Description
This contains two drivers, one that works with VMware's SOAP API through a python client called pyVmomi, and one that works with VMware's newly introduced REST API. The more complete is the former since the REST API is lacking functionality as of yet, still all common driver functions beyond list_images will work. Images for the scope of this work are considered either ovf files or vm templates, some of these templates are not returned by the REST endpoint and thus the choice was made to use the SOAP driver as well, instead of having a lacking method. Also snapshot and console methods require the SOAP driver. If the pyVmomi dependency is missing, using these methods will result in an ImportError. The driver also contains async code, during the driver's design and testing, an environment with over 3k VM-s was used among others. This lead to the realization that the usual way of writing list_nodes gets slow very fast for large number of nodes, especially when there is a need of a _to_node helper method that does more requests for each node. There is a "flag" argument in the list_nodes method for whether to proceed asynchronously or not, but the existence of async code requires python 3.5+. The included tests are checking only the REST driver methods. Despite these drawbacks, the authors decided to publish this driver, since vSphere is popular and it works very well. 
 

### Status

- done, ready for review

- [ ] Documentation


Kami:
Thanks for the contribution.

For now, this in-line async code should probably work, but we should come up with a standardized convention for handling async across different drivers in a consistent manner.

Perhaps another set of methods which have ``async_`` (e.g. ``async_list_nodes``, ``async_reboot_node``, etc.). And we also need a way for user to specify the event loop to use.

This could either be done on the global level (e.g. ``libcloud.async.set_event_loop(loop)``) and perhaps also on driver level, if user wishes to use different loop for different drivers (probably that likely).

What do you think?

Eis-D-Z:
I personally guess that ideally at some point even the requests logic will become async for libcloud, until then yes, I agree that `list_nodes()` should have an `async_list_nodes` and the event loop better to be set at driver level, since the driver might not be on the main thread, or someone might want two drivers on separate threads etc...

libcloud/compute/drivers/vsphere.py
```diff
@@ -0,0 +1,1992 @@
...
+        }
+        return VSphereNetwork(id=data.name, name=data.name, extra=extra)
+
+    def list_sizes(self):
+        """
+        Returns sizes
+        """
+        return []
+
+    def list_images(self, location=None, folder_ids=[]):
```
Kami: Please don't default to a mutable value and do ``folder_ids=None`` and then inside the method ``folder_ids = folder_ids or []``.

libcloud/compute/drivers/vsphere.py
```diff
@@ -0,0 +1,1992 @@
...
+
+    def create_node(self, **kwargs):
+        """
+        Creates and returns node.
+
+        :keyword    ex_network: Name of a "Network" to connect the VM to ",
+        :type       ex_network: ``str``
+
+        """
+        name = kwargs['name']
```
Kami: Please directly specify all the required and possible arguments in the method signature instead of using ``**kwargs``.

This makes it much more user-friendly for the end user.

Kami:
@Eis-D-Z

> I personally guess that ideally at some point even the requests logic will become async for libcloud, until then yes, I agree that list_nodes() should have an async_list_nodes and the event loop better to be set at driver level, since the driver might not be on the main thread, or someone might want two drivers on separate threads etc...

Would you mind making that change? Aka adding ``async_list_nodes`` method instead of using ``async_`` kwargs.

Eis-D-Z:
All the issues you point out have been resolved as indicated.
Thank you!

Kami:
@Eis-D-Z Thanks and sorry for the delay - I will have a look today.

Kami:
I finally merged this into trunk - thanks for the contribution!

Merging was quite painful since there were a bunch of conflicts (it appears that your branch is not based directly on top of upstream/trunk).

In addition to that, I added missing providers.py entry (bc7b6fe2c1b01d088232ad52d4bc478c9e88cc3c) and added missing driver argument to ``LibcloudError`` class - c690a0efb333062e6ec3e79cfc86f4e131648868.

Eis-D-Z:
I tried to have a clean upstream branch before adding the driver, I had a look at changes and only driver related files for changed, I'm sorry to hear you had conflicts. Thank you very much!

PR #1485, opened 2020-09-02
Fixes request path for ex_set_image_labels()

petersen-poul:
### Description

A simple fix - the request in ex_set_image_labels was missing the "/images/" path. 

I made this commit against 2.8.x because that's where I needed it, but it's also missing in trunk. 

### Status

- done, ready for review

- [ ] Documentation

### DIFF
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -9088,6 +9088,7 @@ class GCENodeDriver(NodeDriver):
         extra['options'] = volume.get('options')
         extra['labels'] = volume.get('labels', {})
         extra['labelFingerprint'] = volume.get('labelFingerprint')
+        extra['users'] = volume.get('users', [])
 
         if 'licenses' in volume:
             lic_objs = self._licenses_from_urls(licenses=volume['licenses'])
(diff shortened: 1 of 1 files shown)


Kami:
Thanks for the contribution.

The change looks good to me, but it looks like existing tests are failing (can you please look into that?). In addition to that, it would be also good to add a test case for this fix.

libcloud/compute/drivers/gce.py
```diff
@@ -2080,7 +2080,7 @@ def ex_set_image_labels(self, image, labels):
             raise ValueError("Must specify a valid libcloud image object.")
         current_fp = image.extra['labelFingerprint']
         body = {'labels': labels, 'labelFingerprint': current_fp}
-        request = '/global/%s/setLabels' % (image.name)
+        request = '/global/images/%s/setLabels' % (image.name)
```
Kami, line 2083: Just adding this for reference here - https://cloud.google.com/compute/docs/reference/rest/v1/images/setLabels

Kami:
I fixed the failing tests, added an assertion (d9683458e0ec5a57fde9879cb3d2e257e705f7f9) and merged changes into trunk.

@petersen-poul I had some troubles merging changes directly into trunk so I needed to manually cherry pick the commits.

It looks like your branch is not based on trunk, but on top of ``2.8.x`` - I assume you are using v2.8.x release series which still supports Python 2?

At this point, I only merged change into trunk (those changes will be included in the next v3.x.x release), but perhaps if some other bug fixes accumulate, we can make another v2.8.x release.

petersen-poul:
Whoops - my apologies for not noticing your comments earlier. Yes, I'm stuck in v2.8.3 for now due to Python 2.7. Yes, the changes for trunk should be virtually the same. If I find anything else in my migration efforts, I'll update them as well.

PR #1488, opened 2020-09-09
Fixes an issue where HTTP requests cause HTTP 400 errors

### LINKED ISSUE #1487: GET request causes HTTP 400 error (OpenStack & potentially other providers)
## Summary

**HTTP 400 errors** are appearing in the logs of our **OpenStack server**:  `ERROR WSGI: code 400, Bad request syntax ('0')`.

They appear when sending a GET request.

## Wireshark

image

## Detailed Information

**Libcloud version:** v3.0.0
**OS:** Ubuntu 14.04.6 LTS
**Provider:** OpenStack Swift
**Service:** Storage

## Reproducing

I'm not sure how to reproduce this without access to an Openstack server along with the server logs/wireshark to examine the HTTP error.  But I would recommend just making a download request.

I traced down the bug to a single-line of code and in a subsequent post will isolate the reason and illustrate the behavior.

```from libcloud.storage.types import Provider

import libcloud.security
libcloud.security.VERIFY_SSL_CERT = False

if __name__ == '__main__':
    cls = get_driver(Provider.OPENSTACK_SWIFT)

    driver = cls('<USER>', '<PASSWORD>',
                 ex_force_auth_url='http://10.110.5.30/auth/v1.0',
                 ex_force_auth_version='1.0',
                 ex_force_service_name='cloudFiles',
                 ex_force_service_type='object-store')

    container = driver.get_container('bucket01')

    # Get the object
    object = driver.get_object('bucket01', 'test-file.txt')

    # Download object as stream.
    for chunk in driver.download_object_as_stream(object):
        print(chunk)
```

mspagon:
### Description

fixes #1487

Removed a single line of code that was causing problems.

HTTP requests with a body equal to an empty string `''` were incorrectly being chunked and ultimately leading to HTTP 400 errors (in certain providers).

### Status

- done, ready for review

- [ ] Documentation

### DIFF
--- a/libcloud/http.py
+++ b/libcloud/http.py
@@ -241,8 +241,6 @@ class LibcloudConnection(LibcloudBaseConnection):
 
         prepped = self.session.prepare_request(req)
 
-        prepped.body = body
-
         self.response = self.session.send(
             prepped,
             stream=stream,


Kami:
Thanks for the contribution.

The change looks good to me, but it would be good to add a test case for it and also ensure it doesn't negatively affect other drivers.

libcloud/http.py
```diff
@@ -241,8 +241,6 @@ def prepared_request(self, method, url, body=None,
 
         prepped = self.session.prepare_request(req)
 
-        prepped.body = body
```
Kami, line 244: I was wondering where we actually then send a body when a prepared request is used and it looks like it's done here - https://github.com/apache/libcloud/pull/1488/files#diff-445eb619a1749349909b5157c751c3a2R239.

I think it would still be good to add some test cases for all this various functionality.

It may not be a bad idea to add a test case which uses a mock http server and tests all the various scenarios (I might look into that if I get a chance).

Kami:
I'm working on the tests per my comment above and I noticed an "interesting" thing - with previous version of the code, request indeed handles the request as chunked one, but it doesn't send ``Transfer-Encoding: chunked`` header (and Libcloud layer doesn't send it either).

So perhaps one option to solve this would also be to send ``Transfer-Encoding: chunked`` header in the Libcloud layer for chunked requests, but your change is likely safer and has less of a chance to negatively affect other requests.

Kami:
I added some basic test cases in - https://github.com/Kami/libcloud/commit/849e024a0bc02e3c934bdc269ab5a81959c8c39f.

I will go ahead and merge it shortly.

Kami:
I went ahead and merged your changes into master.

Thanks again for detailed research into the problem and a fix.

To be on the safe side, I will keep an eye on future bug reports related to this code, but I think it should be fairly safe and there is small chance of an inadvertent regression or similar.

PR #1494, opened 2020-09-22
[ovh] Support North America and all the regions

danhunsaker:
OVH users in North America need to use the `ca.api.ovh.com` endpoints instead of the `api.ovh.com` ones. This PR adds support for that and any other API datacenters OVH may set up in the future. It is also a backward-compatible change, defaulting to the same settings used without the PR.

While in there anyway, I expanded the region list to include all the currently-listed regions on the OVH website (URL in a comment for future reference), and added the ability to retrieve pricing data. These changes should bring the OVH driver up to date with the current state of the provider.


libcloud/compute/drivers/ovh.py
```diff
@@ -44,7 +44,8 @@ class OvhNodeDriver(NodeDriver):
     VOLUME_STATE_MAP = OpenStackNodeDriver.VOLUME_STATE_MAP
     SNAPSHOT_STATE_MAP = OpenStackNodeDriver.SNAPSHOT_STATE_MAP
 
-    def __init__(self, key, secret, ex_project_id, ex_consumer_key=None):
+    def __init__(self, key, secret, ex_project_id, ex_consumer_key=None,
+                 ex_datacenter=None):
```
Kami: For consistency and compatibility with other drivers which, you should use existing ``region`` argument - AWS and many other drivers which support multiple regions / datacenters support that argument.

Kami:
Thanks for the contribution.

I had one comment, besides that, changes look good to me.

Kami:
It looks like tests also failed under Python 3.5 - https://travis-ci.org/github/apache/libcloud/jobs/729459716:

```
headers = {'Accept-Encoding': 'gzip,deflate', 'Content-type': 'application/json', 'Host': 'None.api.ovh.com', 'User-Agent': 'libcloud/3.2.0 (Ovh) ', ...}
    def _json(self, method, url, body, headers):
        meth_name = '_json%s_%s' % (FORMAT_URL.sub('_', url), method.lower())
>       return getattr(self, meth_name)(method, url, body, headers)
E       AttributeError: 'OvhMockHttp' object has no attribute '_json_1_0_cloud_subsidiaryPrice_ovhSubsidiary_US_flavorId_foo_id_get'
```

danhunsaker:
I hate query strings. Fix inbound.

libcloud/compute/drivers/ovh.py
danhunsaker: That makes sense. Update inbound. (I'd just kept the existing names originally.)

danhunsaker:
Hmm. The current test failures have something to do with segfaults from a recent change to pypy, and happen well before my test has a chance to run... Not sure how to address that.

danhunsaker:
Rebased on `master` to take advantage of the segfault test skipping. Everything should be ready, now.

Kami:
@danhunsaker Thanks.

And, yeah looks like that one of the tests I added recently is causing PyPy to segfault for some reason.

For now I pushed a change so we skip that test under PyPy.

Kami:
While testing those changes, I noticed various edge cases and issues which I fixed - 7b6dd90e363117c18c2e6ec17180f28042e6bd02, ae8245ff9e2e434c0f23c869072a751c10e7e09f, fcb937c2412910bf13fb1927620e80548a65712f, 4b95eeff90817bea3a8c2d99593418b54269cfc1 and added corresponding test cases (JsonResponse class one was already there before your changes).

The changes have also been merged into trunk. Thanks again for your contribution.

PR #1498, opened 2020-10-02
Support GCE instances with no service accounts

### LINKED ISSUE #1497: GCE Driver Does Not Support VMs Without Any Service Accounts
## Summary

There does not appear to be any way to create a VM on GCE without any service accounts.

## Detailed Information

Providing the argument `ex_service_accounts=[]` to `create_node` still results in the same behavior as `ex_service_accounts=None`, where the default GCE service account is used on the instance.

GCE does support VMs without any service accounts attached, so I believe this should be a supported configuration.  I've worked around it with the following "stupid monkeypatch", but I believe a proper solution should be included:

```
driver._build_service_accounts_gce_list = lambda *args, **kwargs: []
```

I believe a small change to accept an empty list is the appropriate fix, and I'm happy to provide it, but filing this bug for discussion on the proposal.

Matir:
### Description

This fixes #1497.  When an empty list (`[]`) is passed in for ex_service_accounts,
a VM with no service accounts is created.  When None is passed in (the default),
the behavior remains unchanged, giving the default service account.

### Status

done, ready for review

- [ ] Documentation

### DIFF
--- a/libcloud/compute/drivers/gce.py
+++ b/libcloud/compute/drivers/gce.py
@@ -4893,7 +4893,7 @@ class GCENodeDriver(NodeDriver):
         :rtype:   ``list`` of ``dict``
         """
         gce_service_accounts = []
-        if not service_accounts:
+        if service_accounts is None:
             gce_service_accounts = [{
                 'email': default_email,
                 'scopes': [self.AUTH_URL + default_scope]


Kami:
LGTM, thanks for the contribution.

It would be good to add some test cases though :)

Kami:
I've added a test case (0eb48647e520ed61fa5f473f3b540be9885c3861, 0616c2e8b99736a63d1bee9c5bb4cbd602373712) and merged changes into trunk.

PR #1499, opened 2020-10-05
Adding external features to Outscale Provider

Aikoh0:
### Description

This PR is adding all of the currently available Outscale API features to the Outscale Provider in libcloud and it's corresponding documentation:

- Public Ips
- Accounts
- Tags
- Regions and Subregions
- Access Keys
- Client Gateways
- Dhcp Options
- Direct Links
- Direct Link Interfaces
- Flexible GPUs
- Internet Services
- Listeners
- Load Balancers
- Load Balancer Policies
- Nat Services
- Nets
- Net Access Points
- Net Peerings
- Nics
- Product Types
- Quotas
- Routes
- Routes Tables
- Server Certificates
- Virtual Gateways
- Security Groups
- Security Group Rules
- Subnets
- Tasks
- VPN Connections
- Nodes

### Status

- done, ready for review

### Checklist

- [x] Documentation


libcloud/compute/drivers/outscale.py
```diff
@@ -270,7 +312,7 @@ def create_node(self,
                     ex_keypair_name: str = None,
                     ex_max_vms_count: int = None,
                     ex_min_vms_count: int = None,
-                    ex_nics: dict = None,
+                    ex_nics: [dict] = None,
```
Kami: Hm, does ``[foo]`` type annotation work correctly? I never really seen it used before.

I mostly use ``List[type]`` or ``list[type]``).

libcloud/compute/drivers/outscale.py
```diff
@@ -412,23 +457,80 @@ def create_node(self,
...
+            data = {
+                "DryRun": ex_dry_run,
+                "ResourceIds": [node.id],
+                "Tags": {
+                    "Key": "Name",
+                    "Value": name
+                }
+            }
+            data = json.dumps(data)
+            if self._call_api(action, data).status_code != 200:
```
Kami: To make troubleshooting easier, it would probably be better to throw an exception with the underlying error message /   response.

libcloud/compute/drivers/outscale.py
```diff
@@ -412,23 +457,80 @@ def create_node(self,
...
+        :param      node: the  VM(s)
+                    you want to start (required)
+        :type       node: ``Node``
+
+        :return: the rebooted instances
+        :rtype: ``dict``
+        """
+        action = "StartVms"
+        data = json.dumps({"VmIds": [node.id]})
+        if self._call_api(action, data).status_code == 200:
```
Kami: Same here.

Just returning ``False`` will make troubleshooting in such scenarios much harder.

libcloud/compute/drivers/outscale.py
```diff
@@ -457,6 +559,275 @@ def destroy_node(self, node: Node):
...
+
+        :param      dry_run: If true, checks whether you have the required
+        permissions to perform the action.
+        :type       dry_run: ``bool``
+
+        :return: The Admin Password of the specified Node.
+        :rtype: ``str``
+        """
+        action = "ReadAdminPassword"
+        print(node)
```
Kami: I assume this print statement can get removed.

libcloud/compute/drivers/outscale.py
Kami: EDIT: I just confirmed, it indeed doesn't appear to work:

```bash
libcloud/compute/base.py:216: error: Bracketed expression "[...]" is not valid as a type
```

Please update all ``[foo]`` type annotations to use ``List[foo]`` instead (``List`` is available in ``typing`` module).

Kami:
It's a pretty massive PR so it's a bit hard to review.

Currently two main things I would like to see fixed / improved are:

1. Correct notation for type annotations for lists / container type
2. Better error handling / propagation - currently in most failure situations code returns ``False`` and doesn't log the error or the actual response anyway. At the very least we should log the response + error or throw an exception with the underlying error message to make troubleshooting such scenarios easier.

Aikoh0:
Hi @Kami,
Thanks for the review, and sorry for the delay
I will look into it and come back to you as soon as possible

libcloud/compute/drivers/outscale.py
```diff
@@ -73,6 +76,23 @@ def __init__(self,
...
+        :param      ex_dry_run: If true, checks whether you have the required
+        permissions to perform the action.
+        :type       ex_dry_run: ``bool``
+        :return: locations details
+        :rtype: ``dict``
+        """
+        action = "ReadLocations"
+        data = json.dumps({"DryRun": ex_dry_run})
+        response = self._call_api(action, data)
+        if response.status_code == 200:
```
Kami, line 91: Eventually we should move some of this error handling code to the Response class - this way we can avoid all this duplicated lines, but that's fine for now.

Kami:
Thanks for addressing the feedback. The changes have now been merged into trunk.

Aikoh0:
Thanks for the review @Kami,
I will probably look into adding some exceptions later but I will be quite busy for now with some other task,
I will make a new PR in time.

PR #1504, opened 2020-10-13
Add support for Linode's API v4

dimgal1:
### Description

This Pull Request adds  Linode Compute Driver compatible with[ Linode's API v4](https://developers.linode.com/api/v4/) while simultaneously supporting the previous API version.

### Status

done, ready for review

- [x] Documentation


Kami:
Sorry for the delay.

This is a relatively large PR so it will take a while to review.

micafer:
HI @dimgal1 and @Kami,

I'm interested in this new version.

I tested some of the functionality of the compute driver:

* list_nodes
* list_sizes
* list_locations
* list_volumes
* create_node
* create_volume
* ex_resize_node
* start_node
* stop_node
* reboot_node
* destroy_node
* destroy_volume

I found some minor issues I will put inline.

libcloud/compute/drivers/linode.py
```diff
@@ -728,6 +752,912 @@ def _to_nodes(self, objs):
...
+        'provisioning': NodeState.STARTING,
+        'offline': NodeState.STOPPED,
+        'booting': NodeState.STARTING,
+        'rebooting': NodeState.REBOOTING,
+        'shutting_down': NodeState.STOPPING,
+        'deleting': NodeState.PENDING,
+        'migrating': NodeState.MIGRATING,
+        'rebuilding': NodeState.UPDATING,
+        'cloning': NodeState.MIGRATING,
+        'restoring': NodeState.PENDING,
```
micafer, line 772: There is a missing state 'resizing'. I checked the API documentation and it does not appear. But I made some tests and it does exist. May be a good map value could be:
'resizing': NodeState.RECONFIGURING

libcloud/compute/drivers/linode.py
```diff
@@ -728,6 +752,912 @@ def _to_nodes(self, objs):
...
+            extra=extra)
+
+    def _to_volume(self, data):
+        extra = {
+            'created': self._to_datetime(data['created']),
+            'tags': data['tags'],
+            'location': data['region'],
+            'linode_id': data['linode_id'],
+            'linode_label': data['linode_label'],
+            'state': self.LINODE_VOLUME_STATES[data['status']]
```
micafer: filesystem_path' is an important value that is not returned in the extra fields:

'filesystem_path': data['filesystem_path']

dimgal1:
Hi @micafer,
made the recommendations you mentioned.
Thanks for the feedback!

Kami:
@micafer Since you originally reviewed this PR, can you please have a look again when you get a chance?

And if everything looks OK, feel free to add a CHANGELOG entry and merge it into master.

micafer:
> @micafer Since you originally reviewed this PR, can you please have a look again when you get a chance?
> 
> And if everything looks OK, feel free to add a CHANGELOG entry and merge it into master.

OK, I will take a look.

libcloud/common/linode.py
```diff
@@ -163,3 +176,119 @@ def add_default_params(self, params):
...
+        return super(LinodeResponseV4, self).parse_body()
+
+    def parse_error(self):
+        """
+        Parse the error body and raise the appropriate exception
+        """
+        status = int(self.status)
+        data = self.parse_body()
+        # Use only the first error, as there'll be only one most of the time
+        error = data['errors'][0]
```
micafer, line 207: Why not:
error = " ".join(data['errors'])
In this case all error messages will be returned, not only the first one.

libcloud/compute/drivers/linode.py
```diff
@@ -728,6 +752,914 @@ def _to_nodes(self, objs):
...
+
+    def attach_volume(self, node, volume, persist_across_boots=True):
+        """Attaches a volume to a node.
+        Volume and node must be located in the same region
+
+        :param volume: Volume to be attached (required)
+        :type volume: :class:`StorageVolume`
+
+        :param volume: Node to attach the volume to(required)
+        :type volume: :class:`Node`
```
micafer: It should be:
:param node:
:type node:

libcloud/compute/drivers/linode.py
```diff
@@ -728,6 +752,914 @@ def _to_nodes(self, objs):
...
+
+        :keyword persist_across_boots: Node to attach the volume to(required)
+        :type volume: :class:`Node`
```
micafer: It should be:
:type persist_across_boots: `bool`

libcloud/compute/drivers/linode.py
```diff
@@ -728,6 +752,914 @@ def _to_nodes(self, objs):
...
+        return self._to_addresses(response)
+
+    def ex_allocate_private_address(self, node, address_type='ipv4'):
+        """Allocates a private IPv4 address to node.Only ipv4 is currently supported
+
+        :param node: Node to attach the IP address
+        :type node: :class:`Node`
+
+        :keyword address_type: Type of IP address
+        :type address_type: `bool`
```
micafer: It should be:
:type address_type: `str`

libcloud/compute/drivers/linode.py
```diff
@@ -728,6 +752,914 @@ def _to_nodes(self, objs):
...
+            'type': address_type
+        }
+
+        response = self.connection.request('/v4/linode/instances/%s/ips'
+                                           % node.id,
+                                           data=json.dumps(attr),
+                                           method='POST').object
+        return self._to_address(response)
+
+    def ex_share_address(self, node, addresses=[]):
```
micafer: Why setting addresses as an optional keyword?
It is required to share an address.
I think it shoul be better:

 def ex_share_address(self, node, addresses):

libcloud/compute/drivers/linode.py
```diff
@@ -728,6 +752,914 @@ def _to_nodes(self, objs):
...
+        The following requirements must be met:
+        - The node must not have a pending migration
+        - The account cannot have an outstanding balance
+        - The node must not have more disk allocation than the new size allows
+
+        :param node: the Linode to resize
+        :type node: :class:`Node`
+
+        :param size: the size of the new node
+        :type node: :class:`NodeSize`
```
micafer: It should be:
:type size:

micafer:
Adding other methods as:
ex_get_node or ex_get_volume will be very useful.

libcloud/common/linode.py
dimgal1: data['errors'] is a list of dictionaries so I think this will not work.
Maybe it could be something like this?
```suggestion
        error_list = []
        for error in data['errors']:
            reason = error.get('reason')
            # The field in the request that caused this error
            field = error.get('field')
            if field is not None:
                msg = '%s-%s' % (reason, field)
            else:
                msg = reason
            error_list.append(msg)
        error_msg = ' '.join(error_list)
```

dimgal1:
Thanks for the feedback @micafer . 
Fixed the issues and added these two methods you mentioned.

libcloud/common/linode.py
micafer: Yes, you are right. May be you can leave the original one. As you said in the first error is the most important one.

micafer:
LGTM.

Kami:
Looks like that merging this "broke" automatically generated provider tables - it shows Linode DNS driver doesn't implement any methods.

I believe that's related to the way driver is now structured - our table generation script only changes methods on the base class.

I will work on a workaround.

Kami:
Here we go - 85d6bbbd :)

PR #1513, opened 2020-10-26
Fix error when creating SSHFP DNS records with CloudFlare

### LINKED ISSUE #1512: Cannot create SSHFP records with Cloudflare DNS driver
## Summary

When attempting to create an SSHFP DNS record using the Cloudflare driver, an error like `1004: DNS Validation Error (error chain: 9101: algorithm is a required data field.)` is raised and the record is not created.

## Detailed Information

It is not terribly well documented, but the Cloudflare API does not allow you to create SSHFP records using the "normal" `<algorithm> <type> <fingerprint>` format as a value to the `content` property like you would do with other records, instead you must leave `content` as `null`, and pass a dictionary like:

```json
{
  "algorithm": "2",
  "type": "1",
  "fingerprint": "..."
}
```

as the value of the `data` property.

I have a modified version of the driver working locally, I will try to open an MR to fix this soon

insertjokehere:
### Description

Fixes #1512 

### Status

Ready for review

- [ ] Documentation

### DIFF
--- a/libcloud/dns/drivers/cloudflare.py
+++ b/libcloud/dns/drivers/cloudflare.py
@@ -317,14 +317,20 @@ class CloudFlareDNSDriver(DNSDriver):
         or iodef.
 
         For example: 0 issue test.caa.com
+
+        NOTE: For SSHFP RecordType, data need to be in the format:
+        <algorithm> <type> <fingerprint>
         """
         url = '{}/zones/{}/dns_records'.format(API_BASE, zone.id)
 
-        data = self._normalize_record_data_for_api(type=type, data=data,)
+        content, data = self._normalize_record_data_for_api(
+            type=type, data=data
+        )
         body = {
             'type': type,
             'name': name,
-            'content': data,
+            'content': content,
+            'data': data
         }
 
         merge_valid_keys(body, RECORD_CREATE_ATTRIBUTES, extra)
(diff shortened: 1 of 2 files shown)


libcloud/dns/drivers/cloudflare.py
```diff
@@ -317,14 +317,20 @@ def create_record(self, name, zone, type, data, extra=None):
...
-        data = self._normalize_record_data_for_api(type=type, data=data,)
+        content, data = self._normalize_record_data_for_api(
+            type=type, data=data
+        )
         body = {
             'type': type,
             'name': name,
-            'content': data,
+            'content': content,
+            'data': data
```
Kami, line 333: I assume this will work fine and API won't return an error if ``data`` is empty?

Or perhaps, to be on the safe side, we should just exclude ``data`` attribute if it's an empty dict?

libcloud/dns/drivers/cloudflare.py
```diff
@@ -436,18 +445,28 @@ def ex_disable_ipv6_support(self, zone):
...
-            return data
+            return data, cf_data
 
         if type == RecordType.CAA:
             # Replace whitespace with \t character which CloudFlare API
             # expects
             data = data.replace(' ', '\t')
 
-        return data
+        elif type == RecordType.SSHFP:
```
Kami, line 460: How does CloudFlare API returns this data in the list / get records response?

If it returns it inside the ``data`` attribute ant not inside the ``content`` one, ``_normalize_record_data_from_api`` method needs to be updated as well.

Kami: I quickly verified with their API and it looks like in this case data is indeed returned in the ``content`` field (``"content":"2 1 ABCDEF12345"``) so "de-normalization" is not needed on the response.

I will add a comment to the code to clarify that.

Kami:
I've pushed some small changes and merged it into trunk (54276d3857a4e64c11bacd52bc4f48acdecc35b0, ec9bd4f02bee197cec1208af0e9e6d4223ff8f27, d9c1701b856b93d875599523158b3d2273f6f762). Thanks for the contribution.

If you think that "only include data if not empty" change is not a good idea, let me know.

insertjokehere:
Hi,

Thanks for merging this change. Not including data if it is empty seems reasonable. Thanks for the additional test as well

Will

PR #1519, opened 2020-11-07
 Optimize and speed up imports in the EC2 driver and elsewhere

Kami:
This pull request updates EC2 driver and moves all the various constants from ``libcloud/compute/constants.py`` module into three different files / modules in ``libcloud/compute/constants/`` directory / package.

Only the relatively small ``ec2_region_details_partial`` module is imported directly when importing EC2 driver module.

Other two modules are imported lazily when needed (aka when calling ``list_sizes()`` methods).

## Background, Context

I was doing some import time and memory profiling and noticed that importing some of that Libcloud modules is relatively small and also has relatively large memory footprint.

After digging in, I found that one of the main culprits was ``libcloud.compute.constants`` module.

## Open Questions

I also removed existing ``libcloud/compute/constants.py`` file. If someone feels strongly about it, I can leave it in place for a while longer for backward compatibility reasons.

It's also worth noting that this module also had a bad name since it only contained EC2 related constant and that was not obvious from the module name.

## TODO

- [x] Add some tests and instrumentation which ensures we don't (accidentally) introduce slow or large imports
- [x] Also optimizing pricing data import and caching - only import and cache pricing data for the requested provider and not for all the providers


Kami:
@c-w While analyzing the imports, I noticed ``requests`` import is quite slow (in the range of 50-100ms).

In most cases, we will indeed want to import requests and perform actual HTTP requests, but in some situations that's not the case and it would be great if we could avoid the import in such scenarios (or at least do it lazily on first request).

https://github.com/apache/libcloud/pull/1332 added a check in ``libcloud/__init__.py`` which means we will always perform that import (even if user will never perform any HTTP requests with Libcloud).

Do you have any idea / suggestion on how we could handle that better? I do think it's good we throw and warn user early in situation such as the one #1332 is supposed to fix.

Kami:
@Eis-D-Z Since you recently modified pricing code, you might have some feedback / input on this change - d7ae996e509e71a9d148227c9d32613b09d6af59.

Imo, it's a right change and something we should have done a while ago when ``pricing.json`` file started growing...

In theory, if needed, we could also add new module level constant which can be changed by the user (e.g. ``libcloud.pricing.CACHE_ALL_PRICING_DATA = True``).

c-w:
@Kami Would it be possible to make `_init_once()` lazy, i.e. run it when we do the first actual piece of work with the library as opposed to when the library is imported? If there is a central place to which we could defer `_init_once()` then we could move the `import requests` statement into the function so that the import isn't executed eagerly.

Kami:
@c-w We don't really have a concept of bootstrap / setup phase in Libcloud so it wouldn't be easy to achieve something like that.

The closest thing we could probably do is perform that check on the first actual HTTP request. So it's probably fine to just leave it as-is :)

Eis-D-Z:
> @Eis-D-Z Since you recently modified pricing code, you might have some feedback / input on this change - [d7ae996](https://github.com/apache/libcloud/commit/d7ae996e509e71a9d148227c9d32613b09d6af59).
> 
> Imo, it's a right change and something we should have done a while ago when `pricing.json` file started growing...
> 
> In theory, if needed, we could also add new module level constant which can be changed by the user (e.g. `libcloud.pricing.CACHE_ALL_PRICING_DATA = True`).

I agree as well the file is quite big and is only going to get larger as more clouds are added, or even as AWS expands in this case. I foresee that in the future the file will get so large that the next logical decision will be to split it, so we will have a pricing file for each cloud separately. Caching the data will save significant time for the users who end up invoking the method a lot.
Good job guys!

Kami:
@Eis-D-Z Thanks for your feedback.

I pushed a change so this behavior can also be changed using module level variable - 3c73f668.

I think that's a good idea in addition to the function level argument, because most users don't call that function directly, but just utilize various drivers which call this function.

PR #1526, opened 2020-12-13
[WIP] Add Github actions workflows for CI/CD

Kami:
This represents another attempt to move to Github actions for CI/CD.

In the past, we were blocked by not being able to secret repository secrets, but those have now been set for this repo by the ASF infra team so we can now hopefully finish the workflow and migrate to Github actions.

PR #1527, opened 2020-12-13
Add Github actions workflows for CI/CD #1526 

Kami:
This represents another attempt to move to Github actions for CI/CD.

In the past, we were blocked by not being able to secret repository secrets, but those have now been set for this repo by the ASF infra team so we can now hopefully finish the workflow and migrate to Github actions.


Kami:
Confirmed it's all working and green now (including steps which utilize secrets).

@micafer @tonybaloney @vdloo @c-w and others - does any one you have any objections to switching to Github actions?

I think we should do it for multiple reasons:

1. Better end user experience
2. Faster build run times
3. (for now) free for open source projects

Run times have been especially problematic for the past year or so - over week when a lot of Apache projects were triggering builds it could take may hours for a job to start (IIRC, there were around 50 builders shared for all the Apache projects).

Travis CI also recently announced changes which are not good for open source projects (https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing).

Having said that - even though those changes are bad for open source projects, we should still thank and be grateful to Travis which provided a free service which we successfully utilized for almost 7 years.

I believe Travis CI does have some special free builders for Apache (or it had them at some point), but speed is still an issue.

micafer:
I fully agree with the change to github actions due to the recent changes in Travis CI policies.

tonybaloney:
@Kami 👍  the change. GitHub Actions does more than we need and is much faster

Issue #1536, opened 2020-12-29
Inconsistency in specifying TTL in DNS records between example and tests


macfreek:
## Summary

Right now, the TTL for individual DNS Records is stored either in the `Record.ttl` attribute, or in the `Record.extra` dict, or in both.

Some of the code stores or access the `Record.ttl` value, other code the `Record.extra['ttl']` value. I'll give two examples of either.

This inconsistency causes bugs. E.g. if you follow the documentation for [Create a record with a custom TTL](https://libcloud.readthedocs.io/en/stable/dns/examples.html#create-a-record-with-a-custom-ttl), and subsequently call `libcloud.dns.base.DNSDriver.export_zone_to_bind_format()`, it will **not** output the specified TTL, but the default TTL of the zone.

## Detailed Information

First of all, note that this only concerns a custom TTL in a `libcloud.dns.base.Record`. This is not commonly set. Usually, records used the default TTL as specified in the associated `libcloud.dns.base.Zone`.

### TTL stored in the `Record.ttl` attribute

In the documentation, at [Create a record with a custom TTL](https://libcloud.readthedocs.io/en/stable/dns/examples.html#create-a-record-with-a-custom-ttl), a custom TTL is given as parameter to Record constructor:

```
ttl = 900
record = zone.create_record(name='www', type=RecordType.A, data='127.0.0.1',
                            ttl=ttl)
```

In the [`Record.__init__()` method, on line 163](https://github.com/apache/libcloud/blob/6b588346083a510396c63b5d18db3012aa083071/libcloud/dns/base.py#L163) this ttl is stored in the `Record.ttl` attribute:

```
self.ttl = ttl
```

### TTL stored in `Record.extra["ttl"]`

On the other hand, in the [libcloud/test/dns/test_base.py unit test, on line 38](https://github.com/apache/libcloud/blob/6b588346083a510396c63b5d18db3012aa083071/libcloud/test/dns/test_base.py#L38), the TTL is not set using the `ttl` parameter, but using the `extra` parameter.

    values = {'id': 3, 'name': 'www', 'type': RecordType.A, 'data': '127.0.0.1',
     'extra': {'ttl': 123}},
    # ...
    record = Record(**values)

[...]

PR #1537, opened 2020-12-29
Deprecate Record.ttl in favour of Record.extra['ttl']

macfreek:
## Deprecate `Record.ttl` in favour of `Record.extra['ttl']`

Please read #1536 carefully. The current implementation stores a TTL for a DNS Record at two places, depending where you look in the code. This pull requests picks one of them.

I added @property methods to maintain backwards compatibility, so the API remains consistent. However, it DOES change the internal data structure of a `libcloud.dns.base.Record`.

### Description

See #1536.

### Status

- needs careful review

Also see comment below about the failing unit test. (In my opinion, two of the unit tests are flawed, not the behaviour of this PR).

- [X] Documentation (fixed that :) )
- [X] [Tests](http://libcloud.readthedocs.org/en/latest/testing.html).

### DIFF
--- a/docs/examples/dns/create_record_custom_ttl.py
+++ b/docs/examples/dns/create_record_custom_ttl.py
@@ -8,6 +8,6 @@ driver = cls(*CREDENTIALS_ZERIGO)
 
 zone = [z for z in driver.list_zones() if z.domain == 'example.com'][0]
 
-ttl = 900
+extra = {'ttl': 900}
 record = zone.create_record(name='www', type=RecordType.A, data='127.0.0.1',
-                            ttl=ttl)
+                            extra=extra)
--- a/libcloud/dns/base.py
+++ b/libcloud/dns/base.py
@@ -161,7 +161,8 @@ class Record(object):
         self.zone = zone
         self.driver = driver
         self.extra = extra or {}
-        if ttl is not None:
+        # extra['ttl'] overrides ttl
+        if ttl is not None and self.extra.get('ttl') is None:
             self.extra['ttl'] = ttl
 
     def update(self,
--- a/libcloud/dns/base.py
+++ b/libcloud/dns/base.py
@@ -160,8 +160,9 @@ class Record(object):
         self.data = data
         self.zone = zone
         self.driver = driver
-        self.ttl = ttl
         self.extra = extra or {}
+        if ttl is not None:
+            self.extra['ttl'] = ttl
 
     def update(self,
                name=None,  # type: Optional[str]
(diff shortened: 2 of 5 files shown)


macfreek:
This seems to open a little can of worms in the unit tests.

There were 3 failing unit tests. One was fixed by making a slight alteration in the code (DurableDNS). Two unit tests seemed erratic to me. I've changed the LuaDNS unit test to make more sense. Finally, the DNSPod behaviour does not make sense to me at all, and is still failing. I've created a separate issue (#1538) to deal with that. I'll add a [WIP] (work in progress) tag there until I've find a good solution there.

## DurableDNS

DurableDNSTests.test_list_records() sets both the `ttl` and `extra['ttl']` parameter when creating a new record, but with different values:

Previous behaviour:
```
record.ttl == '3600'
record.extra['ttl'] == 3600
```

Since this PR only stores the TTL value once, not twice, one of these values has to take precedence.

Behaviour in original PR:
```
record.ttl == '3600'
record.extra['ttl'] == '3600'
```

However, after some considerations, I decided that the value of `record.extra['ttl']` should take precedence over `record.ttl` if both are set.

Behaviour in final PR:
```
record.ttl == 3600
record.extra['ttl'] == 3600
```

Note that this PR still only changes how the TTL is stored. The DurableDNSDriver still sets both the `ttl` and `extra['ttl']` parameter when creating a new record, but only casts the later to an `int`, while keeping the former a `str`. That might be a minor bug in the DurableDNSDriver. I have not (yet) fixed that.

## LuaDNS

TTL is set to 13, but is [expected to be None when read](https://github.com/apache/libcloud/blob/6b588346083a510396c63b5d18db3012aa083071/libcloud/test/dns/test_luadns.py#L202):

    def test_create_record_success(self):
        LuadnsMockHttp.type = 'CREATE_RECORD_SUCCESS'
        record = self.driver.create_record(name='test.com.',
                                           zone=self.test_zone,
                                           type='A',
                                           data='127.0.0.1',
                                           extra={'ttl': 13})
        # ...
        self.assertIsNone(record.ttl)

To me, this is not very logical. As a user of the library I would expect that the ttl would be set to 13, not be None. So I've updated the unit test with the new behaviour.

        self.assertEqual(record.ttl, 13)

Old behaviour:
```
record.ttl == None
record.extra['ttl'] == 13
```

New behaviour:
```
record.ttl == 13
record.extra['ttl'] == 13
```

## DNSPod

The behaviour of DNSPod seems a bit erratic to me. See #1538 for details.

The following unit test now fails:

    def test_create_record_success(self):
        DNSPodMockHttp.type = 'CREATE_RECORD_SUCCESS'
        record = self.driver.create_record(name='@', zone=self.test_zone,
                                           type='A', data='96.126.115.73',
                                           extra={'ttl': 13,
                                                  'record_line': 'default'})

Old behaviour
```
record.ttl == None
record.extra['ttl'] == '600'
```

New behaviour
```
record.ttl == '600'
record.extra['ttl'] == '600'
```

Where `'600'` is the default TTL for the zone. The extra parameter (with `ttl: 13`) seems to be completely ignored. This PR exposes this erratic behaviour. Hence, the failed unit test. I'm hesitant to fix this particular bug in this PR (since I'm not familiar how the DNSPod expects the TTL), but I don't want to remove the unit test either.

Kami:
Thanks for opening this PR. I added my main comment here - https://github.com/apache/libcloud/issues/1536#issuecomment-752412165.

Overall, I agree with you - there should only be one way of specifying it, but I think that should be ``record.ttl`` (as pointed out in that comment, that was actually a plan a while back, but we never realized it).

If you agree about ``record.ttl`` being a better API, are you willing to make changes describes here https://github.com/apache/libcloud/issues/1536#issuecomment-752412165?

As describes there, if we want to give our users enough heads up before a breaking change, first release will just need to include a change where all the drivers + code correctly utilize ``record.ttl``, ``record["extra"]`` still works, but it emits a deprecation warning and then in the next release fully removing it.

macfreek:
Removed comment -- I intended to reply in #1536.

Kami:
Thanks for your contribution.

Issue #1545, opened 2021-01-22
Can't access EC2 zones requiring signature version 4


olegrtecno:
## Summary

The latest version (3.3.0) always uses a default signature version 2 for API authentication. Accessing some EC2 regions (e.g. us-east-2) requires a signature version 4.

## Detailed Information

libcloud 3.3.0
python 3.6.8
centos 7.7

The list_nodes() operation fails for 'us-east-2' with an error message
libcloud.common.types.InvalidCredsError: 'AuthFailure: AWS was not able to validate the provided access credentials'

Probably introduced by https://github.com/apache/libcloud/commit/fe725bda928dc5ea4e51d30b8deb60dd9d3c6b73

Kami:
Thanks for reporting this.

Quickly glancing at this code, I don't see how it could affect it, but I need to dig in.

Looking at the code it seems like we default to signature version 2. Did we ever default to version 4?

* https://github.com/apache/libcloud/blob/trunk/libcloud/compute/drivers/ec2.py#L1618
* https://github.com/apache/libcloud/blob/trunk/libcloud/common/aws.py#L54

It seems like we only force version 4 if token auth is used - https://github.com/apache/libcloud/blob/trunk/libcloud/compute/drivers/ec2.py#L4696.

I wonder if this could actually be related to AWS finally deprecating signature v2 for some of those regions (https://aws.amazon.com/blogs/aws/amazon-s3-update-sigv2-deprecation-period-extended-modified/)?

If that's the case, it may make sense to default to v4 for everything, but we need to make sure it don't break anything.

olegrtecno:
Thanks for looking at this.

The code defaults to version 2, which is OK. However, the override for regions, that require version 4, which was taken from the regions settings doesn't work anymore, after the settings are taken from the partial config, where the signature version is missing: https://github.com/apache/libcloud/blob/trunk/libcloud/compute/drivers/ec2.py#L5799

Kami:
@olegrtecno I was able to reproduce the issue - it's indeed a regression.

I will open a PR with a bug fix shortly. As part of that change, I will also make it possibly to specify ``signature_version`` via driver constructor argument and override default value from region constants file.

As a temporary workaround, you could do something like this:

```python
import libcloud.common.aws
libcloud.common.aws.DEFAULT_SIGNATURE_VERSION = "4"
```

This should work as long as you are only using drivers which support signature version 4.

[...]

PR #1546, opened 2021-01-22
Fix authentication related regression in EC2 driver introduced in v3.3.0

Kami:
This pull request fixes EC2 driver related authentication regression / bug which was inadvertently introduced in v3.3.0, originally reported in #1545 .

## Background, Context

#1519 introduced a bug which would cause EC2 driver to always use signature version 2 for authentication.

This won't work with some newer regions which require signature version 4.

This pull request fixes that and also adds a change which allows user to explicitly specify / override which signature version to use for authentication by passing ``signature_version`` argument to the driver constructor.

## Thanks

Thanks to @olegrtecno for reporting this issue.

## TODO

- [x] Changelog entry
- [x] Add a release note entry on potential workaround


Kami:
I'll go ahead and merge this since I hope to start voting thread for a bug fix release today.

Issue #1544, opened 2021-01-19
Google Storage Driver does not always return content-length


RunOrVeith:
## Summary

The google storage driver crashes with a KeyError in `get_object` because google does not send the `content-length` header in some cases.

```
  File "/home/veith/Projects/python-flexi-path/venv/lib/python3.8/site-packages/libcloud/storage/drivers/s3.py", line 370, in get_object
    obj = self._headers_to_object(object_name=object_name,
  File "/home/veith/Projects/python-flexi-path/venv/lib/python3.8/site-packages/libcloud/storage/drivers/s3.py", line 1066, in _headers_to_object
    obj = Object(name=object_name, size=headers['content-length'],
KeyError: 'content-length'

Process finished with exit code 1

```

## Detailed Information

Libcloud version 3.3.0
Ubuntu 18.04
Python 3.8

Google Storage does not send the content-length header in the following case:

- File has `content-encoding: gzip`
- Request header has `Accept-Encoding: gzip`
- File is **not** publicly accessible (allUsers have read rights inside the bucket)

This causes the driver to crash as soon as you call `get_object` on such a file, because the HEAD request that is sent does not return `content-length`, which is directly queried in `_headers_to_object`.

**Code to reproduce**:

```python
import gzip
from libcloud.storage.providers import Provider, get_driver


def create_gzipped_file():
    content = b"I am a test file with super nice content that is so interesting to read"
    encoded_bytes = gzip.compress(content)
    file = "example_gzipped_file.txt"
    with open(file, mode="wb") as f:
        f.write(encoded_bytes)
    return file


# Set your credentials here
email = ""  # Add a service account email here 
key = ""   # Add a key here

# Make sure this container is not publicly readable!
container_name = "ml-integration-tests"

# Upload a dummy file with gzip encoding
file_path = create_gzipped_file()
driver = get_driver(Provider.GOOGLE_STORAGE)(key=email, secret=key)
container = driver.get_container(container_name=container_name)
driver.upload_object(file_path=file_path, container=container, object_name=file_path,  extra={"content_type": "text/plain"},  headers={"Content-Encoding": "gzip"})

# KeyError raised here when we try to access the file
# Note that connection.request sets the headers to accept gzip
#     headers.update({'Accept-Encoding': 'gzip,deflate'})
# in common/base.py line 574
driver.get_object(container_name=container_name, object_name=file_path)
```

**Workaround**:
[...]

PR #1547, opened 2021-01-22
Fix a bug with edge-case in get_object() method in Google Storage Driver

Kami:
This pull request fixes issue reported in #1544.

## Background, Context

In some scenarios, Google Storage S3 compatible API returns object size in ``x-goog-stored-content-length`` header of in ``content-length`` one.

Before this change, our code didn't correctly handle this case. updated the code so ``object.size`` attribute populated in ``get_object()`` is an integer and not a string - this way it's consistent with ``list_objects()`` method.

As part of this PR, I also

## Thanks

Thanks to @RunOrVeith for reporting this issue.


Kami:
I'll go ahead and merge this since I hope to start voting thread for a bug fix release today.

PR #1549, opened 2021-01-26
Make libcloud compliant with the latest Outscale API.

outscale-fne:
### Description

Added the API Acess Rules and Client Certificate Authority (CA) enpoints.

### Status

- done, ready for review

- [x] Documentation

### DIFF
--- a/libcloud/compute/drivers/outscale.py
+++ b/libcloud/compute/drivers/outscale.py
@@ -7711,7 +7711,7 @@ class OutscaleNodeDriver(NodeDriver):
 
     def ex_create_certificate_authority(
         self,
-        ca_perm: str = None,
+        ca_perm: str,
         description: str = None,
         dry_run: bool = False
     ):
@@ -7727,11 +7727,12 @@ class OutscaleNodeDriver(NodeDriver):
         :param      dry_run: If true, checks whether you have the required
         permissions to perform the action.
         :type       dry_run: ``bool``
+
+        :return: the created Ca.
+        :rtype: ``dict``
         """
         action = "CreateCa"
-        data = {"DryRun": dry_run}
-        if ca_perm is not None:
-            data.update({"CaPerm": ca_perm})
+        data = {"DryRun": dry_run, "CaPerm": ca_perm}
         if description is not None:
             data.update({"Description": description})
         response = self._call_api(action, json.dumps(data))
@@ -7741,7 +7742,7 @@ class OutscaleNodeDriver(NodeDriver):
 
     def ex_delete_certificate_authority(
         self,
-        ca_id: str = None,
+        ca_id: str,
         dry_run: bool = False
     ):
         """
@@ -7755,9 +7756,7 @@ class OutscaleNodeDriver(NodeDriver):
         :type       dry_run: ``bool``
         """
         action = "DeleteCa"
-        data = {"DryRun": dry_run}
-        if ca_id is not None:
-            data.update({"CaId": ca_id})
+        data = {"DryRun": dry_run, "CaId": ca_id}
         response = self._call_api(action, json.dumps(data))
         if response.status_code == 200:
             return True
@@ -7786,6 +7785,9 @@ class OutscaleNodeDriver(NodeDriver):
         :param      dry_run: If true, checks whether you have the required
         permissions to perform the action.
         :type       dry_run: ``bool``
+
+        :return: a list of all Ca matching filled filters.
+        :rtype: ``list`` of  ``dict``
         """
         action = "ReadCas"
         data = {"DryRun": dry_run, "Filters": {}}
@@ -7802,7 +7804,7 @@ class OutscaleNodeDriver(NodeDriver):
 
     def ex_update_certificate_authority(
         self,
-        ca_id: str = None,
+        ca_id: str,
         description: str = None,
         dry_run: bool = False
     ):
@@ -7819,11 +7821,12 @@ class OutscaleNodeDriver(NodeDriver):
         :param      dry_run: If true, checks whether you have the required
         permissions to perform the action.
         :type       dry_run: ``bool``
+
+        :return: a the created Ca or the request result.
+        :rtype: ``dict``
         """
         action = "UpdateCa"
-        data = {"DryRun": dry_run}
-        if ca_id is not None:
-            data.update({"CaId": ca_id})
+        data = {"DryRun": dry_run, "CaId": ca_id}
         if description is not None:
             data.update({"Description": description})
         response = self._call_api(action, json.dumps(data))
@@ -7884,14 +7887,15 @@ class OutscaleNodeDriver(NodeDriver):
 
     def ex_delete_api_access_rule(
         self,
-        api_access_rule_id: str = None,
+        api_access_rule_id: str,
         dry_run: bool = False,
     ):
         """
         Delete an API access rule.
         You cannot delete the last remaining API access rule.
 
-        :param      api_access_rule_id: The id of the targeted rule.
+        :param      api_access_rule_id: The id of the targeted rule
+        (required).
         :type       api_access_rule_id: ``str``
 
         :param      dry_run: If true, checks whether you have the required
@@ -7966,7 +7970,7 @@ class OutscaleNodeDriver(NodeDriver):
 
     def ex_update_api_access_rule(
         self,
-        api_access_rule_id: str = None,
+        api_access_rule_id: str,
         ca_ids: List[str] = None,
         cns: List[str] = None,
         description: str = None,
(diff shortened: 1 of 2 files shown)


outscale-fne:
Up!

libcloud/compute/drivers/outscale.py
```diff
@@ -7709,6 +7709,318 @@ def ex_list_vpn_connections(
             return response.json()["VpnConnections"]
         return response.json()
 
+    def ex_create_certificate_authority(
+        self,
+        ca_perm: str = None,
```
Kami: If the argument is required, it should not default to None.

libcloud/compute/drivers/outscale.py
```diff
@@ -7709,6 +7709,318 @@ def ex_list_vpn_connections(
...
+        :param      description: The description of the CA.
+        :type       description: ``bool``
+
+        :param      dry_run: If true, checks whether you have the required
+        permissions to perform the action.
+        :type       dry_run: ``bool``
+        """
+        action = "CreateCa"
+        data = {"DryRun": dry_run}
+        if ca_perm is not None:
```
Kami: Hm, I'm a bit confused, docstring says it's required, but code makes it look like it's not.

libcloud/compute/drivers/outscale.py
```diff
@@ -7709,6 +7709,318 @@ def ex_list_vpn_connections(
...
+        if description is not None:
+            data.update({"Description": description})
+        response = self._call_api(action, json.dumps(data))
+        if response.status_code == 200:
+            return response.json()["Ca"]
+        return response.json()
+
+    def ex_delete_certificate_authority(
+        self,
+        ca_id: str = None,
```
Kami: I assume id is required so it should not default to None.

libcloud/compute/drivers/outscale.py
```diff
@@ -7709,6 +7709,318 @@ def ex_list_vpn_connections(
...
+        :param      ca_id: The ID of the CA you want to delete. (required)
+        :type       ca_id: ``str``
+
+        :param      dry_run: If true, checks whether you have the required
+        permissions to perform the action.
+        :type       dry_run: ``bool``
+        """
+        action = "DeleteCa"
+        data = {"DryRun": dry_run}
+        if ca_id is not None:
```
Kami: If ca_id is required, this if check should not be there.

libcloud/compute/drivers/outscale.py
```diff
@@ -7709,6 +7709,318 @@ def ex_list_vpn_connections(
...
+        ca_fingerprints: List[str] = None,
+        ca_ids: List[str] = None,
+        descriptions: List[str] = None,
+        dry_run: bool = False
+    ):
+        """
+        Returns information about one or more of your Client Certificate
+        Authorities (CAs).
+
+        :param      ca_fingerprints: The fingerprints of the CAs.
```
Kami, line 7776: Which arguments in this method are required? I assume either ca_ids or ca_fingersprints? And perhaps they are even mutually exclusive? Or what happens if both are provided? Same with descriptions.

libcloud/compute/drivers/outscale.py
```diff
@@ -7709,6 +7709,318 @@ def ex_list_vpn_connections(
...
+        :param      descriptions: The descriptions of the CAs.
+        :type       descriptions: ``list`` of ``str``
+
+        :param      dry_run: If true, checks whether you have the required
+        permissions to perform the action.
+        :type       dry_run: ``bool``
+        """
+        action = "ReadCas"
+        data = {"DryRun": dry_run, "Filters": {}}
+        if ca_fingerprints is not None:
```
Kami, line 7794: I assume at least one of those filters needs to be provided, right? If so, we should check that in the code and throw if it's not true.

libcloud/compute/drivers/outscale.py
```diff
@@ -7709,6 +7709,318 @@ def ex_list_vpn_connections(
...
+        if descriptions is not None:
+            data["Filters"].update({"Descriptions": descriptions})
+        response = self._call_api(action, json.dumps(data))
+        if response.status_code == 200:
+            return response.json()["Cas"]
+        return response.json()
+
+    def ex_update_certificate_authority(
+        self,
+        ca_id: str = None,
```
Kami: Same here, required argument should not default to None.

libcloud/compute/drivers/outscale.py
```diff
@@ -7709,6 +7709,318 @@ def ex_list_vpn_connections(
...
+        description: str = None,
+        ip_ranges: List[str] = None,
+        ca_ids: List[str] = None,
+        cns: List[str] = None,
+        dry_run: bool = False,
+    ):
+        """
+        Create an API access rule.
+        It is a rule to allow access to the API from your account.
+        You need to specify at least the CaIds or the IpRanges parameter.
```
Kami, line 7848: Please remove default None value for required arguments.

libcloud/compute/drivers/outscale.py
```diff
@@ -7709,6 +7709,318 @@ def ex_list_vpn_connections(
...
+        if cns is not None:
+            data["Cns"] = cns
+        response = self._call_api(action, json.dumps(data))
+        if response.status_code == 200:
+            return response.json()["ApiAccessRule"]
+        return response.json()
+
+    def ex_delete_api_access_rule(
+        self,
+        api_access_rule_id: str = None,
```
Kami: Same here, None should be removed.

libcloud/compute/drivers/outscale.py
```diff
@@ -7709,6 +7709,318 @@ def ex_list_vpn_connections(
...
+            filters["IpRanges"] = ip_ranges
+        data = {"Filters": filters, "DryRun": dry_run}
+        response = self._call_api(action, json.dumps(data))
+        if response.status_code == 200:
+            return response.json()["ApiAccessRules"]
+        return response.json()
+
+    def ex_update_api_access_rule(
+        self,
+        api_access_rule_id: str = None,
```
Kami: Please remove None from a required argument.

Kami (changes requested):
Thanks for the contribution.

I've added some in-line comments.

Mostly needs some argument default value clean up and adding some additional checks to the method where at least one filter argument needs to be provided.

outscale-fne:
> Thanks for the contribution.
> 
> I've added some in-line comments.
> 
> Mostly needs some argument default value clean up and adding some additional checks to the method where at least one filter argument needs to be provided.

Thank you sir for reviewing! It seems we misunderstood a part af the conding rules... I am patching right now!

libcloud/compute/drivers/outscale.py
outscale-fne: Here all arguments are not required. You would just get all the certificate authorities you have. If you use exery args, you will get all the certificate authorities matching your requirements. All these params are filters.

libcloud/compute/drivers/outscale.py
outscale-fne: As said before, no args required here.

libcloud/compute/drivers/outscale.py
outscale-fne: `You need to specify at least the CaIds or the IpRanges parameter.`
I don't knbow what to do here.
If I remove the default None for both, the user will have to set one to none himself right? Is that what you are asking me to do?
To be sure I understood everything!

libcloud/compute/drivers/outscale.py
Kami: Thanks for clarifying - I thought at least one filter must be provided.

libcloud/compute/drivers/outscale.py
Kami: Yeah, in this case you should default them to None and then you can add check similar to this one to the code:

```python
if not ca_ids and not ip_ranges:
    raise ValueError("Either ca_ids or ip_ranges argument must be provided")
```

Also, are those arguments mutually exclusive? If they are, we should also add another check:

```python
if ca_ids and ip_ranges:
    raise ValueError("Either ca_ids or ip_ranges argument must be provided, but not both")
```

Kami:
Thanks for addressing the feedback - there is one small last comment which still need some work.

libcloud/compute/drivers/outscale.py
outscale-fne: Thank you for the explanations. I am addressing this right now.

outscale-fne (commented):
Seems ok now.

Kami (approved):

Kami:
Thanks. Merged into trunk.

In the future, please also make sure you add corresponding unit tests (including for all the various edge cases - e.g. ensuring ValueError is thrown, etc.).

PR #1552, opened 2021-02-07
Add support for Azure Government to Azure Blobs storage driver

### LINKED ISSUE #1551: Azure blobs - setting AzureBlobsStorageDriver  `host` does not work for non-emulated requests
## Summary

I'm using libcloud to connect to Azure Gov, so I need to change `host` in my AzureBlobsStorageDriver connection object (which uses `<account>.blob.core.usgovcloudapi.net` instead of the default `<account>.blob.core.windows.net`). This does not work - the call fails with `Unexpected status code: 400`.

This broke after https://issues.apache.org/jira/browse/LIBCLOUD-1033 was merged ~2 years ago.

## Detailed Information

https://issues.apache.org/jira/browse/LIBCLOUD-1033 adds support for Azurite, which largely assumes libcloud users would be setting `host` only if using an Azure emulator. Thus, it will always set an `account_prefix`=`key` IFF a `host` is specified, which will break non-emulated Azure calls ([see how a list containers call has a separate Request URI for Emulated calls](https://docs.microsoft.com/en-us/rest/api/storageservices/list-containers2)). 
A workaround for those not using emulators was proposed [here](https://github.com/apache/libcloud/blame/fe4f81b5a31a8f470ff6ba5af7451cccfc423c2b/libcloud/storage/drivers/azure_blobs.py#L167), to set the key to '' (empty string). **This workaround also does not work** - the call also fails with `Unexpected status code: 400`.

After a lengthy debugging session I figured out why the proposed workaround does not work: Setting `key` to the empty string does not prevent `account_prefix` from being set (it also gets set to the empty string). Further, doing this adds an extra `/` to the Request URI ([code here](https://github.com/apache/libcloud/blame/fe4f81b5a31a8f470ff6ba5af7451cccfc423c2b/libcloud/storage/drivers/azure_blobs.py#L193)), which will cause the Azure request to fail. Setting `key=None` fixes this issue.
However that is not the only issue with the proposed workaround - not having a `key` also prevents Azure authentication from working. Therefore I had to add a `driver.connection.user_id = key_variable` statement after the AzureBlobsStorageDriver object was initialized to fix this.

TL;DR - my current workaround is to do this:
```
AzureBlobsStorageDriver(key=None, secret=<secret>, host=<storageAccount>.blob.core.usgovcloudapi.net )
driver.connection.user_id = <storageAccount>
```

My fix recommendation is that an `emulated` boolean flag parameter be added to the AzureBlobsStorageDriver initializer, which will only set the `account_prefix` if it's True. This should also allow the `key` parameter to be set with the `host`.

OS: Mac OS
Python version: 3.6.8
Libcloud version: 3.3.1

c-w:
### Description

This pull request adds support to the Azure Blobs storage driver for deployments of Azure Storage that use custom hostnames such as Azure Government, Azure China, and Azure Private Link. To target these deployments, the driver can be initialized as follows:

```py

cls = get_driver(Provider.AZURE_BLOBS)

host = 'privatelink.blob.core.windows.net'  # for Azure Private Link
host = 'blob.core.chinacloudapi.cn'  # for Azure China
host = 'blob.core.usgovcloudapi.net'  # for Azure Government

driver = cls(key='your storage account name',
             secret='your access key',
             host=host)
```

The change maintains backwards compatibility with the previous approach of using a custom host argument to target Azurite or IoT Edge Storage by explicitly checking for the Azure China, Azure Government, and Azure Private Link well-known endpoint constants.

Resolves https://github.com/apache/libcloud/issues/1551

### Status

- done, ready for review

### Checklist

- [x] [Code linting](http://libcloud.readthedocs.org/en/latest/development.html#code-style-guide) **[CI passed](https://github.com/apache/libcloud/actions/runs/545677170)**
- [x] Documentation **[Added new docs section](https://github.com/apache/libcloud/blob/505894fa9e2b2776889b673789651ff46734c7d7/docs/storage/drivers/azure_blobs.rst#connecting-to-azure-government)**
- [x] [Tests](http://libcloud.readthedocs.org/en/latest/testing.html) **[libcloud-tests passed](https://clewolff.visualstudio.com/libcloud-tests/_build/results?buildId=1589&view=logs&s=ee3800fd-6e81-525f-e564-94108585217d)**
- [x] [ICLA](http://libcloud.readthedocs.org/en/latest/development.html#contributing-bigger-changes) **[committer](http://people.apache.org/phonebook.html?uid=clewolff)**

### DIFF
--- a/libcloud/storage/drivers/azure_blobs.py
+++ b/libcloud/storage/drivers/azure_blobs.py
@@ -177,7 +177,7 @@ class AzureBlobsConnection(AzureConnection):
     implements an auditing proxy or similar, the driver can be instantiated
     with ``host='theaccount.somewhere.tld'`` and ``key=''``.
 
-    :param account_prefix: Optional prefix identifying the sotrage account.
+    :param account_prefix: Optional prefix identifying the storage account.
                            Used when connecting to a custom deployment of the
                            storage service like Azurite or IoT Edge Storage.
     :type account_prefix: ``str``
--- a/libcloud/storage/drivers/azure_blobs.py
+++ b/libcloud/storage/drivers/azure_blobs.py
@@ -68,6 +68,9 @@ AZURE_LEASE_PERIOD = int(
 )
 
 AZURE_STORAGE_HOST_SUFFIX = 'blob.core.windows.net'
+AZURE_STORAGE_HOST_SUFFIX_CHINA = 'blob.core.chinacloudapi.cn'
+AZURE_STORAGE_HOST_SUFFIX_GOVERNMENT = 'blob.core.usgovcloudapi.net'
+AZURE_STORAGE_HOST_SUFFIX_PRIVATELINK = 'privatelink.blob.core.windows.net'
 
 AZURE_STORAGE_CDN_URL_DATE_FORMAT = '%Y-%m-%dT%H:%M:%SZ'
 
@@ -173,9 +176,10 @@ class AzureBlobsConnection(AzureConnection):
     these deployments, the parameter ``account_prefix`` must be set on the
     connection. This is done by instantiating the driver with arguments such
     as ``host='somewhere.tld'`` and ``key='theaccount'``. To specify a custom
-    host without an account prefix, e.g. for use-cases where the custom host
-    implements an auditing proxy or similar, the driver can be instantiated
-    with ``host='theaccount.somewhere.tld'`` and ``key=''``.
+    host without an account prefix, e.g. to connect to Azure Government or
+    Azure China, the driver can be instantiated with the appropriate storage
+    endpoint suffix, e.g. ``host='blob.core.usgovcloudapi.net'`` and
+    ``key='theaccount'``.
 
     :param account_prefix: Optional prefix identifying the storage account.
                            Used when connecting to a custom deployment of the
@@ -206,7 +210,7 @@ class AzureBlobsStorageDriver(StorageDriver):
 
     def __init__(self, key, secret=None, secure=True, host=None, port=None,
                  **kwargs):
-        self._host_argument_set = bool(host)
+        self._host = host
 
         # B64decode() this key and keep it, so that we don't have to do
         # so for every request. Minor performance improvement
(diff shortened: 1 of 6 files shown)


libcloud/storage/drivers/azure_blobs.py
```diff
@@ -217,15 +221,34 @@ def __init__(self, key, secret=None, secure=True, host=None, port=None,
                                                       port=port, **kwargs)
 
     def _ex_connection_class_kwargs(self):
```
Kami, line 223: Some unit tests for this method would be good.

libcloud/storage/drivers/azure_blobs.py
```diff
@@ -217,15 +221,34 @@ def __init__(self, key, secret=None, secure=True, host=None, port=None,
...
+        # targeting the default Azure Storage endpoints
+        if self._host is None:
+            return {'host': '%s.%s' % (self.key, AZURE_STORAGE_HOST_SUFFIX)}
+
+        # connecting to a special storage region like Azure Government or
+        # Azure China requires setting a custom storage endpoint but we
+        # still use the same scheme to identify a specific account as for
+        # the standard storage endpoint
+        try:
+            host_suffix = next(
```
Kami, line 234: Is there a situation where we may want to allow end user to specify full host as-is - aka so we don't add any prefix to it, but user already includes that in the ``host`` argument?

I think this would give users more flexibility in some situations.

Kami: To clarify - if a new endpoint ever gets added or similar, this would require use to update the code to support it.

If we allowed user to specify a full arbitrary host, they may be able to use that instead (until a new version is released / similar).

For us to be able to support that, we would likely need to add new argument to the constructor - e.g. ``ex_use_host_as_is`` / ``ex_dont_prefix_host`` (or some better name, can't come up with anything better atm).

libcloud/storage/drivers/azure_blobs.py
c-w: Added tests for the Azurite path in c5815533 and for the new GovCloud path in 739a6a33.

libcloud/storage/drivers/azure_blobs.py
c-w: The only current Azure Storage version that I'm aware of which wouldn't be covered by the approach proposed in this PR is [Azure Stack Hub Storage](https://docs.microsoft.com/en-us/azure-stack/user/azure-stack-storage-overview?view=azs-2008). However the API version that libcloud uses (`2018-11-09`) is newer than the [latest supported version](https://docs.microsoft.com/en-us/azure-stack/user/azure-stack-profiles-azure-resource-manager-versions?view=azs-2008#overview-of-the-2019-03-01-hybrid-profile) of Azure Stack (`2017-11-09`) so libcloud doesn't support it anyways.

I get the appeal of an `ex_force_host` or similar argument, but I wonder if the simplicity of this explicit approach will be more user-friendly as it doesn't introduce an argument that's specific to just this driver and instead manages the complexity internally. Thoughts?

Kami: Yeah, I didn't really mean it as an alternative, but in addition to your proposed changes (so the interface for the common case would still be the same, but in case there is a need to use a fully custom host, there is an easy way to do it).

I also don't think that's a blocker for this PR so if you think it's a good idea, feel free to open a new PR with that change in the future.

Kami (approved):
LGTM.

Once both of the PRs are merged you can also work on a new minor release with those two changes, if you wish.

libcloud/storage/drivers/azure_blobs.py
c-w: In that case I'd say to defer the addition of `ex_force_host` for now.

PR #1553, opened 2021-02-08
Respect common HTTP headers when uploading Azure Storage blob via stream

### LINKED ISSUE #1550: Azure Blob Storage does not return Content-Encoding when blob was uploaded via stream
## Summary

When uploading a blob to Azure Blob Storage that is gzip encoded using `upload_object_via_stream` and then try to read it again, the response does not contain the `Content-Encoding: gzip` header. It works when uploading using `upload_object`.

## Detailed Information

Ubuntu 18.04, Libcloud 3.3.1, python 3.8

**Code to Reproduce**:

```python
import gzip
from libcloud.storage.providers import Provider, get_driver


def create_gzipped_file():
    content = b"I am a test file with super nice content that is so interesting to read"
    encoded_bytes = gzip.compress(content)
    file = "example_gzipped_file.txt"
    with open(file, mode="wb") as f:
        f.write(encoded_bytes)
    return file


account = ""  # Add a service account here
key = ""  # Add a key here
container_name = ""  # Add a container name here


driver = get_driver(Provider.AZURE_BLOBS)(key=account, secret=key)
container = driver.get_container(container_name=container_name)

# Upload a dummy file with gzip encoding via stream
content = b"I am a test file with super nice content that is so interesting to read"
encoded_bytes = gzip.compress(content)
streaming_file = "example_gzipped_file_stream_upload.txt"
driver.upload_object_via_stream(
    iterator=iter([encoded_bytes]),
    container=container,
    object_name=streaming_file,
    extra={"content_type": "text/plain"},
    headers={"Content-Encoding": "gzip"},
)

# Upload a dummy file with gzip encoding regularly
regular_file = create_gzipped_file()
driver.upload_object(
    regular_file,
    container=container,
    object_name=regular_file,
    extra={"content_type": "text/plain"},
    headers={"Content-Encoding": "gzip"},
)

# Get the headers, this is a HEAD request, but also not present in a GET request
regular_response_headers = driver.get_object(container_name=container_name, object_name=regular_file).extra
streaming_response_headers = driver.get_object(container_name=container_name, object_name=streaming_file).extra
print(regular_response_headers["content_encoding"])  # gzip
print(streaming_response_headers["content_encoding"])
# --> content_encoding is not present (or rather defaults to None)

```

As you can see, the file that was uploaded via stream does not have a valid Content-Encoding header.
I looked for a possible cause, but could not really find one.

What I found is:
- `AzureBlobStorageDriver._commit_block` does not set any headers. This is called after all chunks have been uploaded. However, setting the header in that request manually does not help
- The Azure documentation for both [PUT](https://docs.microsoft.com/en-us/rest/api/storageservices/put-blob) and [GET](https://docs.microsoft.com/en-us/rest/api/storageservices/get-blob) mention that the Content-Encoding header should be returned
- Difference in execution path for `upload_object` and `upload_object_via_stream` is a call to `AzureBlobStorageDriver._upload_directly` for the former, and `AzureBlobStorageDriver._upload_in_chunks` for the later
- Similar to #1544, this might be a bug within Azure itself, but here I could not verify it completely
- I have not found a workaround other than just trying if it was gzip encoded if I can't read it normally
- Not sure if `upload_object_via_stream` uses [Append Block](https://docs.microsoft.com/en-us/rest/api/storageservices/append-block) internally, but that at least does not accept the Content-Encoding header. 

**Possible Solutions**
- if this is a bug that can be fixed, the Content-Encoding should be present when retrieving the blob regardless of upload method
- If this is a limitation of streaming upload, an error (or at least a warning) should be in place if the user supplied the `Content-Encoding` header when uploading

c-w:
### Description

Resolves https://github.com/apache/libcloud/issues/1550

As per the [documentation](https://docs.microsoft.com/en-us/rest/api/storageservices/put-block-list#request-headers), when uploading an object via stream to Azure Storage, common HTTP headers (e.g. `Content-Encoding`) must be provided via an Azure-specific name (e.g. `x-ms-blob-content-encoding`). The driver was previously ignoring user-provided headers when uploading objects via stream and wasn't adapting the header names to their Azure-specific equivalent, which are both issues fixed by this pull request.

### Status

- done, ready for review

- [x] [Code linting](http://libcloud.readthedocs.org/en/latest/development.html#code-style-guide) **[CI passed](https://github.com/apache/libcloud/actions/runs/546750410)**
- [x] Documentation **n/a: bugfix**
- [x] [Tests](http://libcloud.readthedocs.org/en/latest/testing.html) **[libcloud-tests passed](https://clewolff.visualstudio.com/libcloud-tests/_build/results?buildId=1591&view=logs&j=1a900d4b-8046-5265-156b-217ea8bd79e8)**
- [x] [ICLA](http://libcloud.readthedocs.org/en/latest/development.html#contributing-bigger-changes) **[committer](http://people.apache.org/phonebook.html?uid=clewolff)**

### DIFF
--- a/libcloud/storage/drivers/azure_blobs.py
+++ b/libcloud/storage/drivers/azure_blobs.py
@@ -803,6 +803,7 @@ class AzureBlobsStorageDriver(StorageDriver):
         response = self._commit_blocks(object_path=object_path,
                                        chunks=chunks,
                                        lease=lease,
+                                       headers=headers,
                                        meta_data=meta_data,
                                        content_type=content_type,
                                        data_hash=data_hash,
@@ -822,7 +823,7 @@ class AzureBlobsStorageDriver(StorageDriver):
             'bytes_transferred': bytes_transferred,
         }
 
-    def _commit_blocks(self, object_path, chunks, lease,
+    def _commit_blocks(self, object_path, chunks, lease, headers,
                        meta_data, content_type, data_hash,
                        object_name, file_path):
         """
@@ -837,7 +838,7 @@ class AzureBlobsStorageDriver(StorageDriver):
 
         data = tostring(root)
         params = {'comp': 'blocklist'}
-        headers = {}
+        headers = headers or {}
 
         lease.update_headers(headers)
         lease.renew()
@@ -856,6 +857,8 @@ class AzureBlobsStorageDriver(StorageDriver):
         headers['Content-MD5'] = data_hash.decode('utf-8')
 
         headers['Content-Length'] = len(data)
+
+        headers = self._fix_headers(headers)
 
         response = self.connection.request(object_path, data=data,
                                            params=params, headers=headers,
@@ -928,6 +931,31 @@ class AzureBlobsStorageDriver(StorageDriver):
                                           object_name=obj.name)
 
         return False
+
+    def _fix_headers(self, headers):
+        """
+        Update common HTTP headers to their equivalent in Azure Storage
+
+        :param headers: The headers dictionary to be updated
+        :type headers: ``dict``
+        """
+        to_fix = (
+            'cache-control',
+            'content-encoding',
+            'content-language',
+        )
+
+        fixed = {}
+
+        for key, value in headers.items():
+            key_lower = key.lower()
+
+            if key_lower in to_fix:
+                fixed['x-ms-blob-%s' % key_lower] = value
+            else:
+                fixed[key] = value
+
+        return fixed
 
     def _update_metadata(self, headers, meta_data):
         """
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,18 @@
 ﻿Changelog
 =========
 
+Changes in Apache Libcloud 3.3.2
+--------------------------------
+
+Storage
+~~~~~~~
+
+- [Azure Blobs] Respect Content-Encoding, Content-Language and Cache-Control
+  headers when uploading blobs via stream.
+
+  Reported by Veith Röthlingshöfer - @RunOrVeith.
+  (GITHUB-1550)
+
 Changes in Apache Libcloud 3.3.1
 --------------------------------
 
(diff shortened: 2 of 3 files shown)


libcloud/storage/drivers/azure_blobs.py
```diff
@@ -929,6 +932,31 @@ def delete_object(self, obj):
 
         return False
 
+    def _fix_headers(self, headers):
```
Kami, line 958: Can you please add some unit tests for this method / functionality?

Kami (approved):
LGTM, just some unit tests would be good (I assume libcloud-tests covers the integration / end to end part) :)

libcloud/storage/drivers/azure_blobs.py
c-w: Fitting this into the mock request/response framework was proving challenging so I added an isolated test for the new normalization functionality in 98cd383. I also added an integration test for this scenario in https://github.com/c-w/libcloud-tests/commit/e305b9d34276f50543b8c3a5f342eb18f04fe945.

c-w: Investigating why the new e2e test fails for Azurite v3 only ([see test run](https://clewolff.visualstudio.com/libcloud-tests/_build/results?buildId=1593&view=logs&j=2be97940-aaf3-5b4e-dc43-b88617d20841&t=741988fb-4d4e-5131-94f3-05b5b69d36ed)).

Kami: Yeah, direct unit tests for that method are fine :)

Maybe something to do with auth signature? Since, IIRC, we use header values when generating the signature.

Kami: On that note - I wonder if we could some merge some of that end to end tests in that repo (if it makes sense and how we could do that).

c-w: I believe the test failure may be linked to an issue in AzuriteV3, see https://github.com/Azure/Azurite/issues/629#issuecomment-776258485. Given that the test passes for all other storage targets, I disabled the test for AzuriteV3 in https://github.com/c-w/libcloud-tests/commit/b2be083358cfe3898aff50bb380c175cbf8ea2aa.

PR #1558, opened 2021-03-01
Updates on CloudSigma driver

dimgal1:
### Description

- Add all supported regions
- Use example sizes  currently provided by CloudSigma
- Change methods parameters to use actual  cores to define CPUs instead of MHz
- Change Drive methods parameters to use GBs instead of bytes
- Implement SSH key related methods ( `list_key_pairs`, `get_key_pair`, `create_key_pair`, `import_key_pair_from_string`, `delete_key_pair`), add ssh keys param on `create_node` 
- Implement `ex_attach_drive`, `ex_detach_drive`, `ex_destroy_drive`, `reboot_node`, `ex_get_node`
- Make deleting all attached drives on `destroy_node` optional
- Add tests on methods added
- Add volume methods which call the appropriate ex_drive method ( since CloudSigma drives are pretty much volumes in the context of libcloud, I thought this makes sense)

### Status

- done, ready for review

- [ ] Documentation

### DIFF
--- a/libcloud/compute/drivers/cloudsigma.py
+++ b/libcloud/compute/drivers/cloudsigma.py
@@ -22,6 +22,7 @@ import re
 import time
 import copy
 import base64
+import hashlib
 
 try:
     import simplejson as json
@@ -35,13 +36,16 @@ from libcloud.utils.misc import str2dicts, str2list, dict2str
 from libcloud.common.base import ConnectionUserAndKey, JsonResponse, Response
 from libcloud.common.types import InvalidCredsError, ProviderError
 from libcloud.common.cloudsigma import INSTANCE_TYPES
+from libcloud.common.cloudsigma import SPECS_TO_SIZE
 from libcloud.common.cloudsigma import API_ENDPOINTS_1_0
 from libcloud.common.cloudsigma import API_ENDPOINTS_2_0
 from libcloud.common.cloudsigma import DEFAULT_API_VERSION, DEFAULT_REGION
+from libcloud.common.cloudsigma import MAX_VIRTIO_CONTROLLERS, MAX_VIRTIO_UNITS
 from libcloud.compute.types import NodeState, Provider
 from libcloud.compute.base import NodeDriver, NodeSize, Node
 from libcloud.compute.base import NodeImage
 from libcloud.compute.base import is_private_subnet
+from libcloud.compute.base import KeyPair
 from libcloud.utils.iso8601 import parse_date
 from libcloud.utils.misc import get_secure_random_string
 
@@ -94,7 +98,8 @@ class CloudSigmaInsufficientFundsException(Exception):
 
 
 class CloudSigmaNodeSize(NodeSize):
-    def __init__(self, id, name, cpu, ram, disk, bandwidth, price, driver):
+    def __init__(self, id, name, cpu, ram, disk, bandwidth, price,
+                 driver, extra=None):
         self.id = id
         self.name = name
         self.cpu = cpu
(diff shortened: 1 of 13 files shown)


libcloud/compute/drivers/cloudsigma.py
```diff
@@ -1138,10 +1148,13 @@ def create_node(self, name, size, image, ex_metadata=None,
         # ide 0:0
         data = {}
         data['name'] = name
-        data['cpu'] = size.cpu
+        data['cpu'] = size.cpu * 2000
```
Kami, line 1151: Would be good to add a comment on why we need to do ``* 2000`` here.

libcloud/compute/drivers/cloudsigma.py
```diff
@@ -1194,19 +1207,55 @@ def create_node(self, name, size, image, ex_metadata=None,
 
         return node
 
-    def destroy_node(self, node):
+    def destroy_node(self, node, delete_drives=False):
```
Kami: Since it's not a standard argument, please prefix it with ``ex_``.

libcloud/compute/drivers/cloudsigma.py
```diff
@@ -1194,19 +1207,55 @@ def create_node(self, name, size, image, ex_metadata=None,
...
+        """
+        state = node.state
+
+        if state == NodeState.RUNNING:
+            stopped = self.stop_node(node)
+        else:
+            stopped = True
+
+        if not stopped:
+            raise CloudSigmaException(
```
Kami, line 1244: I guess there is no way to retrieve a better error - aka why stopping node failes?

libcloud/compute/drivers/cloudsigma.py
```diff
@@ -1914,15 +2124,47 @@ def _parse_ips_from_nic(self, nic):
...
+        image = None
+        drive_size = 0
+        for item in extra['drives']:
+            if item['boot_order'] == 1:
+                drive = self.ex_get_drive(item['drive']['uuid'])
+                drive_size = drive.size
+                image = '{} {}'.format(drive.extra.get('distribution', ''),
+                                       drive.extra.get('version', ''))
+                break
+        # try to find if node size is from example sizes given by CloudSigma
```
Kami, line 2152: It seems somewhat inefficient to run this on every to ``_to_node()`` method call (even if that ``list_nodes()`` method doesn't result in a HTTP request and just filters a local list).

Is there any way we could handle this better / more efficiently?

Perhaps we could build a module level dictionary constant where the key would be a tuple of (cpus, memory, drive_size) and the value would be ``NodeSize``. Then we can just do a simple dictionary lookup.

I think this should be fine since there isn't a lot of pre-defined sizes,

Kami:
Thanks for the contribution.

I added some comments, nothing major.

libcloud/compute/drivers/cloudsigma.py
dimgal1: I wasn't sure about this one, I used the return value only because the base `NodeDriver` returns a boolean so I handled the case of False.
Practically `stopped` was never False, CloudSigma returns 403 status code if the action cannot be done so an exception is raised or the action is completed and `stopped` is True. 
Let me know if I should remove this completely

libcloud/compute/drivers/cloudsigma.py
dimgal1: I used a dictionary as you said, now we could remove completely `INSTANCE_TYPES` and use only the new dictionary's values to list sizes as they are basically the same. Let me know what you think.

Kami (approved):
Thanks for addressing all the comments. LGTM :+1:

Issue #1560, opened 2021-03-03
Support volume v3 API endpoint in OpenStack driver


micafer:
## Support volume v3 API ednpoint in OpenStack driver

Currently the OpenStack driver only supports v2 API of the volume API.
It should be added the support to v3 as v2 is currently deprecated.
Furthermore in some sites the cinderv2 volume is not yet available (only cinderv3), so all the volume related functions fail with the error 'Could not find specified endpoint'.

https://docs.openstack.org/api-ref/block-storage/v3

PR #1561, opened 2021-03-03
Support volume v3 API endpoint in OpenStack driver

micafer:
## Support volume v3 API endpoint in OpenStack driver

Implements: #1560

### Description

This solution tries to find if the cinderv3 endpoint is available and in this case the volume related functions will use that endpoint, otherwise it will user the cinderv2 one.

### Status

- done, ready for review

- [ ] Documentation


Kami (approved):
Can you please add a changelog entry?

Besides that, LGTM :+1:

PR #1570, opened 2021-03-17
Fix null value in DigitalOcean create a record

karantan:
### Description

This fixes "field could not be unmarshalled" error when trying to send
'null' as a boolean value. It needs to be int or bool, not a string.

Ref: https://developers.digitalocean.com/documentation/v2/#create-a-new-domain-record

### Status

- done, ready for review

- [x] Documentation
- ~[ ] [ICLA](http://libcloud.readthedocs.org/en/latest/development.html#contributing-bigger-changes) (required for bigger changes)~


Kami (approved):
LGTM :+1: 

I wonder if ``null`` string is a left over from previous version of their API or similar.

Kami:
I see ``update_record()`` does the same thing so I will update that code as well.

PR #1571, opened 2021-03-17
Fix NSOne DNS Provider

karantan:
### Description

- handle mx records
- handle root domain

### Status

- done, ready for review

- [x] Documentation

### DIFF
--- a/libcloud/dns/drivers/nsone.py
+++ b/libcloud/dns/drivers/nsone.py
@@ -114,7 +114,7 @@ class NsOneDNSDriver(DNSDriver):
             else:
                 raise e
 
-        zone = self._to_zone(response.objects[0])
+        zone = self._to_zone(response.object)
 
         return zone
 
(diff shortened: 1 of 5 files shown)


Kami:
Thanks for the contribution and sorry for the delay.

The change looks good to me, but it would be great if you could add corresponding unit tests.

libcloud/dns/drivers/nsone.py
```diff
@@ -206,7 +206,7 @@ def delete_record(self, record):
 
         return response.status == httplib.OK
 
-    def create_record(self, name, zone, type, data, extra=None):
+    def create_record(self, name, zone, type_, data, extra=None):
```
Kami: ``type`` argument is part of the standard API (aka is the same for all the provider drivers) so we should use it here as well.

I know it's not ideal since it clashes with a built-int value, but not much we can do at this point (besides changing the standard API + updating all the drivers in the future).

Kami: Besides that, changes look good to me - please just add some tests when you get a chance.

karantan:
hey, sorry for the delay. I was busy with other things ... anyway I added 3 tests and renamed `type_` back to `type`. do we need to do anything else?

karantan:
@Kami gentle ping

Kami (approved):
@karantan Sorry for a very late response.

Thanks for addressing the comments. Change LGTM :+1:

PR #1572, opened 2021-03-20
Add integration tests for Azure Storage driver

c-w:
### Description

As discussed in https://github.com/apache/libcloud/pull/1553#discussion_r572431845, this pull request ports the Azure Storage driver integration tests from [libcloud-tests](https://github.com/c-w/libcloud-tests) to trunk.

The tests are split into a generic storage integration test base (which uses the high level storage driver API to run various scenarios) and specific instantiations which exercise various Azure Storage backends. The intent is that the test base in the future can be re-used to quickly add integration tests for other storage backends.

Notes:
- Transient errors with the Azure Ubuntu package mirrors required adding a call to `apt-get update` to fix the CI, similar to what's described in https://github.com/actions/virtual-environments/issues/675.
- The new storage integration have been added as a new tox target to avoid slowing down the unit test suite, re-using some existing patterns laid out by integration tests of the compute API.
- ~~While libcloud-tests ran against containerized and cloud-based storage accounts, for now this pull request only ports the former as I will soon loose access to my free cloud credits. However, if someone wants to add me to an Azure subscription that is funded for this project, or provide me with the credentials for a service principal, I'm more than happy to add the required code to run the test suite against the live cloud storage backends.~~

### Status

- done, ready for review

### Checklist

- [x] Documentation


integration/storage/base.py
```diff
@@ -0,0 +1,359 @@
...
+            for optional in 'host', 'port', 'secure':
+                value = getattr(self, optional, None)
+                if value is not None:
+                    kwargs[optional] = value
+
+            driver_class = providers.get_driver(self.provider)
+            self.driver = driver_class(**kwargs)
+
+        def tearDown(self):
+            for container in self.driver.list_containers():
```
Kami, line 65: Should we wrap this with try / except and still proceed if there is an error (but I guess we should at least log an error if there is an exception and instruct user to manually delete those containers or similar).

Kami:
Thanks for working on this.

> While libcloud-tests ran against containerized and cloud-based storage accounts, for now this pull request only ports the former as I will soon loose access to my free cloud credits. However, if someone wants to add me to an Azure subscription that is funded for this project, or provide me with the credentials for a service principal, I'm more than happy to add the required code to run the test suite against the live cloud storage backends.

@tonybaloney Do you happen to know if Azure has some kind of free credits for OSS program, similar to the AWS one which we could use for running end to end tests?

c-w:
@Kami @tonybaloney Given that ASF is a non-profit, the [Azure for non-profits grant](https://www.microsoft.com/en-us/nonprofits/azure) could be an avenue to get some credits.

integration/storage/base.py
c-w: Done in d9f577e8.

tonybaloney:
I've put in an internal request at Microsoft for some credit.

I'll ping back on here when/if it gets approved

c-w:
@tonybaloney @Kami Any thoughts on next steps for this integration test PR?

tox.ini
```diff
@@ -212,10 +212,15 @@ commands =
...
+[testenv:integration-compute]
+deps = -r{toxinidir}/integration/compute/requirements.txt
 
-commands = python -m integration
+commands = python -m integration.compute
+
+[testenv:integration-storage]
+deps = -r{toxinidir}/integration/storage/requirements.txt
+
+commands = python -m integration.storage
```
Kami, line 230: This is fine for now, although in the future I would prefer to use ``pytest`` directly - this way common pytest flags (e.g. ``-s``) and plugins can be used.

Kami: EDIT: Sorry, I just see you followed an existing pattern since original compute integration tests code is quite old :)

Kami (approved):
LGTM, thanks for working on this :+1: 

(I'll push a small change in the near future to utilize pytest instead of standard unit test test loader for consistency with other test jobs and so we utilize various pytest plugins such as test timing info, etc.)

Issue #1575, opened 2021-04-26
LibcloudConnection Session timeout  does not work


dimgal1:
## Summary

Based on this issue https://github.com/psf/requests/issues/3341, the requests module has changed the way timeout is set,  so this
https://github.com/apache/libcloud/blob/cc5987b5f79d4bca9c9139f24878686224b75425/libcloud/http.py#L205 
has no effect.

PR #1576, opened 2021-04-26
Fix LibcloudConnection Session timeout issue

dimgal1:
### Description

See issue #1575

### Status

- done, ready for review

- [ ] Documentation

### DIFF
--- a/libcloud/http.py
+++ b/libcloud/http.py
@@ -229,7 +229,8 @@ class LibcloudConnection(LibcloudBaseConnection):
             headers=headers,
             allow_redirects=ALLOW_REDIRECTS,
             stream=stream,
-            verify=self.verification
+            verify=self.verification,
+            timeout=self.session.timeout,
         )
 
     def prepared_request(self, method, url, body=None,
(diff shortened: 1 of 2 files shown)


c-w (approved):

Kami:
Thanks for the contribution.

The change looks good to me, but can you please also add a test case for it?

dimgal1:
done @Kami.

Kami (approved):

Kami:
After digging in some more and to be on the safe side, I decided to add some additional test cases / asserts (32653fe64a045f8c03d127544be86edc74dbfa74), since the one which was added in this PR (https://github.com/apache/libcloud/pull/1576/files#diff-3ad0002be7627437faad5c60d1fadd3674be25d07619a8c21e0f56942965381eR190) really just asserts that the ``timeout`` variable is set on the session object and not that the timeout is actually passed to the underlying requests ``send()`` method (which is what was not happening previously due to the code not passing that argument).

I know another test case was also added in that PR, but that one verifies ConnectionTimeout whereas the one I added verifies ReadTimeout.

PR #1579, opened 2021-05-02
Use pytest for running integration tests

Kami:
Small change and attempt for integration tests to also use pytest for running the tests.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -74,4 +74,4 @@ jobs:
 
       - name: Run tox target
         run: |
-          tox -e integration-storage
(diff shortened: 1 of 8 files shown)


Kami:
I will also try to get external action (skip duplicate jobs) to work again by cloning 3d party action as part of the GHA workflow since ASF doesn't allow to directly utilize other external actions, but this cloning a specific revision seems to be accepted workaroud (https://github.com/apache/calcite/pull/2318/files).

And since GHA still doesn't allow us to only re-run specific jobs on failure, I will probably move integration tests in a separate workflow (this way we can re run just the unit tests / integration tests, depending on what has failed instead of everything).

@c-w wdyt?

PR #1580, opened 2021-05-07
Add integration tests for S3 storage driver

c-w:
### Description

This pull request adds integration tests for the S3 storage driver.

Given that there are existing buckets in the test AWS account, this pull request also updates the base tests to ignore any containers that weren't created as part of the test.

Unlike for Azure and MinIO, on S3, `create_container` for an already existing container seems to not throw `ContainerAlreadyExistsError` so the assertion has been disabled in this driver's test suite. In the future we may want to consider fixing all these minor inconsistencies that are brought to light by the tests.

### Status

- done, ready for review

### DIFF
--- a/integration/storage/base.py
+++ b/integration/storage/base.py
@@ -45,6 +45,9 @@ class Integration:
         account = None
         secret = None
 
+        container_name_prefix = 'lcsit'
+        container_name_max_length = 63
+
         def setUp(self):
             for required in 'provider', 'account', 'secret':
                 value = getattr(self, required, None)
@@ -63,6 +66,9 @@ class Integration:
 
         def tearDown(self):
             for container in self.driver.list_containers():
+                if not container.name.startswith(self.container_name_prefix):
+                    continue
+
                 for obj in container.list_objects():
                     try:
                         obj.delete()
@@ -86,7 +92,7 @@ class Integration:
 
         def test_containers(self):
             # make a new container
-            container_name = random_container_name()
+            container_name = self._random_container_name()
             container = self.driver.create_container(container_name)
             self.assertEqual(container.name, container_name)
             container = self.driver.get_container(container_name)
@@ -98,7 +104,7 @@ class Integration:
 
             # check that the new container can be listed
             containers = self.driver.list_containers()
-            self.assertEqual([c.name for c in containers], [container_name])
+            self.assertIn(container_name, [c.name for c in containers])
 
             # delete the container
             self.driver.delete_container(container)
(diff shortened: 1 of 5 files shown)


integration/storage/base.py
```diff
@@ -284,6 +292,16 @@ def _create_tempfile(self, prefix='', content=b''):
             self.addCleanup(os.remove, path)
             return path
 
+        @classmethod
+        def _random_container_name(cls):
+            suffix = random_string(cls.container_name_max_length)
+            name = cls.container_name_prefix + suffix
```
Kami, line 298: Yeah, I think including a special prefix on which we can filter on in the "clean up phase" is a good idea.

integration/storage/test_s3.py
```diff
@@ -0,0 +1,78 @@
...
+        for key, value in config.items():
+            if not value:
+                raise unittest.SkipTest('missing environment variable %s' % key)
+
+        cls.account = config['AWS_ACCESS_KEY_ID']
+        cls.secret = config['AWS_ACCESS_KEY_SECRET']
+
+    @classmethod
+    def tearDownClass(cls):
+        client = boto3.Session(
```
Kami, line 53: Just curious why we use boto3 for the clean up phase and not the Libcloud itself?

Kami (approved):

Kami:
I see the overall integration tests duration is ~14 minutes and ``test_objects_stream_iterable`` takes ~ 90 seconds.

Is there any way we could speed that up a bit? E.g. use a smaller object for that test or similar. Or we are intentionally testing large object size and we can't avoid slower duration?

Kami:
> Unlike for Azure and MinIO, on S3, create_container for an already existing container seems to not throw ContainerAlreadyExistsError so the assertion has been disabled in this driver's test suite. In the future we may want to consider fixing all these minor inconsistencies that are brought to light by the tests.

Yeah I agree, this seems like something we should eventually fix for consistency across drivers. Please feel free to open an issue for it if you get a chance.

On that note, what is the current behavior? It just returns an existing container with that name?

integration/storage/test_s3.py
c-w: I figured that if we have a bug/regression in libcloud, we don't want to keep stray resources around so might as well use the official library to perform the cleanup.

c-w:
> On that note, what is the current behavior? It just returns an existing container with that name?

Yes, `create_container` on an existing container returns a container with the same properties as the existing one. Here's a repro:

```py

cls = get_driver(Provider.S3)
driver = cls(AWS_ACCESS_KEY_ID, AWS_ACCESS_KEY_SECRET)

containers = driver.list_containers()
print(driver.create_container(containers[0].name).name == containers[0].name)
```

PR #1584, opened 2021-05-25
Optimized 'iterate_container_objects' by filtering the objects before…

Ido-Levi:
… creating the iterator

### Description
In the local driver - if a prefix is present, till now all of the objects in storage would've been serialized and just then filtered, changed it so the filtering would be beforehand.

### Status
done, ready for review

- [v] [Code linting](http://libcloud.readthedocs.org/en/latest/development.html#code-style-guide) (required, can be done after the PR checks)
- [v] Documentation
- [v] [Tests](http://libcloud.readthedocs.org/en/latest/testing.html)

### DIFF
--- a/libcloud/storage/drivers/local.py
+++ b/libcloud/storage/drivers/local.py
@@ -246,10 +246,11 @@ class LocalStorageDriver(StorageDriver):
 
         cpath = self.get_container_cdn_url(container, check=True)
 
+        fpath = cpath
         if prefix:
-            cpath = os.path.join(cpath, prefix)
+            fpath = os.path.join(cpath, prefix)
 
-        for folder, subfolders, files in os.walk(cpath, topdown=True):
+        for folder, subfolders, files in os.walk(fpath, topdown=True):
             # Remove unwanted subfolders
             for subf in IGNORE_FOLDERS:
                 if subf in subfolders:
@@ -279,8 +280,7 @@ class LocalStorageDriver(StorageDriver):
         """
         prefix = self._normalize_prefix_argument(prefix, ex_prefix)
 
-        objects = self._get_objects(container, prefix)
-        return self._filter_listed_container_objects(objects, prefix)
+        return self._get_objects(container, prefix)
 
     def get_container(self, container_name):
         """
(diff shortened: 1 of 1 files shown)


libcloud/storage/drivers/local.py
```diff
@@ -276,7 +279,7 @@ def iterate_container_objects(self, container, prefix=None,
         """
         prefix = self._normalize_prefix_argument(prefix, ex_prefix)
 
-        objects = self._get_objects(container)
+        objects = self._get_objects(container, prefix)
         return self._filter_listed_container_objects(objects, prefix)
```
c-w: If we now do the filtering natively in the driver I don't we need the fallback call to `_filter_listed_container_objects` anymore.

Ido-Levi: Aight, removed the call to `_filter_listed_container_objects`

c-w (approved):

louis-van-der-stam:
Hi, it seems I'm experiencing the same issue. Seeing the PR is open, will it go into a release? Are there still issues that need to be resolved? Can I help?

Kami (approved):

Kami (approved):

Kami:
Sorry for the delay - the change has now been merged.

louis-van-der-stam:
After upgrading to version 3.4.1 from version 3.3.1, the unit tests of my project broke! After diving into the code, it turns out that currently the local driver takes the prefix and appends it as a directory. Effectively this means that the prefix is not used as a file prefix anymore, but instead as a sub-directory name of the container main directory.

Looking a bit further, I was unable to find a os level function that does filtering when requesting a directory listing. Please unmerge this PR!

Kami:
@louis-van-der-stam Thanks for reporting this issue.

@Ido-Levi It looks like that the change was probably inadvertently backward incompatible and breaking (prefix filtering should apply to the object name and not to the sub-directory / bucket name). It sounds like we should revert it. What do you think?

Kami:
@louis-van-der-stam #1631 should fix this regression. Sadly we didn't have an existing test cases for this scenario which would have caught this issue early. Can you please verify the fix?

@Ido-Levi Thanks for your original contribution and good intention with the performance optimization, but sadly the change needs to be reverted since it breaks the expected behavior (prefix filtering being applied against the object name - which may include actual bucket name, but that's provider / implementation specific).

EDIT: Thinking about it some more, we may still be able to introduce optimization for some scenarios in the future (perhaps by checking if the name contains path separator or similar, but that would only apply to some scenarios and likely also has more edge cases which we need to handle correctly).

Issue #1577, opened 2021-04-27
Retry decorator on RateLimitExceeded issue


dimgal1:
## Summary

Enabling retry logic does not work for rate limit exceeded responses.

## Detailed Information

retry decorator expects a `RateLimitExceeded` error to be raised, though `session.request` method does not raise an exception if a response is returned by the server no matter the status code.

I've tested it on azure_arm driver by setting `RETRY_FAILED_HTTP_REQUESTS` to True and trying:
```python
while True:
    nodes = driver.list_nodes()
    print(nodes)
```

Hitting the throttling limit results in:

```python
  Traceback (most recent call last):
  File "test.py", line 23, in <module>
    nodes = driver.list_nodes()
  File "/home/dimgal/libcloud/libcloud/compute/drivers/azure_arm.py", line 395, in list_nodes
    r = self.connection.request(action,
  File "/home/dimgal/libcloud/libcloud/common/azure_arm.py", line 227, in request
    return super(AzureResourceManagementConnection, self) \
  File "/home/dimgal/libcloud/libcloud/common/base.py", line 655, in request
    response = responseCls(**kwargs)
  File "/home/dimgal/libcloud/libcloud/common/base.py", line 165, in __init__
    raise exception_from_message(code=self.status,
libcloud.common.exceptions.RateLimitReachedError: 429 Rate limit exceeded
```
urllib3 supports [retries](https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#urllib3.util.Retry) based on HTTP status codes, maybe use this? Let me know how you think this should be handled

RunOrVeith:
I implemented a pull request that goes into the direction of making this more customizable

PR #1588, opened 2021-05-31
Make it easier to customize retry logic

RunOrVeith:
### Description

The old implementation of retrying is fixed for all connection classes and is not customizable.
For example the exception types that are retried are not available as a parameter from within the connection, so
you had to live with what the default value is.
Unfortunately, that is not always enough, e.g. for S3 storage you can [receive 503 Slow Down Responses](https://aws.amazon.com/premiumsupport/knowledge-center/s3-troubleshoot-503-cross-region/) that should be retried with exponential back-off. Google Storage also has a similar mechanism.
These errors lead to exceptions at the moment.

This pull request takes a *first step* in fixing this by making the retry logic explicitly configurable within the connection.

If you deem this reasonable we can make the change to retry S3 slow down requests.
This requires differentiating HTTP exceptions by their status code, similar to what is mentioned in #1577.

In case you do not want to add more exceptions to retry, this change here would still enable end-users to customize the behavior without major hacks:

```python
from libcloud.utils.retry import MinimalRetry

class MyFancyRetry(MinimalRetry):

    def should_retry(self, exception):
        retry = ...  # some logic
        return retry


driver = get_driver(Provider.SOME_PROVIDER)(...)
driver.connection.retryCls = MyFancyRetry
```

I tried my best to be fully backward compatible and no actual behavior is changed yet.
If you are open to breaking changes we could also think about using urllib.Retry as suggested in #1577, and/or unifying the retry parameters that are passed around into some parameter class.

### Status

Done, but could be extended with some feedback

- [x] [Code linting](http://libcloud.readthedocs.org/en/latest/development.html#code-style-guide) (required, can be done after the PR checks) (done via commit hook)
- [x] Documentation
- [x] [Tests](http://libcloud.readthedocs.org/en/latest/testing.html) (locally)

### DIFF
--- a/libcloud/utils/retry.py
+++ b/libcloud/utils/retry.py
@@ -18,6 +18,7 @@ import ssl
 import time
 from datetime import datetime, timedelta
 from functools import wraps
+import logging
 
 from libcloud.utils.py3 import httplib
 from libcloud.common.exceptions import RateLimitReachedError
@@ -26,7 +27,7 @@ __all__ = [
     'Retry'
 ]
 
-
+_logger = logging.getLogger(__name__)
 # Error message which indicates a transient SSL error upon which request
 # can be retried
 TRANSIENT_SSL_ERROR = 'The read operation timed out'
@@ -42,6 +43,7 @@ class TransientSSLError(ssl.SSLError):
 DEFAULT_TIMEOUT = 30  # default retry timeout
 DEFAULT_DELAY = 1  # default sleep delay used in each iterator
 DEFAULT_BACKOFF = 1  # retry backup multiplier
+DEFAULT_MAX_RATE_LIMIT_RETRIES = float("inf")  # default max number of times to retry on rate limit
 RETRY_EXCEPTIONS = (RateLimitReachedError, socket.error, socket.gaierror,
                     httplib.NotConnected, httplib.ImproperConnectionState,
                     TransientSSLError)
@@ -50,15 +52,18 @@ RETRY_EXCEPTIONS = (RateLimitReachedError, socket.error, socket.gaierror,
 class MinimalRetry:
 
     def __init__(self, retry_delay=DEFAULT_DELAY,
-                 timeout=DEFAULT_TIMEOUT, backoff=DEFAULT_BACKOFF):
+                 timeout=DEFAULT_TIMEOUT, backoff=DEFAULT_BACKOFF,
+                 max_rate_limit_retries=DEFAULT_MAX_RATE_LIMIT_RETRIES):
         """
         Wrapper around retrying that helps to handle common transient exceptions.
         This minimalistic version only retries SSL errors and rate limiting.
-        :param retry_exceptions: types of exceptions to retry on.
 
         :param retry_delay: retry delay between the attempts.
         :param timeout: maximum time to wait.
         :param backoff: multiplier added to delay between attempts.
+        :param max_rate_limit_retries: The maximum number of retries to do when being rate limited by the server.
+                                       Set to `float("inf")` if retrying forever is desired.
+                                       Being rate limited does not count towards the timeout.
 
         :Example:
 
@@ -72,12 +77,15 @@ class MinimalRetry:
             timeout = DEFAULT_TIMEOUT
         if backoff is None:
             backoff = DEFAULT_BACKOFF
+        if max_rate_limit_retries is None:
+            max_rate_limit_retries = DEFAULT_MAX_RATE_LIMIT_RETRIES
 
         timeout = max(timeout, 0)
 
         self.retry_delay = retry_delay
         self.timeout = timeout
         self.backoff = backoff
+        self.max_rate_limit_retries = max_rate_limit_retries
 
     def __call__(self, func):
         def transform_ssl_error(function, *args, **kwargs):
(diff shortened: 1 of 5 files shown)


Kami:
Sorry for the delay - I'll try to look at this PR this weekend.

RunOrVeith:
No problem, any updates yet?

libcloud/utils/retry.py
```diff
@@ -0,0 +1,150 @@
...
+                try:
+                    return transform_ssl_error(func, *args, **kwargs)
+                except Exception as exc:
+                    if isinstance(exc, RateLimitReachedError):
+                        time.sleep(exc.retry_after)
+
+                        # Reset retries if we're told to wait due to rate
+                        # limiting
+                        current_delay = self.retry_delay
+                        end = datetime.now() + timedelta(
```
Kami, line 116: Should we have some kind of "ultimate" retry deadline timestamp / counter? Otherwise it looks like we could technically end up retrying for ever in case we are keep getting ``RateLimitReachedError`` error back, right?

While at it, it may also not hurt to add some debug statements around the retry code path. In Libcloud most actions (except deployment / SSH layer) are short running so we don't really have logging in most places, but here it might be useful.

Kami (approved):
Small comment about having another hard time / counter limit on how many times we retry on ``RateLimitReachedError`` error to prevent from potentially getting stuck in that loop for ever and a comment about debug logging.

Besides that, LGTM.

libcloud/utils/retry.py
RunOrVeith: You wont retry forever due to the timeout parameter. If the timeout (30 seconds by default) is exceeded while retrying, the retry is aborted.

RunOrVeith:
there is already a timeout parameter for hard timeouts.

libcloud/utils/retry.py
Kami: But we only perform ``datetime.now() >= end:`` check if an exception is not of type ``RateLimitReachedError``. In case we get ``RateLimitReachedError`` back, we set ``end`` to ``retry_after + timeout``.

So unless I'm misunderstanding this piece of code, we could end up retrying for ever or until we stop receiving ``RateLimitReachedError`` exception back from the API (if for some reason API will end up returning rate limit reached for a long time).

So unless I'm misunderstanding it, I believe we still need some modifications to avoid this edge case (e.g. maximum number of rate limit reached retries or similar - even if we move ``if datetime.now() >= end`` before the rate limit reached check, I believe that won't work as expected because we update ``end`` inside that if branch and we will never reach this condition in case we keep getting ``RateLimitReachedError`` back since we will be keep resetting end time).

Kami:
Since it's a bit larger change, can you please also look into submitting an ICLA - https://www.apache.org/licenses/icla.pdf.

libcloud/utils/retry.py
RunOrVeith: Yes you are right, I will make it so that retrying the rate limit will also count towards the timeout (maybe with a parameter to turn that behavior off).
I think if there is another "counter" parameter, it will be hard to get an intuitive behavior about how that interacts with the timeout parameter

Kami: Thanks and yeah, I agree - utilizing the existing ``timeout`` argument makes the most sense.

RunOrVeith:
I sent the ICLA out and will probably be able to add the fix for the missing final timeout on retries this week

Kami:
@RunOrVeith Sounds good, thanks.

RunOrVeith:
I implemented it with a counter now that defaults to infinite retries on being rate limited since that's how it was before. But you can turn it off now

Kami:
@RunOrVeith Thanks. Any specificreason why you ended up going with counter approach instead of utilizing the existing timeout parameters (which we both agree would make make sense and also be simpler)?

Kami:
@RunOrVeith And if you still agree that using timeout argument makes the most sense, I can adjust the code (+ add tests) to do that so we don't drag merging this any longer (since I would like to do a new release with all the accumulated changes as soon as possible).

Kami:
@RunOrVeith Here is a proposed change where we respect timeout argument - e86b374af0261d87f423147e40bb6d999edb7da1.

I personally think that's a better / less surprising API. If for some reason, there is a need for different handling of retry exceptions, I would tentatively be OK with a special case for that (counter), as long as it's an opt-in behavior.

I guess one issue with the timeout change is that it's different behavior from previous version, but I think it's better and safer to have a safe guard and not just default to retry for ever on rate limit reached errors.

If we do decide to go with timeout approach, I will document this in upgrade notes (and perhaps a feature flag / module level variable to revert to old behavior, if needed).

Let me know what you think.

EDIT: And another change I made is to default to 2 seconds sleep delay in case ``retry_after`` attribute is not case on the exception class (I can make it a module level constant / similar).

If retry-after attribute is not present in the headers, that value defaults to ``0``. This could be problematic, because in previous version of the code, we would not just try to retry for ever, but we could also be busy waiting (aka hogging all the CPU) if ``0`` was used for ``time.sleep()``

RunOrVeith:
I did it with the counter to not break the old behavior of retrying forever, but if a breaking change is fine then I agree that your suggestion is the preferred way.
There is just an outdated comment left in the code, otherwise it looks good to me.

Kami:
@RunOrVeith Thanks.

Yeah, that makes sense. Having said that - this code is quite old so it's probably about time to improve it a bit.

I will document this behavior changes in upgrade notes and add module level constant to revert to old behavior (if someone has a valid reason to do that).

Kami:
Here we go - 25475003a0b581aecf85f653b384316ec0a4f9f4.

While at it, I also refactored a bunch of tests since they used unncessary complex patterns which added two levels of indentation which made code harder to read - af7f4c4bbd3f8f89973e9ad286e9b0afdbc35ab0.

I also added test cases for bunch of scenarios we were missing (which did catch a regression in my change which I fixed).

Kami:
PR has been merged. Thanks.

PR #1596, opened 2021-07-27
EC2: Add support for gp3 and io2 volumes

palashgandhi:
### Description

The ec2 driver currently supports `standard`, `io1`, `gp2`, `st1` and `sc1` volume types. This change adds support for `gp3` and `io2` volume types. The driver currently maintains a list of volume types and raises a `ValueError` if an unsupported volume type is passed. The solution is to simply add the new types to this list. This change also enhances the `create_volume` unit test to verify that a `ValueError` is raised for a volume type that is not in the list.

### Status

- done, ready for review

- [x] Documentation

### DIFF
--- a/libcloud/compute/drivers/ec2.py
+++ b/libcloud/compute/drivers/ec2.py
@@ -1179,7 +1179,7 @@ VOLUME_MODIFICATION_ATTRIBUTE_MAP = {
 }
 
 VALID_EC2_REGIONS = REGION_DETAILS_PARTIAL.keys()
-VALID_VOLUME_TYPES = ['standard', 'io1', 'gp2', 'st1', 'sc1']
+VALID_VOLUME_TYPES = ['standard', 'io1', 'io2', 'gp2', 'gp3', 'st1', 'sc1']
 
 
 class EC2NodeLocation(NodeLocation):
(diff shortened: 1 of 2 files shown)


tonynguien (approved):
Thanks for making this change, Palash.

libcloud/compute/drivers/ec2.py
```diff
@@ -2157,6 +2163,9 @@ def create_volume(self, size, name, location=None, snapshot=None,
             if ex_kms_key_id is not None:
                 params['KmsKeyId'] = ex_kms_key_id
 
+        if ex_volume_type == 'gp3' and ex_throughput:
```
tonynguien, line 2166: I believe we want to allow caller to specify throughput for io1 and io2 volume types, similar to iops.

tonynguien (changes requested):
Thanks for making the change.

libcloud/compute/drivers/ec2.py
palashgandhi: According to the [EC2 API reference](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateVolume.html):

>  Throughput
> The throughput to provision for a volume, with a maximum of 1,000 MiB/s.
> This parameter is valid only for gp3 volumes. 

Are you seeing different behavior i.e. are you able to actually set the throughput for a volume type other than `gp3`? If so, the API reference might be outdated.

Regarding iops, my change already has the following logic that should allow you to set `Iops` for io1, io2 and gp3 volumes:
```
    if ex_volume_type in ['io1', 'io2', 'gp3'] and ex_iops:
            params['Iops'] = ex_iops
```

tonynguien: You're right. We can't set throughput for io1 and io2 volumes, i.e. AWS Console volume modify doesn't have a throughput change option.

tonynguien (approved):

Kami (approved):
Thanks for the contribution.

The change looks good to me.

Kami:
Merged, thanks.

PR #1602, opened 2021-09-14
fix linux command deployment when running libcloud on windows

atarbalouti:
### Description

When running libcloud on Windows to deploy a linux machine in the cloud with a ScriptDeployment, it had always promted me an error which was like "stderr: bash: /home/googlelibcloud_deployment_b20d6e1e.sh: No such file or directory". So, the path was not right formatted. The error was in the os.path.join method, which produces a false path on windows like "/home/google\\libcloud_deployment_b20d6e1e.sh". So I had to add the .replace("\\","/") method to the os.path.join.

### Status

- done, ready for review

- [ ] Documentation

### DIFF
--- a/libcloud/compute/deployment.py
+++ b/libcloud/compute/deployment.py
@@ -212,7 +212,7 @@ class ScriptDeployment(Deployment):
         if self.name and (self.name[0] not in ['/', '\\'] and
                           not re.match(r"^\w\:.*$", file_path)):
             base_path = os.path.dirname(file_path)
-            name = os.path.join(base_path, self.name)
+            name = os.path.join(base_path, self.name).replace("\\", "/")
         elif self.name and (self.name[0] == '\\' or
                             re.match(r"^\w\:.*$", file_path)):
             # Absolute Windows path
--- a/libcloud/compute/ssh.py
+++ b/libcloud/compute/ssh.py
@@ -688,7 +688,7 @@ class ParamikoSSHClient(BaseSSHClient):
                 # Windows path
                 file_path = cwd + '\\' + file_path
             else:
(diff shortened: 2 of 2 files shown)


libcloud/compute/deployment.py
```diff
@@ -212,7 +212,7 @@ def run(self, node, client):
         if self.name and (self.name[0] not in ['/', '\\'] and
                           not re.match(r"^\w\:.*$", file_path)):
             base_path = os.path.dirname(file_path)
-            name = os.path.join(base_path, self.name)
+            name = os.path.join(base_path, self.name).replace("\\", "/")
```
Kami, line 215: Hm, I believe this will also replace slashes on Linux, right?

If so, we should make sure this replace only runs when running deployment step on Windows target.

libcloud/compute/ssh.py
```diff
@@ -688,7 +688,7 @@ def _sanitize_file_path(self, cwd, file_path):
                 # Windows path
                 file_path = cwd + '\\' + file_path
             else:
-                file_path = pjoin(cwd, file_path)
+                file_path = pjoin(cwd, file_path).replace("\\", "/")
```
Kami, line 691: Same here, I believe this will also run on Linux target which will likely cause issues.

Kami:
Thanks for the contribution.

I added some comments - I believe those changes may also affect Linux targets which is likely not what we want. Ideally there would also be some unit tests to begin with.

I know we had some Windows target deployment related issues in the past, but I also know we have deployment stuff being used on Windows targets without any issues so it's likely some edge case which current code is not correctly handling.

Kami:
Closing due to inactivity. In case this issue is still relevant, feel free to re-open it.

PR #1616, opened 2021-10-22
Integration tests changes and fixes

Kami:
Just trying to figure out what is going on with integration tests and why they are failing / timing out.

Changes included in this PR:

- Add a deadline / timeout and fail the tests if container doesn't start and become ready in defined timeout
- Ensure containers are always killed / stopped on exit (sometimes it happened locally that for some reason ``tearDownClass`` was not called).
- Decrease default HTTP connection and request timeout

### DIFF
--- a/integration/storage/base.py
+++ b/integration/storage/base.py
@@ -24,6 +24,7 @@ import sys
 import tempfile
 import time
 import unittest
+import atexit
 
 import requests
 
@@ -32,8 +33,12 @@ try:
 except ImportError:
     docker = None
 
+import libcloud.http
+
 from libcloud.common.types import LibcloudError
 from libcloud.storage import providers, types
+
+libcloud.http.DEFAULT_REQUEST_TIMEOUT = 10
 
 
 MB = 1024 * 1024
@@ -317,6 +322,8 @@ class Integration:
         container = None
         verbose = False
 
+        container_ready_timeout = 30
+
         @classmethod
         def setUpClass(cls):
             if docker is None:
@@ -341,9 +348,17 @@ class Integration:
                 environment=cls.environment,
             )
 
+            # We register atexit handler to ensure container is always killed, even if for some
+            # reason tearDownClass is sometimes not called (happened locally a couple of times)
+
+            atexit.register(cls._kill_container)
+
             wait_for(cls.port, cls.host)
 
             container_ready = cls.ready_message is None
+
+            start_ts = int(time.time())
+            timeout_ts = start_ts + cls.container_ready_timeout
 
             while not container_ready:
                 time.sleep(1)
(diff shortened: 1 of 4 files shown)


Kami:
@c-w CI run didn't finish yet, but locally tests seem to be failing with:

```bash

________________________________________________________________ ERROR at setup of MinioTest.test_objects_stream_iterable _________________________________________________________________

cls = <class 'integration.storage.test_minio.MinioTest'>

>   ???
E   ValueError: Container fff2b0bdc3 failed to start and become ready in 30 seconds (did not find "IAM initialization complete" message in container logs).
E   
E   Container Logs:
E   API: http://172.17.0.2:9000  http://127.0.0.1:9000 
E   
E   Console: http://172.17.0.2:42801 http://127.0.0.1:42801 
E   
E   Documentation: https://docs.min.io
E   
E   WARNING: Console endpoint is listening on a dynamic port (42801), please use --console-address ":PORT" to choose a static port.
```

I didn't dig in much yet. Do you think that's related to the upstream change in the Docker image (in this case we should pin it to a known working tag and to that + update ready message) or something else?

EDIT: CI run completed, looks like it contains the same error - https://github.com/apache/libcloud/runs/3976839689?check_suite_focus=true.

Kami:
It looks like that the Docker image behavior has indeed changed and it appears 24c46851374ac7637238abca545f3d9cdd6960c7 will do the trick (at least it did the trick locally, still waiting for CI).

As part of this PR we now also have some timeouts in place which will hopefully make it easier and faster to troubleshoot timeout related issues (previously build was hanging for hours until default GHA job timeout kicked in and no step output was available).

Kami:
@c-w It looks like now tests start to fail with a different error:

```bash
E       azure.core.exceptions.ResourceExistsError: (MaxStorageAccountsCountPerSubscriptionExceeded) Subscription d6d608a6-e0c8-42ae-a548-2f41793709d2 already contains 251 storage accounts in location eastus2 and the maximum allowed is 250.
E       Code: MaxStorageAccountsCountPerSubscriptionExceeded
E       Message: Subscription d6d608a6-e0c8-42ae-a548-2f41793709d2 already contains 251 storage accounts in location eastus2 and the maximum allowed is 250.
```

Do we need some additional subscription clean up step or similar which runs before each test run?

c-w:
@Kami I'll take a look over the weekend.

Kami:
@c-w Thanks. For now, I went ahead and manually deleted storage accounts in the Azure Portal.

Going forward, as a improvement I propose adding a cleanup step which runs before each test run and also as part of a cron job (in case tests are non-gracefully terminated and tearDownClass cleanup step doesn't run).

To be on the safe side, we should probably prefix all the resources created by tests using ``libcloud-itests-`` or similar and only delete resources with that prefix (I know all the resources already use libcloud prefix, but some of the accounts already contain a bucket with libcloud name which shouldn't get deleted so this would potentially conflict).

Kami:
@c-w I think this should probably do it for now - fcc0d72730c94fdf55229dacf5971b3121064f13.

Will see how it behaves going forward.

PR #1617, opened 2021-10-22
Clean up old / stray Azure resources before the test run, speed up tests

Kami:
This pull request updates Azure Blobs integration test class so we clean any old stray resource groups left from previous test runs before we actually run the tests.

This should ensure we don't have any limits and we delete any stray resources from previous runs (we can end up with stray resources if the CI build is terminated non-gracefully and ``tearDownClass`` has no chance to run).

### DIFF
--- a/integration/storage/test_azure_blobs.py
+++ b/integration/storage/test_azure_blobs.py
@@ -19,6 +19,7 @@ import string
 import sys
 import unittest
 import time
+import datetime
 
 try:
     from azure import identity
@@ -137,20 +138,21 @@ class StorageTest(Integration.TestBase):
         # couple (6) of hours
         resource_groups = resource_client.resource_groups.list()
         now_ts = int(time.time())
-        delete_threshold_ts = now_ts - (6 * 60 * 60)
+        delete_threshold_ts = now_ts - int(datetime.timedelta(hours=6).total_seconds())
 
         for resource_group in resource_groups:
-            resource_create_ts = resource_groups.tags.get('create_ts', now_ts)
+            resource_create_ts = resource_group.tags.get('create_ts', now_ts)
+            resource_create_ts = int(resource_group.tags.get('create_ts',
+                                                             delete_threshold_ts - 100))
 
             if resource_group.name.startswith(name) and resource_group.location == location and \
-               'test' in resource_group.tags:
-               #'test' in resource_group.tags and resource_create_ts <= delete_threshold_ts:
+               'test' in resource_group.tags and resource_create_ts <= delete_threshold_ts:
                 print("Deleting old stray resource group: %s..." % (resource_group.name))
 
                 try:
                     resource_client.resource_groups.begin_delete(resource_group.name)
                 except Exception as e:
-                    print("Failed to delete resource group: %s" % (str(e)))
+                    print("Failed to delete resource group: %s" % (str(e)), file=sys.stderr)
 
         group = resource_client.resource_groups.create_or_update(
             resource_group_name=name,
@@ -158,7 +160,7 @@ class StorageTest(Integration.TestBase):
                 location=location,
                 tags={
                     'test': cls.__name__,
-                    'create_ts': now_ts,
+                    'create_ts': str(now_ts),
                     'run': os.getenv('GITHUB_RUN_ID', '-'),
                 },
             ),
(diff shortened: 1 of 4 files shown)


integration/storage/test_azure_blobs.py
```diff
@@ -128,12 +129,35 @@ def setUpClass(cls):
...
         timeout = float(os.getenv('AZURE_TIMEOUT_SECONDS', DEFAULT_TIMEOUT_SECONDS))
 
+        # We clean up any left over resource groups from previous runs on setUpClass. If tests on
+        # CI get terminated non-gracefully, old resources will be left laying around and we want
+        # to clean those up to ensure we dont hit any limits.
+        # To avoid deleting groups from concurrent runs, we only delete resources older than a
+        # couple (6) of hours
+        resource_groups = resource_client.resource_groups.list()
+        now_ts = int(time.time())
+        delete_threshold_ts = now_ts - (6 * 60 * 60)
```
c-w: Nit: this may be easier to read using `from datetime import timedelta; timedelta(hours=60).total_seconds()`

integration/storage/test_azure_blobs.py
```diff
@@ -128,12 +129,35 @@ def setUpClass(cls):
...
+
+        for resource_group in resource_groups:
+            resource_create_ts = resource_groups.tags.get('create_ts', now_ts)
+
+            if resource_group.name.startswith(name) and resource_group.location == location and \
+               'test' in resource_groups.tags and resource_create_ts <= delete_threshold_ts:
+               #'test' in resource_groups.tags and resource_create_ts <= delete_threshold_ts:
```
c-w: Nit: remove comment before merge.

integration/storage/test_azure_blobs.py
```diff
@@ -128,12 +129,35 @@ def setUpClass(cls):
...
+                print("Deleting old stray resource group: %s..." % (resource_group.name))
+
+                try:
+                    resource_client.resource_groups.begin_delete(resource_group.name)
+                except Exception as e:
+                    print("Failed to delete resource group: %s" % (str(e)))
```
c-w: Nit: suggest to log to stderr so it gets highlighted in the logs: `from sys import stderr; print("Failed to delete resource group: %s" % (str(e)), file=stderr)`

c-w (approved):

Kami:
@c-w Thanks for the review.

I pushed some more changes and fixes and I believe it should be working now. I also reverted the testing change to allow for concurrent runs.

integration/storage/test_azure_blobs.py
```diff
@@ -124,17 +128,46 @@ def setUpClass(cls):
...
+        now_ts = int(time.time())
+        delete_threshold_ts = now_ts - int(datetime.timedelta(hours=6).total_seconds())
+
+        for resource_group in resource_groups:
+            resource_create_ts = resource_group.tags.get('create_ts', now_ts)
+
+            if resource_group.name.startswith(RESOURCE_GROUP_NAME_PREFIX) and \
+               resource_group.location.lower() == location.lower() and \
+               'test' in resource_group.tags and resource_create_ts <= delete_threshold_ts:
+                assert resource_group.name.startswith(RESOURCE_GROUP_NAME_PREFIX)
```
c-w, line 152: Is this necessary given that we have the check on line 149 which means that this assert should always pass?

c-w (approved):

integration/storage/test_azure_blobs.py
Kami: It's jut an additional safe guard in case there was a fat finger in if or similar which could potentially result in objects with a different prefix to be deleted (granted, it's probably an over kill since we have additional checks and right now that account is only used for those tests).

Kami:
I noticed some test cases took almost 100 seconds, so I pushed a change which should speed it up - d31e3ee65768dbaa05bf85f85feb90c20f64749b, bbcaa2748fa96c607b6ac8252aff415ac30facbe.

This will substantially reduce number of iterations (previously for some test cases we used 1 * 1024 * 1024 iterations which will be slow), but there will still be multiple ones so that iteration / chunk upload functionality will still be tested.

Kami:
https://github.com/apache/libcloud/runs/3980290729?check_suite_focus=true

Looks like the integration tests run with that change now takes ~1 minute instead of ~8 minutes.

PR #1618, opened 2021-10-24
Fix publish pricing CI job

Kami:
Just trying to figure out why this job suddenly started failing on CI, it works fine locally (perhaps AWS blocked GHA IPs or similar?).


Kami:
Looks like the issue is related to using older version of demjson and 2to3 during setup phase not working anymore with new version of setuptools.

I'm trying to find a working alternative to demjson for parsing Javascript objects, it appears jsonnet may work.

Issue #1591, opened 2021-06-10
Signing in with OAuth is unnecessarily slow when enabeling retries


RunOrVeith:
## Summary

When you create a google driver with OAuth2 credentials, there is code to determine which auth type is being used.
This checks if it can reach the metadata server from Google Compute Engine to see if it is some kind of GCE login.
When using OAuth credentials, this will always fail. If you have retries enabled it will non the less retry with exponential back off until the timeout is reached and then proceed to say that OAuth is being used. 

## Detailed Information

The following snippet reproduces the issue on libcloud 3.3.1 (Ubuntu 18.04, Python 3.8)

```python
import os
from libcloud.storage.providers import Provider, get_driver


def demonstrate_slow_oauth_login(client_id: str, secret: str):
    os.environ["LIBCLOUD_RETRY_FAILED_HTTP_REQUESTS"] = "1"
    driver_type = get_driver(provider=Provider.GOOGLE_STORAGE)
    driver = driver_type(key=client_id, secret=secret)
```
What now happens is this:

1. Up in the call chain of `__init__` when the driver gets created, the `BaseDriver.__init__` creates a connection. In this case it is `GoogleStorageConnection`.
2. The `GoogleStorageConnection.__init__` calls this:
```python
    @classmethod
    def guess_type(cls, user_id):
        if cls._is_sa(user_id):
            return cls.SA
        elif cls._is_gcs_s3(user_id):
            return cls.GCS_S3
        elif cls._is_gce():  # <-- This slows everything down
            return cls.GCE
        else:
            return cls.IA   # <-- we just want to go here with OAuth
```
3. The check for `_is_gce` works as follows:
```python
def _get_gce_metadata(path=''):
    try:
        url = 'http://metadata/computeMetadata/v1/' + path.lstrip('/')
        headers = {'Metadata-Flavor': 'Google'}
        response = get_response_object(url, headers=headers)
        return response.status, '', response.body
    except Exception as e:
        return -1, str(e), None
```
Will end up making a request with retries (inside `Connection.request`):
```python
retry_request = retry(timeout=self.timeout,
                      retry_delay=self.retry_delay,
                      backoff=self.backoff)
retry_request(self.connection.request)(method=method,
                                       url=url,
                                       body=data,
                                       headers=headers,
                                       stream=stream)
```

4. Calling this when not using gce always results in an error: 
```
[...]

PR #1621, opened 2021-10-25
[Google] Don't retry failed requests on auth when determining if GCE metadata server is available

Kami:
This pull request should help with an edge case / issue reported in #1591.

## Background, context

Right now if retrying failed HTTP requests is enabled globally (either via module level constant or via environment variable value), Libcloud will also try to retry a request inside ``_is_gce()`` on driver instantiation / authentication where we try to determine if GCE metadata server is available and what kind of auth credentials are provided by the user.

This is not desired when GCE metadata server is not available and a different type of credentials are used (e.g. oAuth 2).

## Proposed Approach

In this PR, I simply disable retrying failed requests for that code piece where we determine if metadata server is available.

I think this should be sufficient, but if needed / desired, we can add another argument to the driver constructor which would cause driver to skip this step all together (e.g. ``ex_don_check_gce_metadata_server``).

## TODO

- [x] Unit tests

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,19 @@ Common
 
 - Update setup.py metadata and indicate we also support Python 3.10.
 
+- [Google] Update Google authentication code so we don't try to retry failed
+  request when trying to determine if GCE metadata server is available when
+  retrying is enabled globally (either via module level constant or via
+  environment variable value).
+
+  This will speed up scenarios when trying is enabled globally, but GCE
+  metadata server is not available and different type of credentials are used
+  (e.g. oAuth 2).
+
+  (GITHUB-1591)
+
+  Reported by Veith Röthlingshöfer - @RunOrVeith.
+
 Storage
 ~~~~~~~
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,14 @@ Common
 
 - Update setup.py metadata and indicate we also support Python 3.10.
 
+- [Google] Update Google authentication code so so we don't try to contact
+  GCE metadata server when determining auth credentials type when oAuth 2.0 /
+  installed app type of credentials are used.
+
+  (GITHUB-1591, GITHUB-1621)
+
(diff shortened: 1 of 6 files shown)


RunOrVeith:
sounds reasonable to me, but I actually think we could do something that doesn't require interaction.
According to [this documentation](https://developers.google.com/identity/sign-in/web/server-side-flow) Step 3, the IA user_id for the guess_type method always ends with "apps.googleusercontent.com",

so I would propose something like 
```python

class GoogleAuthType:
    ...
 
    @classmethod
    def guess_type(cls, user_id):
         ...
         elif cls._is_installed_application(user_id):
               return cls.IA
        elif cls._is_gce():
              # it's important that this is the last check, because it may take a long time if it fails and is retrying is enabled
              return cls.GCE
        else:
              # Either rise an error (not sure which one) or fall back to cls.IA again
              raise RuntimeError("Could not guess authentication scheme")

    @staticmethod
    def _is_installed_application(user_id):
        return user_id.endswith("apps.googleusercontent.com")

```

I tried it with an instance were I want to authenticate with IA and the ending is there in the user_id

Kami:
@RunOrVeith Yeah, if that's the case and we can avoid that branch all together, that would be ideal.

I will dig into the docs and into making that change.

Kami:
Re throwing exception in the else clause - I think in general that's a good idea (fail fast and loudly), but I'm slightly worried this may break some edge case since right now we fall back to ``cls.IA``.

So perhaps to be on the safe side, it's the best to leave ``else`` clause as is for now.

RunOrVeith:
yes that was also my concern with the exception

Kami:
@RunOrVeith This should probably do it - 0b8b22307526c185ebc372cd4d1cc5543eedbb6d.

To be on the safe side, I left else clause as it was before.

RunOrVeith:
Looks good, except that the CI complains about the unsanitized url. Not sure how to fix that though.

Kami:
@RunOrVeith Thanks for the review. I already marked that as "non an issue" in the Github UI.

PR #1625, opened 2021-11-06
Run unit tests in parallel

Kami:
This PR attempts to run unit tests in parallel using ``pytest-xdist`` plugin.

To avoid issues with some of the test functions not being "cross process safe" we run all the tests in a single test file in the same worker process.

Locally (with 16 cores) unit tests now finish ~5-10x faster, but it depends on number of available cores.


Kami:
This change uncovered some issues with some test files relying on shared state (aka when running all the test cases together in the same process some other import unintentionally setting some state which test relies on) which I'm trying to fix.

Kami:
On CI, tests now finish 30-40 seconds faster (2 min 20 vs 1 min 40). Not a massive speed up, but GHA runner likely doesn't have as many cores and it's likely more heavily over provisioned with other jobs.

Issue #1594, opened 2021-07-11
Migrate to the next version of Python requests when released


potiuk:
Hello Maintainers, 

I am a PMC member of Apache Airflow, and I wanted to give you a bit of heads-up with rather important migration to the upcoming version of requests library in your Python release. 

Since you are using requests library in your project, you are affected.

As discussed at length in https://issues.apache.org/jira/browse/LEGAL-572 we found out that the `chardet` library used by `requests` library was a mandatory dependency to requests and since it has LGPL licence, we should not release any Apache Software with it. 

Since then (and since in Airflow we rely on requests heavily) we have been working with the requests maintainers and "charset-normalizer" maintainer to make it possible to replace chardet with MIT-licensed `charset-normalizer` instead so that requests library can be used in Python releases by Apache projects.

This was a bumpy road but finally the PR by @ashb has been merged: https://github.com/psf/requests/pull/5797 and we hope soon a new version of requests library will be released. 

This is just a heads-up. I will let you know when it is released, but I have a kind requests as well - I might ask the maintainers to release a release candidate of requests and maybe you could help to test it before it is released, that would be some re-assurance for the maintainers of requests who are very concerned about stability of their releases.

Let me know if you need any more information and whether you would like to help in testing the candidate when it is out.

potiuk:
Requests 2.26.0 released without the mandatory chardet dependency:

https://pypi.org/project/requests/2.26.0/

Instead of chardet, use the MIT-licensed charset_normalizer for Python3 to remove license ambiguity for projects bundling requests. If chardet is already installed on your machine it will be used instead of charset_normalizer to keep backwards compatibility. (#5797)

You can also install chardet while installing requests by specifying [use_chardet_on_py3] extra as follows:

shell pip install "requests[use_chardet_on_py3]" 

Python2 still depends upon the chardet module.

Kami:
Sorry for the delay and thank for notifying us about this issue / change.

I will look into updating setup.py metadata so we require ``requests>=2.26.0`` and including that change in the next release (I will bump a minor version to be on the safe side).

We don't support Python 2 anymore so hopefully there won't be any problems, but I will also go over requests changelog to make sure we avoid any possible surprises.

Kami:
Looks like latest version if requests also dropped support for Python 3.5 which means this change will be more involved.

For now, I will add a conditional to setup.py to use older version of requests under Python 3.5, but in the future we should probably also consider dropping support for Python 3.5 to simplify things.

[...]

PR #1627, opened 2021-11-11
[v3.4.x] Fix installation under Python 3.5, don't use strict requests version pinning to avoid issues with conflicting dependency version requirements

Kami:
v3.4.0 release introduced a setup.py metadata regression which brakes installation under Python 3.5 when using pip.

Since that release should still support Python 3.5 we need to fix the issue and release v3.4.1 bug fix release.

This pull request fixes the setup.py metadata issue. Sadly none of the existing install checks caught the issue (and it's basically impossible for them to catch it - to catch the issue we would need to build sdist file under Python >= 3.6 and then install it under Python 3.5).


Kami:
I tested and confirmed the fix using the following approach:

```bash
# Create distribution under Python 3.py
python setup.py dist
# Create Python 3.5 venv and try to install previously created dist
rm -rf py35-test/ ; python3.5 -m venv py35-test ; source py35-test/bin/activate
pip install dist-v340/apache-libcloud-3.4.0.tar.gz
```

To be double sure it indeed fixes the issue, I will also upload it to test PyPi server and test installing it from there using Python 3.5.

Kami:
Tested it with Test PyPi server and it seems to be picking up and trying to install correct version of ``requests`` library now (but installation fails since that version is not available on test PyPi server.

```bash
$ python3.5 -m pip install --index-url https://test.pypi.org/simple/ apache-libcloud
Requirement already satisfied: apache-libcloud in /home/kami/w/libcloud/libcloud
Collecting requests>=2.25.1 (from apache-libcloud)
  Could not find a version that satisfies the requirement requests>=2.25.1 (from apache-libcloud) (from versions: 2.5.4.1)
No matching distribution found for requests>=2.25.1 (from apache-libcloud)
```

Previously it would incorrectly try to use >= 2.26.0 which won't work under Python 3.5

Kami:
Technically, since Libcloud is a library and not an application it may actually be the best to simply unpin requests.

I know that is not ideal because of #1594, but it's probably the safest since this >= 2.26.0 actually has potential to brake more user installations in case they already require conflicting version of requests.

So for now I will actually do just that, so we don't need to do another bug fix release in the near future when more people start reporting it's causing issues in their environment due to the conflicting dependency version.

PR #1631, opened 2021-12-01
Fix object name prefix based filtering in the local storage driver

Kami:
This pull request fixes object name prefix based filtering which was accidentally broken by the change in #1584.

It was reported by a user (https://github.com/apache/libcloud/pull/1584#issuecomment-983712951), that the change breaks the behavior.

I confirmed that myself by adding a test case which was failing with changes from #1584, but should pass if the behavior was correct - prefix based filtering is based on the object name and not on sub-directory names so optimization from #1584 is not a good idea (it's true though that in some cases, bucket / sub-directory name is part of the object name, but that's an implementation detail).

It achieves that by reverting changes from #1584.

Sadly we didn't have good existing tests which would have caught this issue early, but I added some regression tests to help us in the future.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -203,6 +203,48 @@ jobs:
         run: |
           script -e -c "tox -e black-check,checks,import-timings,lint,pylint"
 
+  micro-benchmarks:
+    name: Micro Benchmarks
+    runs-on: ubuntu-latest
+
+    needs: pre_job
+    if: ${{ needs.pre_job.outputs.should_skip == 'false' || github.ref == 'refs/heads/trunk' }}
+
+    strategy:
+      matrix:
+        python_version: [3.7]
+
+    steps:
+      - uses: actions/checkout@master
+        with:
+          fetch-depth: 1
+
+      - name: Use Python ${{ matrix.python_version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python_version }}
+
+      - name: Install OS / deb dependencies
+        run: |
+          sudo DEBIAN_FRONTEND=noninteractive apt-get update
+          sudo DEBIAN_FRONTEND=noninteractive apt-get install -yq graphviz gcc libvirt-dev
+
+      - name: Cache Python Dependencies
+        uses: actions/cache@v2
+        with:
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-pip-${{ hashFiles('requirements-tests.txt') }}
+          restore-keys: |
+            ${{ runner.os }}-pip-
+
+      - name: Install Python Dependencies
+        run: |
+          pip install "tox==3.24.4"
+
+      - name: Run Micro Benchmarks
+        run: |
+          script -e -c "tox -e micro-benchmarks"
+
   docs:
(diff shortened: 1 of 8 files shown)


Kami:
@louis-van-der-stam I added another somewhat related change which makes the output of ``list_container_objects`` sorted based on the object name which should result in consistent and stable order across Python versions and implementations - 6a24f1e5941c10535e8dad7fa2def92bc92ee5de, 4cdf818a5a501c6dd06c8491b8cbbee3d316c4d5.

Do you think that change could cause issues?

I mean, it's possible that some code or change which relied on the previous order would break, but before this change, the order wasn't guaranteed anyway and was "arbitrary" and could be different across different Python versions and implementations.

From Python docs on ``os.listdir`` on which ``os.walk`` relies on:

> The list is in arbitrary order. It does not include the special entries '.' and '..' even if they are present in the directory.

louis-van-der-stam:
@Kami I don't think sorting impacts the functionality, but can imagine that it impacts the runtime of the call.

In my case there are huge and ever increasing numbers of objects. My case has no dependency on order -> my preference would be to leave sorting of the results to the caller.

louis-van-der-stam:
@Kami Thanks for your quick response and action!

Kami:
@louis-van-der-stam How many objects we are talking about? Should be pretty straight forward to write a micro-benchmark to see how much overhead it adds and if that overhead is negligible or not (I can look into it a bit later).

louis-van-der-stam:
> @louis-van-der-stam How many objects we are talking about? Should be pretty straight forward to write a micro-benchmark to see how much overhead it adds and if that overhead is negligible or not (I can look into it a bit later).

Right now 150000 objects, adding several 1000s every day. To be honnest since it seems at the os level there is no way arround fetching all and matching for a prefix, I'm thinking about caching the results in a container wrapper...

Kami:
@louis-van-der-stam Thanks for the info. I will add some micro-benchmarks when I get a chance, just to get some idea on the sorting overhead.

But yeah, local storage driver was never built with performance in mind - it was designed and built for testing purposes.

Having said that, per my comment [here](https://github.com/apache/libcloud/pull/1584#issuecomment-983881324), I do think there are some possible performance optimizations for some scenarios. 

In short, something similar to what was originally done in #1584, but making sure all the edge cases are handled correctly and only filtering by sub-directory if the actual prefix maps to a sub-directory and not to an object.

But yeah, in the end, we are limited by the OS primitives available and we will still need to do some late filtering in the Python code no matter what (depending on what you are working on and your use case, inotify based approach may offer advantages, but that's of course out of scope of Libcloud).

Kami:
OK, I had a look and the approach I suggested above won't work so we are stuck with the current implementation and there isn't tons we can do to improve it.

Kami:
@louis-van-der-stam I added a simple benchmark which measures how much overhead sorting adds to that method - 230a094ab72ec0a6f8da76c5f31259240c034ae5.

Screenshot from 2021-12-02 19-39-23

Per results, the overhead is pretty much negligible so I think it's fine to leave that code in place.

louis-van-der-stam:
Thanks for all your work!

PR #1633, opened 2021-12-05
Scaleway object storage driver

reixd:
### Description

This PR adds the capability to connect to Scaleway's Object Storage using libcloud.

### Status

- done, ready for review

- [X] Documentation


Kami:
Thank you for the contribution. Can you please work on getting an ICLA signed - libcloud.readthedocs.org/en/latest/development.html#contributing-bigger-changes?

libcloud/test/storage/test_scaleway.py
```diff
@@ -0,0 +1,56 @@
...
+import unittest
+
+from libcloud.common.types import LibcloudError
+from libcloud.storage.drivers.scaleway import ScalewayStorageDriver
+from libcloud.storage.drivers.scaleway import ScalewayConnectionAWS4
+
+from libcloud.test.secrets import STORAGE_S3_PARAMS
+
+
+class ScalewayStorageDriverTestCase(unittest.TestCase):
```
Kami: Should inherit base s3 tests class here so it verifies that all the base S3 driver related functionality works as expected. IIRC, some existing drivers such as one of the Aurora objects already do that (libcloud/test/storage/test_aurora.py).

libcloud/storage/drivers/scaleway.py
```diff
@@ -0,0 +1,92 @@
...
+
+        super(ScalewayStorageDriver, self).__init__(key=key,
+                                                 secret=secret,
+                                                 secure=secure,
+                                                 host=host,
+                                                 port=port,
+                                                 url=url)
+
+
+class ScalewayFRParConnection(ScalewayConnectionAWS4):
```
Kami: Single driver with "region" argument for the region should be enough.

We moved away from multiple classes to a single driver class + ``region`` constructor argument quite a while ago. There should be some examples on how to do that in S3 and other drivers.

libcloud/storage/drivers/scaleway.py
```diff
@@ -0,0 +1,92 @@
...
+__all__ = [
+    "ScalewayStorageDriver"
+]
+
+
+class ScalewayConnectionAWS4(SignedAWSConnection, BaseS3Connection):
+    service_name = 's3'
+    version = API_VERSION
+
+    def __init__(self, user_id, key, secure=True, host=None, port=None,
```
Kami: I think it may be enough to inherit from ``S3SignatureV4Connection`` class, no need to duplicate the constructor and override signature version there.

Kami:
Thanks again for the contribution. I added some review comments in-line.

libcloud/storage/providers.py
```diff
@@ -95,6 +95,10 @@
         "DigitalOceanSpacesStorageDriver",
     ),
     Provider.MINIO: ("libcloud.storage.drivers.minio", "MinIOStorageDriver"),
+    Provider.SCALEWAY: ("libcloud.storage.drivers.scaleway", "ScalewayStorageDriver"),
+    Provider.SCALEWAY_FR_PAR: ("libcloud.storage.drivers.scaleway", "ScalewayFRParStorageDriver"),
```
Kami: Same as above, those region specific constants should be removed and the main class should take ``region`` constructor argument.

libcloud/test/storage/test_scaleway.py
reixd: Wilco

reixd:
Thank you for the review, will make the changes in the next days.
I will also sign the ICLA.

reixd:
@Kami 
* ICLA signed and sent per mail
* Suggestions worked out and commited

I hope the test "suite" is enough ;)

docs/examples/storage/scaleway/upload_example.py
```diff
@@ -0,0 +1,16 @@
+from libcloud.storage.types import Provider
+from libcloud.storage.providers import get_driver
+
+FILE_PATH = "/home/user/myfile.tar.gz"
+
+cls = get_driver(Provider.SCALEWAY)
+driver = cls("api key", "api secret key", host="s3.fr-par.scw.cloud")
```
Kami: I believe this example should be updated so a ``region`` instead of ``host`` argument is specified, right?

libcloud/storage/drivers/scaleway.py
```diff
@@ -0,0 +1,86 @@
...
+            url=url,
+            **kwargs,
+        )
+
+    @classmethod
+    def list_regions(self):
+        return REGION_TO_HOST_MAP.keys()
+
+    def get_object_cdn_url(self, *argv):
+        raise LibcloudError(NO_CDN_SUPPORT_ERROR, driver=self)
```
Kami: May want to throw ``NotImplementedError`` instead.

reixd:
@Kami ICLA is through. 
Are we free to go? ;)

Kami:
@reixd Thanks for the reminder - I will have a look as soon as I get a chance :)

libcloud/common/aws.py
```diff
@@ -347,14 +347,15 @@ def _get_credential_scope(self, dt):
         )
 
     def _get_signed_headers(self, headers):
-        return ";".join([k.lower() for k in sorted(headers.keys())])
+        return ";".join([k.lower()
```
Kami: I would like to understand reasoning behind this and make sure it doesn't break the generated signature. Can you please provide more context and confirm that?

libcloud/storage/drivers/scaleway.py
```diff
@@ -0,0 +1,87 @@
...
+    region_name = "fr-par"
+
+    def __init__(
+        self,
+        key,
+        secret=None,
+        secure=True,
+        host=None,
+        port=None,
+        region=None,
```
Kami: Please also default ``region`` keyword argument to ``fr-par`` here.

libcloud/storage/types.py
```diff
@@ -106,6 +106,10 @@ class Provider(object):
     S3_RGW = "s3_rgw"
     S3_RGW_OUTSCALE = "s3_rgw_outscale"
     MINIO = "minio"
+    SCALEWAY = "scaleway"
+    SCALEWAY_FR_PAR = "scaleway_fr_par"
```
Kami: Please remove region specific constants and leave just ``SCALEWAY`` here - as discussed, we moved away from region specific constants and old ones are here for backward compatibility reasons.

Kami:
@reixd Added a couple of more small comments and approved GHA run. Can you please also fix failing lint GHA job?

libcloud/common/aws.py
reixd: Lets ask @murlock 
Apparently without this change, the S3 Storage in Scaleway does not work.

libcloud/storage/drivers/scaleway.py
reixd: good catch.

libcloud/storage/types.py
reixd: It slipped away...

libcloud/common/aws.py
Kami: I would hope we have tests for various edge cases for those signature classes, but I would need to dig in.

But yeah, the docs do seem to claim the headers need to be lowecase - https://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html.

Would be good to add some test cases which confirm that.

I'll also quickly test this change with S3 when I get a chance, just to confirm this change doesn't break something.

Kami: I did a very quick and non extensive test with the S3 driver with those code changes and it seems to work fine.

But yeah, I didn't really dig in much.

murlock: Hi everyone
We use only lower case header name:
The data to check signature are setup at https://github.com/openstack-archive/swift3/blob/v1.11/swift3/request.py#L251
and since there is mismatch while preparing data at https://github.com/openstack-archive/swift3/blob/53614c2da837a9a53016b7d3e9911cf96ab39c77/swift3/request.py#L255 a SignatureMismatch is raised

reixd:
> @reixd Added a couple of more small comments and approved GHA run. Can you please also fix failing lint GHA job?

I think I solved the issues. Please start the tests again

Kami:
@reixd Sorry for the delay, I missed your last comment.

LGTM, I will go ahead and merge your change. Thanks again for your contribution!

Issue #1636, opened 2021-12-10
OpenStack: Move floating IP functions to use network service instead of nova


micafer:
## OpenStack: Move floating IP functions to use network service instead of nova

All the Network calls are deprecated and will fail in future versions:
```
These APIs are proxy calls to the Network service. Nova has deprecated all the proxy APIs and users should use the native 
APIs instead. These will fail with a 404 starting from microversion 2.36. See: Relevant Network APIs.
```
https://docs.openstack.org/api-ref/network/v2/?expanded=list-floating-ips-detail#floating-ips-floatingips

PR #1637, opened 2021-12-10
Avoid raising exception if ip is not found: Fix #1595

### LINKED ISSUE #1595: OpenStack_1_1_NodeDriver.ex_get_floating_ip raises exception when not found
## Summary

Calling ex_get_floating_ip with an IP not in the list raises an exception, but feels like it should return None instead to be consistent w/ ex_get_node_details.

Honestly, not sure this is a bug.  Seems like there are a lot of methods in this driver that will throw an exception if the requested resource is not found.  Figured I'd raise this anyway so we can discuss.  I wound up wrapping this method in my own that has the behavior I want, and that feels wrong.

## Detailed Information

Using apache-libcloud 3.3.1.

```
>>> d = get_driver('openstack', ...)
>>> d.ex_get_floating_ip('1.2.3.4')
  File "/home/vagrant/source/libcloud-drivers/.venv/lib/python3.8/site-packages/libcloud/compute/drivers/openstack.py", line 2580, in ex_get_floating_ip
    ip_obj, = [x for x in floating_ips if x.ip_address == ip]
ValueError: not enough values to unpack (expected 1, got 0)
```

micafer:
### Description

See: #1595.
I also fix similar issues in other functions.
There are no tests for this functions.
I will not add them as this functions are deprecated (see #1636).

### Status

- done, ready for review

- [ ] Documentation

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -25,6 +25,10 @@ Compute
   (GITHUB-1629)
   [Miguel Caballer - @micafer]
 
+- [OpenStack] Avoid raising exception if ip is not found.
+  (GITHUB-1595)
+  [Miguel Caballer - @micafer]
+
 Storage
 ~~~~~~~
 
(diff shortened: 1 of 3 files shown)


Kami (approved):
LGTM, but please add a changelog entry.

PR #1638, opened 2021-12-13
OpenStack: Move floating IP functions to use network service instead of nova

micafer:
### Description

See #1636.
Also some code improvements have been made.

### Status

- done, ready for review

- [ ] Documentation


Kami:
It would be good if you could also add an upgrade notes entry on this change with a short example of how the updated / new code looks like.

libcloud/compute/drivers/openstack.py
```diff
@@ -4441,6 +4510,74 @@ def __repr__(self):
...
+    def __init__(self, id, ip_address, pool, node_id=None, driver=None, extra=None):
+        self.id = str(id)
+        self.ip_address = ip_address
+        self._pool = pool
+        self._node_id = node_id
+        self.driver = driver
+        self.extra = extra if extra else {}
+
+    @property
+    def pool(self):
```
Kami: Hm, looks like this method will always try to perform the API call (``ex_get_network()``) on subsequent calls even if the code throws an exception and ``self._pool`` is not populated.

It looks like it's similar story with ``node_id`` property.

I wonder if there is something we could to improve that to avoid surprises - if it was regular method, I would be fine with it, but with a variable / property, this call time overhead is kinda hidden and could be substantial in case operation which is triggered when ``self._pool`` is not populated yet gets called on each property access because that internal variable wasn't populated for some reason (exception or similar).

I guess one simple option would be to simplify set that variable to None on error and return that in case of an error (and use some other special marker value for "not set"). That would only really work though if the underlying operation is synchronous and there is no chance for network errors or other similar exceptions.

Kami: Another even simpler option may be to not use a property but a method named ``get_pool()`` and ``get_node_id()`` - this makes it more obvious that the method call likely has a side affect.

Kami (approved):
I have some concerns / questions on the property potentially having expensive side affects on each property access during some edge cases, besides that, LGTM.

libcloud/compute/drivers/openstack.py
micafer: Finally I used simpler option.

libcloud/compute/drivers/openstack.py
```diff
@@ -4330,6 +4331,74 @@ def ex_del_server_group(self, server_group):
...
+    def ex_create_floating_ip(self, ip_pool):
+        """
+        Create new floating IP. The ip_pool attribute is optional only if your
+        infrastructure has only one IP pool available.
+        :param      ip_pool: name or id of the floating IP pool
+        :type       ip_pool: ``str``
+        :rtype: :class:`OpenStack_2_FloatingIpAddress`
+        """
+        for pool in self.ex_list_floating_ip_pools():
+            if ip_pool == pool.name or ip_pool == pool.id:
```
dpeschman: If ip_pool is None, as the docstring suggests should be possible if the infrastructure has only one pool, I do not think this implementation would create the IP.

CHANGES.rst
```diff
@@ -25,6 +25,20 @@ Compute
...
 
+- [OpenStack] OpenStack: Move floating IP functions to use network service
+  instead of nova.
+
+  This change affects all the floating ip related functions of the
+  ``OpenStack_2_NodeDriver`` class. Two new classes have been added
+  ``OpenStack_2_FloatingIpPool`` and ``OpenStack_2_FloatingIpAddress``.
+  The main change applies to the FloatingIP class where ``node_id``
+  property cannot be directly obtained from FloatingIP information and it
+  must be get from the related Port information with the ``get_node_id``
```
dpeschman: I think "must be get" should be "must be got" or "must be gotten".

libcloud/compute/drivers/openstack.py
```diff
@@ -4441,6 +4510,64 @@ def __repr__(self):
...
+                net = None
+            if net:
+                self.pool = OpenStack_2_FloatingIpPool(
+                    net.id, net.name, self.driver.network_connection
+                )
+        return self.pool
+
+    def get_node_id(self):
+        if not self.node_id:
+            # if not is not set, get it from port_details
```
dpeschman: "if not" > "if node_id"

libcloud/compute/drivers/openstack.py
micafer: You are right. Fixed in new commit.

CHANGES.rst
micafer: You are right. Fixed in new commit.

libcloud/compute/drivers/openstack.py
micafer: I don't get your point here ...

dpeschman: Sorry. This phrase "if not is not set" - I think you meant "if node_id is not set".

Issue #1644, opened 2022-01-14
Add Azure create node OS disk delete option


rgharris:
## Feature Request

Azure VM OS disks can be configured at the time of VM creation to be deleted when the VM is deleted. It would be helpful if libcloud supported this option. There is not an option in libcloud to do this on delete and it is convenient in some use cases to delete the disk regardless of the method (CLI, portal, libcloud, etc.).

Supporting this requires a `deleteOption` of `Delete` under `OSDisk` when creating a VM.

See the [OSDisk spec](https://docs.microsoft.com/en-us/rest/api/compute/virtual-machines/create-or-update#osdisk) for details. I plan create a pull request for this functionality soon as I have implemented the changes already to support my use case.

Issue #1643, opened 2022-01-14
Create node from Azure Compute Gallery images


rgharris:
## Feature Request

[Azure Compute Gallery](https://docs.microsoft.com/en-us/azure/virtual-machines/shared-image-galleries) is the recommended way to create custom images for Azure VMs. It would be very helpful if libcloud supported creating nodes from these images.

Supporting these works just like the `AzureImage` in libcloud today except you provide a single `id` under `imageReference` (instead of multiple key/values).

See this [example request](https://docs.microsoft.com/en-us/rest/api/compute/virtual-machines/create-or-update#create-a-vm-from-a-custom-image) and the [corresponding spec](https://docs.microsoft.com/en-us/rest/api/compute/virtual-machines/create-or-update#imagereference). I plan create a pull request for this functionality soon as I have implemented the changes already to support my use case.

PR #1645, opened 2022-01-14
Create Azure node from Compute Gallery image, OS disk delete

rgharris:
### Description

Source issues:
- #1643
- #1644

[Azure Compute Gallery](https://docs.microsoft.com/en-us/azure/virtual-machines/shared-image-galleries) is the recommended way to create custom images for Azure VMs. Supporting these works just like the `AzureImage` in libcloud today except you provide a single `id` under `imageReference` (instead of multiple key/values). See this [example request](https://docs.microsoft.com/en-us/rest/api/compute/virtual-machines/create-or-update#create-a-vm-from-a-custom-image) and the [corresponding spec](https://docs.microsoft.com/en-us/rest/api/compute/virtual-machines/create-or-update#imagereference). 

Azure VM OS disks can be configured at the time of VM creation to be deleted when the VM is deleted. It would be helpful if libcloud supported this option. There is not an option in libcloud to do this on delete and it is convenient in some use cases to delete the disk regardless of the method (CLI, portal, libcloud, etc.). Supporting this requires a `deleteOption` of `Delete` under `OSDisk` when creating a VM. See the [OSDisk spec](https://docs.microsoft.com/en-us/rest/api/compute/virtual-machines/create-or-update#osdisk) for details. 

A newer `api-version` is required for `create_node` (`deleteOption` not accepted otherwise).

### Status

Done, ready for review. 

The need to change the `api-version` for the Azure create node API may be problematic. I have tested `deleteOption` with Azure VMs using Compute Gallery Images with user code. I have added simple tests as well.

- [x] Documentation


libcloud/compute/drivers/azure_arm.py
```diff
@@ -703,7 +739,7 @@ def create_node(
 
         r = self.connection.request(
             target,
-            params={"api-version": RESOURCE_API_VERSION},
```
Kami, line 706: As you pointed out in the PR description, changing this version may be problematic.

To be on the safe side, I think it would be good to do the following:

1. Add new ``api_version`` to the driver constructor and allow user to override default API version (we already use similar pattern in many other drives).
2. Document this change in the upgrade_nodes.rst and show an example how user can utilize ``api_version`` constructor argument to set it to the previous default value.

Kami:
Thanks for the contribution.

I added a comment on how we could make the API version change a bit safer in case that change brakes the behavior for some user / setup, besides that, LGTM.

libcloud/compute/drivers/azure_arm.py
rgharris: Thanks for the feedback @Kami. 

There are various hardcoded `api-version` values (e.g., "2019-06-01") in the driver in addition to many references to `RESOURCE_API_VERSION`.

Which solution would fit best with the existing patterns for changes like this?

1. Leave the existing hardcoded version values. Update all using `RESOURCE_API_VERSION` to use the updated value via `self.api_version` (allow override).
2. Update all using `RESOURCE_API_VERSION` or hardcoded version values to use the updated value via `self.api_version` (allow override).
3. Move just `create_node` to use the new API version and add an extra parameter allowing the version to be overridden. This doesn't align with your suggestion of course but does minimize the potential impact to other methods.

1 & 2 do cause some potential issues: we will be affecting many methods, not just `create_node` that I have modified in this pull request. Updating the version for many methods is likely helpful as the current default is an older preview version. But further testing would be needed.

Kami: We could start by trying with updating all the places to use the latest version and what's specified via ``api_version`` argument and then go from there - depending on if end to end tests pass or not. So in short, what you proposed via 2).

Of course end to end tests do not cover every single possible scenario and setup, but it should be a start and give us at least some basic confidence.

rgharris: I gave that a try. It looks like Azure has different supported versions per service. For example, `2021-07-01` is supported for [VMs](https://docs.microsoft.com/en-us/rest/api/compute/virtual-machines) but not [disks](https://docs.microsoft.com/en-us/rest/api/compute/disks). [Public IP addresses](https://docs.microsoft.com/en-us/rest/api/virtualnetwork/public-ip-addresses) have differing versions as well. 

So, using a common API version across all requests will not work. This does explain why newer code has used hardcoded version numbers. Some variation of solution (3) will likely be necessary. Or we could do a deeper dive into differences in the API versions and move to the new version with no override (for `create_node` and optionally other VM APIs) if no breaking changes. What are your thoughts on those options? 

An error when using `2021-07-01` for `ex_get_volume`:
> [NoRegisteredProviderFound] No registered resource provider found for location 'westus2' and API version '2021-07-01' for type 'disks'. The supported api-versions are '2016-04-30-preview, 2017-03-30, 2018-04-01, 2018-06-01, 2018-09-30, 2019-03-01, 2019-07-01, 2019-11-01, 2020-05-01, 2020-06-30, 2020-09-30, 2020-12-01, 2021-04-01, 2021-08-01, 2021-12-01'. The supported locations are 'southeastasia, eastus2, centralus, westeurope, eastus, northcentralus, southcentralus, westus, northeurope, eastasia, brazilsouth, westus2, westcentralus, ukwest, uksouth, japaneast, japanwest, canadacentral, canadaeast, centralindia, southindia, australiaeast, australiasoutheast, koreacentral, westindia, francecentral, southafricanorth, uaenorth, australiacentral, switzerlandnorth, germanywestcentral, norwayeast, jioindiawest, westus3, swedencentral'.

Kami: Thanks for checking.

In this case, it sounds like it may be the best to only update api version used with the ``create_node()`` method.

And to make code a bit easier to maintain going forward (and in the case user needs to override the API version for some reason), we should probably define those versions as part of module level constants - something along the lines of ``VMS_API_VERSION, DISKS_API_VERSION``, etc.

If you want to try updating versions for other resources as well, that's fine, but it will likely require more work testing everything.

rgharris: In 83e6f51d03ad9010093f44b29491a84cabb09858 I added constants per resource type API version. This required a few methods to use updated API versions due to inconsistencies within some resources. I have done some minimal testing of all methods. I would be okay with using this pattern or just adjusting `create_node()` to simplify testing, let me know what you think.

These had version changes in the commit:
```
# updated to latest version due to changes in create_node
list_nodes 
create_node 
reboot_node
destroy_node 
ex_get_node
start_node
stop_node  

# updated to version used by ex_resize_volume
create_volume 
list_volumes
ex_get_volume 

# updated to version used by ex_delete_public_ip
ex_get_public_ip 
ex_list_public_ips
ex_create_public_ip  

# updated to version used by ex_update_nic_properties
ex_create_network_interface
```

rgharris: In c3b12ecb20872ed5b0af9391d4dc539f017c4844 I have reverted the version number change for the volume methods due to `properties.accountType` in the request changing to `sku.name` in the newer version. Found this after further manual testing. I've added a comment noting that ideally down the road all volume operations would be updated to use the same version. 

@Kami do you think this PR is ready to be merged? If not, any suggestions on this API version consolidation?

Updated list of version changes:
```
# updated to latest version due to changes in create_node
list_nodes 
create_node 
reboot_node
destroy_node 
ex_get_node
start_node
stop_node 

# updated to version used by ex_delete_public_ip
ex_get_public_ip 
ex_list_public_ips
ex_create_public_ip  

# updated to version used by ex_update_nic_properties
ex_create_network_interface
```

Kami (approved):
Sorry for the delay, LGTM - I will go ahead and merge the changes.

Issue #1647, opened 2022-01-27
Enable to set microversion in OpenStack driver


micafer:
## Feature Request

OpenStack compute API enables to set the microversion to use to interact with API:
https://docs.openstack.org/api-guide/compute/microversions.html

In fact some functionality does not work if it is not set e.g. https://bugs.launchpad.net/nova/+bug/1835042

Kami:
Resolved via #1647.

PR #1648, opened 2022-01-27
Enable to set microversion in OpenStack driver

micafer:
### Description

Enable to set microversion in OpenStack driver: see #1647

### Status

- done, ready for review

- [x] Documentation


libcloud/compute/drivers/openstack.py
```diff
@@ -1551,7 +1561,9 @@ def create_node(
         if ex_os_scheduler_hints:
             data["os:scheduler_hints"] = ex_os_scheduler_hints
 
-        resp = self.connection.request("/servers", method="POST", data=data)
+        resp = self.connection.request(
+            "/servers", method="POST", data=data, headers=self._set_microversion()
```
Kami: Hm, so we only want to set this header for "POST /servers" requests and nothing else? Shouldn't it be set for all the outgoing API requests?

libcloud/compute/drivers/openstack.py
```diff
@@ -1449,9 +1449,19 @@ class OpenStack_1_1_NodeDriver(OpenStackNodeDriver):
     _networks_url_prefix = "/os-networks"
 
     def __init__(self, *args, **kwargs):
+        self.ex_force_microversion = str(kwargs.pop("ex_force_microversion", ''))
```
Kami: For consistency with ``ex_force_api_version``, it may be better to default it to ``None``.

libcloud/compute/drivers/openstack.py
Kami: And in case we do want to set it for all the outgoing requests, ``add_default_headers()`` on the connection class may be a better place to do that.

Kami (approved):

Issue #1646, opened 2022-01-26
GCE driver list_images method does not return all the images


micafer:
## Summary

GCE driver list_images method does not return all the images.

## Detailed Information

GCE driver list_images method does not return all the images.
Only first 500 images are returned per project, and in some cases (as ubuntu-os-cloud) there are more than 900.

micafer:
One easy solution is setting:
```py
self.connection.gce_params = {'maxResults': 1500}
```
before this line:
https://github.com/apache/libcloud/blob/trunk/libcloud/compute/drivers/gce.py#L2630

micafer:
This problem is not only related with images.
In many API call there is no management of the pagination, so when the number of results is over the default maxResults value (500) the same problem will appear.

PR #1655, opened 2022-02-02
Add paginated request method to GCEConnection

micafer:
### Description

Source issues:
- #1646

The problem is not only related with images.
In many API call there is no pagination management, so when the number of results is over the default maxResults value (500) the same problem will appear.
In this PR I have added a new paginated_request function to the GCEConnection class and use it in the list_images function to solve #1646. But other calls to simple "request" call may be replaced with the same function.

### Status

- done, ready for review

- [ ] Documentation


Kami:
Thanks for working on this.

I was like 80% sure pagination is already implemented in GCE driver, but I guess I was wrong :)

Kami:
EDIT: It looks like it's indeed already implemented for most operations - https://github.com/apache/libcloud/blob/1e3a15d287a37f4acf4fb27866502da961ee85c5/libcloud/compute/drivers/gce.py#L170, https://github.com/apache/libcloud/blob/1e3a15d287a37f4acf4fb27866502da961ee85c5/libcloud/compute/drivers/gce.py#L2777, etc.

@micafer Can we utilize the same method for images as well (or perhaps make changes to that method in case any changes are needed to support image resource as well)?

micafer:
> EDIT: It looks like it's indeed already implemented for most operations -
> 
> https://github.com/apache/libcloud/blob/1e3a15d287a37f4acf4fb27866502da961ee85c5/libcloud/compute/drivers/gce.py#L170
> 
> ,
> https://github.com/apache/libcloud/blob/1e3a15d287a37f4acf4fb27866502da961ee85c5/libcloud/compute/drivers/gce.py#L2777
> 
> , etc.
> @micafer Can we utilize the same method for images as well (or perhaps make changes to that method in case any changes are needed to support image resource as well)?

We cannot use it "as is" we have to modifify it.
Let me see if it the changes are no "traumatic"

micafer:
> > EDIT: It looks like it's indeed already implemented for most operations -
> > https://github.com/apache/libcloud/blob/1e3a15d287a37f4acf4fb27866502da961ee85c5/libcloud/compute/drivers/gce.py#L170
> > 
> > ,
> > https://github.com/apache/libcloud/blob/1e3a15d287a37f4acf4fb27866502da961ee85c5/libcloud/compute/drivers/gce.py#L2777
> > 
> > , etc.
> > @micafer Can we utilize the same method for images as well (or perhaps make changes to that method in case any changes are needed to support image resource as well)?
> 
> We cannot use it "as is" we have to modifify it. Let me see if it the changes are no "traumatic"

Hi @Kami ,

I have reviewed it and I really think that is better to have a new function. The request_aggregated_items has a different behavior (merge results from different zones)  and we can change it to make it work also with images, but the code will not be so readeable.

Kami:
@micafer Sounds good.

In this case, let's please just add a comment to the new function which clarifies how it differs from the ``request_aggregated_items``. Something along the lines that both methods handle pagination, but ``request_aggregated_items`` is meant to be used with operations which operate on a zone or in an aggregated fashion across all the zones whereas the former is used for operations which don't operate on a zone.

Kami:
On a related note - can you please add a test case and a changelog entry?

Ideally we would have a unit test which exercises the pagination code path which means mock http method being called multiple times and returning a different page each time.

libcloud/compute/drivers/gce.py
```diff
@@ -224,14 +243,13 @@ def _merge_response_items(self, list_name, response_list):
...
-                # example k would be a zone or region name
-                # example v would be { "disks" : [], "otherkey" : "..." }
-                for k, v in resp["items"].items():
-                    if list_name in v:
-                        merged_items.setdefault(k, {}).setdefault(list_name, [])
-                        # Combine the list with the existing list.
-                        merged_items[k][list_name] += v[list_name]
+            # example k would be a zone or region name
+            # example v would be { "disks" : [], "otherkey" : "..." }
+            for k, v in resp["items"].items():
```
Kami: I see you removed `if "items" in resp`, after merging that PR, I will make the following change (`for k, v in resp.get("items", {}).items()`) to be on the safe side (in case some response for some reason doesn't contains `items` attribute - probably quite unlikely, but just to be on the safe side).

Kami: EDIT: Sorry, now I see that where this code is used and called it ensures that the item always contains ``items`` attribute so that check is indeed not needed.

Issue #1659, opened 2022-03-05
Missing parameter in an OpenStack Connection class


LanderOtto:
I am using Openstack with Application Credential. 
Running a simple code like this (Ubuntu 20.04, python 3.8.10)

```python
Driver = get_driver(Provider.OPENSTACK)
conn = Driver(ID, SECRET, ex_force_auth_url=ADDRESS, ex_force_auth_version= '3.x_appcred')
list_images = conn.list_images()
print(list_images)
```

libcloud (3.4.1) throws this exception
```
Traceback (most recent call last):
  File "test.py", line 17, in <module>
    list_images = conn.list_images()
  File "/usr/local/lib/python3.8/dist-packages/libcloud/compute/drivers/openstack.py", line 438, in list_images
    self.connection.request('/images/detail').object, ex_only_active)
  File "/usr/local/lib/python3.8/dist-packages/libcloud/common/openstack.py", line 245, in request
    return super().request(action=action, params=params, data=data,
  File "/usr/local/lib/python3.8/dist-packages/libcloud/common/base.py", line 567, in request
    action = self.morph_action_hook(action)
  File "/usr/local/lib/python3.8/dist-packages/libcloud/common/openstack.py", line 319, in morph_action_hook
    self._populate_hosts_and_request_paths()
  File "/usr/local/lib/python3.8/dist-packages/libcloud/common/openstack.py", line 332, in _populate_hosts_and_request_paths
    osa = self.get_auth_class()
  File "/usr/local/lib/python3.8/dist-packages/libcloud/common/openstack.py", line 220, in get_auth_class
    self._osa = cls(auth_url=auth_url,
TypeError: __init__() got an unexpected keyword argument 'auth_cache'
```

I fixed it by adding in **common/openstack_identity.py** in the constructor of class _OpenStackIdentity_3_0_Connection_AppCred_  the missing parameter _auth_cache=None_

Kami:
Thanks for reporting this issue.

It seems likely that this regression has been introduced as part of https://github.com/apache/libcloud/pull/1557, but it's weird that none of the tests caught it.

I was just preparing v3.5.0 release, but it sounds like I will need to wait until this issue is figured out and fixed (if confirmed).

Kami:
@dpeschman Any chance you will be able to look into this in the near future?

Kami:
I was able to confirm and reproduce the issue locally (and the proposed fix indeed appears to be working, but I didn't test it end to end).

[...]

PR #1557, opened 2021-02-26
OpenStack authentication token cache

dpeschman:
### Description

This change adds support to the OpenStack drivers for saving authentication tokens in an external cache that can be shared among multiple processes. This is to address the performance degradation in Keystone that can occur when tokens are allocated quickly and not reused (see apache#1460).

Parameter ex_auth_cache is added to the OpenStack drivers. If supplied, this cache will be queried for an active authentication token prior to requesting a new one from OpenStack. When a new token is acquired from OpenStack, it is inserted into the cache for later reuse.

In the event that a token is revoked, causing a 401 Unauthorized response while using it, that token is removed from the cache to prevent future uses.

### Status

- done, ready for review

### Checklist (tick everything that applies)

- [x] [Code linting](http://libcloud.readthedocs.org/en/latest/development.html#code-style-guide) (required, can be done after the PR checks)
- [x] Documentation
- [x] [Tests](http://libcloud.readthedocs.org/en/latest/testing.html)
- [x] [ICLA](http://libcloud.readthedocs.org/en/latest/development.html#contributing-bigger-changes) (required for bigger changes)

PR #1660, opened 2022-03-05
Fix OpenStack authentication regressions

Kami:
This PR fixes inadvertent regressions introduced in #1557 which broke some of the OpenStack authentication methods.

Issue reported by @LanderOtto via #1659.

### DIFF
--- a/libcloud/test/compute/test_openstack.py
+++ b/libcloud/test/compute/test_openstack.py
@@ -4105,9 +4105,10 @@ class OpenStack_AuthVersions_Tests(unittest.TestCase):
         for auth_version in AUTH_VERSIONS_WITH_EXPIRES:
             driver_kwargs = {}
 
-            if auth_version == "1.1":
-                # 1.1 is old and deprecated so we skip it
-                pass
+            if auth_version in ["1.1", "3.0"]:
+                # 1.1 is old and deprecated, 3.0 is not exposed directly to the endu ser
+                continue
+
 
             user_id = OPENSTACK_PARAMS[0]
             key = OPENSTACK_PARAMS[1]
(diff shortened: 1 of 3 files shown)


Kami:
All the new unit tests I added pass with two small changes / fixes in `libcloud/common/openstack_identity.py`, but I haven't actually tested all the authentication methods end to end with an actual OpenStack installation.

Kami:
It looks like that after making those other tests pass, new auth cache related tests are now failing (it seems like auth cache is not being correctly used in some scenarios / auth version, but didn't dig in too much yet).

 Probably best to wait @dpeschman since he's the most familiar with that piece of code.

Kami:
OK, I managed to get all the tests to pass - the reason why that test was failing was my fault - I removed a version constant I thought it's unused, but it turns out it's actually used internally, just not directly exposed to the end user.

libcloud/test/compute/test_openstack.py
```diff
@@ -3989,6 +4076,76 @@ def test_auth_user_info_is_set(self):
...
+        Test case which verifies that the driver can be correctly instantiated using all the
+        supported API versions.
+        """
+        cls = get_driver(Provider.OPENSTACK)
+
+        for auth_version in AUTH_VERSIONS_WITH_EXPIRES:
+            driver_kwargs = {}
+
+            if auth_version in ["1.1", "3.0"]:
+                # 1.1 is old and deprecated, 3.0 is not exposed directly to the endu ser
```
dpeschman: "endu ser" should be "end user".

dpeschman (approved):
Looks good to me.

libcloud/test/compute/test_openstack.py
Kami: Fixed, thanks.

PR #1663, opened 2022-03-11
Authenticate with Azure Ad for Provider AZURE_BLOBS

brookewhite9:
### Description

Add the ability to authenticate Azure Blob Storage Accounts via Azure AD service principal secrets and certificates.
This feature would be following this feature from jclouds (https://issues.apache.org/jira/browse/JCLOUDS-1558)
It would be using the rest api to access azure ad

section of our documentation.

### Status
- done, ready for review

- [x] Documentation


libcloud/common/azure.py
```diff
@@ -99,6 +105,96 @@ class AzureRawResponse(RawResponse):
...
+        host=None,
+        port=None,
+        tenant_id=None,
+        identity=None,
+        cloud_environment=None,
+        **kwargs
+    ):
+        super(AzureActiveDirectoryConnection, self).__init__(identity, secret, **kwargs)
+        if not cloud_environment:
+            cloud_environment = "default"
```
Kami: Would be better to just default ``cloud_environment`` argument in the method signature.

Kami:
Thanks for the contribution.

Overall the change seems reasonable to me, but I'm not an Azure expert so I would prefer @c-w to review it as well.

In addition to that, since it's a large change, end to end tests would also be good - we already do have some azure end to end tests in place (ones based on Azurite docker image and other ones talking directly to Azure API).

c-w:
Happy to take a look and thanks for the contribution!  This is great. There
were a fair few times that I myself had to fall back to the first party SDK
due to lack of Azure AD auth so this will be a game changer.

Just a heads up: my baby daughter was born yesterday so life is a bit
hectic right now, but I’ll try to get to this ASAP.

Kami:
@c-w Thanks and congrats!

brookewhite9:
@c-w Congrats! No rush, enjoy the new addition to your family :)
@Kami I will add the end to end test and address the default comment, thanks for your help!

brookewhite9:
@Kami for the end to end test you mentioned, how do you get it to run? I've tried following the testing guide, but it only seems to run the unit tests, not the integration ones
As for tox, I seem to be getting stuck on a loop of being unable to install libvirt-python(I am on a windows)

Kami:
@brookewhite9 Only a single test file depends in ``libvirt-python`` so if you can't install it, you should still be able to run rest of the tests (you can comment out that requirement in tox.ini).

Sadly we don't have docs on running integration tests at this point - it relies on credentials to be set for a couple of providers, something which most uses don't have.

Here is a link to how those tests run on CI - https://github.com/apache/libcloud/blob/trunk/.github/workflows/integration-tests.yml#L78.

You should be able to run Azure Storage tests using this command:

```bash
pytest -rsx -vvv --capture=tee-sys -o log_cli=True --durations=10 integration/storage/test_azure_blobs.py 
```

By default, this will only run tests which don't need access to Azure, but utilize local Azurite containers. If you want to run tests against actual Azure endpoints, you need to set the following environment variables:

* AZURE_CLIENT_SECRET
* AZURE_CLIENT_ID
* AZURE_TENANT_ID
* AZURE_SUBSCRIPTION_ID

Technically, you could also push your changes (new tests) to this branch and rely on the CI runs, but this process may be very slow and inefficient.

c-w (commented):
Changes look great 👍 I’d suggest to add some integration tests (look [here](https://github.com/apache/libcloud/blob/b1e40057c0283a498ab370658a75dc5ecb65d0b1/integration/storage/test_azure_blobs.py) for a starting point) and then this is ready to merge. In the integration tests you may also want to add a new subclass of the [Azurite tests](https://github.com/apache/libcloud/blob/b1e40057c0283a498ab370658a75dc5ecb65d0b1/integration/storage/test_azure_blobs.py#L61) as they added support for OAuth credentials which should be similar enough to the AAD flow ([see docs](https://docs.microsoft.com/en-us/azure/storage/common/storage-use-azurite?tabs=visual-studio#authorization-for-tools-and-sdks)).

brookewhite9:
@c-w I am currently trying to develop the integration tests, but even the ones that were there before fail with a 403 forbidden error on my local machine. Do you have any suggestions for how I can validate my tests?

brookewhite9:
@c-w  could I get some guidance on the integration tests? I can't test that they work on my local and I don't have permissions for the workflows to test it remotely

c-w:
@brookewhite9 Apologies for the delay. This is now at the top of my open source todo list. I'll aim to get to this as soon as I can.

Kami:
Any chance we could get those changes ready in time for the next release (some time in the next two weeks)?

brookewhite9:
@Kami  @c-w  I am still unable to do anything with the integration tests. Other than that, I am working on getting all the workflows passing. But, other than that my changes should be good to go. Can I have some guidance on how to get this merged before the next release?

Kami:
If we feel confident that those changes are unlikely to break any of the existing functionality, one option would be to merge them and then work on the integration tests afterwards / as soon as someone gets a chance to work on it.

Kami:
@c-w ^ How comfortable do you feel about something like that?

c-w:
Agreed. Between work (no longer at Microsoft/Azure) and growing family unfortunately I'm really struggling to find the time to support this properly. I'm 👍 for merge.

brookewhite9:
So would this mean that I would need to come back and work on the integration tests when I have the time? Or would that become a backlog issue?

Kami:
@brookewhite9 It would become a backlog issue - hopefully someone will have a chance to get to it in the near future.

Kami:
I went ahead and merged it into trunk so it can make it into the upcoming release.

Hopefully someone will have time in the future to add some integration tests :)

Thanks for the contribution.

PR #1664, opened 2022-03-12
Enable Retrying of Raw Requests parse_error

Germandrummer92:
### Description

We are currently again running into some errors again unfortunately when using libcloud and the storage buckets scaling up/returning 500s under load.

How does this fix retrying? The RawResponse now ALSO parses the error as the base Response class does. At best we would actually call the super init in the RawResponse class, but that is not possible due to the overriding of mutable properties with computed ones, I might look into that some other time.

Since the actual error is only thrown on `parse_error` calls and that method was before never called for raw responses in the retry loop, this will now be caught and retriggered.

Imho this makes the flow clearer anyways  since it's the same for both Response types, and removes the "duplicate" error parsing in the case of a non-raw-response.

### Status

- done, ready for review

- [X] Documentation


Germandrummer92:
@RunOrVeith  @Kami  fyi would be great to get this merged/released

libcloud/test/storage/test_local.py
```diff
@@ -42,6 +42,29 @@
     LocalStorageDriver = None
 
 
+class PickleableAcquireLockInSubprocess:
```
Germandrummer92, line 45: I am not sure about this one, for me the local test was failing with "unpickleable" errors, I am not sure how it worked before/on CI? Maybe it's also just a mac problem

Kami: Yeah, IIRC there is a known issue with some Python 3.x version and pickling in some multiprocessing scenarios on OS X (encounter this issue in a different project).

EDIT: https://github.com/spack/spack/issues/14102

Kami:
Thanks for the contribution.

The change seems reasonable to me and I think it should be safe to merge it as long as ll the tests pass :)

Germandrummer92:
this one good to go @Kami? Would be great for us to get it out quickly 🙂

Kami:
Merged, thanks!

Germandrummer92:
sorry to bother you again @Kami: but any chance we can get a patch release with this change? It's proving surprisingly difficult on our side unfortunately to introduce a git dependency due to some poetry intricacies it turns out, an official release would make this much easier

Kami:
@Germandrummer92 will try to do it this week, but sadly can't promise anything.

Having said that, if you can't install it directly from git using pip, could you perhaps install it using pip from zip archive - https://github.com/apache/libcloud/archive/8c4f229c98f521281a468070507f5387f6255bef.zip?

Germandrummer92:
We managed now, thanks anyways!

Kami:
Glad to hear you sorted out.

One thing we could still do is to build dev .tar.gz and .whl bundle on each push to master and upload it as an artifact as a part of Github Actions build.

That wouldn't be for long term production use, but mostly for testing and similar.

RunOrVeith:
Sadly installing directly from git kills the dependency resolving of poetry for no apparent reason (and induces a slow down of factor 100x before crashing), so it would be really great to get a patch release
image

Kami:
@RunOrVeith I will work on it this weekend.

Having said that, how about installing it from a .zip / .tar.gz bundle as proposed above? I quickly tested it locally in isolation and it seems to be working fine.

```toml
[tool.poetry.dependencies]
python = "^3.8"
poetry = "init"
apache-libcloud = {url = "https://github.com/apache/libcloud/archive/8c4f229c98f521281a468070507f5387f6255bef.zip"}
```

Also pretty curious why poetry git install takes so long. Libcloud repository is small and even if poetry doesn't use a shallow clone (which it appears it doesn't), cloning the repo should be fast. I haven't installed git dependencies using poetry yet, but I installed tons of dependency from git directly via pip without issues.

Kami:
And yet another short term option would be to use pre-built artifact from https://github.com/apache/libcloud/actions/runs/2033897262.

But since those builds artifacts are not stored permanently that means you would need to bundle it with your app git repo or similar.

RunOrVeith:
I tried it with the zip and it seems to reduce the resolving time by about 25-30%, however it produced a wrong dependency with the cryptography library that broke the build. So I will just deal with the longer resolving time and some manual hacking the lock file until it is released

Kami:
That's rather weird since Libcloud doesn't directly depend on ``cryptography``? Or is it a transitive dependency of the ``requests`` library or similar?

In any case, we will try to do a patch release as soon as possible.

PR #1666, opened 2022-03-14
Fix codecov

Germandrummer92:
### Description

I saw as part of [1664](https://github.com/apache/libcloud/pull/1664) that coverage is currently being collected also for the tests folder, this ignores them

### Status
- done, ready for review

- [X] Documentation

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - trunk
+      - fix_codecov
       - pyjion_cicd_tests
   pull_request:
     branches:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - trunk
-      - fix_codecov
       - pyjion_cicd_tests
   pull_request:
     branches:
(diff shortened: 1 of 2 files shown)


Kami:
Thanks for the contribution.

Just curious on the context for this change. A lot of projects (including Libcloud) intentionally include tests in the coverage report - the idea is to see which tests are executed (aka tests code coverage should be close too 100% otherwise that likely indicates an issue with dead / unreachable tests).

Germandrummer92:
Huh. That is a very good point. I don't think I have seen coverage used that way before, but you learn something new every day.  🙂 

Then two questions however: 

1) would it make sense to split the test package coverage + production code coverage into two numbers? Both would represent the intention better (one is to find "dead" test code) one to really check how well the production code is tested.
2) Why I started with this anyways: The Codecov comment on [1664](https://github.com/apache/libcloud/pull/1664#issuecomment-1066069396) looks to me like the coverage of `test_connection.py` and `test_local.py` went down, however on the codecov page itself it doesn't show those files as having [Coverage Changes](https://app.codecov.io/gh/apache/libcloud/compare/1664/changes), am I misunderstanding the table from codecov?

Kami:
1) Yeah, if we could split actual package and tests code in two separate reports that would be ideal, but I'm not sure if it's possible to do that with codecov.

2) The diff page does also seem to list those files, but it's hidden / non-expanded by default (I assume that's because the files weren't directly modified by this PR, but the coverage still changed) so I also missed it at first - https://app.codecov.io/gh/apache/libcloud/compare/1664/diff

Screenshot from 2022-03-15 10-27-03

Germandrummer92:
okay thanks, makes sense 👍 Closing this then.

PR #1668, opened 2022-03-18
Build and store dev release artifacts on each push to trunk

Kami:
This pull request adds a new Github Actions workflow which builds .tar.gz and .whl release artifacts and stores it as a Github Actions artifact.

Those artifacts are not stored long term and are primary meant to be used for testing and similar purposes.


Kami:
Confirmed it's working - https://github.com/Kami/libcloud/actions/runs/2006767284.

Issue #1670, opened 2022-03-22
Error getting node_id in OpenStack_2_FloatingIpAddress


lionel-van-bemten:
There seems to be a regression between libcloud 3.4.1 and 3.5.0. Not sure whether related to #1411.

> \>\>\> libcloud.__version__
> '3.5.0'
> \>\>\> from libcloud.compute.providers import get_driver
> \>\>\> from libcloud.compute.types import Provider
> \>\>\> dr = get_driver(Provider.OPENSTACK)("###", "###", ex_force_auth_url="###", ex_tenant_name="###", api_version='2.0', ex_force_auth_version='3.x_password')
> \>\>\> for ip in dr.ex_list_floating_ips():
> ...   print(ip.node_id)
> ...
> None
> None
> None
> None
> None
> None
> None
> None
> None
> None
> \>\>\> print(nesc.ex_list_floating_ips()[0])
> <OpenStack_2_FloatingIpAddress: id=2e0ad6dd-c134-4434-bd1d-6409bbde6376, ip_addr=###, pool=None, driver=<libcloud.compute.drivers.openstack.OpenStack_2_NodeDriver object at 0x7f02c8cb7b80>>

> \>\>\> libcloud.__version__
> '3.4.1'
> \>\>\> dr = get_driver(Provider.OPENSTACK)("###", "###", ex_force_auth_url="###", ex_tenant_name="###", api_version='2.0', ex_force_auth_version='3.x_password')
> \>\>\> for ip in dr.ex_list_floating_ips():
> ...   print(ip.node_id)
> ...
> f2e4a54e-6f85-48b7-a523-8548032dce32
> a92b9f8a-5f81-4f68-8331-a15ef99bf446
> None
> None
> None
> None
> None
> None
> None
> None

Kami:
Thanks for reporting this.

I would imagine that this regression is likely related to #1638. @micafer could you please have a look when you get a chance? Thanks.

micafer:
Using new network API the value of `node_id` is not returned in the floating ip list call, so you need to call the `get_node_id()'  method to get it:

https://github.com/apache/libcloud/blob/trunk/libcloud/compute/drivers/openstack.py#L4546

It is commented in the changelog of the issue GITHUB-1638:

[CHANGES.rst](https://github.com/apache/libcloud/blob/trunk/CHANGES.rst)

lionel-van-bemten:
[...]

PR #1671, opened 2022-03-24
Fix upgrade nodes: #1670 #1638

micafer:
### Description

As the OpenStack compute driver has moved the floating ip related functions from nova to neutron. There are one change that affect the users: node_id property cannot be directly obtained from FloatingIP information and it must be gotten from the related Port information with the ``get_node_id`` method.

### Status

- done, ready for review

- [x] Documentation


Kami (approved):
Thanks :+1:

PR #1676, opened 2022-04-06
Fix #1675: Error in volume api calls if microversion is set in OpenStack

### LINKED ISSUE #1675: Error in volume api calls if microversion is set in OpenStack
## Summary

Error in volume api calls if microversion is set.
If `ex_force_microversion` param is set some volume api calls fails with a 400 error.

## Detailed Information

The error is caused by setting the service fixed to compute in this line:

https://github.com/apache/libcloud/blob/trunk/libcloud/common/openstack.py#L347

micafer:
### Description

See #1675

Fix error in volume api calls if microversion is set in OpenStack.
In previous version if `ex_force_microversion` is set, it is assumed to set
it to the compute service. Now if only a version is set `2.67`, compute
service is assumed but it can be also set the service name `volume 3.21`.

### Status

- done, ready for review

- [ ] Documentation

### DIFF
--- a/libcloud/common/openstack.py
+++ b/libcloud/common/openstack.py
@@ -355,8 +355,10 @@ class OpenStackBaseConnection(ConnectionUserAndKey):
                 raise LibcloudError("Invalid microversion format.")
 
             if self.service_type and self.service_type.startswith(service_type):
-                headers["OpenStack-API-Version"] = "%s %s" % (service_type,
-                                                              microversion)
+                headers["OpenStack-API-Version"] = "%s %s" % (
+                    service_type,
+                    microversion,
+                )
         return headers
 
     def morph_action_hook(self, action):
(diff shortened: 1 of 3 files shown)


libcloud/common/openstack.py
```diff
@@ -343,9 +343,22 @@ def add_default_headers(self, headers):
...
+            # If service not set in microversion, asume compute
+            microversion = self._ex_force_microversion.strip().split()
+            if len(microversion) == 2:
+                service_type = microversion[0]
+                microversion = microversion[1]
+            elif len(microversion) == 1:
+                service_type = "compute"
+                microversion = microversion[0]
+            else:
+                raise LibcloudError("Invalid microversion format.")
```
Kami: Would be good to include expected / valid format in the error message.

Kami (approved):

PR #1683, opened 2022-05-05
Fix crash on missing etag in s3 storage driver

### LINKED ISSUE #1682: Google storage sometimes does not send etag
## Summary

When requesting a blob from google storage that is public (it works for private buckets), Google seems to have suddenly stopped sending the `etag` header. Unfortunately, this is required in `_headers_to_object`, which now crashes.
~~This means no files on a public bucket can be accessed any more at the moment.~~
This only affects some files that have been uploaded in a specific way, I am currently trying to reproduce it.

## Detailed Information
```python
from libcloud.storage.providers import Provider, get_driver
key = ""  # add your key here
secret = "" # add your secret here

# Make sure this points to a public blob, it works for private buckets
container_name = ""  # add a container name here
object_name = ""  #  add the object name here

driver_type = get_driver(provider=Provider.GOOGLE_STORAGE)
driver = driver_type(key=key,  secret=secret)
driver.get_object(
    container_name=container_name,
    object_name=object_name
)
```

Results in the following error:

```
  File "/home/veith/.config/JetBrains/PyCharm2022.1/scratches/scratch_1.py", line 12, in <module>
    driver.get_object(
  File "/home/veith/Projects/.venv/lib/python3.8/site-packages/libcloud/storage/drivers/s3.py", line 412, in get_object
    obj = self._headers_to_object(
  File "/home/veith/Projects/.venv/lib/python3.8/site-packages/libcloud/storage/drivers/s3.py", line 1195, in _headers_to_object
    hash = headers["etag"].replace('"', "")
KeyError: 'etag'
```

The problem are the first 2 lines in s3.py:

```python
    def _headers_to_object(self, object_name, container, headers):
        hash = headers["etag"].replace('"', "")  # expects etag to be present and will crash
        extra = {"content_type": headers["content-type"], "etag": headers["etag"]}
        meta_data = {}
```

Libcloud 3.5.1, Ubuntu 18.04, Python 3.8

I will try to create a fix ASAP

RunOrVeith:
### Description

Replaces the required access to the `etag` key with optional access.
I am not sure what implications this has on other components, but you can at least read public blobs again.
This fixes #1682 

### Status

- done, ready for review

- [X] [Code linting](http://libcloud.readthedocs.org/en/latest/development.html#code-style-guide)
- [X] Documentation (no update required)
- [X] [ICLA](http://libcloud.readthedocs.org/en/latest/development.html#contributing-bigger-changes) (I think I signed this already some time ago)

### DIFF
--- a/libcloud/storage/drivers/s3.py
+++ b/libcloud/storage/drivers/s3.py
@@ -1192,8 +1192,10 @@ class BaseS3StorageDriver(StorageDriver):
         return content_length
 
     def _headers_to_object(self, object_name, container, headers):
-        hash = headers["etag"].replace('"', "")
-        extra = {"content_type": headers["content-type"], "etag": headers["etag"]}
+        hash = headers.get("etag", "").replace('"', "")
+        extra = {"content_type": headers["content-type"]}
+        if "etag" in headers:
+            extra["etag"] = headers["etag"]
         meta_data = {}
 
         if "content-encoding" in headers:
--- a/libcloud/storage/drivers/s3.py
+++ b/libcloud/storage/drivers/s3.py
@@ -1196,6 +1196,7 @@ class BaseS3StorageDriver(StorageDriver):
         extra = {"content_type": headers["content-type"]}
         if "etag" in headers:
             extra["etag"] = headers["etag"]
+
         meta_data = {}
 
         if "content-encoding" in headers:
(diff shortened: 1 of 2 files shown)


Kami:
Thanks for reporting and fixing this issue.

Do you happen to know how it affects object uploads? Since, IIRC, we use e-tag returned on upload to verify that checksum matches and that there was no corruption on upload. Or this only affects get requests?

EDIT: I had a better look and it indeed appears it only affects GET object requests so it should be fine (only the ``hash`` field will be empty in case etag header is not returned, but that's probably fine).

Kami (approved):

Kami:
Merged into trunk, thanks!

PR #1690, opened 2022-05-10
test/test_http.py: stop HTTP server (thread) properly

ogayot:
### Description

Hello,

The `HttpLayerTestCase` test class creates a `HTTPServer` instance and runs it in a separate thread.

After running all the test-cases, we attempt to close the server and the thread by killing the thread. Unfortunately, the code that does that is unreachable because the tear-down member function is called `tearDownCls` intead of `tearDownClass` [1].

Moreover, there is no `threading.Thread.kill function`. This was undetected because the code was unreachable.

The proper way to clean things up is to:

1. Stop the HTTP server using `HTTPServer.shutdown()`
2. Join the thread using `threading.Thread.join()`

Thanks!

[1] https://docs.python.org/3/library/unittest.html#unittest.TestCase.tearDownClass

### Status

- done, ready for review

- [ ] Documentation
- [ ] [Tests](http://libcloud.readthedocs.org/en/latest/testing.html) _changes are in the tests only_

### DIFF
--- a/libcloud/test/test_http.py
+++ b/libcloud/test/test_http.py
@@ -116,8 +116,9 @@ class HttpLayerTestCase(unittest.TestCase):
         cls.mock_server_thread.start()
 
     @classmethod
-    def tearDownCls(cls):
-        cls.mock_server_thread.kill()
+    def tearDownClass(cls):
+        cls.mock_server.shutdown()
+        cls.mock_server_thread.join()
 
     def test_prepared_request_empty_body_chunked_encoding_not_used(self):
         connection = LibcloudConnection(host=self.listen_host, port=self.listen_port)


Kami:
Thanks for the contribution and a good catch.

The change looks good to me, but funny enough, I get this warning with your change, but I don't get it without it:

```bash
sys:1: ResourceWarning: unclosed <socket.socket fd=3, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=0, laddr=('127.0.0.1', 57898)>
```

Will do some more digging in. It may actually be related to outgoing client connection.

EDIT: Looks like that adding:

```python
self.mock_server.socket.close()
```

After ``mock_server.shutdown()`` solves that issue.

Kami:
Added server.socket.close() and merged it into trunk, thanks!

ogayot:
> Added server.socket.close() and merged it into trunk, thanks!

Oh, good catch as well @Kami, thanks!
Looking back, I think `server.server_close()` was the expected call after `server.shutdown()`.

```python
server.shutdown()
server.server_close()
```

But it seems that `server.socket.close()` calls `server.socket.close()` as you did, so I'm sure it's no big deal :)
https://github.com/python/cpython/blob/ecfff63e06e77e22035a7f7caa26986f033f3aea/Lib/socketserver.py#L483

PR #1694, opened 2022-05-14
Use tuple to compare versions for test against Paramiko.

havardAasen:
### Description

For the test test_key_file_non_pem_format_error() we want to compare and
see if Paramiko version is greater than '2.7.0'. The version is compared as a
string. This breaks when we test against version '2.10' and above.

We change from comparing a string, to comparing a tuple.
This is identical to this commit: 470dc46014e5b89a61a74cb6fd62493c49483a15

### Status

- done, ready for review

- [ ] Documentation


Kami (approved):
Good catch, thanks.

havardAasen:
I did a force-push. Create a tuple of the version string when Paramiko isn't installed, instead of a list.

Kami:
Merged, thanks.

PR #1697, opened 2022-05-20
tests: add way to skip tests that require access to the network/Internet

ogayot:
### Description
Hello,

This patch introduces two new variables which allow to skip some of the tests run:

 * `NO_INTERNET` -> skips the tests that require access to the Internet
 * `NO_NETWORK` -> skips the tests that require the network to be up.

`NO_NETWORK` implies `NO_INTERNET`.

Tests that are skipped by `NO_NETWORK` include tests that bind sockets to the loopback interface.

As for the two tests that are skipped with `NO_INTERNET`, they respectively:

 * send DNS queries and expect a NXDOMAIN response
 * attempt to send packets to 10.255.255.1 and expect a timeout (this requires a default route)

#### Rationale

In Ubuntu, we don't have full access to the Internet when running the test suite as part of autopkgtest, so we will set the `NO_INTERNET` variable to skip the tests that we cannot run.
To simulate a lack of Internet access, I ran the tests in a new, clean, network namespace (i.e., `# ip netns add test ; ip netns exec test /bin/bash`). Unfortunately, more tests started to fail (including the ones that bind to the loopback interface).
Therefore, I also added the `NO_NETWORK` variable, that disables more tests.

Thanks,
Olivier

### Status

- done, ready for review

- [ ] Documentation

### DIFF
--- a/libcloud/test/__init__.py
+++ b/libcloud/test/__init__.py
@@ -16,6 +16,7 @@
 import unittest
 import random
 import requests
+import os
 from libcloud.common.base import Response
 from libcloud.http import LibcloudConnection
 from libcloud.utils.py3 import PY2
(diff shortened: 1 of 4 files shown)


Kami (approved):
LGTM :+1:

PR #1699, opened 2022-05-25
Get GCE image price

Eis-D-Z:
## Add new method to get gce image price
     Add tests

### Description
Getting GCE image price right now with 
  `libcloud.pricing.get_pricing(driver_type='compute', driver_name='gce_images')`
is broken. I added a separate method to do that since gce images price dictionary veers off the pattern of the rest of the pricing data. I added some tests that use directly the pricing data and not the mock pricing data, on purpose, the main reason is to detect any changes in the format of the pricing data by Google.

### Status
- done, ready for review

- [x] Documentation


Kami:
Short therm this change looks good to me, but long term we should figure out how we can integrate it into standard ``get_size_pricing()`` method.

libcloud/pricing.py
```diff
@@ -200,6 +202,71 @@ def get_size_price(driver_type, driver_name, size_id, region=None):
     return price
 
 
+def get_gce_image_price(image_name, size_name, cores=1):
```
Kami: Would it be possible to make this "private" and call it inside ``get_size_price()`` method when GCE pricing is requested?

Since the whole idea behind Libcloud is to try to use common / standard API as much as possible. If we create a provider specific methods everywhere it kinda defeats the purpose (yes, I know it's not always possible to do it in common / standard way, but we should try our best).

Kami:
@Eis-D-Z Thanks for the contribution. Would it be possible to implement something like this https://github.com/apache/libcloud/pull/1699/files#r889676868?

Eis-D-Z:
Hey Tomaz,
Of course, sorry for the delay. I am on it.

Eis-D-Z:
Hi Tomaz,
I'm very sorry, I got some issues and only just now resolved them. I can
continue to improve the PR to be inline with the overall philosophy.
I do have some questions though:
1) get_size_price has `size` in its name so it might be misleading to
return image price, do you want to rename the method or are you cool with
this.
2) I will need to add two arguments, `image_name` and `cores`, and naming
is my weak point. Are those names alright?

3) I am preparing another PR with prices for EC2, because AWS have added to
their official endpoint prices. (The same endpoint where we get the sizes)
I will open it within the day on a separate PR, I believe you wouldn't want
to add it in this PR.

Cheers
Alex

Kami:
@Eis-D-Z 

1) In this case, to reduce confusion it would probably be better to introduce a new function, e.g. ``get_image_price`` or similar.

2) If we go with 1), I assume those arguments will then be added to this new function? I guess it may still be somewhat confusing since "cores" implies size to some extent, but I know that some image licensing is based on number of cores, so it probably also fine / make sense to add it there.

Eis-D-Z:
Ok I will go with 1.

Kami:
@Eis-D-Z Changes look good to me, but unit tests appear to be failing. Can you please have a look when you get a chance? Thanks.

Kami:
@Eis-D-Z How is this PR looking? Would be great if we could get it wrapped up merged. Thanks.

Eis-D-Z:
@Kami Excuse me for the delay.

Kami (approved):

Kami (approved):
LGTM, thanks!

PR #1713, opened 2022-06-08
test: alleviate race conditions in test_retry_*

enr0n:
### Description

The retry tests (test_retry_*) are inherently racy because they assume mock side effects
will always finish before the timeout. If for some reason the execution is a bit slow,
this can lead to the timeout beating the mocks, and the tests will fail. Adjust these tests
to alleviate this issue.

### Status

- done, ready for review

- [ ] Documentation

### DIFF
--- a/libcloud/test/test_connection.py
+++ b/libcloud/test/test_connection.py
@@ -429,7 +429,7 @@ class ConnectionClassTestCase(unittest.TestCase):
         con.connection = Mock()
 
         mock_connect.side_effect = socket.gaierror("")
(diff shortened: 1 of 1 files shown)


Kami:
Thanks for the contribution.

We intentionally tried to use as low timeout as possible to speed up the test run, but as you said, that is not the most robust thing and may result in failures on slower machines and similar.

Think PR will only add 5 seconds or so to the overall unit tests run time so I think it should be fine.

Kami:
Merged, thanks!

PR #1715, opened 2022-06-20
Use official endpoint for EC2 prices

Eis-D-Z:
### Use official endpoint to get EC2 prices to get as many prices and as acurate as possible. This enables to add prices for instances with suse, rhel and rhel with high availability. 
This like the previous script targets only OnDemand instance prices

### Status
- done, ready for review

- [x] Documentation


Kami:
Great, thanks!

Just curious is this official pricing endpoint any smaller? Or it's perhaps even bigger? Just curious since previous JSON we used to download was massive (multi GB) which meant that download could take a while (but we already downloaded it in chunks to avoid excessive memory usage + added progress bar).

EDIT: I can answer that myself - appears that file is 3 GB.

contrib/scrape-ec2-prices.py
```diff
@@ -161,82 +48,155 @@
...
+        with open(TEMPFILE, "wb") as fo:
+            for chunk in response.iter_content(chunk_size=2**20):
+                if chunk:
+                    fo.write(chunk)
+    return open(TEMPFILE, "r")
+
+
+def get_json():
+    try:
+        return open(TEMPFILE, "r")
```
Kami: I assume this requires us to cache whole massive JSON dictionary in memory, right?

I guess it's fine and we probably already did the same before.

Kami: May be a good idea to use with context manager to ensure file is closed (then again, this script is short lived so any file handle leaks are not a big deal)

contrib/scrape-ec2-prices.py
```diff
@@ -161,82 +48,155 @@
...
+
+
+# Prices and sizes are in different dicts and categorized by sku
+def get_all_prices():
+    # return variable
+    # prices = {sku : {price: int, unit: string}}
+    prices = {}
+    current_sku = ""
+    current_rate_code = ""
+    amazonEC2_offer_code = "JRTCKXETXF"
```
Kami, line 106: What is this code used for? Just trying to understand how it works.

Kami:
I've added some inline comments / questions.

contrib/scrape-ec2-prices.py
Eis-D-Z: Yes, this code was taken from the size scraping file, I will update to the latest version though.

contrib/scrape-ec2-prices.py
Eis-D-Z: The file has two big keys, "products" and "terms". Under "products" are the technical details of each sku. 
Under "terms" there are pricing data. A sample entry is like this:
{ "terms": { "onDemand": { sku: { sku: {"JRTCKXETXF": { "priceDimensions": { rateCode: {"unit": "$", "value": "0.17" } } } } } } }

It is fortunate that this ijson parser has been used since it makes it easy to find the "rateCode" number (with ijson you parse every line and it tells you if what you encounter is a key or a value). Otherwise it would be more difficult. I guess that amazon has a way of determining the rateCode based on the sku or some other data but I haven't found their relation.

The JRTCKXETXF is the same for all EC2 items and I named it amazonEC2_offer_code.

Thus what the code does is parse the "terms" dict line by like, when it encounters a dict key under terms["OnDemand"] set it as the current_sku. Then search the next lines for the rateCode of the item and after that finally get the price and the unit of the sku. Save those two in the dict to be returned under the sku. ({sku1: {"price": "xx", "unit": "$"}, sku2:{... }

contrib/scrape-ec2-prices.py
Eis-D-Z: Right. I will fix this.

Kami:
I've pushed some small lint and ``tox.ini`` fixes, but it looks like the script still fails with empty cache:

```
  File "contrib/scrape-ec2-prices.py", line 301, in <module>
    main()
  File "contrib/scrape-ec2-prices.py", line 294, in main
    pricing_data = scrape_ec2_pricing()
  File "contrib/scrape-ec2-prices.py", line 137, in scrape_ec2_pricing
    prices = get_all_prices()
  File "contrib/scrape-ec2-prices.py", line 104, in get_all_prices
    with open(json_file, 'r') as f:
TypeError: expected str, bytes or os.PathLike object, not tuple
```

I pushed a small fix for that, but didn't have time to dig in if there are other issues present.

On top of that, it looks like script takes a very long time to complete even when cache is available (I never got it to finish - not sure if it just takes a really long time or there is an infinite loop or similar).

What kind of run times have you seen on your computer?

If it takes a really long time, it might also not be a bad idea to add some kind of basic progress bar (e.g. print "." every now and then to give some feedback that script didn't get stuck or even better, periodically print how many items we have processed so far out of the total number).

Kami:
EDIT: It looks like ``ijson.parse()`` takes a very long time already and then iteration over the parsed content.

I don't think it's possible to add progress bar for ``ijson.parse()`` step itself (we would need to modify ijson or similar), but we might be able to add it for actual code where we process parsed response.

Kami:
I pushed some additional changes to increase buffer size to speed up parsing a bit and to print up a progress bar - since ``ijson`` handles parsing in an incremental manner and returns a generator there is no easy way for us to determine how many more items we need to parse / how far along we are so we can only show that parsing is being performed, but we can't display how many items there are left.

Kami:
@Eis-D-Z Script seems to be working now and other checks passing, but unit tests one is still failing - https://github.com/apache/libcloud/runs/7481787304?check_suite_focus=true

It appears it's failing because the pricing.json file format has changed. Or is something else going on? Can you please have a look when you get a chance? Thanks!

Kami:
@Eis-D-Z How is this PR looking? Would be great if we could get it wrapped up merged. Thanks.

Eis-D-Z:
@Kami Ready, sorry for the delay.

Kami:
@Eis-D-Z Thanks! Since this change basically doubled the pricing file (from 2 MB to 3 MB), I will take some time to verify that it has no negative implications on memory usage and similar (we had similar issues with sizing file in the past and to avoid those issues, we split some of the stuff into different files and made some imports lazy and only exercised in case EC2 driver is used).

Kami:
I had a look and by default we only cache data in memory for drivers for which pricing data is requested by the using so this will only affect people using EC2 driver.

I will still add a note about that to the changelog.

Kami:
I added a note on pricing file size + potential small memory usage increase (ce5b74479f4d21f2928d2d2b56d1cf734c4ee38a) and also updated the script to include progress bar when parsing sku pricing data (7bffd095b151d451bf6f723e89a88745c52399a8).

Kami:
Merged, thanks!

PR #1736, opened 2022-08-06
azure-arm: Add support to create disks in specific availability zones

palashgandhi:
### Description

Azure ARM's compute driver does not provide the ability to pass zones when creating a new volume. This change adds a new parameter that allows passing zones as a list of strings. 
Some open questions for the maintainers:
1. I have also merged the disk API versions because the `zones` parameter was only supported in newer API versions. I do not fully understand the impact of bumping the API versions, but if tests pass, I think it should be fine. Thoughts?
2. The new API version also caused request and response payloads to differ i.e. the volume types are now passed as [DiskSku](https://docs.microsoft.com/en-us/rest/api/compute/disks/create-or-update?tabs=HTTP#disksku) objects instead of strings. How do we handle API changes? Is there documentation around making such API changes?

I have tested the following disk operations:
1. Create empty disks.
2. Attach disks to a VM.
3. Resize disks.
4. Detach disks.
5. List disks.
6. Delete disks.

### Status
- done, ready for review

- [ ] Documentation


palash-gandhi:
@Kami when you get a chance, can you please take a look at this change? Thanks in advance.

Kami:
@pgandhi-delphix Sorry for the delay, I will look into it this weekend.

libcloud/compute/drivers/azure_arm.py
```diff
@@ -917,8 +914,9 @@ def create_volume(
         location=None,
         snapshot=None,
         ex_resource_group=None,
-        ex_account_type=None,
```
Kami, line 920: So this option has been deprecated / removed?

Kami (approved):
LGTM :+1:

Kami:
As far as your two questions go:

> I have also merged the disk API versions because the zones parameter was only supported in newer API versions. I do not fully understand the impact of bumping the API versions, but if tests pass, I think it should be fine. Thoughts?

Integration tests would increase the confidence here, but sadly we don't have integration tests for Azure ARM.

 >  The new API version also caused request and response payloads to differ i.e. the volume types are now passed as [DiskSku](https://docs.microsoft.com/en-us/rest/api/compute/disks/create-or-update?tabs=HTTP#disksku) objects instead of strings. How do we handle API changes? Is there documentation around making such API changes?

One option would be to allow user to pass ``api_version`` or similar argument to the constructor and then when user specifies a new version which uses new ``DISK_API_VERSION`` value, otherwise it would default to the previous value. This could / would of course increase complexity a lot so it only means sense to do it for major breaking API changes.

PR #1744, opened 2022-08-15
Azure: Add the ability to use Ultra SSDs

jwk404:
This change adds a function to update additional_capabilities so that
Ultra SSDs can be added. Additional changes allow specific provisioned
IOPS and throughput.

### Description
To add UltraSSD disks in Azure, a flag must be toggled on the VM
itself. Once the flag is enabled, a disk of this type may be
provisioned with a specific cap on IOPS and throughput. This commit
adds a function to toggle the compatibility flag, and adds the ability
to specify IOPS and throughput numbers.

### Status
- done, ready for review

- [ ] Documentation

### DIFF
--- a/libcloud/compute/drivers/azure_arm.py
+++ b/libcloud/compute/drivers/azure_arm.py
@@ -898,6 +898,8 @@ class AzureNodeDriver(NodeDriver):
         ex_sku_name=None,
         ex_tags=None,
         ex_zones=None,
+        ex_iops=None,
+        ex_throughput=None,
     ):
         """
         Create a new managed volume.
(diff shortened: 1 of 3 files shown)


libcloud/compute/drivers/azure_arm.py
```diff
@@ -2065,6 +2079,41 @@ def ex_create_tags(self, resource, tags, replace=False):
...
+            }
+        }
+
+        self.connection.request(
+            target,
+            data=data,
+            params={"api-version": VM_API_VERSION},
+            method="PUT",
+        )
+
```
Kami, line 2078: Does this API operation return some kind of response? Would it make sense to return something from this method (depending also on what API operation returns)? Perhaps True on success?

Kami:
Thanks for the contribution.

I added a comment / question on the method return value, besides that, LGTM.

libcloud/compute/drivers/azure_arm.py
jwk404: Thanks for catching that. I updated the method as suggested.

Kami:
I've fixed a small typo in docstring (c48b456af7dc243a6ccf00d53408f0494e961f79) and merged changes into trunk. Thanks!

PR #1752, opened 2022-09-02
Add new pip audit check to Github Actions Workflow

Kami:
This pull request adds new step to the existing Github Actions workflows which runs pip audit check - https://github.com/pypa/gh-action-pip-audit.

NOTE: Right now we only run pip audit on direct dependency of the library (we don't run audit check on any test / dev dependencies).


Kami:
Closing in favor of #1753 which is not part of a fork so new check will actually run.

PR #1765, opened 2022-09-09
Reformat code using pyupgrade and Python >= 3.6 compatible syntax

Kami:
This pull request reformats the code using pyupgrade using Python 3.6 and higher compatible Python syntax.

It also removes some (but not all) of the Python 3 compatibility code which is not needed anymore.

In addition to that, it also applies black formatting to the ``integration/`` directory (looks like that was missed in the past).


Kami:
Thanks to @tonybaloney for mentioning pyupgrade on the mailing list.

PR #1781, opened 2022-10-10
Update the new US based locations.

mohsen-hassani-cs:
### Description

The new locations are wdc.alpha3cloud.com and sjc.alpha3cloud.com

### DIFF
--- a/libcloud/common/cloudsigma.py
+++ b/libcloud/common/cloudsigma.py
@@ -45,12 +45,12 @@ API_ENDPOINTS_2_0 = {
     "sjc": {
         "name": "San Jose, CA",
         "country": "United States",
-        "host": "sjc.cloudsigma.com",
+        "host": "sjc.alpha3cloud.com",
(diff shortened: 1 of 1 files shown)


Kami:
Thanks for the contribution. The change looks good to me, but it would be good to also add / update test cases.

Having said that - should this driver also be renamed from CloudSigma to Alpha2Cloud? Or should the driver name be left as is and we perhaps add an alias for Alpha2Cloud -> Cloud Sigma (or vice versa).

mohsen-hassani-cs:
Thanks @Kami

The driver should not be renamed. It should be left as CloudSigma.
We have just changed the URLs for two locations but the driver and the functionality should be the same.

Kami:
@mohsen-hassani-cs Thanks for confirming and sorry for the delay (I've missed your comment). I will go ahead and merge you PR.

Kami (approved):

PR #1796, opened 2022-11-05
Fix Aliyun OSS storage upload_object KeyError: 'ETag' issue

shengwubin:
### Environment

```
Distributor ID:	Ubuntu
Description:	Ubuntu 22.04.1 LTS
Release:	22.04
Codename:	jammy
Python:         3.9.12
```

```
pip install apache-libcloud
# apache-libcloud-3.6.1
```

### Description

```python

key='your key'
secret='your secret'
oss_endpoint='oss-cn-hangzhou.aliyuncs.com'
oss_region='cn-hangzhou'
bucket_name='your bucket'
source='/path/to/source'
target='/path/to/target'

cls = get_driver(Provider.ALIYUN_OSS)
driver = cls(key, secret, host=oss_endpoint, region=oss_region)
container=driver.get_container(bucket_name)
obj=container.upload_object(source, target)
print(obj)
```

The above code will have below output on my computer:
```python
  File "/anaconda3/lib/python3.9/site-packages/libcloud/storage/base.py", line 203, in upload_object
    return self.driver.upload_object(
  File "/anaconda3/lib/python3.9/site-packages/libcloud/storage/drivers/oss.py", line 454, in upload_object
    return self._put_object(
  File "/anaconda3/lib/python3.9/site-packages/libcloud/storage/drivers/oss.py", line 642, in _put_object
    server_hash = headers["etag"].replace('"', "")
KeyError: 'etag'
```
### Changes

The problem here is that when sending uploading requests to OSS, the bucket name cannot be inserted into the url, for example, the final url should be like `https://your-bucket.oss-cn-hangzhou.aliyuncs.com/...`, but the current code will send a request with `https://oss-cn-hangzhou.aliyuncs.com/...`, then the bucket information will not be included in the `upload_object` function.

Below is what I change:
1. Add a `container` argument in the `_upload_object` function and pass `container` to the request.
2. Use `.upper()` when comparing the `data_hash` and `server_hash` to avoid hash mismatch

Note that the changes only happen in the `libcloud/storage/drivers/oss.py` in my latest pull request.

Chinese document: [https://help.aliyun.com/document_detail/31978.html?spm=a2c4g.11186623.0.0.140a26c0FKn8dD](https://help.aliyun.com/document_detail/31978.html?spm=a2c4g.11186623.0.0.140a26c0FKn8dD)

### Status

done


Kami (commented):
Thanks for the contribution, I will add some comments in-line.

libcloud/common/base.py
```diff
@@ -527,6 +527,7 @@ def request(
         stream=False,
         json=None,
         retry_failed=None,
+        container=None,
```
Kami: Overall, the changes except this one look good to me.

I don't think we should add ``container`` argument this function since it's a generic function not tied to any specific API so we should create a dependency on the storage API / container argument here.

Kami: Also, did you push all the changes? Since I don't see ``container`` argument you added in a bunch of places actually being used (or did I miss something)?

shengwubin: The reason I changed this line is to avoid influencing other storage providers, for example, s3.
My previous changes were not a good idea because they added several lines into `libcloud/storage/base.py` so that other providers' upload functions will throw exceptions.
I have reverted all the changes that lead to this situation, and the new solution will only affect `libcloud/storage/drivers/oss.py`. The possible side effect is that if the method `_upload_object` of class `StorageDriver` in `libcloud/storage/base.py` changes, the OSS-related method also needs to be changed if necessary.

Kami: Thanks - this new approach indeed looks much better and safer :)

Kami:
Looks like this still needs more work since tests are now failing with:

```bash
>       result_dict = self._upload_object(
            object_name=object_name,
            content_type=content_type,
            request_path=request_path,
            request_method=method,
            headers=headers,
            file_path=file_path,
            stream=stream,
            container=container,
        )
E       TypeError: upload_file() got an unexpected keyword argument 'container'
```

shengwubin:
> Looks like this still needs more work since tests are now failing with:
> 
> ```shell
> >       result_dict = self._upload_object(
>             object_name=object_name,
>             content_type=content_type,
>             request_path=request_path,
>             request_method=method,
>             headers=headers,
>             file_path=file_path,
>             stream=stream,
>             container=container,
>         )
> E       TypeError: upload_file() got an unexpected keyword argument 'container'
> ```

@Kami I add `container=container` to the oss storage test cases. The issue here is that when replacing the original `_upload_object` with `upload_file`, the argument `container` is missing

Kami (approved):

Kami:
Merged into trunk. Thanks.

PR #1807, opened 2022-12-01
Fix google oauth desktop

### LINKED ISSUE #1806: OAuth2 Desktop Client login is broken for base google driver
## Summary

You can no longer sign into the google driver using a OAuth2 Desktop Client, [since out-of-band authentication is deprecated and has been removed in October.
](https://developers.google.com/identity/protocols/oauth2/resources/oob-migration)

## Detailed Information

Code to reproduce (you need an OAuth2 Client Id with Desktop type).
If you already have a token you are fine until it expires, to reproduce you need to rename or delete the file `~/.google_libcloud_auth.<your client id>` 

```python

if __name__ == '__main__':
    driver_type = get_driver("google_storage")
    driver_type(key="YOUR CLIENT ID", secret="YOUR CLIENT SECRET")
```
This will now display an URL for you do click on and sign in, where you used to receive a code to copy-paste and would then proceed to be logged in. 
Now, you just get a window that your request is invalid.

The issue is in `libcloud.common.google.GoogleInstalledAppAuthConnection.get_code` (line 423 in libcloud 3.6.1).
There, the redirect url is set to self.redirect_url, which can not be changed and is currently the default value defined in 
`libcloud.common.google.GoogleBaseAuthConnection.__init__`: `"urn:ietf:wg:oauth:2.0:oob"`, which is used for the out-of-band authentication.

The new workflow works without copy and pasting if you change the redirect_uri to `"http://127.0.0.1:8087` and then have a local webserver running to receive the redirect.

I will post a pull request to fix this as soon as I can.

## Workaround

If you can not wait for the release of the new version with this issue fixed, here is a workaround:

1. Go to `libcloud.common.google` and find the method `GoogleInstalledAppConnection.get_code`
2. Within that method, there is a parameter `"redirect_uri": self.redirect_uri,` . Replace that line with `"redirect_uri": "http://127.0.0.1:8087/",`
3. Run the following snippet
 ```python

if __name__ == '__main__':
    driver_type = get_driver("google_storage")
    driver_type(key="YOUR CLIENT ID", secret="YOUR CLIENT SECRET")
 ```
4. It will print out an URL. Follow the instructions there. IMPORTANT: before you finish the sign in, you need to open the development console in your browser and go to the network tab
5. Finish sign in. In the network tab you will find a GET request to the domain 127.0.0.1:8087. Click on it, there should be a parameter in the url that reads code=<SOME-LETTERS-AND-NUMBERS> . You need to copy this code (without the code=)
7. On your terminal, make the following request using cURL; don't forget to  paste your code in the marked spot!
```
curl -X POST "https://oauth2.googleapis.com/token" --data-urlencode "code=<PASTE YOUR COPIED CODE HERE>" --data-urlencode "client_id=<YOUR CLIENT ID>" --data-urlencode "client_secret=<YOUR CLIENT SECRET>" --data-urlencode "redirect_uri=http://127.0.0.1:8087/" --data-urlencode "grant_type=authorization_code" > $HOME/.google_libcloud_auth.<YOUR CLIENT ID>
```
8. You should now be able to use your app again since you already have a token and a different execution path is taken.

RunOrVeith:
### Description

Fixes #1806 by introducing sign in through local loopback instead of out of band.

### Status

Done, ready for review

- [x] Documentation

### DIFF
--- a/libcloud/common/google.py
+++ b/libcloud/common/google.py
@@ -64,7 +64,6 @@ Setting up Installed Application authentication:
 
 Please remember to secure your keys and access tokens.
 """
-import contextlib
 import os
 import sys
 import time
--- a/libcloud/common/google.py
+++ b/libcloud/common/google.py
@@ -64,8 +64,7 @@ Setting up Installed Application authentication:
 
 Please remember to secure your keys and access tokens.
 """
-
-
+import contextlib
 import os
 import sys
 import time
@@ -73,6 +72,8 @@ import errno
 import base64
 import logging
 import datetime
+import urllib.parse
+from http.server import BaseHTTPRequestHandler, HTTPServer
 from typing import Optional
 
 from libcloud.utils.py3 import b, httplib, urlparse, urlencode
@@ -314,13 +315,14 @@ class GoogleBaseAuthConnection(ConnectionUserAndKey):
     name = "Google Auth"
     host = "accounts.google.com"
     auth_path = "/o/oauth2/auth"
+    redirect_uri_port = 8087
 
     def __init__(
         self,
         user_id,
         key=None,
         scopes=None,
-        redirect_uri="urn:ietf:wg:oauth:2.0:oob",
+        redirect_uri="http://127.0.0.1",
         login_hint=None,
         **kwargs,
     ):
@@ -352,6 +354,7 @@ class GoogleBaseAuthConnection(ConnectionUserAndKey):
         self.scopes = " ".join(scopes)
         self.redirect_uri = redirect_uri
         self.login_hint = login_hint
+        self._state = "Libcloud Request"
 
         super().__init__(user_id, key, **kwargs)
 
@@ -420,9 +423,9 @@ class GoogleInstalledAppAuthConnection(GoogleBaseAuthConnection):
         auth_params = {
             "response_type": "code",
             "client_id": self.user_id,
-            "redirect_uri": self.redirect_uri,
+            "redirect_uri": self._redirect_uri_with_port,
             "scope": self.scopes,
-            "state": "Libcloud Request",
+            "state": self._state,
         }
         if self.login_hint:
             auth_params["login_hint"] = self.login_hint
@@ -432,7 +435,7 @@ class GoogleInstalledAppAuthConnection(GoogleBaseAuthConnection):
         url = "https://{}{}?{}".format(self.host, self.auth_path, data)
         print("\nPlease Go to the following URL and sign in:")
         print(url)
-        code = input("Enter Code: ")
+        code = self._receive_code_through_local_loopback()
         return code
 
     def get_new_token(self):
@@ -443,14 +446,13 @@ class GoogleInstalledAppAuthConnection(GoogleBaseAuthConnection):
         :return:  Dictionary containing token information
         :rtype:   ``dict``
         """
-        # Ask the user for a code
         code = self.get_code()
 
         token_request = {
             "code": code,
             "client_id": self.user_id,
             "client_secret": self.key,
-            "redirect_uri": self.redirect_uri,
+            "redirect_uri": self._redirect_uri_with_port,
             "grant_type": "authorization_code",
         }
 
@@ -479,6 +481,54 @@ class GoogleInstalledAppAuthConnection(GoogleBaseAuthConnection):
         if "refresh_token" not in new_token:
             new_token["refresh_token"] = token_info["refresh_token"]
         return new_token
+
+    @property
+    def _redirect_uri_with_port(self):
+        return self.redirect_uri + ":" + str(self.redirect_uri_port)
+
+    def _receive_code_through_local_loopback(self):
+        # See https://developers.google.com/identity/protocols/oauth2/native-app#redirect-uri_loopback
+        access_code = None
+
+        class AccessCodeReceiver(BaseHTTPRequestHandler):
+
+            # noinspection PyMethodParameters,PyPep8Naming
+            def do_GET(self_):
+                query = urlparse.urlparse(self_.path).query
+                query_components = dict(qc.split("=") for qc in query.split("&"))
+                if "state" in query_components and query_components["state"] != urllib.parse.quote(
+                    self._state
+                ):
+                    raise ValueError(
+                        "States do not match: {} != {}, can't trust authentication".format(
+                            self._state, query_components["state"]
+                        )
+                    )
+                nonlocal access_code
+                access_code = query_components["code"]
+                self_.send_response(200)
+                self_.send_header("Content-type", "text/html")
+                self_.end_headers()
+                self_.wfile.write("<html><head><title>Libcloud Sign-In</title></head>".encode())
+                self_.wfile.write("<body><p>You can now close this tab</p>".encode())
+
+        if (
+            "127.0.0.1" in self.redirect_uri
+            or "[::1]" in self.redirect_uri
+            or "localhost" in self.redirect_uri
+        ):
+            # HTTPServer does not understand localhost unless you explicitly call it so
+            server_address = "localhost", self.redirect_uri_port
+        else:
+            server_address = self.redirect_uri, self.redirect_uri_port
+
+        server = HTTPServer(server_address=server_address, RequestHandlerClass=AccessCodeReceiver)
+        server.handle_request()
+        if access_code is None:
+            raise RuntimeError(
+                "Could not receive OAuth2 code: could not extract code though loopback"
+            )
+        return access_code
 
 
 class GoogleServiceAcctAuthConnection(GoogleBaseAuthConnection):
--- a/libcloud/common/google.py
+++ b/libcloud/common/google.py
@@ -355,7 +355,6 @@ class GoogleBaseAuthConnection(ConnectionUserAndKey):
         self.scopes = " ".join(scopes)
         self.redirect_uri = redirect_uri
         self.login_hint = login_hint
-        self._state = "Libcloud Request"
 
         super().__init__(user_id, key, **kwargs)
 
@@ -410,6 +409,8 @@ class GoogleBaseAuthConnection(ConnectionUserAndKey):
 
 class GoogleInstalledAppAuthConnection(GoogleBaseAuthConnection):
     """Authentication connection for "Installed Application" authentication."""
+
+    _state = "Libcloud Request"
 
     def get_code(self):
         """
--- a/libcloud/common/google.py
+++ b/libcloud/common/google.py
@@ -64,6 +64,8 @@ Setting up Installed Application authentication:
 
 Please remember to secure your keys and access tokens.
 """
+
+
 import os
 import sys
 import time
@@ -72,8 +74,8 @@ import base64
 import logging
 import datetime
 import urllib.parse
-from http.server import BaseHTTPRequestHandler, HTTPServer
 from typing import Optional
+from http.server import BaseHTTPRequestHandler, HTTPServer
 
 from libcloud.utils.py3 import b, httplib, urlparse, urlencode
 from libcloud.common.base import BaseDriver, JsonResponse, PollingConnection, ConnectionUserAndKey
--- a/libcloud/common/google.py
+++ b/libcloud/common/google.py
@@ -411,8 +411,7 @@ class GoogleInstalledAppAuthConnection(GoogleBaseAuthConnection):
 
     def get_code(self):
         """
-        Give the user a URL that they can visit to authenticate and obtain a
-        code.  This method will ask for that code that the user can paste in.
+        Give the user a URL that they can visit to authenticate.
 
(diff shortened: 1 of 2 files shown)


RunOrVeith:
The checks that are failing seem not to have been caused by my changes

Kami:
Good catch, thanks.

libcloud/common/google.py
```diff
@@ -480,6 +482,62 @@ def refresh_token(self, token_info):
             new_token["refresh_token"] = token_info["refresh_token"]
         return new_token
 
+    @property
+    def _redirect_uri_with_port(self):
+        return self.redirect_uri + ":" + str(self.redirect_uri_port)
+
+    def _receive_code_through_local_loopback(self):
```
Kami, line 490: It would be great if you could add a test case for this method. I know it's a bit more involved to test the whole method, but IIRC we already have existing patterns with mock http server and threads in some places.

RunOrVeith: I added a test case, I had to move one of the mocks in GoogleTestCase out of the parent class and into the child classes that actually require it, because that mocked the behavior I actually want to test here.
The test is more or less end to end, I am not mocking the HTTP server, just the user interaction of signing in is replaced with a fake response from google

RunOrVeith:
I might need some help to make the pipeline pass because I am not sure what is happening here.
The test file that I modified works fine on my machine in less than a second (I can not run all tests locally since I don't have the secrets file).

Kami:
Thanks for adding tests. I've fixed test failure related issues in trunk when upgrading various dependencies, etc.

Will fix lint issue and merge it into trunk. Thanks again for your contribution!

Kami:
Re tests hanging - I noticed that as well, but it only happens some times. I assume it's some cross test pollution / race condition which only happens sometimes (because it likely depends on how tests are distributed across worker proceesses by pytest-xdist plugin which runs tests in multiple processes).

Kami:
I did some digging in and added timeout for hanging tests and looks like the problematic test is the one which was added in this PR (It's also easy to reproduce it by just running ``libcloud/test/common/test_google.py`` test file directly):

```bash
======================================================================================== FAILURES =========================================================================================
__________________________________________________ GoogleInstalledAppAuthConnectionFirstLoginTest.test_it_aborts_if_state_is_suspicious ___________________________________________________
[gw8] linux -- Python 3.8.0 /home/kami/w/libcloud/libcloud/.tox/py3.8/bin/python

self = <libcloud.test.common.test_google.GoogleInstalledAppAuthConnectionFirstLoginTest testMethod=test_it_aborts_if_state_is_suspicious>

    def test_it_aborts_if_state_is_suspicious(self):
>       received_code = self._do_first_sign_in(
            expected_code="1234ABC", state=self.conn._state + "very suspicious"
        )

libcloud/test/common/test_google.py:292: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
libcloud/test/common/test_google.py:324: in _do_first_sign_in
    fake_sign_in_thread.join()
../../../.pythonz/pythons/CPython-3.8.0/lib/python3.8/threading.py:1011: in join
    self._wait_for_tstate_lock()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <Thread(Thread-1, started 140236942829248)>, block = True, timeout = -1

    def _wait_for_tstate_lock(self, block=True, timeout=-1):
        # Issue #18808: wait for the thread state to be gone.
        # At the end of the thread's life, after all knowledge of the thread
        # is removed from C data structures, C code releases our _tstate_lock.
        # This method passes its arguments to _tstate_lock.acquire().
        # If the lock is acquired, the C code is done, and self._stop() is
        # called.  That sets ._is_stopped to True, and ._tstate_lock to None.
        lock = self._tstate_lock
        if lock is None:  # already determined that the C code is done
            assert self._is_stopped
>       elif lock.acquire(block, timeout):
E       Failed: Timeout >15.0s

../../../.pythonz/pythons/CPython-3.8.0/lib/python3.8/threading.py:1027: Failed
---------------------------------------------------------------------------------- Captured stdout call -----------------------------------------------------------------------------------

Please Go to the following URL and sign in:
https://accounts.google.com/o/oauth2/auth?response_type=code&client_id=email%40developer.gserviceaccount.com&redirect_uri=http%3A%2F%2F127.0.0.1%3A8087&scope=https%3A%2F%2Fwww.googleapis.com%2Fauth%2Ffoo&state=Libcloud+Request
---------------------------------------------------------------------------------- Captured stderr call -----------------------------------------------------------------------------------

+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++ Timeout +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ Stack of Thread-1 (140236942829248) ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  File "/home/kami/.pythonz/pythons/CPython-3.8.0/lib/python3.8/threading.py", line 890, in _bootstrap
    self._bootstrap_inner()
  File "/home/kami/.pythonz/pythons/CPython-3.8.0/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/home/kami/.pythonz/pythons/CPython-3.8.0/lib/python3.8/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/home/kami/w/libcloud/libcloud/libcloud/test/common/test_google.py", line 310, in _get_code
    received_code = self.conn.get_code()
  File "/home/kami/w/libcloud/libcloud/libcloud/common/google.py", line 439, in get_code
    code = self._receive_code_through_local_loopback()
  File "/home/kami/w/libcloud/libcloud/libcloud/common/google.py", line 535, in _receive_code_through_local_loopback
    server.handle_request()
  File "/home/kami/.pythonz/pythons/CPython-3.8.0/lib/python3.8/socketserver.py", line 294, in handle_request
    ready = selector.select(timeout)
  File "/home/kami/.pythonz/pythons/CPython-3.8.0/lib/python3.8/selectors.py", line 415, in select
    fd_event_list = self._selector.poll(timeout)

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ Stack of <unknown> (140237122565824) ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  File "/home/kami/w/libcloud/libcloud/.tox/py3.8/lib/python3.8/site-packages/execnet/gateway_base.py", line 285, in _perform_spawn
    reply.run()
  File "/home/kami/w/libcloud/libcloud/.tox/py3.8/lib/python3.8/site-packages/execnet/gateway_base.py", line 220, in run
    self._result = func(*args, **kwargs)
  File "/home/kami/w/libcloud/libcloud/.tox/py3.8/lib/python3.8/site-packages/execnet/gateway_base.py", line 967, in _thread_receiver
    msg = Message.from_io(io)
  File "/home/kami/w/libcloud/libcloud/.tox/py3.8/lib/python3.8/site-packages/execnet/gateway_base.py", line 432, in from_io
    header = io.read(9)  # type 1, channel 4, payload 4
  File "/home/kami/w/libcloud/libcloud/.tox/py3.8/lib/python3.8/site-packages/execnet/gateway_base.py", line 400, in read
    data = self._read(numbytes - len(buf))

+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++ Timeout +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
______________________________________ GoogleInstalledAppAuthConnectionFirstLoginTest.test_it_receives_the_code_that_google_sends_via_local_loopback ______________________________________
[gw8] linux -- Python 3.8.0 /home/kami/w/libcloud/libcloud/.tox/py3.8/bin/python

self = <libcloud.test.common.test_google.GoogleInstalledAppAuthConnectionFirstLoginTest testMethod=test_it_receives_the_code_that_google_sends_via_local_loopback>

    def test_it_receives_the_code_that_google_sends_via_local_loopback(self):
        expected_code = "1234ABC"
        received_code = self._do_first_sign_in(expected_code=expected_code, state=self.conn._state)
>       self.assertEqual(received_code, expected_code)
E       AssertionError: None != '1234ABC'

libcloud/test/common/test_google.py:289: AssertionError

```

I will dig in more later this week and try to fix it.

Kami:
Race condition is indeed related to this new test - we need to ensure "send_code" thread is started before "get_thread" to avoid race condition / dead lock (get code trying to retrieve code before server is running).

I've added a temporary fix / workaround in - 97ba9b5386eba5bdefc44da053b5d3ebcf3b60e8 . Proper fix would not rely on wait / sleep, but on a state change.

Kami:
And there is yet another issue related to re-using the same port each time which causes issues when running multiple tests in parallel using multiple processes - https://github.com/apache/libcloud/actions/runs/3652800376/jobs/6171593038.

```bash
Exception in thread Thread-3:
  File "/opt/hostedtoolcache/PyPy/3.7.13/x64/lib-python/3/threading.py", line 926, in _bootstrap_inner
    self.run()
  File "/opt/hostedtoolcache/PyPy/3.7.13/x64/lib-python/3/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/home/runner/work/libcloud/libcloud/libcloud/test/common/test_google.py", line 311, in _get_code
    received_code = self.conn.get_code()
  File "/home/runner/work/libcloud/libcloud/libcloud/common/google.py", line 439, in get_code
    code = self._receive_code_through_local_loopback()
  File "/home/runner/work/libcloud/libcloud/libcloud/common/google.py", line 533, in _receive_code_through_local_loopback
    server = HTTPServer(server_address=server_address, RequestHandlerClass=AccessCodeReceiver)
  File "/opt/hostedtoolcache/PyPy/3.7.13/x64/lib-python/3/socketserver.py", line 452, in __init__
    self.server_bind()
  File "/opt/hostedtoolcache/PyPy/3.7.13/x64/lib-python/3/http/server.py", line 137, in server_bind
    socketserver.TCPServer.server_bind(self)
  File "/opt/hostedtoolcache/PyPy/3.7.13/x64/lib-python/3/socketserver.py", line 466, in server_bind
    self.socket.bind(self.server_address)
OSError: [Errno 98] Address already in use
```

Pushed a quick fix in ea0bf7c3cb740f26eb107df80a13c374065b30e2, but it may require more work (will need to dig in to see if any other tests exercise that code as well, since those would be also affected).

RunOrVeith:
Ok, I see the problem. I only ran the tests locally without any parallel execution so I didn't notice this. Thanks for the fixes!
I think the proper fix for the second problem would be to use [Event](https://docs.python.org/3/library/threading.html#threading.Event), so basically (untested):

```python
        ...
        sign_in_started = threading.Event()
        received_code = None

        def _get_code():
            nonlocal received_code
            sign_in_started.set()
            received_code = self.conn.get_code()

        def _send_code():
            target_url = self.conn._redirect_uri_with_port
            params = {"state": state, "code": expected_code}
            params = urllib.parse.urlencode(params, quote_via=urllib.parse.quote)
            requests.get(url=target_url, params=params)

        fake_sign_in_thread = threading.Thread(target=_get_code)
        fake_google_response = threading.Thread(target=_send_code)

        fake_sign_in_thread.start()
        sign_in_started.wait()
        fake_google_response.start()
        ...
```

Kami:
@RunOrVeith Yeah, I agree that Event (+timeout) is likely indeed the most robust approach.

Kami:
I quickly tried to get it working using Event without success (requires more work and digging in), so I will leave it as-is.

Well actually, I did get it to work if I added timeout to ``wait()`` but the only reason why that worked is because that introduced the delay in the same way as ``time.sleep()`` does and not because of the actual event signaling itself.

PR #1847, opened 2023-01-17
optimize read_in_chunks

Tobi995:
### Description

We noticed that utils.files.read_in_chunks has significant memory & CPU cost when up or downloading large files with mismatching chunk sizes. This PR attempts to fix that by potentially yielding multiple chunks per read and by avoiding large slicing ops. This seems to remove some more than quadratic scaling, bringing massive speed ups. 

```python
import tracemalloc

import numpy as np
import time

from libcloud.utils.files import read_in_chunks
import numpy as np

def run_scenario_1(data: bytes):
    # similar to calling _upload_multipart_chunks with one large array of bytes 
    for a in read_in_chunks(iter([data]), chunk_size=5 * 1024 * 1024, fill_size=True):
        ...

def run_scenario_2(data: bytes):
    # as in download_object_as_stream
    response_chunk = 5 * 1024 * 1024
    for a in read_in_chunks(iter([data[i:i+response_chunk] for i in range(0, len(data), response_chunk)]), chunk_size=8096, fill_size=True):
        ...

if __name__ == "__main__":
    tracemalloc.start()
    data = "c".encode("utf-8") * (40 * 1024 * 1024)
    times = []
    for i in range(10):
        start_time = time.time()
        run_scenario_1(data)
        times.append(time.time() - start_time)
    print("scenario 1", np.median(times))

    times = []
    for i in range(10):
        start_time = time.time()
        run_scenario_2(data)
        times.append(time.time() - start_time)
    print("scenario 2", np.median(times))

    current_size, peak = tracemalloc.get_traced_memory()
    print(f'Current consumption: {current_size / 1024 / 1024}mb, '
                f'peak since last log: {peak / 1024 / 1024}mb.')
``` 
Gives the following stats:
Without this PR:
```
scenario 1 0.08405923843383789
scenario 2 30.04505753517151
Current consumption: 40.009538650512695mb, peak since last log: 159.9015874862671mb.
```
With this PR:
```
scenario 1 0.013306617736816406
scenario 2 0.060915589332580566
Current consumption: 40.01009464263916mb, peak since last log: 85.03302574157715mb.
```
Scenario 2 with different data sizes:

- 20Mb: Before 3.97s, After 0.028s
- 30Mb: Before 8.4s, After 0.04s
- 40Mb: Before 30.0s, After 0.05s
- 50Mb: Before 54s, After 0.07s
- 60Mb: Before 102s, After 0.096s

### Status
done, ready for review

- [ ] Documentation

### DIFF
--- a/libcloud/test/test_utils.py
+++ b/libcloud/test/test_utils.py
@@ -215,7 +215,7 @@ class TestUtils(unittest.TestCase):
             iterator(), chunk_size=10, fill_size=True
         ):
             chunk_count += 1
-            self.assertEqual(result, b("aaaaaaaaaa"))
+            self.assertEqual(result, b("a") * 10)
         self.assertEqual(chunk_count, 200)
 
     def test_read_in_chunks_large_iterator_batches(self):
@@ -236,7 +236,7 @@ class TestUtils(unittest.TestCase):
             iterator(), chunk_size=10, fill_size=True
         ):
             chunk_count += 1
-            self.assertEqual(result, b("aaaaaaaaaa"))
+            self.assertEqual(result, b("a") * 10)
         self.assertEqual(chunk_count, 10_000)
 
     def test_read_in_chunks_filelike(self):
--- a/libcloud/test/test_utils.py
+++ b/libcloud/test/test_utils.py
@@ -202,15 +202,42 @@ class TestUtils(unittest.TestCase):
             for x in range(0, 1000):
                 yield "aa"
 
+        chunk_count = 0
         for result in libcloud.utils.files.read_in_chunks(
             iterator(), chunk_size=10, fill_size=False
         ):
+            chunk_count += 1
             self.assertEqual(result, b("aa"))
+        self.assertEqual(chunk_count, 1000)
 
+        chunk_count = 0
         for result in libcloud.utils.files.read_in_chunks(
             iterator(), chunk_size=10, fill_size=True
         ):
+            chunk_count += 1
             self.assertEqual(result, b("aaaaaaaaaa"))
+        self.assertEqual(chunk_count, 200)
+
+    def test_read_in_chunks_large_iterator_batches(self):
+        def iterator():
+            for x in range(0, 10):
+                yield "a" * 10_000
+
+        chunk_count = 0
+        for result in libcloud.utils.files.read_in_chunks(
+            iterator(), chunk_size=10, fill_size=False
+        ):
+            chunk_count += 1
+            self.assertEqual(result, b("a") * 10_000)
+        self.assertEqual(chunk_count, 10)
+
+        chunk_count = 0
+        for result in libcloud.utils.files.read_in_chunks(
+            iterator(), chunk_size=10, fill_size=True
+        ):
+            chunk_count += 1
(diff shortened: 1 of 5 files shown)


libcloud/test/test_utils.py
```diff
@@ -202,15 +202,42 @@ def iterator():
...
+            chunk_count += 1
+            self.assertEqual(result, b("a") * 10_000)
+        self.assertEqual(chunk_count, 10)
+
+        chunk_count = 0
+        for result in libcloud.utils.files.read_in_chunks(
+            iterator(), chunk_size=10, fill_size=True
+        ):
+            chunk_count += 1
+            self.assertEqual(result, b("aaaaaaaaaa"))
```
RunOrVeith: If you make this b("a") * 10 it's clearer to see what is going on

libcloud/utils/files.py
```diff
@@ -75,7 +75,13 @@ def read_in_chunks(iterator, chunk_size=None, fill_size=False, yield_empty=False
             return
 
         if fill_size:
-            if empty or len(data) >= chunk_size:
+            chunk_start = 0
```
RunOrVeith: maybe add a quick explanation why it is smarter to do it this way than the old way

RunOrVeith (approved):

libcloud/utils/files.py
Germandrummer92: not sure if it makes it better just as an idea, but to encapsulate the comment also you could move it into a method now?

```python
def _optimized_chunked_generator(data: bytes, chunk_size: int) -> Generator[bytes, None, bytes]:
       # We want to emit chunk_size large chunks, but chunk_size can be larger or smaller than the chunks returned
       # by get_data. We need to yield in a loop to avoid large amounts of data piling up.
       # The loop also avoids copying all data #chunks amount of times by keeping the original data as is.
       chunk_start = 0
       while chunk_start + chunk_size < len(data):
                yield data[chunk_start : chunk_start + chunk_size]
                chunk_start += chunk_size
               data = data[chunk_start:]

     return data

...

data = yield from _optimized_chunk_generator(data=data, chunk_size=chunk_size)
```

might make this method itself a tiny bit more readable?

Tobi995: Done

Kami:
Awesome, thanks for this contribution.

I think it would also be good to add micro benchmarks for this function / code path to `libcloud/test/benchmarks/` so we can track performance over time, avoid regressions, etc.

Kami:
I've added a basic baseline micro benchmark (to help us avoid regressions in the future) in 523838cd5e5a7f0c0dcab59651ebf4dee3fe6d1b.

I was indeed able to reproduce your results (large performance speed up). Right now this benchmark only measures CPU time and not memory usage. Measuring memory usage would be a nice addition, but that would be a bit more involved change since pytest-benchmark doesn't support doing something like that natively.

Screenshot from 2023-01-28 12-34-51

As you mentioned, the performance gets much worse with older implementation with larger object / input data size, but to prevent micro benchmarks for running too long, I used smaller data sizes.

Thanks again and great work :tada:

Kami (approved):

Issue #1824, opened 2022-12-19
Azure arm driver does not list all nodes


jan-mue:
## Summary

The Azure ARM driver only lists 50 nodes in a subscription.

## Detailed Information

The function `list_nodes()` of the Azure ARM driver only lists 50 nodes when the Azure Portal shows more than 50 VMs.

This is probably caused by missing pagination here:
https://github.com/apache/libcloud/blob/trunk/libcloud/compute/drivers/azure_arm.py#L474

According to the documentation, there should be a property `nextLink` in the response:
https://learn.microsoft.com/en-us/rest/api/compute/virtual-machines/list-all

Libcloud version: 3.6.1
Python version: 3.10
Operating system: Ubuntu 20.04

Kami:
Resolved via #1850.

PR #1850, opened 2023-01-25
Fix list_nodes in the Azure arm driver

jan-mue:
### Description

The Azure API returns an attribute `nextLink` for pagination that is currently not used in the Azure ARM driver.
Documentation: https://learn.microsoft.com/en-us/rest/api/compute/virtual-machines/list-all?tabs=HTTP

This PR fixes the issue where not all VMs are returned for subscriptions with a lot of VMs (#1824)

### Status

ready for review

- [ ] Documentation


Kami:
Thanks for the contribution.

It would be great if you could also add a unit test for this change.

jan-mue:
@Kami sorry for the long delay, but I've added some unit tests now. Could you please have another look at the PR?

Kami:
@jan-mue Thanks for adding a test case. I will review it shortly and if everything looks fine, merge changes into trunk.

libcloud/compute/drivers/azure_arm.py
```diff
@@ -468,11 +468,17 @@ def list_nodes(self, ex_resource_group=None, ex_fetch_nic=True, ex_fetch_power_s
...
                 self.subscription_id
             )
-        r = self.connection.request(action, params={"api-version": VM_API_VERSION})
-        return [
-            self._to_node(n, fetch_nic=ex_fetch_nic, fetch_power_state=ex_fetch_power_state)
-            for n in r.object["value"]
-        ]
+        params = {"api-version": VM_API_VERSION}
+        nodes = []
+        while True:
```
Kami: I think this code is far for now, but to be on the safe side and avoid possible infinite loop, I think it would be good to add additional guard here - e.g. bail out if we are still stuck in this loop for more than X seconds.

Kami:
@jan-mue Can you please also sign an ICLA when you get a chance (https://libcloud.readthedocs.io/en/latest/development.html#contributing-bigger-changes)? Thanks.

Kami:
I pushed a change so we don't end up in an infinite loop in case server returns a bad response, there is a bug in the code or similar - c49ecd0a4b8fe1e3820d8764f9aded21e5dee11c.

I picked that 60 seconds limit arbitrarily, but if needed or wanted, it can be changed (e.g. increased) - @jan-mue Can you please advise if 60 seconds is enough or the limit should be higher?

I imagine we may want a higher limit since depending on the method arguments, each ``_to_node()`` call will result in additional HTTP request(s).

For now, in case deadline is reached we simply return, but we may also want to throw or at least log a warning in such scenario.

Kami:
EDIT: I went ahead and bumped that limit to 300 seconds for now. This way I can wrap up and merge this PR.

And then based on your future feedback @jan-mue, we can adjust this value.

Kami (approved):

PR #1877, opened 2023-03-21
Fix list_volumes, and list_nodes in Outscale provider

outscale-mgo:
### Description

the way `list_nodes` was listing `privates_ips` and `public_ips` was buggy. and either empty, or crash in a few case. (if `"Nics" in vm` where true)
the only was to get IPs was using `node.extra['PublicIp'/'PrivateIp']`, this patch enable user to get ip, using `node.public_ips` (but list only main ip)

also it seems `_to_volumes` was broken due to it calling `_to_volumes` instead of `_to_volume`

### Status

- done, ready for review

- [ ] Documentation

### DIFF
--- a/libcloud/compute/drivers/outscale.py
+++ b/libcloud/compute/drivers/outscale.py
@@ -7747,7 +7747,12 @@ class OutscaleNodeDriver(NodeDriver):
 
     def _to_node(self, vm):
         name = ""
-        private_ips = [vm["PrivateIp"]]
+        private_ips = []
+        if "PrivateIp" in vm:
+            private_ips = [vm["PrivateIp"]]
+        public_ips = []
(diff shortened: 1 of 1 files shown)


outscale-mgo:
ping :)

Kami:
Thanks for the contribution.

The change looks good to me, but (regression) test for those changes would be good.

Kami:
@outscale-mgo I went ahead and merged this PR into trunk, but I think some tests are strongly needed, especially since it appears there are no existing tests for this driver right now (so no wonder this issue was not caught earlier).

If you plan to contribute bigger changes, please also make sure you sign an ICLA - https://libcloud.readthedocs.io/en/latest/development.html#contributing-bigger-changes.

Thanks.

PR #1891, opened 2023-04-14
Fix list_volumes in the Azure arm driver

jan-mue:
### Description

Not all disks are returned when listing disks on Azure. The same change as for list_nodes (#1850) is necessary for list_volumes.

### Status

ready for review

- [ ] Documentation

### DIFF
--- a/libcloud/compute/drivers/azure_arm.py
+++ b/libcloud/compute/drivers/azure_arm.py
@@ -25,7 +25,7 @@ import base64
 import binascii
 
 from libcloud.utils import iso8601
-from libcloud.utils.py3 import basestring
+from libcloud.utils.py3 import basestring, urlparse, parse_qs
 from libcloud.common.types import LibcloudError
 from libcloud.compute.base import (
     Node,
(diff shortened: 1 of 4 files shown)


libcloud/compute/drivers/azure_arm.py
```diff
@@ -1008,11 +1008,17 @@ def list_volumes(self, ex_resource_group=None):
         action = action.format(
             subscription_id=self.subscription_id, resource_group=ex_resource_group
         )
-
-        response = self.connection.request(
-            action, method="GET", params={"api-version": DISK_API_VERSION}
-        )
-        return [self._to_volume(volume) for volume in response.object["value"]]
+        params = {"api-version": DISK_API_VERSION}
```
Kami, line 1011: Since similar code is duplicated in two places, it would be nice to refactor common functionality into a utility function.

Kami:
To be able to merge this PR I applied similar guard I added in the list_nodes() PR (d9f6268a2ec0211219e21b1e4f56813d265e62f1) - https://github.com/apache/libcloud/pull/1850#issuecomment-1658422834

It's not ideal and it would need some refactor love to handle pagination in a standard manner across all the methods and to avoid duplication, but it's better than noting having any guard at all.

If you think timeout should be adjusted (e.g. increased), please let me know.

PR #1920, opened 2023-07-31
Also run unit tests under Python 3.12 beta

Kami:
This pull request updates tox and GHA config to also run tests under Python 3.12 beta.


Kami:
NOTE: Tests will be failing until https://github.com/pypa/pip/issues/11501 is resolved.

PR #1924, opened 2023-08-01
Replace "skip-duplicate-action" GHA jobs pre-check with native GHA concurrency functionality

Kami:
This pull request updates all the workflows and switches from ``skip-duplicate-actions`` GHA action to native "concurrency" functionality.

Ability to cancel old and redundant builds is now part of native GHA functionality. Utilizing native functionality allows us to simplify the workflows and remove that ``pre_job`` check. It also allows us to get rid of additional write permission which was required by that action.


Kami:
I confirmed the change works as expected (https://github.com/apache/libcloud/actions/runs/5725752694?pr=1924), so I will go ahead and implement it for all the workflows.

PR #1925, opened 2023-08-01
Remove deprecated and unused ApiDocs command from setup.py

Kami:
This pull request removes deprecated and unused "apidocs" command from setup.py (apidocs are being generated using sphinx for a long time already).

This way it also removes dependency on ``distutils.core.Command`` which is not available anymore in Python 3.12.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,7 +17,6 @@ import os
 import re
 import sys
 import fnmatch
-from distutils.core import Command
 
 import setuptools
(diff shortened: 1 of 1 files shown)


Kami:
Keep in mind that this is just the first step. To get dist steps to pass under Python 3.12, we will need to get rid of setup.py all together and replace it with packaging metadata in setup.cfg / pyproject.toml file.

PR #1929, opened 2023-08-02
Migrate from setup.py to pyproject.toml

Kami:
This pull request is a work in progress attempt at trying to migrate package and build metadata from setup.py to pyproject.toml.

It will take some time to polish out all the kinks to ensure produced sdist and wheel artifacts match what would have been produced by the previous setup.oy config.

As part of this change, running tests directly via ``python setup.py test`` will be deprecated and removed (https://docs.pytest.org/en/7.2.x/explanation/goodpractices.html#do-not-run-via-setuptools) -  I will document that in the changelog and update the affected readme.

## TODO

- [x] Verify new .tar.gz and wheel artifacts match what would be generated with setup.py in the past
- [x] Update dist.sh script to utilize build package (https://github.com/pypa/build)
- [x] Research potentially using hatchling instead of setuptools for the build backend (may want to avoid this for now to keep the scope of the changes smaller)
- [x] Update development docs (release management section)
- [ ] To reduce build artifact sizes, possibly consider excluding tests + test fixture files again (we've did that a long time ago and then changed to including tests with the distribution artifact - so the end user can run the tests on the downloaded artifact and verify the artifact is in order)


Kami:
^ @rsnk96 Just a heads up since you made a similar change recently by moving most of the tooling config into pyproject.toml.

This change is still very much work in progress, but I just wanted to give you a heads up.

Kami:
I checked and tests the artifacts and they appear to be looking and working fine.

For the time being, I will leave simple mainimal ``setup.py`` file there so we can still rely on ``python setup.py --version``.

Alternative without that minimal setup.py file would be much more complicated (especially since we rely on dynamic version functionality in pyproject.toml).

PR #1934, opened 2023-08-08
Test RTD PR build #1

Kami:
Test PR to see if it correctly triggers RTD PR build.


Kami:
It looks like it only works when changes are part of the main git remote (https://gitbox.apache.org/repos/asf/libcloud) and not part of a fork.

PR #1935, opened 2023-08-08
Test RTD PR build #2

Kami:
Test PR to see if it correctly triggers RTD PR build.


Kami:
It indeed does (https://libcloud--1935.org.readthedocs.build/en/1935/), but PR comment is not added since ASF only supports webhook and not Github App based integration.

PR #1940, opened 2023-08-17
Fix inappropriate length comparison

munahaf:
In file: nttcis.py, the comparison of collection length creates a logical short circuit. Since we are comparing two elements, the length should be greater than 1. This is moved to the outside if. The change should be reviewed so that it captures the intended business logic.

Sponsorship and Support:

This work is done by the security researchers from OpenRefactory and is supported by the Open Source Security Foundation (OpenSSF)(https://openssf.org/): Project Alpha-Omega(https://alpha-omega.dev/). Alpha-Omega is a project partnering with open source software project maintainers to systematically find new, as-yet-undiscovered vulnerabilities in open source code - and get them fixed – to improve global software supply chain security.

The bug is found by running the Intelligent Code Repair (iCR) tool by OpenRefactory and then manually triaging the results.


Kami:
Thanks for the contribution.

The original logic indeed looks weird / wrong, but I didn't originally implement the code and I'm not sure about the logic and the reasoning behind it.

Having said that, some tests for that logic (including various edge cases) would be a good start (and of course making sure that existing tests / if any, still pass).

munahaf:
I am not familiar with the business logic of the code. Unfortunately, I would not be able to come up with the test cases. We are just following the anomaly as identified by a static analysis tool.

Kami:
@munahaf Thanks.

Similar here and that's also why I don't feel that comfortable with merging this code - there are no tests and I don't fully understand the original developer reasoning.

munahaf:
At least the change I proposed appears less confusing and still retains the original logic of the code. Perhaps we can start with that. 

Per git blame, the code was initiated by this effort: https://github.com/apache/libcloud/commit/b19a16539c34d22c3d7582c2a32f6a613b78bb16

What is the appropriate way of getting some stakeholders (committer, reviewer) involved?

Kami:
@munahaf Agreed.

You could try pinging the person who originally added that code change (@snmpboy), but no idea if they are still involved with that provider.

Kami (approved):
LGTM.

I think it's fine and safe to merge this change. And if we ever hear back from the person who originally implemented this change, we can make any adjustments then, if needed.

PR #1944, opened 2023-08-21
Make classes inheriting from Type hashable

ricardobranco777:
### Description

Attributes of classes that inherit from `libcloud.common.types.Type` are not hashable, so they can't be tested for membership in sets, unlike those that inherit from `libcloud.container.types.Type` which work as intended.

```
$ python
Python 3.11.4 (main, Jun  9 2023, 07:59:55) [GCC 12.3.0] on linux
>>> from libcloud.compute.types import NodeState
>>> NodeState.RUNNING
running
>>> NodeState.RUNNING in {"running"}
False
>>> str(NodeState.RUNNING) in {"running"}
True
>>> from libcloud.container.types import ContainerState
>>> ContainerState.RUNNING in {"running"}
True
```

### Status

done, ready for review.

- [ ] Documentation

### DIFF
--- a/libcloud/common/types.py
+++ b/libcloud/common/types.py
@@ -82,7 +82,7 @@ class Type(str, Enum):
         return self.value
 
     def __hash__(self):
-        return id(self)
+        return hash(self.value)
 
 
 class LibcloudError(Exception):
--- a/libcloud/test/common/test_types.py
+++ b/libcloud/test/common/test_types.py
@@ -1,0 +1,32 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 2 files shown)


Kami:
Thanks for the contribution.

The change looks good to me, but can you add a a couple of simple unit tests for it so we can get it merged?

Kami:
I also had a quick look at the container Type implementation and it appears it doesn't define ``__hash__`` and all the other special methods defined by the ``libcloud.common.Type``.

It seems like that for consistency reasons, `libcloud.container.Type` should eventually be made to inherit from `libcloud.common.type`.

An alternative could perhaps also be to completely remove those methods since it appear they are only present for backward compatibility reasons (https://github.com/apache/libcloud/blob/abba8c1719a8bda6db8efde2d46fd1b423ae4304/libcloud/container/types.py#L19). But before doing that, we would of course first need to analyze and understand the impact of such change.

ricardobranco777:
> The change looks good to me, but can you add a a couple of simple unit tests for it so we can get it merged?

Done.

> It seems like that for consistency reasons, `libcloud.container.Type` should eventually be made to inherit from `libcloud.common.type`.

I'm very much in favour of doing that if we remove the dunder methods.

Kami:
@ricardobranco777 

Thanks.

> I'm very much in favour of doing that if we remove the dunder methods.

To be on the safe side, I think it's better to do this in a separate add-on PR once this one is merged.

Kami:
The PR has been merged.

> It seems like that for consistency reasons, libcloud.container.Type should eventually be made to inherit from libcloud.common.type.

Once you get a chance, it would be great if you could make this change and also get rid of all the dunder methods which are there for backward compatibility reasons with older versions of Libcloud before ``Type`` inherited from Enum. We just need to test it and ensure that this change won't break anything.

Thanks.

PR #1946, opened 2023-08-25
Updated Linode (Akamai Connected Cloud) support (including cloud-init)

mraygalaxy2:
### Description

1. cloud-init support is a new feature available in 2023.
2. The main entry point (create_node) had an arrangement of non-keyword parameters that were not consistent with other libcloud drivers. This has been fixed.
3. One remaining function (already available in the API) was exposed to also be consistent with other drivers.

### Status

- done, ready for review

- [x] Documentation

### DIFF
--- a/libcloud/compute/drivers/linode.py
+++ b/libcloud/compute/drivers/linode.py
@@ -37,6 +37,7 @@ from datetime import datetime
 from libcloud.utils.py3 import httplib
 from libcloud.compute.base import (
     Node,
+    KeyPair,
     NodeSize,
     NodeImage,
     NodeDriver,
@@ -861,6 +862,33 @@ class LinodeNodeDriverV4(LinodeNodeDriver):
         data = self._paginated_request("/v4/images", "data")
         return [self._to_image(obj) for obj in data]
 
+    def create_key_pair(self, name, public_key=""):
+        """
+        Creates an SSH keypair
+
+        :param name: The name to be given to the keypair (required).\
+        :type name: `str`
+
+        :keyword public_key: Contents of the public key the the SSH key pair
+        :type public_key: `str`
+
+        :rtype: :class: `KeyPair`
+        """
+        attr = {"label": name, "ssh_key": public_key}
+        response = self.connection.request(
+            "/v4/profile/sshkeys", data=json.dumps(attr), method="POST"
+        ).object
+        return self._to_key_pair(response)
+
+    def list_key_pairs(self):
+        """
+        Provide a list of all the SSH keypairs in your account.
+
+        :rtype: ``list`` of :class: `KeyPair`
+        """
+        data = self._paginated_request("/v4/profile/sshkeys", "data")
+        return [self._to_key_pair(obj) for obj in data]
+
     def list_locations(self):
         """
         Lists the Regions available for Linode services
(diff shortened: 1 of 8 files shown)


Kami:
Thanks for the contribution.

When you get a chance, can you please add some test case for those changes? Thanks.

Kami:
It also seems like that it may be a good idea to rename the driver in the future (Akamai Connected Cloud)?

We can still leave (deprecated) "linode" alias in place for the foreseeable future to make the change backward compatible.

mraygalaxy2:
Acknowledged @Kami . I'll work on that.

Regarding the name change, we haven't been directed to do that right now by the company, particularly as the API endpoint (api.linode.com) is unable to change for the forseeable future.

Dorthu (approved):
Ran though all of the changes and tested them against the API and Metadata Service, everything worked like I expected it to 👍

mraygalaxy2:
okie dokie @Kami  I've pushed those tests. Take a 2nd look?

mraygalaxy2:
NOTE: I do want to remove (or cleanup) this comment: https://github.com/apache/libcloud/pull/1946/commits/d8add44eec876cae1103b99d545a747a0e33a4bc#diff-e2e26eb0ef50c4cf2bf6b12658cd644d7ca1fe06cd2a3bc307d607149c83a2a4R976

Looking for feedback before I do that.

libcloud/compute/drivers/linode.py
```diff
@@ -945,15 +973,28 @@ def reboot_node(self, node):
...
-        image=None,
-        name=None,
+        # Previously, the following 3 parameters did not match the rest of the libcloud
+        # codebase drivers. They should be in the same order as other compute drivers.
+        # Previously, it looked like this:
+        #         size,
+        #         image=None,
+        #         name=None,
+        #
+        # Comments welcome on how backwards compatibility (if any) should work here.
```
Kami, line 983: I think it's fine to leave it as-in and document this backward incompatible / breaking change in the upgrade notes file (https://github.com/apache/libcloud/blob/trunk/docs/upgrade_notes.rst), ideally with some concrete before / after code examples.

Kami:
@mraygalaxy2 Thanks for adding the tests and addressing some very old API inconsistency in the ``create_node()`` method.

I think it would be good to document that breaking API change in the upgrade notes file (https://github.com/apache/libcloud/commit/d8add44eec876cae1103b99d545a747a0e33a4bc#r127271125).

Besides that, LGTM.

mraygalaxy2:
@Kami Thank you, sir. I'll push a change right now to update that document.

libcloud/compute/drivers/linode.py
```diff
@@ -997,6 +1038,12 @@ def create_node(
         :keyword ex_private_ip: whether or not to request a private IP
         :type    ex_private_ip: ``bool``
 
+        :keyword ex_userdata: add cloud-config compatible userdata to be
```
Kami, line 1041: I think it would actually be better to let user pass in a raw string and then we can take care of base64 encoding the string inside the method ourselves.

This way the method abstracts away this implementation detail from the end user (we already do a similar thing in a bunch of places).

Kami:
@mraygalaxy2 Thanks. There is one more small thing which I thought I already commented on in the past (https://github.com/apache/libcloud/pull/1946/files#r1324941487), but maybe I forgot to :)

libcloud/compute/drivers/linode.py
mraygalaxy2: @Dorthu Would you like me to make that change? ^^^

Dorthu: I agree that it would be more convenient for the library to handle the encoding, especially if there's precedent elsewhere in libcloud, as long as that behavior is featured prominently in the documentation.

mraygalaxy2: Will do! Thanks guys.

mraygalaxy2: okie dokie @Kami @Dorthu  I have pushed that change (and tested it against Linode).

mraygalaxy2:
OK, base64 changes complete. Also tell me if those those docs/upgrade_notes.rst and CHANGES.rst changes look ok. I'm happy to clean them up if there's a mistake.

Kami:
Thanks for addressing the review feedback.

I made some small changes and merged it into trunk:

* Reword upgrade notes entry a bit - 01c7e61de89144205a2dcc1db31239bb527b725c
* Default `ex_userdata` to `None` instead of `False` since this makes more sense for an argument which is not a boolean - bb0dba7d0fdc24bf4b21209a6e4267dd6c773a6d
* Add test case for `ex_userdata` argument - bb0dba7d0fdc24bf4b21209a6e4267dd6c773a6d

mraygalaxy2:
Thank you so much, @Kami for the changes! Much appreciated.

PR #1821, opened 2022-12-12
Add support for af south 1 in s3 driver

alier350:
### Description

We noticed the following error when running our Ansible automation on af-south-1 VMs on AWS:
```
File "/usr/local/lib/python3.8/dist-packages/libcloud/storage/drivers/s3.py", line 1134, in __init__
        raise ValueError('Invalid or unsupported region: %s' % (region))
    ValueError: Invalid or unsupported region: af-south-1
```
The reason for the error is missing configuration for the `af-south-1` region of AWS in the S3 driver of libcloud, which we have remediated locally and now raising this PR to contribute back to the project.

### Status

- done, ready for review (Used pylint for linting)

### Checklist (tick everything that applies)

- [x] [Code linting](http://libcloud.readthedocs.org/en/latest/development.html#code-style-guide) (required, can be done after the PR checks)
- [ ] Documentation
- [ ] [Tests](http://libcloud.readthedocs.org/en/latest/testing.html)
- [ ] [ICLA](http://libcloud.readthedocs.org/en/latest/development.html#contributing-bigger-changes) (required for bigger changes)

PR #1950, opened 2023-09-01
S3 eu south

H3199:
### Description

EU-South was added similarly as EU-North.
I couldn't connect to EU-South region with K8ssandra's Medusa, which uses this storage driver to establish connection to AWS. After adding the region Medusa's connection worked.

### Status

- done, ready for review

- [ ] Documentation


libcloud/storage/drivers/s3.py
```diff
@@ -1458,6 +1460,14 @@ class S3EUNorth1StorageDriver(S3StorageDriver):
     ex_location_name = "eu-north-1"
     region_name = "eu-north-1"
 
+class S3EUSouth1Connection(S3SignatureV4Connection):
```
Kami, line 1463: There is no need anymore to define a class per region. A long time ago we moved away from a class per region to `region` constructor argument approach. Other classes were just left in places for backward compatibility reasons.

Just adding an entry to `REGION_TO_HOST_MAP` dictionary is sufficient.

Related comment on a similar PR - https://github.com/apache/libcloud/pull/1821#issuecomment-1368977503.

https://github.com/apache/libcloud/blob/9756484101d6570f03542288d2f38fdb381500d1/libcloud/storage/drivers/s3.py#L1253

Kami:
I've removed unnecessary class for the connection and storage driver and merged change into trunk. 

Thanks.

PR #1954, opened 2023-09-17
Add VPC IP and Elastic IP to ECS node as private and public IP

vicwicker:
### Description

As per documentation [1], the InnerIpAddress field is only relevant
when the node network type is set to classic. If the node is within
a VPC, the private IP is returned by the VpcAttributes/PrivateIpAddress.

In addition, elastic IPs in field EipAddress are also public ips.
They differ from other public IPs in that they are static and stay
the same once allocated.

Fields EipAddress and VpcAttributes are already returned in the response
as part of the "extra" field of the node object. This is however not
consistent with other drivers as these fields are also public and
private IP addresses, respectively.

[1] https://www.alibabacloud.com/help/en/ecs/developer-reference/api-describeinstances

### Status

- done, ready for review

- [x] Documentation


Kami (approved):
LGTM, thanks.

PR #1957, opened 2023-09-23
[azure arm] delete VM OS disk if it is a managed disk

vicwicker:
### Description
VMs on Azure can be launched with managed or unmanaged disks.
Unmanaged disks are typically launched from specific VHDs.
Managed disks are managed by Azure and they do not have a VHD
associated that we can remove (e.g., ubuntu image offered by Azure).

Deleting a VM on the Azure Web console allows you to choose whether
you want to delete the OS disk as well, independently of this is
managed or unmanaged. The original code would allow that only for
unmanaged disks via the 'ex_destroy_vhd' parameter but, as said,
this is incomplete as VMs can be launched from Azure managed disks.

This change proposes to introduce a new parameter 'ex_destroy_os_disk'
to delete the VM OS disk if it is a managed disk. Ideally, this
parameter should default to True as (i) that is the Azure's default
option and (ii) it would align with 'ex_destroy_vhd'. However, this
might pose backward compatibility issues if libcloud starts deleting
managed OS disks by default as well. I am not sure what the libcloud
maintainers' take is in this regard but this would be my proposal.

### Status

- done, ready for review

- [x] Documentation


libcloud/compute/drivers/azure_arm.py
```diff
@@ -912,6 +917,19 @@ def destroy_node(
                         raise
                 time.sleep(10)
 
+        # Optionally destroy the OS managed disk
+        managed_disk = node.extra["properties"]["storageProfile"]["osDisk"].get("managedDisk")
```
Kami: Can we be sure that all those nested objects / dictionaries are indeed always present? Or it would be safer to use `.get("key", {})` notation for all the nested attributes?

libcloud/compute/drivers/azure_arm.py
```diff
@@ -808,6 +808,7 @@ def destroy_node(
         ex_destroy_vhd=True,
         ex_poll_qty=10,
         ex_poll_wait=10,
+        ex_destroy_os_disk=False,
```
Kami: I think it's fine to default it to `True` as long as that change is documented well (including an example of how to set it to `False`) in `docs/upgrade_notes.rst` file.

libcloud/compute/drivers/azure_arm.py
vicwicker: You're right. I didn't find anywhere on Azure documentation the properties are always there so I fell back to your proposal to make it safer. Thanks.

libcloud/compute/drivers/azure_arm.py
vicwicker: Thanks. I set it to `True` but I still need to document it in the notes. I understand we want to keep `ex_destroy_vhd` and `ex_destroy_os_disk` separately. The first to delete the OS disk when it is created through a VHD and `ex_destroy_os_disk` when it is a managed disk. Is that correct?

Kami: Yeah, I think that's correct.

Once you will add the upgrade notes entry, I will go ahead and merge the PR. Thanks.

Kami (approved):
LGTM.

Please just add an entry in the upgrade notes document and then I will go ahead and merge it into trunk.

Kami:
@vicwicker It looks like some lint checks (black) failed. Can you please also fix that when you get a chance?

I think running `tox -eisort,black` should to the trick.

Thanks.

vicwicker:
Sorry for the delay. It seems I forgot to run black after addressing the first comments. I have done now so and I have also added the release notes. I created a new section 3.9.0 but I am not sure if that's what you were asking. I have also run one last test against my use-case. Thanks a lot for you help :)

vicwicker:
@Kami It seems the validation is still failing but I am not sure if this is related to the change directly. AFAIU, the error seems to be a 404 while uploading to codecov. Please let me know if I am reading this wrong and / or my PR requires some fixing. Thanks!

Kami:
@vicwicker Sorry for the delay. I believe issue is related to secrets not being passed to forked repo builds for security reasons.

I think it should pass once I merge it into trunk.

Kami:
PR has been merged into trunk. Thanks again for your contribution and patience :)

PR #1967, opened 2023-10-24
Adjust parameters to GoDaddy get_zone method

Circuitsoft:
Other drivers take the ID, not the domain name. This puts the GoDaddy driver in-line with how other drivers work.

### Description

Saltstack has been throwing errors when working with GoDaddy dns entries because the driver takes different parameters than other drivers it was designed around. This fixes the interface to put it more in-line with other drivers.

### Status

done, ready for review

- [ ] Documentation

### DIFF
--- a/libcloud/dns/drivers/godaddy.py
+++ b/libcloud/dns/drivers/godaddy.py
@@ -288,12 +288,13 @@ class GoDaddyDNSDriver(DNSDriver):
         """
         Get a zone (by domain)
 
-        :param  zone_id: The domain, not the ID
+        :param  zone_id: The domain ID
         :type   zone_id: ``str``
 
         :rtype:  :class:`Zone`
         """
-        result = self.connection.request("/v1/domains/%s/" % zone_id).object
+        domain, = [x.domain for x in self.list_zones() if x.id == zone_id]
+        result = self.connection.request("/v1/domains/%s/" % domain).object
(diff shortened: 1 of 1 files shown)


Kami:
@Circuitsoft Thanks for the contribution.

Since this is a breaking change, can you please add an entry to `docs/upgrade_notes.rst` file? You should be able to find some examples on how to word an entry in that file.

Thanks.

Kami:
@Circuitsoft It appears tests are also failing. 

```
 >        domain, = [x.domain for x in self.list_zones() if x.id == zone_id]
 E       ValueError: not enough values to unpack (expected 1, got 0)
```

Can you please also look into this when you get a chance? Thanks

libcloud/dns/drivers/godaddy.py
```diff
@@ -288,12 +288,13 @@ def get_zone(self, zone_id):
...
         Get a zone (by domain)
 
-        :param  zone_id: The domain, not the ID
+        :param  zone_id: The domain ID
         :type   zone_id: ``str``
 
         :rtype:  :class:`Zone`
         """
-        result = self.connection.request("/v1/domains/%s/" % zone_id).object
+        domain, = [x.domain for x in self.list_zones() if x.id == zone_id]
```
Kami, line 296: This will throw in case no matching items are found.

PR #1973, opened 2023-11-17
fix for TB in list_sizes for eqnx

luss:
### Description

Some newer large instance sizes are measured in TB rather than GB


Kami:
@luss Thanks for the contribution. Can you please add a corresponding unit  / mock http based test? Thanks.

Kami:
I've fixed lint (486ca721fe096190cdcb6fe27dc84366e40b64c8) and added tests (59286b46ca1b4b25caaa445d015ebeae52787d02) and merged change into trunk. Thanks.

PR #1983, opened 2024-01-04
Enhance `KubeVirtNodeDriver` Compute Driver

cdfmlr:
### Description

This pull request brings several improvements to the `create_node` method and related functions within the `KubeVirtNodeDriver` (`libcloud/compute/drivers/kubevirt.py`).

**Features added to `KubeVirtNodeDriver.create_node`**:

- Improved compatibility with the base `NodeDriver` class:
    - Added support for the `size: NodeSize` parameter, while retaining legacy compatibility with `ex_cpu` and `ex_memory`.
    - Added support for the `auth: NodeAuthSSHKey|NodeAuthPassword` parameter.
- Support for deployment (`deploy_node`):
    - `KubeVirtNodeDriver` now supports node deployment automatically, benefiting from the above compatibility changes.
- Support for general volume (disks) types that KubeVirt supports:
    - (==Breaking change==) Modified the content of the `ex_disks` parameter to align with the related [KubeVirt API](https://kubevirt.io/user-guide/virtual_machines/disks_and_volumes/), making it compatible with any volume types rather than hardcoded support for limited volume or disk types (previously only `persistentVolumeClaim` was supported).
- Added support for the `ex_template` parameter, allowing users to customize the entire Kubernetes object declaring the virtual machine. This is particularly useful for:
    - Supporting advanced configurations not covered by other parameters.
    - Facilitating the reuse of existing virtual machine configurations.

**Fixes**:

- `_to_node`: Improved the logic for parsing memory, eliminating crashes on virtual machines with more than 1 GiB RAM.
- `create_node`: (==Breaking change==) Renamed the parameter from `ports` to `ex_ports`.
- ~~Addressed various bugs in the `create_node` method, which were eliminated during the refactor and implementation of new features.~~

**Other Changes**:

- `libcloud/compute/drivers/kubevirt.py`: 
    - Added a `KubeVirtNodeSize` function to assist in constructing `NodeSize` instances for `KubeVirtNodeDriver`.
    - Introduced a `KubeVirtNodeImage` function to help construct `NodeImage` instances for `KubeVirtNodeDriver`.
    - Code reorganization: Moved `DISK_TYPES` out of the `create_node` and exported it, enabling users to access a list of supported disk types.
    - Updated docstrings for new features, adhering to sphinx grammar standards.
- `libcloud/test/compute/test_kubevirt.py`:
    - Introduced tests for the `create_node` method: This method was not tested previously.

### Status

- done, ready for review

- [x] Documentation


Kami:
@cdfmlr Thanks for the contribution and good PR description.

I will have a look as soon as I get a chance. In the mean time, would you mind documenting breaking changes (disks, ports) in `docs/upgrades_notes.rst`?

cdfmlr:
> would you mind documenting breaking changes (disks, ports) in `docs/upgrades_notes.rst`?

Sure, I will add it recently.

libcloud/compute/drivers/kubevirt.py
```diff
@@ -179,125 +245,103 @@ def destroy_node(self, node):
...
-                      the ports to be exposed with TCP protocol,
-                      and 'ports_udp' is a list of ints that indicate
-                      the ports to be exposed with UDP protocol.
-        :type  ports: `dict` with keys
-                      'ports_tcp`: `list` of `int`
-                      'ports_udp`: `list` of `int`
+        # check if new node is present
+        # But why not just use the resp from the POST request?
+        # Or self.get_node()?
+        # I don't think a for loop over list_nodes is necessary.
```
Kami, line 317: Yeah, I agree, either using `list_nodes()` or `get_node()` which calls `list_nodes()` underneath is not really great and efficient...

libcloud/compute/drivers/kubevirt.py
```diff
@@ -391,63 +772,162 @@ def create_node(
                     )
                     raise KeyError(msg)
 
+                claim_name = disk["volume_spec"]["claim_name"]
+
+                if claim_name not in self.ex_list_persistent_volume_claims(namespace=namespace):
```
Kami: The method would be a bit more readable if it was refactored into multiple smaller function (e.g. one for creating volume, etc.).

libcloud/compute/drivers/kubevirt.py
```diff
@@ -391,63 +772,162 @@ def create_node(
...
+                    if (
+                        "size" not in disk["volume_spec"]
+                        or "storage_class_name" not in disk["volume_spec"]
+                    ):
+                        msg = (
```
Kami: Do we have test cases which cover this scenario + all other regular + edge cases?

libcloud/compute/drivers/kubevirt.py
```diff
@@ -391,63 +772,162 @@ def create_node(
...
+            # cloud_init_config reference: https://kubevirt.io/user-guide/virtual_machines/startup_scripts/#injecting-ssh-keys-with-cloud-inits-cloud-config
+            if isinstance(auth, NodeAuthSSHKey):
+                public_key = auth.pubkey
+                cloud_init_config = (
+                    """#cloud-config\n""" """ssh_authorized_keys:\n""" """  - {}\n"""
+                ).format(public_key)
+            elif isinstance(auth, NodeAuthPassword):
+                password = auth.password
+                cloud_init_config = (
+                    """#cloud-config\n"""
```
Kami: It would be a bit more readable if we stored cloud init configs in template files and then load those + render them here.

libcloud/compute/drivers/kubevirt.py
```diff
@@ -1231,3 +1719,151 @@ def ex_delete_service(self, namespace, service_name):
         except Exception:
             raise
         return result.status in VALID_RESPONSE_CODES
+
+
+def _deep_merge_dict(source: dict, destination: dict) -> dict:
```
Kami, line 1912: It would be good to add some test cases for this function.

libcloud/compute/drivers/kubevirt.py
```diff
@@ -1231,3 +1719,151 @@ def ex_delete_service(self, namespace, service_name):
...
+            _deep_merge_dict(value, node)
+        elif key not in destination:  # not existing in destination: add it
+            destination[key] = value
+        elif value:  # existing: update if source's value is non-zero
+            destination[key] = value
+
+    return destination
+
+
+def _memory_in_MB(memory):  # type: (Union[str, int]) -> int
```
Kami, line 1953: Same here, some tests would this function would be good.

Kami:
@cdfmlr Thanks.

I had a look again. It mostly looks good, but there are a couple of improvements which can be made:

* `_create_node()` functionality could be refactored into multiple smaller methods. Right now the method is very large and complex.
* Test case situation should be improved. There is a lot of code and conditional branches which are not not exercised / covered (https://app.codecov.io/gh/apache/libcloud/pull/1983?src=pr&el=tree&filepath=libcloud%2Fcompute%2Fdrivers%2Fkubevirt.py#diff-bGliY2xvdWQvY29tcHV0ZS9kcml2ZXJzL2t1YmV2aXJ0LnB5, https://github.com/apache/libcloud/pull/1983#issuecomment-2080694604). Existing test cases exercise only a small amount of "happy" code paths. I know you didn't add all of that code, but since you updated / touched a lot of it, it would be great to also improve the test coverage.

cdfmlr:
@Kami Thank you for the review. Indeed, the `create_node()` looks clumsy. I've actually started refactoring it, but recent higher priority tasks have delayed its completion.

Also, at the moment, I'm unable to add more test cases due to these priority. However, there are some existing test cases for `_deep_merge_dict` and `_memory_in_MB` that I believe were copied from our other projects. I plan to incorporate these tests soon.

Despite the problems with code readability and test coverage, we've been using this code in a production environment for several months. It has been performing well and handling a considerable number of situations that aren't covered by the test cases.

Kami:
@cdfmlr Thanks.

Do you happen to have any ETA when you will be to add some more tests + refactor the code a bit? Since I'm planning to do a v3.9.0 release this week.

cdfmlr:
@Kami sorry for my delay. 

I have added [more tests](https://github.com/apache/libcloud/pull/1983/commits/f765b4299de9770ac49e9f5343ee28a96d99e26b) covering the helper functions (`_deep_merge_dict` and `_memory_in_MB`) and some typical unhappy code paths. The `_create_node()` method has been [refactored](https://github.com/apache/libcloud/pull/1983/commits/40ec5efd603a75ecb315ea5edd562e4e19baa1cb) into smaller functions as well. Meanwhile, a useful new feature to [set cpu/mem requests and limits separately](https://github.com/apache/libcloud/pull/1983/commits/4c622eef1236032b3548470f310ae0e9b6ac2d8c) is introduced and [tested](https://github.com/apache/libcloud/pull/1983/commits/f78d36141939e1e5fe35a94366d4f98143ee5b31).

libcloud/compute/drivers/kubevirt.py
```diff
@@ -316,69 +360,333 @@ def create_node(
...
+
+        :param ex_memory: The amount of memory to allocate to the VM in MiB.
+                          See also the ex_memory parameter in create_node.
+        :type ex_memory: ``int``
+
+        :return: None
+        """
+        # size -> cpu and memory limits / requests
+
+        ex_memory_limit = ex_memory_request = ex_cpu_limit = ex_cpu_request = None
```
Kami, line 458: It would be safer to do:

```python
ex_memory_limit, ex_memory_request, ex_cpu_limit, ex_cpu_request = None, None, None, None
```

Right now the code above works fine since we are defaulting to `None`, but if this ever changed to default to dictionary or a list this could have unintended side affects.

libcloud/compute/drivers/kubevirt.py
```diff
@@ -316,69 +360,333 @@ def create_node(
...
+                )
+            )
+            vm["metadata"]["name"] = name
+
+        return vm
+
+    @staticmethod
+    def _create_node_size(
+        vm, size=None, ex_cpu=None, ex_memory=None
+    ):  # type: (dict, NodeSize, int, int) -> None
```
Kami, line 437: In the future when you are refactoring code, you can move type annotations from comment directly to the function signature (we don't support Python 2 anymore).

libcloud/compute/drivers/kubevirt.py
```diff
@@ -316,69 +360,333 @@ def create_node(
...
+            "name": cloud_init_volume,
+            "cloudInitNoCloud": {"userData": ""},
+        }
+
+        # cloud_init_config reference: https://kubevirt.io/user-guide/virtual_machines/startup_scripts/#injecting-ssh-keys-with-cloud-inits-cloud-config
+        if isinstance(auth, NodeAuthSSHKey):
+            public_key = auth.pubkey
+            cloud_init_config = (
+                """#cloud-config\n""" """ssh_authorized_keys:\n""" """  - {}\n"""
+            ).format(public_key)
```
Kami, line 625: Could there be any surprises if `auth.pubkey` contains a leading or a trailing line break? Aka do we need to call `.strip()` on the value (ideally that would already happen in the base NodeAuth class, but I need to verify that is indeed the case).

libcloud/compute/drivers/kubevirt.py
```diff
@@ -316,69 +360,333 @@ def create_node(
...
+        elif isinstance(auth, NodeAuthPassword):
+            password = auth.password
+            cloud_init_config = (
+                """#cloud-config\n"""
+                """password: {}\n"""
```
Kami, line 631: Same here - do we need to perform any additional cleaning and sanitization of the password value?

To be on the safe side and to prevent possible YAML injections, we should escape / quote those values (password, pubkey).

Since everything except the header looks like yaml, probably the safest way is to define a dictionary for other fields and then calling `yaml.dump()` on it and appending it to the static header value.

Kami: EDIT: I forgot we don't have a dependency on `pyyaml` yet so we would need to add one which is not that great.

One option which would probably work is to just use `json.dumps()` on the actual pubkey / password value to take care of the value escaping / quoting, but we would need to verify it works correctly.

```python
In [4]: s = """#cloud-config\n""" """ssh_authorized_keys:\n""" """  - {}\n"""

In [5]: print(yaml.safe_load(s.format(json.dumps("key with \" quotes ' bar"))))
{'ssh_authorized_keys': ['key with " quotes \' bar']}
```

It looks like it should indeed do the trick, but more unit tests + testing it with the actual cloud init would be better.

Kami:
@cdfmlr Thanks for adding those changes.

I added a couple of more comments. It's mostly a couple of small things, plus the potential security issue with possible YAML injection in case the pub key or password is supplied by the end user and not sanitized before being passed to the Libcloud code.

libcloud/compute/drivers/kubevirt.py
cdfmlr: agree. json.dumps seems a great workaround here. introduced in https://github.com/apache/libcloud/pull/1983/commits/7d7c1021970db7ef8d7deba4305822cb33b204ce. also added test cases to it.

Kami (approved):

Kami:
Merged into trunk. Thanks.

PR #2033, opened 2024-09-03
Only call super() during MockHttp if required

s-t-e-v-e-n-k:
### Description

With pytest 8.2 and above, any class that contains tests is collected, which means they are instantiated, which MockHttp's superclasses do not accept, since they require keyword arguments. To work around this, only call the superclass's __init__ method if we are passed kwargs.

### Status

Done, ready for review

- [x] Documentation (as code comments)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -135,6 +135,11 @@ Other / Development
   (#1994)
   [Tomaz Muraus - @Kami]
 
+- Add a workaround so tests work with pytest >= 8.2. Also use latest version of
+  pytest for running tests.
+  (#2033)
+  [Steve Kowalik - @s-t-e-v-e-n-k]
+
 Changes in Apache Libcloud 3.8.0
 --------------------------------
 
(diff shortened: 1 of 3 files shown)


Kami:
Thanks for the contribution.

I've intentionally help back at upgrading pytest since we rely on test collection logic which has changed in new version of pytest so various tests are not being collected correctly.

I will test and verify this change locally with latest version of pytest.

Kami:
@s-t-e-v-e-n-k I've tested it locally and it indeed appears to be working fine.

Before:

```
==================================== 8156 passed, 13 skipped, 33 warnings in 21.37s =====================================
========================================== 2 passed, 8169 deselected in 1.18s ===========================================
```

After (with pytest > 8.2):

```
==================================== 8157 passed, 13 skipped, 33 warnings in 22.42s =====================================
========================================== 2 passed, 8170 deselected in 1.18s ===========================================
```

I will merge it into trunk shortly. Thanks.

Kami (approved):

PR #2068, opened 2025-04-16
Removal of zone_name from Cloudflare record response following API deprecation

cha0tic87:
Removal of zone_name from Cloudflare record response following API deprecation

### Description

Cloudflare deprecated zone_name from individual DNS records on 2024-11-30, as outlined at https://developers.cloudflare.com/fundamentals/api/reference/deprecations/:

---------------------------------------------------------------------

Zone information in individual DNS records
Deprecation date: November 30, 2024

Currently, each individual DNS record returned by the API contains information about the zone it is on, specifically the zone ID and name.

{
  "result": [
    {
      // ...
      "zone_id": "ab922473c42f4e50819d7c1c9b81b16b",
      "zone_name": "example.com"
    }
  ],
  // ...
}
This information is redundant because both affected API routes are already within the zone scope. In particular, the zone ID will already be known to any user of these routes because it appears in the URL. The zone name can be retrieved by making a GET request to /zones/:zone_id if it is necessary.

After November 30th, 2024, Cloudflare will stop including the zone_id and zone_name fields on individual DNS records in API responses. These fields are currently ignored when sent to the API as part of a request body, so no changes to request bodies are required.

---------------------------------------------------------------------

This is a breaking change in libcloud when fetching records inside zones managed by Cloudflare. This PR removes the check for zone_name attributes in DNS records returned by the Cloudflare API.

### Status

Ready for review

- [ ] [Code linting](http://libcloud.readthedocs.org/en/latest/development.html#code-style-guide)
- [ ] Documentation


micafer (approved):
LGTM

PR #2088, opened 2026-01-25
Retry checks

tisonkun:
Retest if GitHub Status Checks updated.


tisonkun:
Weird. Let me ask the INFRA team why the status check updated in `.asf.yaml` doesn't apply.

PR #2085, opened 2026-01-18
Adopt uv for dependency management and require Python 3.10

PsiACE:
This is a discussion draft. I propose adopting uv as the primary dependency managerbecause it is currently the best option in the Python ecosystem and lets us lock and unify toolchain dependency versions reliably.

Changes in this draft:

  - Replace legacy `requirements-*.txt` with uv.lock and pyproject.toml extras.
  - Update CI workflows and docs to use uv for installs/locks.
  - Drop Python 3.9 and standardize on Python 3.10+ across tooling.
  - Update Dockerfile to remove 3.9 and use 3.10.

### Checklist (tick everything that applies)

- [ ] [Code linting](http://libcloud.readthedocs.org/en/latest/development.html#code-style-guide) (required, can be done after the PR checks)
- [x] Documentation
- [x] [Tests](http://libcloud.readthedocs.org/en/latest/testing.html)
- [x] [ICLA](http://libcloud.readthedocs.org/en/latest/development.html#contributing-bigger-changes) (required for bigger changes)

PR #2093, opened 2026-01-26
ci: Drop Python 3.9 checks

tisonkun:
This is required by https://github.com/apache/libcloud/pull/2085.


tisonkun:
We may add a `Required` task to gather all necessary tasks for convenient like https://github.com/apache/datasketches-rust/blob/583ac180362a18a4f3f57b651bb99d0000e65729/.github/workflows/ci.yml#L113-L129 later.

PR #2095, opened 2026-01-26
Revert name change in required tests

micafer:
### Description

Revert this change (https://github.com/apache/libcloud/commit/0a37446121e07080a8ea9bb4487e72cce4d1c663#diff-b4c2a69650f9ac84008ad9f745859a645c9466350ffdfe5f919655039ee2e2c3L51) as it sets an incorrect name for required tests:

### Status

- done, ready for review

- [ ] Documentation

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -48,8 +48,8 @@ github:
           - "Unit Tests (Python 3.12)"
           - "Unit Tests (Python 3.13)"
           - "Dependency Review"
-          - "Run Various Lint and Other Checks"
-          - "Build and upload Documentation"
+          - "Run Various Lint and Other Checks (3.10)"
+          - "Build and upload Documentation (3.10)"
 
 notifications:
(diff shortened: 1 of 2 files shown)


.asf.yaml
tisonkun, line 1: We may even keep the version-less one and drop python version in name. This way, when we bump toolchain the next time, we don't need to work around it again.

micafer: but it seems that if you use a matrix, the value of the item in the matrix will be always appended so, we should remove the matrix from the tasks

tisonkun: Yes. It seems the matrix has always only one argument and we can inline it.

If there is a real matrix, we can use an aggregation trick like https://github.com/fast/logforth/blob/7cad34bdc63a9692040143497ddbf346f5d10264/.github/workflows/ci.yml#L108-L124

PR #2128, opened 2026-03-23
Add verify_ssl option

micafer:
### Description

Add verify_ssl option, to enable the user to avoid SSL verification explicitly.
Also fix version of setup-uv and update vulneable deps to pass all the tests.

### Status

- Ready for review

- [ ] Documentation


micafer:
@PsiACE could you take a look, it seems that "astral-sh/setup-uv" action is not allowed now.

PsiACE:
hi @micafer , astral-sh/setup-uv@v7 needs to be pinned to a specific hash value, like `astral-sh/setup-uv@37802adc94f370d6bfd71619e3f0bf239e1f3b78`

Would you like to give it a try?

micafer:
Same error.

tisonkun:
https://github.com/apache/infrastructure-actions/blob/72e2dbde7dec469a2aad118a8e931f5d7d9e65f2/actions.yml#L21-L24

I assume either the tag or hash linked here should work.

Background can be https://infra.apache.org/blog/trivy_security_incident.html.

micafer:
Now the tests are running, thanks @PsiACE and @tisonkun !

micafer:
please @PsiACE and @tisonkun review this, the change is minor but important (commit e68658a3aac6378268e8b521e7532e59ccd5cb0c).
The rest of commits are done update some deps to pass the tests.

libcloud/compute/drivers/vsphere.py
tisonkun, line 1: This is the non-trival change.

tisonkun (commented):
Since I don't see any change that calls `VSphereNodeDriver(...)` with different `verify_ssl` param, I expect that it's always `true` now? Or users would initialize their own `VSphereNodeDriver` with `verify_ssl=false` if desired?

micafer:
> Since I don't see any change that calls `VSphereNodeDriver(...)` with different `verify_ssl` param, I expect that it's always `true` now? Or users would initialize their own `VSphereNodeDriver` with `verify_ssl=false` if desired?

Exactly, only if the user sets explicitly `verify_ssl=false` the certificates will not be validate, el the rest of cases they will.

tisonkun (approved):
LGTM. Thanks for working on this!

PR #2159, opened 2026-07-19
Fix Azure ARM create_node JSON error with ex_customdata (#1893)

### LINKED ISSUE #1893: Azure arm ex_customdata causes json error
## Summary

For Azure arm, create_node, the ex_customedata parameter is not properly decoded after base64 encoding. This results in invalid json. When the driver creates the request the error `Object of type bytes is not JSON serializable` occurs.

## Detailed Information

libcloud 3.7.0
Python 3.10.11
Linux ubuntu/debian

To replicate, call azure arm driver.create_node with parameter: ` ex_customdata=str.encode(deployment_script)`,

suggested fix: https://github.com/foospidy/libcloud/commit/20df1c327da71d364132ff8b6e8152dd4554aaad

Sanjays2402:
### Description

Fixes #1893.

For the Azure ARM driver, `create_node` failed when `ex_customdata` was provided:

- `base64.b64encode()` returns `bytes`, which is not JSON serializable, so building the request body raised `Object of type bytes is not JSON serializable`.
- Passing `ex_customdata` as a `str` also failed because `b64encode` requires bytes-like input.

The fix normalizes `ex_customdata` to `bytes`, then decodes the base64 result to a UTF-8 `str` so `customData` is JSON-serializable. Both `str` and `bytes` inputs are now accepted.

### Status

- done, ready for review

- [x] [Code linting](http://libcloud.readthedocs.org/en/latest/development.html#code-style-guide) (required, can be done after the PR checks) — `black --check` and `flake8` clean on changed files
- [ ] Documentation
- [x] [Tests](http://libcloud.readthedocs.org/en/latest/testing.html) — added `test_create_node_ex_customdata` (str + bytes); fails without the fix, passes with it; full `test_azure_arm.py` = 46 passed
- [ ] [ICLA](http://libcloud.readthedocs.org/en/latest/development.html#contributing-bigger-changes) (required for bigger changes) — small bugfix

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -21,6 +21,12 @@ Compute
   (#2135)
   [Steve Kowalik - @s-t-e-v-e-n-k]
 
+- [Azure ARM] Fix ``create_node`` failing with a JSON serialization error when
+  ``ex_customdata`` is provided. The base64 encoded custom data is now decoded
+  to a ``str`` and both ``str`` and ``bytes`` inputs are accepted.
+  (GITHUB-1893)
+  [Sanjay Santhanam - @Sanjays2402]
+
 Changes in Apache Libcloud 3.9.1
 --------------------------------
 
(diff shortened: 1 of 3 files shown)


micafer (approved):
LGTM

PR #2165, opened 2026-07-21
Upgrade Sphinx docs toolchain (6.2.1 → 8.1.3)

ChihweiLHBird:
### What
Modernize the `docs` dependency group and sync the lockfile:

| Package | Before | After |
|---|---|---|
| sphinx | 6.2.1 | 8.1.3 |
| sphinx_rtd_theme | 2.0.0 | 3.1.0 |
| docutils (transitive) | 0.19 | 0.21.2 |

### Why
- `sphinx==6.2.1` cannot run on Python 3.12+ (`ast.Str`, removed in 3.12), blocking local doc builds on modern interpreters.
- `8.1.3` is the newest Sphinx that still supports the project's minimum Python (3.10), so `requires-python` is unchanged.
- `sphinx_rtd_theme==3.1.0` supports Sphinx 6–9 and constrains `docutils<0.23`.

### Changes
- **`pyproject.toml`** — bump the `docs` group as above.
- **`uv.lock`** — regenerated to match (RTD/CI use `uv sync --frozen`, which fails on a stale lock).
- **`.readthedocs.yml`** — removed the obsolete note about pinning an older Sphinx to fix bullet-point rendering. That was a `sphinx_rtd_theme < 1.0` / `docutils >= 0.17` issue, long fixed in the current theme.


ChihweiLHBird:
~~Wait, there's a `v8.2.3`, let me try that version~~ Never mind, `v8.2` doesn't support Python 3.10

.readthedocs.yml
```diff
@@ -24,6 +24,3 @@ sphinx:
 formats:
    - pdf
    - epub
-
-# NOTE: We need to use older version of sphinx otherwise bullet points won't be rendered correctly
-# on RTD using rtd theme.
```
tisonkun, line 29: @ChihweiLHBird Do you verify this note is no longer relevant? If so, may you provide a screenshot or recording as evidence.

ChihweiLHBird: Yes, I verified this is no longer an issue. And double checked today for the bullet points thing:

Image

PR #2168, opened 2026-07-23
[S3] Respect chunk_size in download_object_as_stream

### LINKED ISSUE #1798: chunk_size is ignored for download_object_as_stream with s3 provider
## Summary

When downloading an object from S3 using download_object_as_stream or as_stream, the chunk_size argument is ignored.  Instead the hardcoded CHUNK_SIZE in libcloud.storage.drivers.s3 dictates the number of bytes returned from the iterator with each next call.

## Detailed Information

This is using libcloud 3.6.1 with Python 3.10.8 on an M1 MacBook Pro running Montery 12.6.1

Here's an outline for how to reproduce:

```

driver = get_driver(Provider.S3_US_EAST2)("Your account key", "Your secret")
bucket = driver.create_container("bucket")
data = bucket.get_object("a file").as_stream(chunk_size=1024 * 1024 * 10)

print(len(next(data)))
```

If the file is large enough, the length printed is 5MB despite the 10MB chunk_size.

I can work around this by patching CHUNK_SIZE on the module, but that seems not great.

I understand that the chunk size must be at least 5MB for upload, but that restriction doesn't appear to apply to downloads.

Sanjays2402:
Closes #1798

### Description

`download_object_as_stream` and `download_object_range_as_stream` in the S3
driver passed the hardcoded module-level `CHUNK_SIZE` to
`response.iter_content()` instead of the `chunk_size` given by the caller, so a
user-specified chunk size was silently ignored. They now forward `chunk_size`
(falling back to `CHUNK_SIZE`), matching the backblaze_b2 and cloudfiles
drivers. Added a regression test.

### Status

- done, ready for review

- [x] Documentation

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -34,6 +34,16 @@ Compute
   (#2147)
   [Miguel Caballer - @micafer]
 
+Storage
+~~~~~~~
+
+- [S3] Fix ``chunk_size`` argument being ignored by
+  ``download_object_as_stream`` and ``download_object_range_as_stream``. The
+  requested chunk size is now passed through to the underlying iterator
+  instead of the hardcoded default.
+  (GITHUB-1798)
+  [Sanjay Santhanam - @Sanjays2402]
+
 Changes in Apache Libcloud 3.9.1
 --------------------------------
 
--- a/libcloud/storage/drivers/s3.py
+++ b/libcloud/storage/drivers/s3.py
@@ -521,7 +521,7 @@ class BaseS3StorageDriver(StorageDriver):
             callback=read_in_chunks,
             response=response,
             callback_kwargs={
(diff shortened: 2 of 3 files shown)


micafer:
Looks good for me but could you add also a test for `download_object_range_as_stream`

Sanjays2402:
Good catch, `download_object_range_as_stream` goes through the same path so it should be covered too. Will add a test for it and push.

Sanjays2402:
Added `test_download_object_range_as_stream_uses_chunk_size`, mirroring the existing one. Sanity checked it by reverting the driver change locally - both tests fail without the fix and pass with it. Full `test_s3.py` is green (53 passed, 1 skipped).

micafer (approved):
LGTM

PR #2169, opened 2026-07-23
[Azure Blobs] Forward chunk_size in download_object_as_stream

### LINKED ISSUE #1698: download_object_as_stream for Azure (azure_blobs.py) silently ignores caller's chunk_size without raising an exception.
## Summary

The download_object_as_stream for Azure located in azure_blobs.py will silently ignore the caller's specified chunk_size. It should probably raise an exception if the caller's value is not acceptable. Generally, much of chunk-based upload/download seems dependent on knowing all chunks will equal chunk_size until the last chunk. If libcloud silently changes the caller's expectation, it can lead to confusion (see example in details below).

## Detailed Information

[Below is snippet from azure_blobs.py](https://github.com/apache/libcloud/blob/f3aa980ee55ac914b340c5d129435357c30082c8/libcloud/storage/drivers/azure_blobs.py#L717) ...

<pre>
AZURE_DOWNLOAD_CHUNK_SIZE = (
    int(os.getenv("LIBCLOUD_AZURE_DOWNLOAD_CHUNK_SIZE_MB", "4")) * 1024 * 1024 # 4194304 bytes
)
...
def download_object_as_stream(self, obj, chunk_size=None): 
    """
    @inherits: :class:`StorageDriver.download_object_as_stream`
    """
    obj_path = self._get_object_path(obj.container, obj.name)
    response = self.connection.request(
        obj_path, method="GET", stream=True, raw=True
    )
    iterator = response.iter_content(<b>AZURE_DOWNLOAD_CHUNK_SIZE</b>)
...
drv.download_object_as_stream(storage_object, <b>chunk_size=5*1024*1024</b>) # as one example
</pre>

In the above example, the caller specified 5MB but immediately receives a smaller chunk given silent chunk_size downgrade by libcloud.

Sanjays2402:
### Description

Closes #1698. `download_object_as_stream` and `download_object_range_as_stream` in the Azure Blobs driver passed the hardcoded `AZURE_DOWNLOAD_CHUNK_SIZE` to `iter_content` instead of the caller's `chunk_size`, so a requested size was silently ignored. Both now forward `chunk_size` (falling back to the default when `None`), matching the S3 driver's behavior.

### Status

- done, ready for review

- [ ] Documentation

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -34,6 +34,16 @@ Compute
   (#2147)
   [Miguel Caballer - @micafer]
 
+Storage
+~~~~~~~
+
+- [Azure Blobs] Fix ``chunk_size`` argument being ignored by
+  ``download_object_as_stream`` and ``download_object_range_as_stream``. The
+  requested chunk size is now forwarded to the underlying iterator instead of
+  always using ``AZURE_DOWNLOAD_CHUNK_SIZE``.
+  (GITHUB-1698)
+  [Sanjay Santhanam - @Sanjays2402]
+
 Changes in Apache Libcloud 3.9.1
 --------------------------------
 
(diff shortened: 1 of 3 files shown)


micafer (approved):
LGTM

Issue #2133, opened 2026-03-29
Method "create_node" overrides class "NodeDriver" in an incompatible manner in multiple drivers


ChihweiLHBird:
## Summary

Most drivers `create_node` function is not compatible with the base class' function and causing some IDE warnings.
For example, in `azure`:

```
Method "create_node" overrides class "NodeDriver" in an incompatible manner
  Parameter 5 name mismatch: base parameter is named "location", override parameter is named "ex_cloud_service_name"
  Parameter 6 name mismatch: base parameter is named "auth", override parameter is named "ex_storage_service_name" Pylance (reportIncompatibleMethodOverride)
```

Fixing all of them will definitely introduce breaking changes. So more discussion would be needed apparently.

micafer:
Hi @ChihweiLHBird,

You are totally right.
We must review the code carefully, and probably we need to wait to release a major version to fix all of them adding and ordering the NodeDriver.create_node parameters properly.

ChihweiLHBird:
@micafer Thanks for looking into this! What's the timeline for the next major version? Do you need help to fix them? I can put up a draft PR

PR #2170, opened 2026-07-24
Unify NodeDriver functions

micafer:
### Description

Many drivers overrides class `NodeDriver` functions in an incompatible manner ( #2133)
This PR tries to unify them using the function definition of the `NodeDriver` class.

### Status

- done, ready for review

- [x] Documentation


micafer:
@ChihweiLHBird could you take a look?

PsiACE (approved):
lgtm

libcloud/compute/drivers/linode.py
```diff
@@ -705,24 +726,33 @@ def ex_get_volume(self, volume_id):
...
+        name,
+        description=None,
+    ):
         """Creates a private image from a LinodeDisk.
          Images are limited to three per account.
 
-        :param disk: LinodeDisk to create the image from (required)
-        :type disk: :class:`LinodeDisk`
+        :param node: LinodeDisk to create the image from (required)
+        :type node: :class:`LinodeDisk`
```
ChihweiLHBird: Technically the node should be a Node class... but in Linode this is complicated because every node can have multiple config profiles for booting different disks and the disk is required for image creation API call, and there isn't a very obvious way to determine which config profile or disk booted the node most recently... I guess we can have this called `node` but typed with `LinodeDisk` for now. I can look into the way to fully resolve it in the future

ChihweiLHBird:
- `outscale.create_volume`, required `name` is never used
- `openstack.create_image`, `description` is accepted and never read

I don't think these are blockers, but worth noting for future work.

ChihweiLHBird (approved):

libcloud/compute/drivers/linode.py
micafer: Hi @ChihweiLHBird,
Thanks for cheking this.
I propose I new commit to improve this, could you please take a look?

micafer:
> * `outscale.create_volume`, required `name` is never used
> * `openstack.create_image`, `description` is accepted and never read
> 
> I don't think these are blockers, but worth noting for future work.

I also fix the opentack one to add the description.

libcloud/compute/drivers/linode.py
ChihweiLHBird: Sorry about the delay... yes, it looks good to me! Thank you so much!

PR #2172, opened 2026-07-25
[local storage] Remove IPC lock file after releasing the lock

### LINKED ISSUE #1975: Improper lock usage leads to inode overflow in LockLocalStorage implementation
## Summary

Behaviour of [fasteners.InterProcessLock](https://github.com/harlowja/fasteners/blob/06c3f06cab4e135b8d921932019a231c180eb9f4/fasteners/process_lock.py#L114) is pretty weird: class creates a lockfile by provided path, if it doesn't exist, but not manage to remove it after lock is released.

It may lead to **uncontrolled lockfiles spam** in `/tmp` folder just because libcloud local driver is not removing [this lockfile](https://github.com/apache/libcloud/blob/trunk/libcloud/storage/drivers/local.py#L85) either.

## Detailed Information

This issue encountered in cassandra-medusa `v0.15` and lower, which was using `apache-libcloud<3.4.0,>=3.3.0` as a dependency.

Please see https://github.com/thelastpickle/cassandra-medusa/issues/528 for more details.

---

Seems like the lightweight fix is to run 

```python
                with contextlib.suppress(FileNotFoundError):
                    os.remove(filename)
```

just right in the [exit method](https://github.com/apache/libcloud/blob/trunk/libcloud/storage/drivers/local.py#L114).

Sanjays2402:
Closes #1975

### Description

`fasteners.InterProcessLock` creates its lock file but never removes it, and
`LockLocalStorage.__exit__` did not remove it either. The file name is a sha256
of the locked path, so one file was left behind in the temp directory for every
distinct path ever locked -- a long running process writing many objects can
exhaust inodes. The lock file is now removed after the IPC lock is released.

### Status

done, ready for review

- [x] Documentation

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -37,6 +37,12 @@ Compute
 Storage
 ~~~~~~~
 
+- [Local] Remove the IPC lock file once ``LockLocalStorage`` releases the lock.
+  Previously ``fasteners`` left the file behind, so one file accumulated in the
+  temporary directory for every distinct path which was locked.
+  (GITHUB-1975)
+  [Sanjay Santhanam - @Sanjays2402]
+
 - [Azure Blobs] Fix ``chunk_size`` argument being ignored by
   ``download_object_as_stream`` and ``download_object_range_as_stream``. The
   requested chunk size is now forwarded to the underlying iterator instead of
--- a/libcloud/storage/drivers/local.py
+++ b/libcloud/storage/drivers/local.py
@@ -117,6 +117,14 @@ class LockLocalStorage:
         if self.ipc_lock.exists():
             self.ipc_lock.release()
 
+            # NOTE: fasteners doesn't remove the lock file it creates, so
+            # without this the file accumulates in the temporary directory
+            # for every distinct path which is ever locked.
+            try:
+                os.remove(self.ipc_lock_path)
+            except OSError:
+                pass
+
         if value is not None:
             raise value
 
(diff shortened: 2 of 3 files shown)


micafer (changes requested):
Removing the lock file immediately after releasing InterProcessLock introduces a race which can break mutual exclusion.

For example, process B may acquire the existing file lock immediately after process A releases it. A can then unlink the path, and process C can create and lock a new file at the same path. B and C would consequently hold locks on different inodes and could both enter the critical section.

The new test only verifies that the file is removed in a non-concurrent scenario, so it does not cover this race or verify that mutual exclusion is preserved.

I don't think the lock file can safely be removed as part of __exit__. We likely need a different bounded-lock or out-of-band cleanup strategy.

Sanjays2402:
You're right, and the race is worse than the ordering I had in mind.

Even if I move the `os.remove` to before `release()` so the unlink happens while the lock is still held, it doesn't help: a process already blocked in `acquire()` is blocked on an fd for the *old* inode, so it wakes up holding a lock on an unlinked file while the next process creates a fresh file at the same path and locks that. Two holders, different inodes, exactly the scenario you described. There is no ordering of unlink and release inside `__exit__` that avoids it, because the identity of the lock is the inode and the lookup key is the path.

The only ways I can see to make it safe are all heavier than the leak deserves:

- a second, permanent lock file guarding create+unlink of the real one, which just moves the leak to the guard file (one per path, same cardinality);
- refcounting holders in the file itself, which needs its own atomicity story across processes;
- out-of-band cleanup (age-based sweep of `<tmp>/lock/` at driver init or a maintenance entry point), which is safe but is a different feature from what this PR claims.

And the thing being leaked is a zero-byte file per distinct locked path, bounded by the set of paths the process touches, in a directory the OS already reclaims. That's a poor trade against any risk to mutual exclusion.

So I'll concede this one — the approach in the PR is not fixable in place. Happy to close it. If you think the age-based sweep is worth having as a separate change I can open that instead, but I'd rather you say so than have another speculative PR land in your queue.

PR #2185, opened 2026-09-02
Fix is_valid_ip_address raising ValueError on embedded null byte

Ashishjob:
## Summary

`libcloud.utils.networking.is_valid_ip_address` is a predicate that should return `True`/`False` for whether a string is a valid IP address. It calls `socket.inet_pton` and catches `OSError` to return `False` for malformed input.

However, `socket.inet_pton` raises **`ValueError`** (not `OSError`) when the address string contains an **embedded null byte**, e.g. `"1.2.3.4\x00"`. That `ValueError` was not caught, so instead of returning `False` the function propagated an exception to the caller:

```python
>>> from libcloud.utils.networking import is_valid_ip_address
>>> is_valid_ip_address("1.2.3.4")
True
>>> is_valid_ip_address("1.2.3.4\x00")
ValueError: embedded null character   # expected: False
```

## Fix

Catch `ValueError` as well and return `False`, since a string with an embedded null byte is not a valid address.

## Tests

Added regression cases (IPv4 and IPv6 addresses containing embedded null bytes) to the existing `test_is_valid_ip_address`. Verified they fail before the change (uncaught `ValueError`) and pass after it. Full `libcloud/test/test_utils.py` passes (23 passed, 1 skipped).

## Changelog
Changelog entry added under Common in a follow-up commit referencing this PR number.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,13 @@ Common
   (#2152)
   [Miguel Caballer - @micafer]
 
+- [Utils] Fix ``is_valid_ip_address`` raising ``ValueError`` instead of
+  returning ``False`` for an address containing an embedded null byte
+  (e.g. ``"1.2.3.4\x00"``). ``socket.inet_pton`` raises ``ValueError`` rather
+  than ``OSError`` in that case, which was not caught.
+  (#2185)
+  [Ashish - @Ashishjob]
+
 Compute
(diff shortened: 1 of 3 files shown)


micafer (approved):
LGTM

micafer:
Tests are failing could you check it?

